PR #40, opened 2024-05-05
Sticky header

lkleuver:
Implement a global controllable UI system starting with a fixed compact header when scrolling

Relates to kiesraad/abacus-internal#132


praseodym:
Let's create a new PR when the branch has been updated.

PR #157, opened 2024-07-17
Load polling station and election data from the database for data entry

rnijveld:
Two notes:
- I've changed `id` and `election_id` fields in a `PollingStation` to be `u32`, given that is the input format we've used in several urls, but technically SQLite always stores `i64`. It feels highly unlikely that a single installation of abacus will ever reach the 4 billion rows required for a `u32` to overflow, in which case we would run into conversion errors. We could update urls to accept `i64` instead, but then we do allow negative numbers, which don't make much sense. I don't think that would result in any particularly weird thing, as those negative numbers would just end up being 404 after a database check. If we don't change it this way and keep the `u32` it will mean that from now on we will always need the type coercion syntax that sqlx has and can probably never do a `SELECT * FROM table`
- ~~I've added a new migration that creates the foreign keys, but it repeats a lot from the previous migrations. If we are ok with modifying and reordering the existing migrations that would clean this up a whole lot.~~


backend/src/lib.rs
```diff
@@ -173,7 +173,11 @@ impl From<serde_json::Error> for APIError {
 
 impl From<sqlx::Error> for APIError {
     fn from(err: sqlx::Error) -> Self {
-        APIError::SqlxError(err)
+        if matches!(err, RowNotFound) {
```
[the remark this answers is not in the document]
rnijveld: I missed that! I've removed it!

backend/src/polling_station/structs.rs
```diff
@@ -19,7 +19,8 @@ pub trait Validate {
 /// Polling station of a certain [Election]
 #[derive(Serialize, Deserialize, ToSchema, Debug, FromRow)]
 pub struct PollingStation {
-    pub id: i64,
+    pub id: u32,
+    pub election_id: u32,
     pub name: String,
     pub number: i64,
     pub number_of_voters: Option<i64>,
```
praseodym, line 26: Also change these to `u32`?

rnijveld: Already pushed the change!

rnijveld:
> `u32` is fine, we also use that for election `id`s.

Are we ok with having to use the coercion syntax and always having to iterate all the columns of a table for queries using the `query_as!` macro? If we use `u32` in the struct and want to run a `SELECT * FROM table` we get a type error from the Rust compiler unless we manually add all of the columns in the table. It will probably mean having to repeat ourselves a whole lot.

> You can change existing migrations. This will force everyone to reset their database, but I think that's a fine tradeoff if it means we can keep our migration scripts simple. We don't have any production data so no need to write migrations for that reason either.

I'll update the PR to make it a little more readable!

cikzh (dismissed):

praseodym:
> > `u32` is fine, we also use that for election `id`s.
> 
> Are we ok with having to use the coercion syntax and always having to iterate all the columns of a table for queries using the `query_as!` macro? If we use `u32` in the struct and want to run a `SELECT * FROM table` we get a type error from the Rust compiler unless we manually add all of the columns in the table. It will probably mean having to repeat ourselves a whole lot.

I don't think we'll be doing a lot of `SELECT *` queries, so there shouldn't be a lot of repetition. In any case, I prefer accurate types over terser code.

praseodym:
Looks good! I think we should add a test to check that an invalid polling station id will return a 404 status code.

praseodym (approved):

Lionqueen94 (approved):

jschuurk-kr (dismissed):

Issue #133, opened 2024-07-05
Frontend: update data entry navigation


Lionqueen94:
See https://github.com/kiesraad/abacus/issues/50#issuecomment-2210730723 for behaviour of navigating between different parts of the data entry and https://github.com/kiesraad/abacus/issues/50#issuecomment-2210833946 for the design links. Also see ["Meerdere fouten en waarschuwingen in 1 response"](https://github.com/kiesraad/abacus-internal/blob/main/use-cases/GSB-validatieregels-plausibiliteitschecks.md#meerdere-fouten-en-waarschuwingen-in-1-response).

Routes changes will be made in #184

jschuurk-kr:
Suggestion no.2 change
`<Route path=":electionId/input" element={<InputLayout />}>`
to
`<Route path="elections/:electionId/input" element={<InputLayout />}>`

Issue #99, opened 2024-06-24
Epic: refine validation


praseodym:
Refine data entry validation to implement [all documented rules](https://github.com/kiesraad/abacus-internal/blob/main/use-cases/GSB-validatieregels-plausibiliteitschecks.md), show copy to the user, and make sure all of this is tested.

- [x] #135
- [x] #188 
- [x] #136
- [x] #149 
- [x] #85
- [x] #86
- [x] #185
- [x] #150
- [x] #151
- [x] #186
- [x] #271

PR #166, opened 2024-07-19
Add recounted, voters_recounts and differences fields to api including validation and voters_recounts to frontend

### LINKED ISSUE #110: Backend: add 'recounted', 'differences' fields to API, including validation
See epic (#97) for more details.

### TODO:
- [x] Implement validation rules in backend and texts in frontend: https://github.com/kiesraad/abacus-internal/blob/main/use-cases/GSB-validatieregels-plausibiliteitschecks.md#regels-voor-verschillen
- [x] Validate that field D should be equal to field H, unless the differences are explained in the relevant fields.
- [x] Add mock-api validation

### LINKED ISSUE #113: Frontend: if recounted, add entry of fields A.2 B.2 C.2 D.2 on the number of voters/votes screen
See epic (#97) for more details.

Lionqueen94:
Closes #110 
Closes #113 

### TODO:
- [x] Add rule W.306 to backend and mock and add test


backend/src/polling_station/structs.rs
```diff
@@ -510,27 +764,234 @@ mod tests {
...
+        };
+        let election = election_fixture(&[1]);
+        polling_station_results.validate(
+            &election,
+            &mut validation_results,
+            "polling_station_results".to_string(),
+        );
+        assert_eq!(validation_results.errors.len(), 4);
         assert_eq!(validation_results.warnings.len(), 0);
+        // TODO: Is showing F.11 Incorrect Total still wanted if there was a recount?
```
Lionqueen94, line 830: @praseodym @jorisleker If there was a recount, what do we do with a potential error in the original voters count. I'd say only show the error in voters recount if there is one, but the current implementation will show both and will be blocking.

praseodym: Good point! I think it's weird to block on this (the recount has been done for a reason) but that brings up the question why the original counts need to be entered at all. @jorisleker do you know more about this?

Lionqueen94: Aha good question, I'd say to reflect there actually was a recount, if this is actually showing up in the PV though...

Lionqueen94: I have now changed the code to validate voters_counts if not recouted and to validate voters_recounts if recounted.

backend/src/validation.rs
```diff
@@ -47,6 +51,10 @@ impl fmt::Display for ValidationResultCode {
             ValidationResultCode::AboveThreshold => write!(f, "Above threshold"),
             ValidationResultCode::EqualInput => write!(f, "Equal input"),
             ValidationResultCode::IncorrectCandidatesList => write!(f, "Incorrect candidates list"),
+            ValidationResultCode::MissingRecounts => write!(f, "Missing recounts"),
+            ValidationResultCode::IncorrectDifference => write!(f, "Incorrect difference"),
+            ValidationResultCode::WrongDifferences => write!(f, "Wrong differences"),
```
jschuurk-kr, line 56: The difference between the code `IncorrectDifference` and `WrongDifferences` was not immediately obvious to me.

Suggestion: change `WrongDifferences` to `ConflictingDifferences`.

backend/src/validation.rs
```diff
@@ -47,6 +51,10 @@ impl fmt::Display for ValidationResultCode {
...
+            ValidationResultCode::NoDifference => write!(f, "No difference"),
```
jschuurk-kr, line 57: The other codes state what the problem is with the input. For `NoDifference` that would imply no difference was inputted. However, the validation for this code is that at least one difference field was inputted, when it shouldn't have.

Unfortunately, I don't have a suggestion yet for a better name for the code. Will think some more on it.

jschuurk-kr: Best I can come up with today is to also make this an `IncorrectDifference`, because a value was filled in where that value should be `0`.

jorisleker: Maybe `NoDifferencesExpected`?

backend/src/validation.rs
Lionqueen94: Done in ed1c100539a85fb761d3dd72d650e5481f6854ea! Was struggling to find an understandable code for this

backend/src/validation.rs
Lionqueen94: Done in ed1c100539a85fb761d3dd72d650e5481f6854ea! Was struggling to find an understandable code for this

backend/src/polling_station/structs.rs
praseodym: As discussed with @arjen-ba (and @jschuurk-kr and @marjoleintamis) in the standup:
- A to D (rubriek 1) should not validated if A.2 to D.2 (rubriek 3) is filled in. (Sometimes the original numbers A to C don't even add up to D).
- To make the data entry process easier, it might be easier to follow the order of the original PV and also enter the values for A to D into Abacus when A.2 to D.2 is filled in (@jorisleker will have an opinion on this).
- The numbers A to D are not used when A.2 to D.2 are filled in. However, they might be useful in the EML file to be used for data analysis.

frontend/app/component/form/voters_and_votes/VotersAndVotesForm.tsx
```diff
@@ -222,7 +239,53 @@ export function VotersAndVotesForm() {
...
             isTotal
+            addSeparator={recounted}
           />
+          {recounted && (
+            <>
+              <InputGridRow
+                key="A.2"
+                field="A.2"
+                id="poll_card_recount"
+                title="Stempassen (na hertelling door gemeentelijke stembureau)"
```
jorisleker, line 250: @Lionqueen94 I've made a small tweak to the design to prevent the repetition of labels. Hope you don't mind. See [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3315-15053&t=Se55Zt4P8UonnJcZ-4).

jorisleker (dismissed):

jschuurk-kr:
On the recounted screen, when I check a radiobutton and click "Volgende", "Success" appears at the top of the screen. Sometimes it appears cleanly, i.e. it appears and the rest of the contents moves down. Other times, there's a stutter as if it renders twice. Which reminds me of the discussion about `useEffect()`.

@lkleuver I assume you're removing the "Success" on the pages as part of #133?

backend/src/polling_station/structs.rs
```diff
@@ -73,33 +79,126 @@ impl Validate for PollingStationResults {
...
+                fields: vec![
+                    format!("{field_name}.more_ballots_count"),
+                    format!("{field_name}.fewer_ballots_count"),
+                ],
+                code: ValidationResultCode::ConflictingDifferences,
+            });
+        }
+
+        // W.31 (recounted = false) or W.32 (recounted = true)
+        // validate that no difference should be filled in when there is no difference in the totals
```
jschuurk-kr, line 170: Shouldn't this be an error?

backend/src/polling_station/structs.rs
```diff
@@ -73,33 +79,126 @@ impl Validate for PollingStationResults {
...
+                validation_results.warnings.push(ValidationResult {
+                    fields: vec![format!("{field_name}.more_ballots_count")],
+                    code: ValidationResultCode::NoDifferenceExpected,
+                });
+            }
+            if self.differences_counts.fewer_ballots_count != 0 {
+                validation_results.warnings.push(ValidationResult {
+                    fields: vec![format!("{field_name}.fewer_ballots_count")],
+                    code: ValidationResultCode::NoDifferenceExpected,
+                });
```
jschuurk-kr, line 185: I can't get these to show up in the frontend. We also don't have a test for these.

The response contains `"fields":["data.more_ballots_count"]` while I'd expect `data.differences_counts.more_ballots_count`.

backend/src/polling_station/structs.rs
Lionqueen94: No, because it's possible people filled in the paperwork wrong, so it should be possible to fill in what the paper says, but it should give a warning, something is wrong here. This was discussed before with @jorisleker

backend/src/polling_station/structs.rs
Lionqueen94: Nice catch! I have fixed the field paths, please let me know if there is any other issues

backend/src/polling_station/structs.rs
jschuurk-kr: That line of reasoning can be applied to all current errors, leaving us only with warnings.

I thought the distinction between errors and warnings was: errors for what can't be right, warnings for what seems odd but can be right. Filling in a difference when there is none, sounds to me like the former, not the latter.

jorisleker: Not sure for this one. This section of the proces-verbaal is pretty confusing, and it does not really guide the user in how these numbers should add up (which is something we can fix of course). Are there any rules / checks performed by CSB on these fields?

image

praseodym: The CSB checks for 'onverklaarde verschillen' ('geen verklaring' in this model) in the controleprotocol, but there is no guarantee that these differences add up because OSV2020 doesn't check them.

backend/src/polling_station/structs.rs
```diff
@@ -316,33 +423,205 @@ impl Validate for VotesCounts {
...
+            format!("{field_name}.no_explanation_count"),
+        );
+
+        // F.26 validate that more_ballots_count == too_many_ballots_handed_out_count + other_explanation_count + no_explanation_count
+        if self.more_ballots_count != 0
+            && (self.too_many_ballots_handed_out_count
+                + self.other_explanation_count
+                + self.no_explanation_count
+                != self.more_ballots_count)
+        {
```
jschuurk-kr, line 593: If I leave `more_ballots_count` empty and fill one or more of the other fields, I don't get an error message, because `more_ballots_count` is `0`.

backend/src/polling_station/structs.rs
jschuurk-kr: Good points about the proces-verbaal and the CSB. An additional perspective is the current user experience:

1. You fill in "more ballots counted", while you shouldn't have.
2. You get an IncorrectTotal error.
3. You fix the error.
4. You get a NoDifferenceExpected warning.

That's weird, to have the user correct an error to then give them a warning about the actual problem.

Of course, step 3. in my description is very moot. Either you copied the proces-verbaal correctly, or you didn't. That might take several steps and the errors and warnings are there to help you get there. There is no "fixing the error" until the errors/warnings go away. At most at the end there's an issue with the proces-verbaal and we show a warning/error for those.

So it's more of a question which errors/warnings take precedence over others. I think NoDifferenceExpected is the most important one on this page. Which means it should be an error, because we don't show warnings if there are errors.

backend/src/polling_station/structs.rs
Lionqueen94: @jorisleker should we add another check for this? If any field filled in except more or fewer ballots counted?

jorisleker: maybe @chrismostert can help us here? If we do not do any checks at the level of the CSB, then it does not really make sense to add blocking error here.

jschuurk-kr: I understand the line of reasoning, but it does feel weird that the decision to show an error is not "are these numbers consistent with each other?" but "are the numbers consistent with each other, but only to the degree to which the CSB cares?"

backend/src/polling_station/structs.rs
jschuurk-kr, line 1: The current implemenation of F.26 and F.27 (and their description in the use cases) doesn't allow to combine reasons for both more and fewer ballots counted. If you fill in I, its value is compared with the sum of fields M, N, and O. If you fill in J, its value is compared with the sum of fields K, L, N, O. So field M (too many ballots handed out) is ignored when there are fewer ballots than voters. Fields K (ballots not handed in) and fields L (too few ballots handed out) are ignored when there are more ballots than voters.

I can imagine for example a scenario where there are 3 fewer ballots than voters (field I), explained by 4 too few ballots handed out (field L) and 1 too many ballot handed out (M). That does require some diligent administration from the polling station, but I do think it's the way the proces-verbaal should be filled out. That you're not supposed to do a calculuation and fill field L with a 3 and leave field M empty.

backend/src/polling_station/structs.rs
```diff
@@ -316,33 +423,205 @@ impl Validate for VotesCounts {
...
+        // F.13 validate that total_admitted_voters_recount == poll_card_recount + proxy_certificate_recount + voter_card_recount
+        if self.poll_card_recount + self.proxy_certificate_recount + self.voter_card_recount
+            != self.total_admitted_voters_recount
+        {
+            validation_results.errors.push(ValidationResult {
+                fields: vec![
+                    format!("{field_name}.total_admitted_voters_recount"),
+                    format!("{field_name}.poll_card_recount"),
+                    format!("{field_name}.proxy_certificate_recount"),
+                    format!("{field_name}.voter_card_recount"),
```
jschuurk-kr, line 509: These say `votes_recounts` instead of `voters_recounts` in the response, causing the error not to show up.

backend/src/polling_station/structs.rs
chrismostert: @jorisleker 
At the moment the CSB checks if  `no_explanation_count > 0` and the polling station did not do a recount. 
However, due to OSV not enforcing that the given explanations add up to the difference between admitted voters and counted votes, any difference between these two which does *not* have an explanation is also seen as `no_explanation`.

If F.26 is made blocking, then we would no longer have to check for these 'implicit' unexplained differences, but I don't know for sure if there are situations where you still need to be able to insert totals which do not add up.

jorisleker: @jschuurk-kr totally agreed. Just pointing out that the paper form does not offer very solid guidance as to which number need to add up to which total / difference. Which makes this one a bit harder to enforce then the other totals. You don't want users to get stuck here because their GSB doesn't know how to fill in the paper form.

backend/src/polling_station/structs.rs
Lionqueen94: Fixed in e6bda22191ad310081f5d8140ea216514599e40c. Thanks again for the extensive testing

backend/src/polling_station/structs.rs
jschuurk-kr: @jorisleker Agree with you there. If we make Abacus more strict than OSV, but don't provide enough guidance, we're hindering our users, not helping them.

backend/src/polling_station/structs.rs
jorisleker: yes, ran into this whilst redesigning the paper forms as well. There a basically three things a polling stations member can observe during voting that could lead to differences later on. 

someone:
1. does not put his/her ballot into the ballot box
2, did not get a ballot, but was counted as a voter
3. someone got multiple ballots by accident

1 and 2 lead to a higher number of voters 
3 leads to a higher number of votes. 

1+2-3 should be equal to votes-voters. If not, you've got an unexplained difference. 

Right?

frontend/app/component/form/recounted/RecountedForm.test.tsx
```diff
@@ -1,120 +1,207 @@
...
-    expect(result).toHaveTextContent(/^Error 500_ERROR 500 error from mock$/);
+    const feedbackServerError = await screen.findByTestId("feedback-server-error");
+    expect(feedbackServerError).toHaveTextContent(/^Error500 error from mock$/);
+
+    expect(screen.queryByTestId("result")).not.toBeNull();
+    expect(screen.queryByTestId("result")).toHaveTextContent(/^500 error from mock$/);
+  });
+
+  describe("RecountedForm errors", () => {
+    test("No radio selected", async () => {
```
jschuurk-kr, line 185: The other similar tests mention the code of the error in the name, so would be good to have the F.02 in there.

frontend/app/component/form/differences/DifferencesForm.test.tsx
```diff
@@ -240,13 +217,127 @@ describe("Test DifferencesForm", () => {
...
+      const submitButton = screen.getByRole("button", { name: "Volgende" });
+      await user.click(submitButton);
+
+      const feedbackError = await screen.findByTestId("feedback-error");
+      expect(feedbackError).toHaveTextContent(/^ConflictingDifferences$/);
+      expect(screen.queryByTestId("feedback-warning")).toBeNull();
+      expect(screen.queryByTestId("server-feedback-error")).toBeNull();
+    });
+
+    test("F.26 Incorrect total", async () => {
```
jschuurk-kr, line 335: Tests should use `overrideOnce()` and not rely on the mock implementation.

frontend/app/component/form/differences/DifferencesForm.test.tsx
```diff
@@ -258,13 +349,36 @@ describe("Test DifferencesForm", () => {
       await user.click(submitButton);
 
       const feedbackError = await screen.findByTestId("feedback-error");
-      expect(feedbackError).toHaveTextContent(/^IncorrectTotal,IncorrectTotal$/);
+      expect(feedbackError).toHaveTextContent(/^IncorrectTotal$/);
+      expect(screen.queryByTestId("feedback-warning")).toBeNull();
+      expect(screen.queryByTestId("server-feedback-error")).toBeNull();
+    });
+
+    test("F.27 Incorrect total", async () => {
```
jschuurk-kr, line 357: Tests should use `overrideOnce()` and not rely on the mock implementation.

frontend/app/component/form/voters_and_votes/VotersAndVotesForm.test.tsx
jschuurk-kr, line 1: Am missing a test for error F.13.

jschuurk-kr, line 1: Am missing a test for warning W.30.

frontend/app/component/form/differences/DifferencesForm.test.tsx
jschuurk-kr, line 1: Am missing tests for warning W.31 and W.32.

Probably best to wait until we resolve the [discussion]((https://github.com/kiesraad/abacus/pull/166#discussion_r1695444347)) about whether these should be errors.

backend/src/polling_station/structs.rs
chrismostert: Would it make sense to calculate the "more ballots counted" field in the software based on the difference between the cast votes and admitted voters instead of having the user fill this in? That would solve some of the ways in which the fields can be incorrectly filled.

frontend/app/component/form/recounted/RecountedForm.test.tsx
Lionqueen94: Changed in b90d2a8d33a757b4a0ada6d31a3ade9b59aab4f8

frontend/app/component/form/differences/DifferencesForm.test.tsx
Lionqueen94: Changed in b90d2a8d33a757b4a0ada6d31a3ade9b59aab4f8

frontend/app/component/form/differences/DifferencesForm.test.tsx
Lionqueen94: Changed in b90d2a8d33a757b4a0ada6d31a3ade9b59aab4f8

frontend/app/component/form/voters_and_votes/VotersAndVotesForm.test.tsx
Lionqueen94: Changed in b90d2a8d33a757b4a0ada6d31a3ade9b59aab4f8, however need help to make this test work

Lionqueen94: Changed in b90d2a8d33a757b4a0ada6d31a3ade9b59aab4f8, however need help to make this test work

frontend/app/component/form/voters_and_votes/VotersAndVotesForm.test.tsx
```diff
@@ -336,6 +334,45 @@ describe("Test VotersAndVotesForm", () => {
...
+        },
+      });
+
+      const user = userEvent.setup();
+
+      render(Component);
+
+      // TODO: These fields don't render because recounted is by default undefined
+      //  this value is coming from useState, any idea how to either refactor the code
+      //  or mock useState to be able to test this error?
```
Lionqueen94, line 361: @lkleuver @praseodym @cikzh can any of you assist me here and in the test for W.30 below? See the TODO for the issue.

backend/src/polling_station/structs.rs
praseodym: During CSO, the model doesn't require filling out any differences if the number of voters matches the number of ballots. Only when there are differences and the number of voters is recounted (by recounting the poll cards, voter cards and proxy certificates), the report of the polling station is checked to explain these differences.

For reference, the full section looks like this:

differences section

frontend/app/component/form/voters_and_votes/VotersAndVotesForm.test.tsx
praseodym: I've solved this by making the `PollingStationFormController` more testable by allowing default values to be set through props. See the latest commit!

backend/src/polling_station/structs.rs
jschuurk-kr: @chrismostert I don't think we're supposed to calculate any of the input fields, because the fields need to be copied from the proces-verbaal into the application. If we'd want to consider this, we should check with our legal team.

backend/src/polling_station/structs.rs
jschuurk-kr: > yes, ran into this whilst redesigning the paper forms as well. There a basically three things a polling stations member can observe during voting that could lead to differences later on.
> 
> someone:
> 
>     1. does not put his/her ballot into the ballot box
> 
>     2. did not get a ballot, but was counted as a voter
> 
>     3. someone got multiple ballots by accident
> 
> 
> 1 and 2 lead to a higher number of voters 3 leads to a higher number of votes.
> 
> 1+2-3 should be equal to votes-voters. If not, you've got an unexplained difference.
> 
> Right?

There are also the "andere verklaring" en "geen verklaring" fields which can go either way, depending if "more ballots" or "fewer ballots" has been filled in. Arguably, you could enter negative numbers for "andere verklaring" en "geen verklaring" where needed, but I wouldn't expect everyone to do that consistently.

jschuurk-kr:
It's possible to fill in the differences screen in a way that field `I` gets a warning icon, but no warning is shown, because there's an `IncorrectTotal` error for field `J`:
image

praseodym:
> It's possible to fill in the differences screen in a way that field `I` gets a warning icon, but no warning is shown, because there's an `IncorrectTotal` error for field `J`: [..]
> 
> Error/warning icons will be revisited in Epic: refine validation [#99](https://github.com/kiesraad/abacus/issues/99), so we could decide to fix in that epic.

I think the issue here is that the rendering of errors and warnings for the fields is done in the field component. This means that if the field has an error, no warning is rendered. But if the form page has an error, and a field only has a warning, the warning is still rendered for that field.

The fix would be to not pass any warnings down to the field components when there are errors present, or filter out warnings in the backend when errors are present.

praseodym:
> > It's possible to fill in the differences screen in a way that field `I` gets a warning icon, but no warning is shown, because there's an `IncorrectTotal` error for field `J`: [..]
> > Error/warning icons will be revisited in Epic: refine validation [#99](https://github.com/kiesraad/abacus/issues/99), so we could decide to fix in that epic.
> 
> I think the issue here is that the rendering of errors and warnings for the fields is done in the field component. This means that if the field has an error, no warning is rendered. But if the form page has an error, and a field only has a warning, the warning is still rendered for that field.
> 
> The fix would be to not pass any warnings down to the field components when there are errors present, or filter out warnings in the backend when errors are present.

Should be fixed by e025c3f by only processing warnings in the frontend when no errors are present.

backend/src/polling_station/structs.rs
jschuurk-kr: Taking the unclarity of the current form into account and the expectation that the form will be changed before the first usage of Abacus in production, I think it comes down to the question:

If there are no differences between votes and voters, do we think this is a valid data entry:
image

According to the current implementation it is.

backend/src/polling_station/structs.rs
jschuurk-kr: It also seems odd to me that when there is no difference between voters and votes and you fill both "more" and "fewer", you get a warning, i.e. NoDifferenceExpected. While if there is a difference and you fill both "more" and "fewer", you get an error, i.e. ConflictingDifferences.

jschuurk-kr:
Summary of the three open threads about errors and warnings:
- fields K-O should be used to check more/fewer ballots ([GH thread](https://github.com/kiesraad/abacus/pull/166#discussion_r1696780643)): needs to be fixed, otherwise we can't process valid proces-verbaals
- changing NoDifferenceExpected from a warning into an error ([GH thread](https://github.com/kiesraad/abacus/pull/166#discussion_r1695444347): better to fix, because clearer feedback to the user; at least current implementation does warn te user they filled in something curious
- when there's no difference between votes and voters, check that fields K-O are empty ([GH thread](https://github.com/kiesraad/abacus/pull/166#discussion_r1696666135)): should be fixed in my opinion, otherwise we accept nonsense in those fields when there's no voters vs votes difference

frontend/app/component/form/recounted/RecountedForm.tsx
jschuurk-kr, line 1: The values of `voters_recounts` should be removed from the request, when the user selects the "no" radio button and clicks "Volgende". Otherwise the user might get errors or warnings on fields that are no longer shown in the UI.

Steps to reproduce:
1. Pick "yes" on the recounted page.
2. Fill the recounted voters fields with values that result in an error or warning.
3. Return to the recounted page and pick "no".
4. The error or warning from step 2 is shown, but you can't edit the fields to fix the error/warning.

frontend/app/component/form/voters_and_votes/VotersAndVotesForm.tsx
praseodym: I've removed the label suffix in aa75f9584074e039ee4780e79a71656335436ada.

frontend/app/component/form/recounted/RecountedForm.tsx
praseodym: Should be fixed in d33f6ebf33fc5d912931d4dde363bac05c544eb0.

jschuurk-kr: The issue is fixed, but the values are still cached by the frontend. (My bad for not mentioning that earlier, it did cross my mind.) If you switch back to "yes", the old values are in the request of the recounted page, but the fields are empty on the voters and votes page.

So if I follow the steps above to reproduce, then at step 5. the error/warning is no longer shown. Then continuing:
6. Return to the recounted page.
7. Pick "yes" again.
8. Return to the voters and votes page.
9. The error/warning is shown at the top of the page and there are error/warning icons for the empty fields A.2 to D.2.

frontend/app/component/form/voters_and_votes/VotersAndVotesForm.test.tsx
praseodym: Fixed in f9e20ba, see https://github.com/kiesraad/abacus/pull/166#discussion_r1697521289.

frontend/app/component/form/recounted/RecountedForm.tsx
Lionqueen94: The voters_recounts fields are now cleared when you set recounted to no, which solves the issue of cached warnings/errors.

jschuurk-kr: Went through the steps to reproduce. Neither of the issues occurred.

frontend/app/component/form/differences/DifferencesForm.tsx
jschuurk-kr, line 1: If you fill in the VotersVotes page with a difference between voters and votes, you click "Volgende" and go to the Differences pages, there's an IncorrectDifference error on the page, even though you haven't had the chance yet to enter something in either the more or fewer ballots counted fields.

backend/src/polling_station/structs.rs
```diff
@@ -316,33 +443,211 @@ impl Validate for VotesCounts {
...
+        );
+
+        // W.302 validate that more_ballots_count == too_many_ballots_handed_out_count + other_explanation_count + no_explanation_count - unreturned_ballots_count - too_few_ballots_handed_out_count
+        if self.more_ballots_count != 0
+            && (self.too_many_ballots_handed_out_count
+                + self.other_explanation_count
+                + self.no_explanation_count
+                - self.unreturned_ballots_count
+                - self.too_few_ballots_handed_out_count
+                != self.more_ballots_count)
```
jschuurk-kr, line 614: These are `u32`s, so if `unreturned_ballots_count` and `too_few_ballots_handed_out_count` are larger than the other fields, it results in:
```
thread 'tokio-runtime-worker' panicked at src/polling_station/structs.rs:609:17:
attempt to subtract with overflow
```

I expect the same to be the case for W.303.

backend/src/polling_station/structs.rs
jschuurk-kr: Fixed. All five fields with explanations are used in the IncorrectTotal warnings.

backend/src/polling_station/structs.rs
jschuurk-kr: Decision was made to keep one category of errors: difference between D/D.2 and H should be equal to the more/fewer ballots counted field (F.301 to F.304). All other validations for the differences page result in warnings (W.301 to W.306), not errors.

backend/src/polling_station/structs.rs
praseodym: When adding up those counts you should be using `u64` like in some other places.

backend/src/polling_station/structs.rs
jschuurk-kr: Fixed. Results in a NoDifferenceExpected warning.

backend/src/polling_station/structs.rs
jschuurk-kr: I think that's a different (and equally) valid issue:
- because we're adding up `u32`s, we might get something bigger than a `u32`
- because we're subtracting `u32`s, we might get a negative number

praseodym: Oh right, for the subtraction `i64` is more suitable.

frontend/app/component/form/differences/DifferencesForm.tsx
jschuurk-kr: in scope of #133 Frontend: update data entry navigation

backend/src/polling_station/structs.rs
Lionqueen94: Fixed and tested in 6bc986da52492c8a4ad8a5782d1ee1309855ce5e

praseodym (approved):

jschuurk-kr (approved):

PR #168, opened 2024-07-22
Respond to polling station list call with 404 when the election does not exist

### LINKED ISSUE #167: Return 404 on when querying for non-existing elections
Right now an empty array gets returned, although a 404 response should be given and communicate an invalid `election_id` is given.

cikzh:
Closes #167

### DIFF
--- a/backend/tests/polling_station_integration_test.rs
+++ b/backend/tests/polling_station_integration_test.rs
@@ -104,6 +104,17 @@ async fn test_polling_station_data_entry_valid(pool: SqlitePool) {
     assert_eq!(response.status(), StatusCode::OK);
 }
 
+#[sqlx::test(fixtures(path = "../fixtures", scripts("elections")))]
+async fn test_polling_station_list_invalid_election(pool: SqlitePool) {
+    let addr = serve_api(pool).await;
+    // election ID 1234 does not exist
+    let url = format!("http://{addr}/api/polling_stations/1234");
+    let response = reqwest::Client::new().get(&url).send().await.unwrap();
+
+    let status = response.status();
+    assert_eq!(status, StatusCode::NOT_FOUND);
+}
+
 #[sqlx::test]
 async fn test_polling_station_data_entry_invalid(pool: SqlitePool) {
     let addr = serve_api(pool).await;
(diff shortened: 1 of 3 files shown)


cikzh:
I still need to change the OpenAPI spec, will put it out of Draft when this is done

cikzh:
@jschuurk-kr FYI: this will also be rebased into #146

praseodym (approved):

backend/tests/polling_station_integration_test.rs
jschuurk-kr, line 1: This files combines tests for two endpoints: `api/pollings_stations/<id>` and `api/polling_stations/<id>/data_entries/<id>`. I think we should split these across two files: `polling_station_integration_test.rs` and `data_entries_integration_test.rs`.

rnijveld (approved):

backend/tests/polling_station_integration_test.rs
cikzh: The code that is tested by this file lives in the same Rust module, so I think it makes sense to keep this in the same file.

jschuurk-kr: I'd split integration tests based on the endpoints, not on the files. We're testing from the outside, so it doesn't matter for the tests which files with code are involved.

Relatedly, it looks like the tests are currently sorted on success vs error first, and then on endpoint. That does not map with how I think about integration tests and their coverage. Not that I can't learn a new habit, of course. :-) So glad to have this discussion here, so we an make a decision and then apply it consistently.

cikzh: Shall we create an issue for that? Then we can keep this PR about the 404 response

jschuurk-kr: I was hoping we'd be able to come to a decision fast enough for that not to be necessary. Let's create a different issue for this if we don't have a decision after standup on Tuesday?

praseodym: As discussed in the standup: we split the integration tests into a file per endpoint, not necessarily following the module structure. @rnijveld will create a PR after this one is merged to reorganise the current tests.

Issue #169, opened 2024-07-22
Epic: DOM-to-database testing


praseodym:
Test DOM-to-database, i.e. browser->frontend->backend->database.

- [x] #172
- [x] #177 
- [x] #189 
- [x] #154 
- [x] #218
- [x] #217
- [x] #216
- [x] #318 
-  ~~Decide on locator strategy and update code and tests if needed~~

For frontend+MSW tests, there's also #62 (Playwright tests for clicking through the front-end with MSW).

praseodym:
I think being able to serve the frontend through the backend (#63) will also be helpful here, so we can do the Playwright test using the a fully built frontend instead of a dev server. This doesn't have to depend on the full scope of #63, though.

jschuurk-kr:
References for the pipeline task:
- [reusing workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows): dependencies as reusable workflows, called as jobs in another workflow.
- [uploading(https://github.com/actions/upload-artifact) and [downloading](https://github.com/actions/download-artifact) [artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts)
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow

jschuurk-kr:
Using artifacts from the backend workflow for the dom-to-db tests:
- artifacts: `backend/target/debug/api` and `backend/target/debug/db.sqlite`
- dom-to-db tests need to do their own test data setup, either the files in `backend/fixtures` or a different set of test data specific to the dom-to-db tests

It might be worth it to have data files that are read by the tests and are used to generate the `.sql` files. Once we have the functionality to import that data as EML files, we can migrate to those where sensible.

praseodym:
> References for the pipeline task: [..]

I think it might be easiest to integrate building and testing the backend and frontend and executing the DOM-to-database tests as [multiple jobs in a single workflow](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow). This makes it easy to specify job dependencies while avoiding redoing work.

praseodym:
> Using artifacts from the backend workflow for the dom-to-db tests:
> 
> * artifacts: `backend/target/debug/api` and `backend/target/debug/db.sqlite`

I think we'll want to use a release build and not a debug build. Also, `db.sqlite` is created when starting the API server, so we don't need to ship it as an artefact.

[...]

Issue #63, opened 2024-05-29
Epic: frontend serving via backend


praseodym:
Serve the frontend via the backend so a pre-built frontend can be used together with a running backend.

PR #175, opened 2024-07-24
Add API server CLI option to serve frontend

### LINKED ISSUE #174: Implement frontend server in backend

praseodym:
Closes #174


praseodym:
Just discovered that reloading the frontend won't work because `index.html` is only loaded for `/`. Going to fix that now.

praseodym:
> I'm wondering, given that we want to end up in a scenario where the binary self-contains the entire application (is that still true?), would it not make more sense to (1) include the files in the binary at compile time (and serve them from memory) whenever a feature flag is given for that, and (2) use some proxying setup where the backend just proxies to the vite development server in other cases. We can then use the application in three ways: using the address that the vite development server serves on, which is easiest for frontend development, using the backend address for cases where you want/need the entire application and then using the binary contained files for both CI and in production scenarios so that testing tests on a scenario actually very similar to production setups.

(1) is still the case and is a future task, also part of #63.

(2) doesn't make a lot of sense to me because I think it'll introduce a lot of complexity. The Vite development server uses WebSockets for HMR which we'd also have to proxy. In our current setup the Vite development server already proxies the backend and that works fine with HMR and our backend setup ([npm run dev:server](https://github.com/kiesraad/abacus/tree/main/frontend#running)).

For Playwright tests I'd prefer to not use the Vite development server, but test with the production build. For this we need a simple way to combine the frontend and backend in one web server. I think this PR implements the easiest way to achieve that.

jschuurk-kr (dismissed):
Works on my machine!

If you don't have a `frontend/dist` folder, you get a 404 response at `http://0.0.0.0:8080`. That's definitely fine for now.

praseodym:
> > (1) is still the case and is a future task, also part of #63.
> 
> Maybe we should see if tweede golf's [memory-serve](https://github.com/tweedegolf/memory-serve) crate might actually solve both this and that goal all at once. Using that crate we at least use the same code both during testing and during our eventual production builds.

`memory-serve` looks very suitable for that indeed. Since this PR is enough for what we need at this point (for #169), I propose we don't spend a lot more time on this now and leave integrating `memory-serve` for when we prioritise the work on #63.

> > (2) doesn't make a lot of sense to me because I think it'll introduce a lot of complexity. The Vite development server uses WebSockets for HMR which we'd also have to proxy. In our current setup the Vite development server already proxies the backend and that works fine with HMR and our backend setup ([npm run dev:server](https://github.com/kiesraad/abacus/tree/main/frontend#running)).
> 
> Having to proxy the websockets would definitely be a little harder to do than just using something like tower-http, but I do believe it's doable without enormous effort or complexity. The main reasoning for mentioning it now was that it would increase the similarity between development and production, which I think is always a good thing. Though I would also say that the risk of problems arising from using Vite first and proxying the API instead is pretty low.

The frontend during development is always going to be different from production due to how Vite and React behave during development. How we proxy the backend (or frontend) only adds one minor other difference.

Let's keep using the Vite proxy during development, I feel like that's the least complicated and also low (or zero, it already works) effort.

> > For Playwright tests I'd prefer to not use the Vite development server, but test with the production build. For this we need a simple way to combine the frontend and backend in one web server. I think this PR implements the easiest way to achieve that.
> 
> I strongly agree with not using Vite for playwright tests, but does a production build not mean one with build-in assets already, instead of serving them from file system?

I meant the production build of the frontend, so the result of [`vite build`](https://vitejs.dev/guide/build). For the backend I'd also prefer using the Cargo release profile.

Not compiling the frontend files into the API server binary does mean that we can compile the frontend in parallel with the backend, speeding up our CI times a bit.

backend/src/bin/api.rs
```diff
@@ -1,22 +1,42 @@
...
 /// API server on port 8080.
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
+    let args = Args::parse();
     let pool = create_sqlite_pool().await?;
     let app = router(pool)?;
 
+    let app = if let Some(fd) = args.frontend_dist {
+        app.fallback_service(
+            ServeDir::new(fd.clone()).fallback(ServeFile::new(fd.join("index.html"))),
```
jschuurk-kr, line 34: Now I'm curious: what's the difference between `not_found_service()` and `fallback()`?

praseodym: `not_found_service()` responds with 404 Not Found, `fallback()` with 200 OK.

jschuurk-kr: In which case? If I have an empty `frontend/dist` folder, I get a `404` for both `not_found_service()` and for `fallback()`. If I do have a `frontend/dist/index.html`, I get a `200` for both.

praseodym: With the `not_found_service()`, navigating to `/overview` would serve `index.html` (expected) but with a 404 Not Found status code (not expected). With `fallback()`, this responds with 200 OK instead (expected).

backend/Cargo.toml
```diff
@@ -14,12 +14,14 @@ axum = { version = "0.7", features = ["macros"] }
...
 tokio = { version = "1", features = ["full"] }
 tower = "0.4"
+tower-http = { version = "0.5.0", features = ["fs"] }
 utoipa = { version = "4.2", features = ["axum_extras"] }
 utoipa-swagger-ui = { version = "7.1.0", features = ["axum"], optional = true }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 sqlx = { version = "0.7", features = ["runtime-tokio", "sqlite", "chrono"] }
 chrono = { version = "0.4", features = ["alloc", "std", "serde"] }
+clap = { version = "4.0", features = ["derive"] }
```
cikzh, line 24: Clap pulls in quite some sub dependencies. This could be a quick win to eliminate some. If the interface stays as simple as it is now, we could write our own CLI in the future and lose Clap altogether. 
```suggestion
clap = { version = "4.0", default-features = false, features = ["derive", "std"] }
```

cikzh (changes requested):
Let's indeed not keep #169 waiting on this, though I think it's wise to not wait too long with exploring how we ship (and develop, for that matter) the application eventually.

cikzh (approved):

jschuurk-kr (approved):

PR #176, opened 2024-07-24
single workflow with frontend, backend, and playwright jobs

### LINKED ISSUE #172: Refactor backend and frontend workflow into a single workflow with multiple jobs
Preparatory step to later be able to run Playwright tests against frontend + backend in the pipeline.

jschuurk-kr:
closes #172


.github/workflows/build-lint-test.yml
```diff
@@ -45,3 +45,46 @@ jobs:
         run: cargo clippy --all-targets --all-features -- -D warnings
       - name: Run tests
         run: cargo test --verbose
+  node:
+    name: Node
```
praseodym, line 49: Maybe:
```suggestion
    name: Frontend
```

praseodym: And Backend for the backend job, but that's outside of the diff so I can't suggest it.

.github/workflows/build-lint-test.yml
```diff
@@ -45,3 +45,46 @@ jobs:
...
+        run: npx prettier --check .
+      - name: Lint
+        run: npm run lint
+      - name: Test
+        run: npm test
+      - name: Build
+        run: npm run build
+  playwright:
+    name: Playwright
+    needs: node
```
praseodym, line 72: At this point there is no actual dependency yet, right?

jschuurk-kr: True. My thinking is that the feedback is clearer when only the Frontend job fails, instead of both the Frontend and the Playwright job[^1]. That does come with the cost of the pipeline taking longer. To be perfectly honest, the reason I put the `needs` in there is to have the discussion if that makes sense.

[^1]: Assumption: if at least one Frontend test fails, at least one Playwright test will fail as well.[^2]
[^2]: GitHub **does** do footnotes! :raised_hands:

.github/workflows/build-lint-test.yml
jschuurk-kr: Changed both the key and the `name` of both jobs.

lkleuver (approved):

.github/workflows/build-lint-test.yml
praseodym: I'd prefer the CI runs to be as fast as possible. But we'll also be using the backend as a dependency (https://github.com/kiesraad/abacus/issues/169#issuecomment-2247823109) so there's going to be a dependency on that too, so this frontend dependency doesn't really matter.

praseodym (approved):

praseodym:
I've changed the required status checks for the repository to the new job names.

PR #178, opened 2024-07-24
splits e2e app and e2e lib-ui tests by using separate config files

### LINKED ISSUE #177: Split out Playwright tests: app vs lib
Preparatory step, so it's easy to add DOM-to-database tests later in this epic.

jschuurk-kr:
closes #177


frontend/package.json
```diff
@@ -15,8 +15,8 @@
     "prettier": "prettier --ignore-unknown --write .",
     "preview": "vite preview",
     "test": "vitest",
-    "e2e": "playwright test",
-    "e2eui": "playwright test --ui",
+    "e2e:lib-ui": "playwright test -c playwright.lib-ui.config.ts",
+    "e2e:app": "playwright test -c playwright.app.config.ts",
```
praseodym, line 19: Don't forget to change this in the `README.md`.

jschuurk-kr: thanks! done

praseodym (approved):
Because the config files are TypeScript code you could extract the common config from the two files instead of duplicating it, but this is good enough.

jschuurk-kr:
> Because the config files are TypeScript code you could extract the common config from the two files instead of duplicating it, but this is good enough.

We'll have three files soon enough. :-D

My reasoning for not extracting the common parts is because the files are short enough and I anticipate/hope they'll diverge more in the near future. Happy to revisit this once I add the third file for the DOM-to-database tests.

Lionqueen94 (approved):

Issue #179, opened 2024-07-24
Frontend: application behavior when a non-existing election ID is used


cikzh:
As discussed here: https://github.com/kiesraad/abacus/pull/146#discussion_r1686014231 a user might end up using a non-existing election ID, for example by providing a wrong ID via the URL. This can happen in multiple places in the application, but `/{id}/input` might be an obvious place to test this.

- What should the behavior be when the `election` doesn't exist?
- Determine how to implement this
- Determine how to test this

praseodym:
404 page?

praseodym:
@jorisleker might have a good answer to this :)

jorisleker:
404 page with a text like this?

> Er ging iet mis. De link die je hebt gebruikt is niet (meer) geldig. Ga naar het [overzicht met verkiezingen](#link).

PR #182, opened 2024-07-25
Improve polling station form tests

### LINKED ISSUE #163: Improve polling station form tests
I'd prefer an `overrideOnce()` at the start of this test for the api response instead of relying on what's in `frontend/lib/api-mocks/PollingStationMockData.ts`.

In this case the mock plainly returns the data, so it's unlikely to lead to isses. For the `data_entries` ednpoint it does not. So I'd rather have the rule "always do an `overrideOnce()`", which leaves no doubt what data is used by test, than the rule "it depends on how complicated someone judged the mock implementation to be".

_Originally posted by @jschuurk-kr in https://github.com/kiesraad/abacus/pull/146#discussion_r1683044243_

---

The 404 response of a non-existing election id results in the same behavior as when the API returns an empty list, because of how the PollingStationProvider works. So I would like to have a test that verifies/documents this behavior.
https://github.com/kiesraad/abacus/pull/146/files#r1691007957

cikzh:
Closes #163 

Using `overrideOnce` at the start of every test to make clear what data is used for the test. Also making it more explicit when testing edge cases.

This PR leaves out the 404 test as discussed in the issue, as it is dependent on the decision made in #179

### DIFF
--- a/frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.test.tsx
+++ b/frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.test.tsx
@@ -95,7 +95,7 @@ describe("Test PollingStationChoiceForm", () => {
   });
 
   test("Polling station list no stations", async () => {
-    overrideOnce("get", "/api/polling_stations/1", 200, {
+    overrideOnce("get", "/api/elections/1/polling_stations", 200, {
       polling_stations: [],
     });
     const user = userEvent.setup();
--- a/frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.test.tsx
+++ b/frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.test.tsx
@@ -113,4 +113,24 @@ describe("Test PollingStationChoiceForm", () => {
     // Check if the error message is visible
     expect(screen.getByText("Geen stembureaus gevonden")).toBeVisible();
   });
+
+  test("Polling station request returns 404", async () => {
+    overrideOnce("get", "/api/elections/1/polling_stations", 404, {
+      error: "Resource not found",
+    });
+    const user = userEvent.setup();
+
+    render(
+      <PollingStationProvider electionId={1}>
+        <PollingStationChoiceForm />
+      </PollingStationProvider>,
+    );
+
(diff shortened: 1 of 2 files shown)


praseodym (approved):

jschuurk-kr (changes requested):
> This PR leaves out the 404 test as discussed in the issue, as it is dependent on the decision made in https://github.com/kiesraad/abacus/issues/179

We can cover the current behavior with a test. To me, that too, is part of the zero bug policy.

frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.test.tsx
```diff
@@ -67,6 +71,7 @@ describe("Test PollingStationChoiceForm", () => {
   });
 
   test("Polling station list", async () => {
+    overrideOnce("get", "/api/polling_stations/1", 200, pollingStationMock);
```
Lionqueen94, line 74: Testing the list endpoint should not have an overrideOnce on the detail endpoint. It should be on "/api/elections/{election_id}/polling_stations/". See also the other instances in this file

Lionqueen94 (changes requested):

frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.test.tsx
cikzh: I think I'll fix this by rebasing main and then change it. Thanks!

Lionqueen94, line 74:
```suggestion
    overrideOnce("get", "/api/elections/1/polling_stations", 200, pollingStationMock);
```

cikzh: Thanks, I've changed the overrides in other tests as well, since they're actually all calling the listing endpoint.

Lionqueen94 (dismissed):

praseodym (dismissed):

jschuurk-kr (approved):

Lionqueen94 (approved):

praseodym (approved):

Issue #136, opened 2024-07-08
Frontend: implement validation copy using unique codes


praseodym:
_depends on #135_

Implement [all error/warning copy](https://github.com/kiesraad/opstartfase-team-URN/blob/main/use-cases/GSB-validatieregels-plausibiliteitschecks.md) in frontend. Unique codes are returned by backend when #135 is done.

Tests should also cover the error/warning icons shown in the input fields relevant to the errors/warnings.

jschuurk-kr:
@jorisleker Can you please check the [error/warning copy](https://github.com/kiesraad/opstartfase-team-URN/blob/main/use-cases/GSB-validatieregels-plausibiliteitschecks.md)? With some of the errors for differences changed into warnings, the copy of those probably needs some changes to be less like a blocking error.

PR #183, opened 2024-07-25
Use polling station id instead of number

### LINKED ISSUE #181: Bug: Navigate should use pollingStationId instead of pollingStationNumber
### TODO: ###
- [x] Update list pollingstations route in backend as discussed
- [x] Fix navigate from polling station select page to input page
- [x] Fix implementation and add tests for F.01 on candidateVotesForm and VotersAndVotesForm
- [x] Update overrideOnce responses with changed DataEntryResponse

Lionqueen94:
Closes #181 

- [x] Update list pollingstations route in backend as discussed
- [x] Fix navigate from polling station select page to input page
- [x] Fix implementation and add tests for F.01 on candidateVotesForm and VotersAndVotesForm
- [x] Update overrideOnce responses with changed DataEntryResponse


backend/src/polling_station/structs.rs
```diff
@@ -127,7 +127,7 @@ impl Validate for Count {
         validation_results: &mut ValidationResults,
         field_name: String,
     ) {
-        if self > &1_000_000_000 {
+        if self > &999_999_999 {
```
praseodym, line 130:
```suggestion
        if self >= &999_999_999 {
```

praseodym (approved):
lgtm

frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.tsx
```diff
@@ -15,9 +15,11 @@ export function PollingStationChoiceForm() {
 
   const handleSubmit = () => {
     const parsedStationNumber = parseInt(pollingStationNumber, 10);
-    if (pollingStations.some((ps) => ps.number === parsedStationNumber)) {
-      navigate(`./${pollingStationNumber}/recounted`);
-    }
+    pollingStations.forEach((pollingStation) => {
```
cikzh, line 18: Why changing to a `forEach`? isn't `some` better conveying what is happening here?

cikzh (commented):
Looks good overall! Only one small question

frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.tsx
Lionqueen94: I couldn't use the ps in the navigate when I used some. If this is possible please teach me 🤓

praseodym: `some` doesn't return the polling station, so you can't get the id from it.

Maybe using [`find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) would be better than a `forEach`.

cikzh: Ah, I see! `find` could indeed work. Not tested, but maybe something like this:
```
const pollingStation = pollingStations.find((pollingStation) => pollingStation.number === parsedStationNumber);
if (pollingStation) {
   navigate(`./${pollingStation.id}/recounted`);
}
```

Lionqueen94: Changed in 35bb4e286e598bb6264d02f2c3380180a0b48e60

cikzh (dismissed):

praseodym (approved):

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
```diff
@@ -134,7 +134,7 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
                 key={`list${group.number}-candidate${index + 1}`}
                 field={`${index + 1}`}
                 name="candidatevotes[]"
-                id={`candidate_votes-${candidate.number - 1}.votes`}
+                id={`candidate_votes[${candidate.number - 1}].votes`}
```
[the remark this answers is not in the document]
lkleuver: Also does this still match with errors and warnings?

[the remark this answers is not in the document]
praseodym: > I think this breaks stuff, also [ and ] are not allowed in id fields as far as I can tell https://www.w3.org/TR/html4/types.html#:~:text=ID%20and%20NAME%20tokens%20must,tokens%20defined%20by%20other%20attributes.

That's for HTML 4 (from 1999), HTML 5 doesn't have that restriction: https://mathiasbynens.be/notes/html5-id-class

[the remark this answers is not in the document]
jschuurk-kr: > Also does this still match with errors and warnings?

We don't use this id in any of the errors or warnings. The only (realistic[^1]) error on this page is for the `total` to match the sum of the candidates. That error only refers to `data.political_group_votes[0].total`. And there are no warnings for this page.

[^1]: If you intercept and edit the request, you can get the `OutOfRange` error for a candidate's vote. We have a test for that, but it doesn't check for the existence of the error icon yet. That's part of #136.

jschuurk-kr (dismissed):

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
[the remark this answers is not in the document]
Lionqueen94: What Joep is saying is correct, the OutOfRange error will indeed be the only error on the field, and the `id` now matches the `field` the backend puts the error on. That's the reason I changed it. To make this consistent.

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
```diff
@@ -171,10 +171,9 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
 
 function candidateNumberFromElement(el: HTMLInputElement) {
   const id = el.id;
-  const bits = id.split("[");
-  const numberString = bits[bits.length - 1]?.split("]")[0];
-  if (numberString) {
-    return parseInt(numberString) + 1;
+  const regexMatchArray = id.match(/[(\d*)]/g);
+  if (regexMatchArray) {
+    return parseInt(regexMatchArray[0]) + 1;
```
jschuurk-kr, line 176: Two thoughts, but they apply to the previous implementation as well, so probably better left for another PR.
1. If we change the function's parameter from an HTMLInputElement to the id, we can write some tests for it.
2. If there is no match, we return `0`. And then what happens? What if there are multiple no matches?

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
```diff
@@ -171,10 +171,9 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
 
 function candidateNumberFromElement(el: HTMLInputElement) {
   const id = el.id;
-  const bits = id.split("-");
-  const numberString = bits[bits.length - 1];
-  if (numberString) {
-    return parseInt(numberString) + 1;
+  const regexMatchArray = id.match(/[(\d*)]/g);
```
praseodym, line 174: Maybe?
```suggestion
  const regexMatchArray = id.match(/[(\d+)]/g);
```

praseodym (dismissed):

jschuurk-kr (dismissed):

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
Lionqueen94: Good call on the testing, because I'm terrible at regex. Added tests and fixed a bug in c9bc125fe7102ff7b31d62d00ec7925003ea1b4e

Lionqueen94: And to answer your question 2. Candidate number 0 does not exist, so then there will be an `IncorrectCandidatesList` validation error from the backend. This should never happen of course, just like the OutOfRange validation error, but they are implemented.

jschuurk-kr: We're missing a frontend test for the `IncorrectCandidatesList`. I'll create a PR to add it.

jschuurk-kr (approved):

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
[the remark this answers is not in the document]
Lionqueen94: > Also does this still match with errors and warnings?

It does now @lkleuver I fixed the fieldnamefrompath function used in errorsandwarnings :+1:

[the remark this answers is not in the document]
lkleuver: my preference was to actually use the exact same field name, but was under the wrong impression it was not allowed. great change!

lkleuver (approved):

Issue #108, opened 2024-06-26
Frontend: submit data entry using correct polling station information


praseodym:
Make sure that the data entry page uses the correct polling station information (id, number, name) from the polling station that was selected and also submit to the data entry API endpoint using the that id.

Cases to consider:
- existing polling station number
- non-existing polling station number
- not a number, e.g. "asd", to be handled the same as non-existing polling station number, see [comment](https://github.com/kiesraad/abacus/pull/146#discussion_r1682879672) in earlier PR
- no polling station number (user immediately clicks "Beginnen" without inputting a polling station number)

Submitting through the keyboard should not happen on "Enter", only on "Shift + Enter".

cikzh:
@jorisleker As far as I can tell the design doesn't specify what happens to the "Beginnen" button when an invalid station number is entered. Should we disable the button until a valid and existing polling station is entered? Or should clicking the button just "do nothing"?

praseodym:
> Cases to consider:
> 
> * existing polling station number
> * non-existing polling station number
> * not a number, e.g. "asd", to be handled the same as non-existing polling station number, see [comment](https://github.com/kiesraad/abacus/pull/146#discussion_r1682879672) in earlier PR
> * no polling station number (user immediately clicks "Beginnen" without inputting a polling station number)
> 
> Submitting through the keyboard should not happen on "Enter", only on "Shift + Enter".

@jschuurk-kr you added these items to the issue description but they are not about the data entry form but the polling station select page, which was done in #107. Can you confirm that these points are resolved now?

jschuurk-kr:
>  > \<snipped\>

> @jschuurk-kr you added these items to the issue description but they are not about the data entry form which this issue is about, but the polling station select page that was done in #107. Can you confirm that these points are resolved now?

@praseodym It was not entirely clear to me if clicking "Beginnen" was part of #107 or not, hence that I added those cases here.

As to their status, the first three are ok. If you enter something in the input field, it's either a valid polling station number or not. If it's valid, clicking "Beginnen" gets you to the next page. If it's not valid, nothing happens.

The fourth one is not resolved. If you click "Beginnen" without inputting a polling station number, nothing happens. We should provide some feedback to the user.

cikzh:
I considered the fourth case to be the same as @jorisleker's comment (https://github.com/kiesraad/abacus/issues/108#issuecomment-2247897940) on case 2. 
If it needs fixing I create a separate issue/PR from this one.

[...]

PR #193, opened 2024-07-26
Not found page and routing logic

cikzh:
This PR adds a Not Found page and the logic to trigger the page.

It makes use of the `errorElement` logic in `react-router-dom`. When an unhandled error is thrown, the component defined at `errorElement` will be rendered instead. This way, we can easily define different kinds of error pages in different parts of the application.

Also adds an e2e test to check if the Not Found page is rendered when visiting a page with a non-existing election ID, it also checks if the browser URL remains the same. I'm not sure if it is possible to do this in vitest. If it is, please let me know how and I'll move it.


jschuurk-kr:
> Also adds an e2e test to check if the Not Found page is rendered when visiting a page with a non-existing election ID, it also checks if the browser URL remains the same. I'm not sure if it is possible to do this in vitest.

Should be possible with React's `MemoryRouter` and `router.state.location.pathname`: https://stackoverflow.com/a/73730116

cikzh:
Awesome, thanks! I'll look into it

cikzh:
Thanks @jschuurk-kr ! The test is now done in vitest

frontend/app/routes.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    expect(router.state.location.pathname).toEqual("/");
+    await router.navigate("/1/input");
+    expect(router.state.location.pathname).toEqual("/1/input");
+
+    // Navigate to a non-existing page
+    await router.navigate("/-1/input");
+
+    // NOTE: We're not using the wrapped render function here,
+    // since we want control over our own memory router.
+    render(
```
praseodym, line 35: Maybe clarify that this not the `render` function from `app/test/unit` by renaming the import?

```suggestion
import { render as rtlRender } from "@testing-library/react";
import { describe, expect, test, vi } from "vitest";

import { screen } from "app/test/unit";

import { ApiProvider } from "@kiesraad/api";

import { routes } from "./routes";

const setupTestRouter = () => {
  return createMemoryRouter(routes);
};

describe("routes", () => {
  test("Non existing election id results in not found page", async () => {
    // Since we test what happens after an error, we want vitest to ignore them
    vi.spyOn(console, "error").mockImplementation(() => {
      /* do nothing */
    });

    const router = setupTestRouter();

    // Router sanity checks
    expect(router.state.location.pathname).toEqual("/");
    await router.navigate("/1/input");
    expect(router.state.location.pathname).toEqual("/1/input");

    // Navigate to a non-existing page
    await router.navigate("/-1/input");

    // NOTE: We're not using the wrapped render function here,
    // since we want control over our own memory router.
    rtlRender(
```

praseodym:
I noticed that frontend with MSW doesn't show 404 pages because the mock API doesn't check the election id. I think it should be easy enough to return a 404 from the mock API for invalid election ids, could you add this?

cikzh:
Yes, I noticed too, I got around it by using `overrideOnce`, but changing the MSW makes indeed more sense. Will do.

On a side note: I'm fixing the integer parsing issues in the `Layout` elements. I wish there was a way to do this in the router, as this problem will be present in other `Layout` objects as wel, but as far as I understand from the docs, they removed the ability to do so in a recent version. Is there a way to check if a path segment is numeric in `react-router-dom` that we know?

cikzh:
Yes, I found that too and I'm implementing it similar to that now.

cikzh:
>     6. `http://0.0.0.0:8080/1/input/1/list/` shows the 404 page, meaning we have some dead code in `frontend/app/module/input/PollingStation/CandidatesVotesPage.tsx` trying to display "Geen lijstnummer gevonden" for this case.
> 
> 
> I think 1, 3, 6, need to be addressed. 4 is as expected right now because #108.
> 
> No 3. is caused by `parseInt()`: _"If parseInt encounters a character that is not a numeral in the specified radix, it ignores it and all succeeding characters and returns the integer value parsed up to that point."_

@jschuurk-kr I added test cases for 1 and 3 and fixed them. I think 6 showing the 404 page is correct, as it is an invalid route. The route should always contain at least a list number.

jschuurk-kr:
> I think 6 showing the 404 page is correct, as it is an invalid route. The route should always contain at least a list number.

Agree that showing the 404 page is correct. Was just wondering if that means that the code in `frontend/app/module/input/PollingStation/CandidatesVotesPage.tsx` that displays "Geen lijstnummer gevonden" can no longer be reached and if so, should be removed.

cikzh:
Ah yes, that one is indeed dead as it's handled by the router. I'll remove it :+1:

frontend/app/module/NotFoundProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export const NotFoundContext = createContext<iNotFoundContext>({
+  notFound: false,
+  setNotFound: (notFound: boolean) => {
+    notFound;
+  },
+});
+
+export const NotFoundProvider = (props: object) => {
```
jschuurk-kr, line 15: My IDE can't find any usages of the `NotFoundProvider`.

cikzh: Funny, I would expect the linter to catch this. Since we use the `errorElement` logic of the router, we don't need a provider or context at all. Thanks!

praseodym:
Using a non-existent polling station in the MSW version doesn't lead to the 404 page (scenario 4 in https://github.com/kiesraad/abacus/pull/193#issuecomment-2255553886), is this in scope?

frontend/app/module/NotFoundProvider.tsx
jschuurk-kr: Looks like we need [`eslint-plugin-import`](https://www.npmjs.com/package/eslint-plugin-import) for that, because `eslint` only does unused vars and unused function params.

cikzh:
It looks to be a bug in getting the polling station. I'll fix it in this PR

cikzh:
[4911e73](https://github.com/kiesraad/abacus/pull/193/commits/4911e730b92e61e2225a80759f33b3bde1d82480) removes loading messages as discussed here: https://github.com/kiesraad/abacus/issues/108#issuecomment-2257565164

praseodym (approved):

frontend/app/module/input/PollingStation/PollingStationLayout.tsx
```diff
@@ -1,82 +1,75 @@
...
-            <Button variant="secondary" size="sm" onClick={changeDialog} rightIcon={<IconCross />}>
-              Invoer afbreken
-            </Button>
-            <WorkStationNumber>16</WorkStationNumber>
-          </section>
-        </header>
-        <main>
+    <PollingStationFormController
+      election={election}
+      pollingStationId={parseInt(pollingStationId || "0")}
```
Lionqueen94, line 31:
```suggestion
      pollingStationId={pollingStation.id}
```

frontend/lib/util/hook/useNumericParam.ts
```diff
@@ -0,0 +1,18 @@
+import { useParams } from "react-router-dom";
+
+//
```
Lionqueen94, line 3: Missing comment

cikzh: No comment.

frontend/app/module/input/PollingStation/PollingStationLayout.tsx
cikzh: Yes, good one, we parse this unnecessarily

Lionqueen94 (dismissed):

praseodym (approved):

Lionqueen94 (dismissed):

frontend/lib/util/hook/useNumericParam.ts
```diff
@@ -0,0 +1,23 @@
...
+  const params = useParams();
+  const param = params[parameterKey];
+
+  if (!param) {
+    throw Error(`Parameter ${parameterKey} does not exist`);
+  }
+
+  // Matches only if the whole string is numeric
+  const parsedParam = parseIntStrict(param);
+  if (!parsedParam) {
```
rnijveld, line 18: Technically this doesn't work when the integer is `0` right now. Shouldn't be a problem for all the id params we have right now, given they start at `1`, but for other numeric values it might be an issue.

praseodym: Maybe this then?
```suggestion
  if (parsedParam !== undefined) {
```

cikzh: [8e29b14](https://github.com/kiesraad/abacus/pull/193/commits/8e29b148112e110059702cb33b1cd246bc8dfedf) addresses this

rnijveld (approved):

Lionqueen94 (approved):

PR #208, opened 2024-08-02
Strict parse int

cikzh:
Note: this PR will rebase onto the `not-found` branch.

It creates a `parseInt` function replacement that doesn't allow _any_ non-numeric characters to be present in the string.

Let me know if I should go ahead and replace all relevent `parseInt` calls.

### DIFF
--- a/frontend/lib/util/strings.ts
+++ b/frontend/lib/util/strings.ts
@@ -8,6 +8,6 @@ export function ellipsis(text: string, maxLength: number = 20): string {
 
 // Checks if the _whole_ string is numeric, returning it as a number
 // returns undefined when non-numeric characters are encountered
-export function parseIntStrict(text: string): number | undefined {
-  return /^\d+$/.test(text) ? parseInt(text, 10) : undefined;
+export function parseIntStrict(text: string, radix: number = 10): number | undefined {
+  return /^\d+$/.test(text) ? parseInt(text, radix) : undefined;
 }
--- a/frontend/lib/util/strings.ts
+++ b/frontend/lib/util/strings.ts
@@ -8,6 +8,7 @@ export function ellipsis(text: string, maxLength: number = 20): string {
 
 // Checks if the _whole_ string is numeric, returning it as a number
 // returns undefined when non-numeric characters are encountered
-export function parseIntStrict(text: string, radix: number = 10): number | undefined {
-  return /^\d+$/.test(text) ? parseInt(text, radix) : undefined;
+export function parseIntStrict(text: string): number | undefined {
+  const num = parseInt(text, 10);
+  return !isNaN(num) && num.toString() === text ? num : undefined;
 }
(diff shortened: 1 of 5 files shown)


frontend/lib/util/strings.ts
```diff
@@ -5,3 +5,9 @@ export function ellipsis(text: string, maxLength: number = 20): string {
 
   return text.substring(0, maxLength - 3) + "...";
 }
+
+// Checks if the _whole_ string is numeric, returning it as a number
+// returns undefined when non-numeric characters are encountered
+export function parseIntStrict(text: string, radix: number = 10): number | undefined {
+  return /^\d+$/.test(text) ? parseInt(text, radix) : undefined;
```
praseodym, line 12: Do we have any idea of the performance implications of using a regex here? Regexes are not necessarily fast.

cikzh: No, not exactly. I guess it would partly depend on the browser's JavaScript engine. I went for the easiest implementation first, but I'm happy to write a more straightforward implementation here.

cikzh: Also, we need tests :)

praseodym: Also, with this regex the radix doesn't make sense for radices > 10 (e.g. "7a" is valid hex but doesn't match the regex).

If we set radix = 10, this might be an alternative without regexes:


```typescript
function parseInteger(str: string): number | undefined {
  const num = parseInt(str, 10);
  return !isNaN(num) && num.toString() === str ? num : undefined;
}

// Example usage:
console.log(parseInteger("123")); // 123
console.log(parseInteger("123.45")); // undefined
console.log(parseInteger("abc")); // undefined
console.log(parseInteger("123abc")); // undefined
```

cikzh:
@jschuurk-kr @praseodym @rnijveld in your opinion, should the follow test case pass or fail?

```ts
expect(parseIntStrict("00123")).toBe(123);
```

rnijveld:
> @jschuurk-kr @praseodym @rnijveld in your opinion, should the follow test case pass or fail?
> 
> ```ts
> expect(parseIntStrict("00123")).toBe(123);
> ```

I don't have a strong preference, @praseodym's code would deny this, while your regex one allows this. I would mostly not worry about performance here though, this code won't be in a hot loop, and can be called a million times per second even on my iphone. I would remove the radix argument though, we probably only ever deal with decimal integers anyway. I would say @praseodym's code is a little more readable and keeps it the most strict, so I would go for that one and deny this case.

jschuurk-kr:
> @jschuurk-kr @praseodym @rnijveld in your opinion, should the follow test case pass or fail?
> 
> ```ts
> expect(parseIntStrict("00123")).toBe(123);
> ```

Agree with what @rnijveld said.

Additionally, leaning towards being very strict and having this fail. The purpose is to convert a numerical string to a number. Dealing with leading zeroes is a separate concern.

cikzh:
I also agree. [77db226](https://github.com/kiesraad/abacus/pull/208/commits/77db2263825d9c97724616c96b1d4813b61cc74a) implements @praseodym 's version and adds tests. I'm going to go ahead and merge this to #193 . Thanks for your inputs!

PR #209, opened 2024-08-02
Provide feedback when user clicks "Beginnen"

### LINKED ISSUE #195: Provide feedback when user clicks "Beginnen" without inputting a (valid) polling station number
On the `1/input` page:

- show the alert "Voer een geldig nummer van een stembureau in om te beginnen" in this [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3663-15455&t=nFxASayzDUpV6xHl-11) when
  - The user does not input a polling station number and clicks "Beginnen"
  - The user inputs a non existing polling station number and clicks "Beginnen"
- Hide the alert when the users edits the polling station number input (i.e. the moment the 'aan het zoeken' message is shown.

The design of the alert component is based on the existing alert that is used in several places, ie after a successful login

cikzh:
Closes #195

### DIFF
--- a/frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.test.tsx
+++ b/frontend/app/component/form/polling_station_choice/PollingStationChoiceForm.test.tsx
@@ -133,4 +133,41 @@ describe("Test PollingStationChoiceForm", () => {
     // Check if the error message is visible
     expect(screen.getByText("Geen stembureaus gevonden")).toBeVisible();
   });
+
+  test("Submitting an empty or invalid polling station shows alert", async () => {
+    const user = userEvent.setup();
+    render(
+      <PollingStationListProvider electionId={1}>
+        <PollingStationChoiceForm />
+      </PollingStationListProvider>,
+    );
+
+    const pollingStation = screen.getByTestId("pollingStation");
+    const submitButton = screen.getByRole("button", { name: "Beginnen" });
+
+    await user.click(submitButton);
+
+    // Test that an alert is visible
+    const pollingStationSubmitFeedback = await screen.findByTestId("pollingStationSubmitFeedback");
+    expect(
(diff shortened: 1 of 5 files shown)


jorisleker (changes requested):
@cikzh noticed that the color we used for color-error was not consistent. I've updated it. Can you update the color of the icon and the left border to `#B42318` (all error icons should be this color, so I think you should update the color-error var).

The text should be black instead of color-error, my bad. Updated the design as well.

Please take a look at the paddings of the component:
- padding on the container is `padding: 16px 24px;` (1rem 1.5 rem)
- space between icon and text: 16px (1rem)
- the container should be display:inline-box instead of box

Size of text should be 16px (1rem) iso 20px (1.25rem)

cikzh:
I see #B42318 is equal to the `--color-error-darkest` var, so I use that now

cikzh:
@jorisleker I've updated that paddings as you specified.

>     * the container should be display:inline-box instead of box

I assumed this is because of the width of the box? I've also removed the `width: 100%`. Let me know if that's correct.

jorisleker:
> @jorisleker I've updated that paddings as you specified.
> 
> > ```
> > * the container should be display:inline-box instead of box
> > ```
> 
> I assumed this is because of the width of the box? I've also removed the `width: 100%`. Let me know if that's correct.

Yes, perfect. Now only font-size remains :-)

cikzh:
> Yes, perfect. Now only font-size remains :-)

[93259a4](https://github.com/kiesraad/abacus/pull/209/commits/93259a4940f1eea826b3d92ee68ca5feee9995c0)

jorisleker (approved):

Lionqueen94 (approved):

Issue #101, opened 2024-06-25
UX/design debt 2


jorisleker:
- [x] On /input the link 'Bekijk de lijst met alle stembureaus' should have cursor:pointer (applies to all links)
- [x] The 'Shift + Enter' text should not break over multiple lines. Use `white-space: nowrap;`. Add some padding/margin below the 'Beginnen' button so the 'shift + enter' text is positioned neatly below the button (screenshot of how it should look: .5rem between button and text, text has .5rem margin-left (iso 2rem) <img width="273" alt="Scherm­afbeelding 2024-06-26 om 08 27 17" src="https://github.com/kiesraad/urn-uitslag/assets/3286523/358dac49-f065-4be1-ac9e-98ae2637f980">
- [x] There should be a spacer before the input for the list-totals, that's lacking. Also, there should **not** be a bottom border nor a background on the totals row. (see design below) <img width="683" alt="Scherm­afbeelding 2024-07-16 om 14 37 10" src="https://github.com/user-attachments/assets/efaf3af1-e574-40f4-9599-64163d360aeb">
- [x] ~On /input it looks like the width of the paragraph of text "Klopt de naam van het stembureau met de naam op je papieren proces verbaal? Dan kan je beginnen." is defined as a percentage of the parent column. Please use max-width instead, to  allow the text to occupy more space on smaller viewports. (applies to all paragraph texts: use max-width instead of percentage) <img width="573" alt="Scherm­afbeelding 2024-06-26 om 08 21 10" src="https://github.com/kiesraad/urn-uitslag/assets/3286523/2d23c334-c84f-455b-9332-5f261b05cd42">~
- [x] All pages have 'Abacus' as page title. Pages should have descriptive titles
  - /user/login --> Inloggen - Abacus
  - /user/account/setup --> Account instellen  - Abacus
  - /overview --> Overzicht verkiezingen - Abacus
  - /input --> Kies een stembureau - Abacus
  - /input/030 --> Invoeren `<stembureau nummer>` `<stembureau naam>`  - Abacus
- [x] When navigating to another page, user should be scrolled to top of page (repro: make your window less high, open list of polling stations on '/input', scroll to bottom of page and pick the last polling station. You will end up on a scrolled down version of the data entry screen (tested on https://6b2ce232.urn-uitslag.pages.dev/input/030 with long list of political groups)
[...]

PR #215, opened 2024-08-05
Frontend: A few design fixes

Lionqueen94:
## Done:
- [x] Alert Figma design changes (see https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=985-19666&m=dev#863283465)

Fixed from #101 :
- [x] when a list has exactly n\*25 candidates, the spacer/separator for the next 25 candidates is shown. The separator should only be added if there is a (n\*25)+1'nd  candidate.  Scherm­afbeelding 2024-08-01 om 11 13 17
- [x] The zebra-striping of the rows does not match the striping used in the paper forms. On paper, the first row of each 25 rows is white. In our digital form, row 26 and row 76 have a dark background. Each block of 25 rows should start with a white background.
Scherm­afbeelding 2024-08-05 om 09 49 33


praseodym:
It looks like this PR includes some unintended changes (e.g. to the backend) that might be the result of incorrectly resolved merge conflicts?

Lionqueen94:
> It looks like this PR includes some unintended changes (e.g. to the backend) that might be the result of incorrectly resolved merge conflicts?

Ah shit I created this from my other branch, will correct it

PR #228, opened 2024-08-06
first dom-to-db tests

jschuurk-kr:
This PR adds three dom-to-db tests:
1. input flow, no recount, no differences
1. display error
1. display warning

Purpose of the last two tests is twofold:
- test displaying of errors and warnings;
- ensure the assert in the input flow tests for the invisibility of errors/warnings checks the right things.

I also moved the Playwright tests for app+msw with the dom-to-db tests to `frontend/e2e-tests`. The Playwright Ladle tests remain in `frontend/lib/ui`.

The backend now also takes an optional `--port` argument, so Playwright can start the backend on a different port than the default one.

---

Two things I'd like feedback on:
1\. The page objects are fully in English. Arguably they should be in Dutch. Having the locators in Dutch makes sense. But should convenience functions then also be (partially) in Dutch (`fillCandidate()` or `fillKanidaat()` or `kandidaatInvulen()`)? And either way you'd end up with a weird mix of Dutch and English, e.g. `await verschillenPagina.volgende.click();`

2\. In `VotersVotesPgObj.ts` I defined two `interface`s for the arguments of two convenience methods. Is this the best way to do this? The internet is not providing me a clear answer on `type` vs `interface`.

---

Finally, a comment on the locators. The [Playwright docs recommends](https://playwright.dev/docs/locators#locating-elements) using user-facing attributes, so using `getByRole()` and `getByLabel()` over using `getByTestId()` over using CSS/XPath locators. We can mostly follow that advice. For the `InputGrid` however, I've had to use `getByTestId()`. Writing Playwright tests as part of a feature PR should help with this in the future. As it's easier to make sure there are good locators at that time, then doing it afterwards.


frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
```diff
@@ -0,0 +1,117 @@
...
+    await differencesPage.clickPoliticalGroup("Political Group A"); // TODO: remove once naviation works
+
+    const candidatesListPage_1 = new CandidatesListPage(page);
+    await candidatesListPage_1.fillCandidate(0, 100);
+    await candidatesListPage_1.fillCandidate(1, 22);
+    await candidatesListPage_1.total.fill("122");
+    await candidatesListPage_1.next.click();
+
+    await expect(differencesPage.error).toBeHidden();
+    await expect(differencesPage.warning).toBeHidden();
```
Lionqueen94, line 68:
```suggestion
    await expect(candidatesListPage_1.error).toBeHidden();
    await expect(candidatesListPage_1.warning).toBeHidden();
```

frontend/e2e-tests/page-objects/input/DifferencesPgObj.ts
```diff
@@ -0,0 +1,26 @@
...
+  readonly next: Locator;
+
+  constructor(page: Page) {
+    super(page);
+
+    this.moreBallotsCount = page.getByTestId("more_ballots_count");
+    this.fewerBallotsCount = page.getByTestId("fewer_ballots_count");
+    this.unreturnedBallotsCount = page.getByTestId("unreturned_ballots_count");
+    this.tooFewBallotsHandedOutCount = page.getByTestId("too_few_ballots_handed_out_count");
+    this.tooFewBallotsHandedOutCount = page.getByTestId("too_many_ballots_handed_out_count");
```
Lionqueen94, line 20:
```suggestion
    this.tooManyBallotsHandedOutCount = page.getByTestId("too_many_ballots_handed_out_count");
```

frontend/e2e-tests/page-objects/input/DifferencesPgObj.ts
```diff
@@ -0,0 +1,26 @@
...
+    this.otherExplanationCount = page.getByTestId("other_explanation_count");
+    this.otherExplanationCount = page.getByTestId("no_explanation_count");
```
Lionqueen94, line 22:
```suggestion
    this.noExplanationCount = page.getByTestId("no_explanation_count");
```

frontend/e2e-tests/page-objects/input/DifferencesPgObj.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { InputBasePage } from "./InputBasePgObj";
+
+export class DifferencesPage extends InputBasePage {
+  readonly moreBallotsCount: Locator;
+  readonly fewerBallotsCount: Locator;
+  readonly unreturnedBallotsCount: Locator;
+  readonly tooFewBallotsHandedOutCount: Locator;
+  readonly otherExplanationCount: Locator;
+  readonly next: Locator;
```
Lionqueen94, line 11:
```suggestion
  readonly moreBallotsCount: Locator;
  readonly fewerBallotsCount: Locator;
  readonly unreturnedBallotsCount: Locator;
  readonly tooFewBallotsHandedOutCount: Locator;
  readonly tooManyBallotsHandedOutCount: Locator;
  readonly otherExplanationCount: Locator;
  readonly noExplanationCount: Locator;
  readonly next: Locator;
```

jschuurk-kr: I removed the unused locators instead.

praseodym (commented):
> Two things I'd like feedback on: 1. The page objects are fully in English. Arguably they should be in Dutch. Having the locators in Dutch makes sense. But should convenience functions then also be (partially) in Dutch (`fillCandidate()` or `fillKanidaat()` or `kandidaatInvulen()`)? And either way you'd end up with a weird mix of Dutch and English, e.g. `await verschillenPagina.volgende.click();`

This is why I prefer writing code only English instead of a weird mix of both.

> 2. In `VotersVotesPgObj.ts` I defined two `interface`s for the arguments of two convenience methods. Is this the best way to do this? The internet is not providing me a clear answer on `type` vs `interface`.

Interfaces are fine here. Both `type` and `interface` have pros and cons, and if there is no need for advanced features that only one of the two provides then it comes down to a stylistic choice. There is an optional lint for this: https://typescript-eslint.io/rules/consistent-type-definitions/ by default prefers interfaces, but can also be configured to prefer types instead.

frontend/playwright.d2d.config.ts
praseodym, line 1: I think we should extract the shared config between the three files, I can have a go at that if you agree?

praseodym: See #230!

cikzh (approved):
Looks good!

> The one disadvantage I see of having it all in English is that you have to translate when mapping pages to page objects and vice versa. So it does take a little extra effort to read the tests. But I still prefer that over using a weird hodgepodge of Dutch and English.

I'd say English is fine, we do the mapping all over the place in our application.

praseodym (approved):

PR #232, opened 2024-08-07
Add API command flag to seed data into the database on startup

### LINKED ISSUE #154: Simplify loading fixture data in backend
We should make loading fixture data in the backend easy

_Originally posted by @rnijveld in https://github.com/kiesraad/abacus/issues/109#issuecomment-2220012436_

rnijveld:
Resolves #154 

This also adds a `--reset-database` flag to allow quickly re-initializing the database.

The fixture seeding is implemented as a list of fixtures that needs to be updated manually, this allows setting a different set of fixtures for data loading (i.e. allowing you to skip some fixtures that are only useful in specific tests for example).

### DIFF
--- a/backend/src/bin/api.rs
+++ b/backend/src/bin/api.rs
@@ -25,10 +25,12 @@ struct Args {
     port: u16,
 
     /// Seed the database with initial data using the fixtures
+    #[cfg(feature = "dev-database")]
     #[arg(short, long)]
     seed_data: bool,
 
     /// Reset the database
+    #[cfg(feature = "dev-database")]
     #[arg(short, long)]
     reset_database: bool,
 }
@@ -38,7 +40,7 @@ struct Args {
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
     let args = Args::parse();
-    let pool = create_sqlite_pool(args.reset_database, args.seed_data).await?;
+    let pool = create_sqlite_pool(&args).await?;
     let app = router(pool)?;
 
     let app = if let Some(fd) = args.frontend_dist {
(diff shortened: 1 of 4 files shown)


backend/src/bin/api.rs
```diff
@@ -46,9 +54,53 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    }
+
     Ok(pool)
 }
+
+mod fixtures {
+    macro_rules! load_fixture {
+        ($fixture:literal) => {
+            Fixture {
+                data: include_str!(concat!("../../fixtures/", $fixture, ".sql")),
```
jschuurk-kr, line 80: It would be nice to be able to give the path to the fixtures as an optional argument. That way when the pipeline uploads and then downloads the built executable and the fixtures, they don't have to be in a folder structure matching this path.

praseodym: [`include_str!`](https://doc.rust-lang.org/std/macro.include_str.html) includes the file in the binary, so it doesn't need to be present when running the binary. I think that should resolve this?

rnijveld: Yep! That should just work, you don't need the fixture files anymore after you've produced the binary. I could sort of see the use-case of allowing you to load different sets of fixtures for different use-cases, but I would prefer to only add that if we ever need it.

praseodym (commented):
lgtm, but maybe some doc comments in the fixtures module would be helpful?

praseodym (dismissed):

backend/src/bin/api.rs
jschuurk-kr: Ah cool, that makes it even easier. :-)

That does mean that for now we would ship the fixtures in a release?

rnijveld: I can put it behind a feature flag that is enabled by default, that way we can disable all the database loading and resetting functionality for a build that is intended for release? @praseodym any opinion?

praseodym: > I can put it behind a feature flag that is enabled by default, that way we can disable all the database loading and resetting functionality for a build that is intended for release? @praseodym any opinion?

Sounds good!

praseodym:
I've added a build with `--no-default-features` to CI to make sure the build without default features doesn't accidentally break again, see 3448791.

praseodym (approved):

praseodym:
> I've added a build with `--no-default-features` to CI to make sure the build without default features doesn't accidentally break again, see [3448791](https://github.com/kiesraad/abacus/commit/344879118592f3a040c0ad26a022ec5801e9c67c).

Actually we don't need a full build to check this, we can let Clippy do it instead. See 355f600.

praseodym (approved):

cikzh (approved):

Issue #50, opened 2024-05-21
Data entry for candidate lists


praseodym:
Extend frontend with data entry screens according to [Figma design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---URN?node-id=126-3819), showing candidates from the API (#48, #54). Saving to data entry to the backend is not part of this issue, but next up in #52.

PR #239, opened 2024-08-07
Add document describing logic of data entry forms

jschuurk-kr:
Goal: document the logic in our data entry screens

Questions:
- is this useful?
- is this correct and complete?

### DIFF
--- a/documentatie/flowcharts/data-entry-forms.md
+++ b/documentatie/flowcharts/data-entry-forms.md
@@ -171,7 +171,7 @@ flowchart TD
 Note that if there is a warning and the user changes the input, they should no longer have the option to accept the warning. They need to click "Next" first, to validate the changed input.
 
 
-## Click navigation item
+## Click navigation item or browser back/forward buttons
 
 ```mermaid
 flowchart TD
@@ -183,6 +183,10 @@ flowchart TD
 
     %% steps
     click-nav-item(click navigation item)
+    click-browser-back("click browser
+        back button")
+    click-browser-forward("click browser
+        forward button")
     on-furthest-page{on furthest page?}
     user-made-changes{user made changes?}
     save-changes{save changes?}
@@ -193,6 +197,11 @@ flowchart TD
 
     %% flow
     flow-start --> click-nav-item
+    flow-start --> click-browser-back
+    flow-start --> click-browser-forward
+    click-browser-back --> on-furthest-page
+    %% if you can click the browser forward button, you by definition are not on the furthest page
+    click-browser-forward --> user-made-changes
     click-nav-item --> on-furthest-page
     on-furthest-page -- yes --> cache-input
     cache-input --> go-to-page
--- a/documentatie/flowcharts/data-entry-forms.md
+++ b/documentatie/flowcharts/data-entry-forms.md
@@ -18,13 +18,13 @@ flowchart TD
     %% elements
     flow-start([start])
 
-    disabled-styling([disabled styling])
-    current-styling([current styling])
-    unsubmitted-styling([unsubmitted styling])
-    error-styling([error styling])
-    warning-styling([warning styling])
-    empty-styling([empty styling])
-    green-check-styling([valid styling])
+    styling-disabled([styling: disabled])
+    styling-current([styling: current])
+    styling-unsubmitted([styling: unsubmitted])
+    styling-error([styling: error])
+    styling-warning([styling: warning])
+    styling-empty([styling: empty])
+    styling-valid([styling: valid])
 
     page-visited{page visited before?}
     current-page{current page?}
@@ -36,18 +36,18 @@ flowchart TD
 
     %% flow
     flow-start --> page-visited
-    page-visited -- no --> disabled-styling
+    page-visited -- no --> styling-disabled
     page-visited -- yes --> current-page
-    current-page -- yes --> current-styling
+    current-page -- yes --> styling-current
     current-page -- no --> submitted
-    submitted -- no --> unsubmitted-styling
+    submitted -- no --> styling-unsubmitted
     submitted -- yes --> errors
-    errors -- yes --> error-styling
+    errors -- yes --> styling-error
     errors -- no --> warnings
-    warnings -- yes --> warning-styling
+    warnings -- yes --> styling-warning
     warnings -- no --> empty
-    empty -- yes --> empty-styling
-    empty -- no --> green-check-styling
+    empty -- yes --> styling-empty
+    empty -- no --> styling-valid
 
 ```
 
--- a/documentatie/flowcharts/data-entry-forms.md
+++ b/documentatie/flowcharts/data-entry-forms.md
@@ -13,8 +13,6 @@ An important thing to keep in mind when reading these diagrams is that a user ca
 
 Render happens based on last received API response.
 
-### Diagram
-
 ```mermaid
 flowchart TD
     %% elements
@@ -71,6 +69,8 @@ flowchart TD
     render-submitted-data([render with submitted data])
 
     show-error([show error])
+    hide-checkbox-accepted(["hide checkbox
+        warnings accepted"])
     show-warning([show warning])
     show-checked-accepted(["show checked checkbox
         warnings accepted"])
@@ -81,6 +81,8 @@ flowchart TD
     error-cur-page{error for current page?}
     warning-cur-page{warning for current page?}
     cached-input-available{cached input available?}
+    input-changed{"input changed
+        since submit?"}
     warning-accepted{"warning(s) accepted?"}
 
     %% flow
@@ -98,7 +100,10 @@ flowchart TD
     error-cur-page -- no --> warning-cur-page
 
     warning-cur-page -- yes --> show-warning
-    show-warning --> warning-accepted
+    show-warning --> input-changed
+    input-changed -- yes --> hide-checkbox-accepted
+    hide-checkbox-accepted --> flow-end
+    input-changed -- no --> warning-accepted
     warning-accepted -- yes --> show-checked-accepted
     warning-accepted -- no --> show-unchecked-accepted
     show-checked-accepted --> flow-end
@@ -128,10 +133,6 @@ flowchart TD
 
     user-addresses-warning{user addresses warning}
 
-    
-    render-error(render error)
-    render-warning(render warning)
-
     click-next(click next)
     call-api(call api)
     change-input(change input)
@@ -148,16 +149,14 @@ flowchart TD
 
     error-any-prev-page -- no --> error-cur-page
 
-    error-cur-page -- yes --> render-error
-    render-error --> user-addresses-error
+    error-cur-page -- yes --> user-addresses-error
     
     user-addresses-error -- abort --> abort-input
     user-addresses-error -- resolve --> change-input
     change-input --> click-next
 
     error-cur-page -- no --> warning-cur-page
-    warning-cur-page -- yes --> render-warning
-    render-warning --> warnings-accepted
+    warning-cur-page -- yes --> warnings-accepted
     warnings-accepted -- no --> user-addresses-warning
     warnings-accepted -- yes --> go-to-next-page
     user-addresses-warning -- resolve --> change-input
--- a/documentatie/flowcharts/date-entry-forms.md
+++ b/documentatie/flowcharts/date-entry-forms.md
@@ -66,31 +66,44 @@ flowchart TD
 
     render-empty-fields([render with empty fields])
     render-cached-data([render with cached data])
+    flow-end([end])
+
     render-submitted-data([render with submitted data])
 
-    show-error(show error)
-    show-warning(show warning)
+    show-error([show error])
+    show-warning([show warning])
+    show-checked-accepted(["show checked checkbox
+        warnings accepted"])
+    show-unchecked-accepted(["show unchecked checkbox
+        warnings accepted"])
 
-    cur-page-submitted{current page submitted}
+    page-submitted{page submitted?}
     error-cur-page{error for current page?}
     warning-cur-page{warning for current page?}
     cached-input-available{cached input available?}
+    warning-accepted{"warning(s) accepted?"}
 
     %% flow
-    flow-start --> cur-page-submitted
+    flow-start --> page-submitted
 
-    cur-page-submitted -- no --> cached-input-available
+    page-submitted -- no --> cached-input-available
     cached-input-available -- yes --> render-cached-data
     cached-input-available -- no --> render-empty-fields
-    cur-page-submitted -- yes --> error-cur-page
+    page-submitted -- yes --> render-submitted-data
+    
+    render-submitted-data --> error-cur-page
 
     error-cur-page -- yes --> show-error
-    show-error --> render-submitted-data
+    show-error --> flow-end
     error-cur-page -- no --> warning-cur-page
 
     warning-cur-page -- yes --> show-warning
-    show-warning --> render-submitted-data
-    warning-cur-page -- no --> render-submitted-data
+    show-warning --> warning-accepted
+    warning-accepted -- yes --> show-checked-accepted
+    warning-accepted -- no --> show-unchecked-accepted
+    show-checked-accepted --> flow-end
+    show-unchecked-accepted --> flow-end
+    warning-cur-page -- no --> flow-end
 ```
 
 
--- a/documentatie/flowcharts/date-entry-forms.md
+++ b/documentatie/flowcharts/date-entry-forms.md
@@ -109,7 +109,6 @@ flowchart TD
 
 ## Click "Volgende"
 
-
 ```mermaid
 flowchart TD
 
@@ -140,8 +139,6 @@ flowchart TD
 
     go-to-next-page([go to next page])
 
-
-
     %% flow
     flow-start --> click-next
     click-next --> call-api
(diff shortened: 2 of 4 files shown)


documentatie/flowcharts/date-entry-navigation.md
```diff
@@ -0,0 +1,149 @@
...
+
+    cur-page-submitted{cur-page-submitted?}
+
+    fill-cached-input
+
+    %% flow
+    flow-start --> error-prev-page
+
+    error-prev-page -- yes --> illegal-state
+    error-prev-page -- no --> error-next-page
```
jorisleker, line 46: any next page could have an error. Only one, and there could not be more then one. But it does not have to be the immediate next page. Is this what you mean here?

Also: why do we need to know whether the next page has an error when rendering the current page? We need to know this to set the correct state on the items in the navigation menu, but then we need to know it for all pages, not just prev, current and next?

documentatie/flowcharts/date-entry-navigation.md
```diff
@@ -0,0 +1,149 @@
...
+
+    error-next-page -- yes --> error-icon-next-page
+    error-icon-next-page --> error-cur-page
+    error-next-page -- no --> error-cur-page
+
+    error-cur-page -- yes --> cur-page-submitted
+    cur-page-submitted -- yes --> error-icon-cur-page
+    error-icon-cur-page --> show-error
```
jorisleker, line 54: the current page never has an icon other then --> in the nav menu. No errors, no warnings. Just the -->

documentatie/flowcharts/date-entry-navigation.md
```diff
@@ -0,0 +1,149 @@
...
+    warning-prev-page -- yes --> warning-icon-prev-page
+    warning-icon-prev-page --> warning-next-page
+
+    warning-next-page -- no --> warning-cur-page
+    warning-next-page -- yes --> warning-icon-next-page
+    warning-icon-next-page --> warning-cur-page
+
+    warning-cur-page -- no --> fill-cached-input
+    warning-cur-page -- yes --> warning-icon-cur-page
+    warning-icon-cur-page --> show-warning
```
jorisleker, line 68: the current page never has an icon other then --> in the nav menu. No warnings, no errors. We do ofc show errors, and if there are no errors any warnings.

documentatie/flowcharts/date-entry-navigation.md
```diff
@@ -0,0 +1,149 @@
+# Data Entry Navigation
+
+## Questions
+- Where does "cache input" happen in these flows?
+- Is there a different warning icon for accepted versus not-accepted warnings? Or is it not possible to leave a page without accepting the warning? Also, accepting warnings is not yet included in any of the flows.
```
jorisleker, line 5: - it's impossible to leave the page without accepting the warning.
- there is no other icon. All warnings are accepted warnings and we show the icon regardless.

documentatie/flowcharts/date-entry-navigation.md
```diff
@@ -0,0 +1,149 @@
...
+    save-changes{save-changes?}
+    go-to-page
+    remain-on-page
+    api-call
+
+    errors-or-warnings-cur-page{errors-or-warnings-cur-page?}
+
+    %% flow
+    flow-start --> click-nav-item
+    click-nav-item --> save-changes
```
jorisleker, line 140: not sure anymore whether this was what we agreed upon. This is rather nasty: If I want to navigate to a previous page, because I just discovered I did something wrong there, this logic will force me to complete the current page before I can do that. Not sure we would want this.

jorisleker (changes requested):

documentatie/flowcharts/date-entry-navigation.md
jschuurk-kr: prev-page is indeed any previous page. next-page is any next page. Will add the "any".

documentatie/flowcharts/date-entry-navigation.md
jschuurk-kr: It's how I interpreted https://github.com/kiesraad/abacus/issues/50#issuecomment-2210730723 `save-changes` is the dialog asking the user if they want to save their changes or not. So the user does have the choice to not save their changes and go to the other page.

What is missing in this flowchart is that if the user does not wish to save changes, we should cache the input.

Which means we need to be mindful of our choice of wording in the [save changes dialog](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3151-21014&t=Qt2ouMkmbcb1VKOR-0).

jorisleker:
Flow for accepting warnings (as discussed with Les)

- user submits form to backend
- backend does validations
- if input triggers one or more warnings:
  - user remains on current page
  - warning is shown
  - checkbox 'accept warnings to continue is show'
  - if user does not accept warnings, she cannot continue to next page: show alert 'you have to accept warnings'
- if user changes any of the inputs on current page:
  - hide the 'accept warnings' checkbox (we don't know wether the changed input still contains warnings, or leads to new warnings or erros'
  - if user submits the page, revalidate the users input (return to first bullet)
 
In short:
- only navigate to next step if:
  - input did not contain errors or warnings
  - input did contain warnings but 'accept warnings' was checked
- only show the 'accept warnings' checkbox if:
  - the page contains warnings
  - the input is unchanged compared to what was submitted to the backend

documentatie/flowcharts/date-entry-navigation.md
jschuurk-kr: fixed

documentatie/flowcharts/date-entry-page.md
```diff
@@ -0,0 +1,182 @@
...
+    current-icon-cur-page
+
+    error-icon-next-page
+    %% error-icon-prev-page is illegal state
+
+    warning-icon-next-page
+    warning-icon-prev-page
+
+    error-any-next-page{error-any-next-page?}
+    error-any-prev-page{error-any-prev-page?}
```
jorisleker, line 33: I think there isn't really a reason to differentiate between previous and next pages from here? Just 'any other' pages should suffice?

also: 'next pages' should be interpreted as 'next pages up to the furthest page the user has seen'. Pages beyond what the user has seen are unreachable, and cannot have an icon

jorisleker (changes requested):

documentatie/flowcharts/date-entry-navigation.md
jschuurk-kr: fixed

documentatie/flowcharts/date-entry-navigation.md
jorisleker: @jschuurk-kr ah, that's not how https://github.com/kiesraad/abacus/issues/50#issuecomment-2210730723 was meant. There is a subtle difference between the following cases:

- you are on the furtest page and navigate to any previous page: we save your input, but don't do any validation. This input has an 'incomplete' state, which is fine for the furtest page. Valdiation only happens when you use the 'volgende' button and try to continue to the next step. 
- you are not on the furtest page, have made changes and navigate to any other page: because your input might contain errors that will make any next step an illegal one, we need to validate the input. Now we show the dialog to ask the user whether they want to save their changes.

Shorter:
- When user uses navigation:
  - on furtest page? Submit/cache input, no validation
  - not on furtest page? Aks user if they want to save changes
    - y: submit to backend, validate. Errors or warnings?
      - remain on page, show error/warning and have user fix them or accept warnings
    - n: reset changes, navigate to requested page.

documentatie/flowcharts/date-entry-page.md
```diff
@@ -17,55 +17,92 @@ flowchart TD
...
+    fill-cached-input
+
+    %% flow
+    flow-start --> error-any-prev-page
+
+    error-any-prev-page -- yes --> illegal-state
+    error-any-prev-page -- no -->  error-cur-page
+
+    error-cur-page -- yes --> cur-page-submitted-error
+    cur-page-submitted-error -- yes --> show-error
```
jorisleker, line 82: I think this diagram can be simplified by putting `cur-page-submitted` above `error-cur-page`, because then you won't need to repeat `cur-page-submitted`. If no: you just render the cashed data. If yes, first check for errors, then check for warnings.

documentatie/flowcharts/date-entry-page.md
```diff
@@ -0,0 +1,167 @@
...
+    call-api --> error-any-prev-page
+    error-any-prev-page -- yes --> go-to-prev-page
+    go-to-prev-page --> flow-end
+
+    error-any-prev-page -- no --> error-cur-page
+
+    error-cur-page -- yes --> render-cur-page
+    error-cur-page -- no --> warning-cur-page
+
+    warning-cur-page -- yes --> render-cur-page
```
jorisleker, line 131: this should be extended with the 'accept warning' flow. 

when current page contains a warning, current page contains a checkbox to accept the warning
if checkbox is checked, the warning is dismissed and user navigates to next page

documentatie/flowcharts/date-entry-navigation.md
jschuurk-kr: flowchart aangepast

documentatie/flowcharts/date-entry-page.md
jschuurk-kr: updated based on your comment with the text description https://github.com/kiesraad/abacus/commit/02a1e6bd28f4d580e211ea4af9167c2a284cf0fd#commitcomment-145135420

documentatie/flowcharts/date-entry-page.md
jschuurk-kr: good catch! updated the diagram, also removing the "illegal state" part in favor of a comment at the top of the document

jschuurk-kr:
> if the user navigates back to a page where a warning was previously accepted, we do show the checkbox, in checked state.

@jorisleker Can a user un-accept the warning, as in un-check the checkbox after checking it and submitting the page?

jorisleker:
> @jorisleker Can a user un-accept the warning, as in un-check the checkbox after checking it and submitting the page?

Yes, the UI makes that possible. Would not really make sense to do it though, as it would get you stuck on the current page.

I am not totally sure we will keep the checkbox as the way to accept warnings. We might change this to a dialog after clicking 'volgende'. 

Benefit of that would be that we could only show it when it's relevant (user made changes to the input, and the (new) input has warnings). Downside is that these types of dialogs tend to result in user automatically clicking the default action + it's hard to see the input + the warning when a dialog is displayed over it. 

So: still in doubt whether the checkbox is the ideal UI, but best I can think of now

jorisleker:
@jschuurk-kr Nice updates. I think the only thing still missing is the 'accept warnings' step?

jschuurk-kr:
> @jschuurk-kr Nice updates. I think the only thing still missing is the 'accept warnings' step?

@jorisleker working on it

documentatie/flowcharts/date-entry-page.md
jschuurk-kr: done

praseodym:
What about using the browser's back and forward buttons? How do we deal with that for the "save changes?" and "remain on current page" steps?

documentatie/flowcharts/date-entry-forms.md
```diff
@@ -0,0 +1,211 @@
...
+
+    user-addresses-warning{user addresses warning}
+
+    
+    render-error(render error)
+    render-warning(render warning)
+
+    click-next(click next)
+    call-api(call api)
+    change-input(change input)
```
jorisleker, line 137: the 'hide accept warnings checkbox' item should be added between this and 'click next'

jorisleker (changes requested):

documentatie/flowcharts/date-entry-forms.md
```diff
@@ -0,0 +1,211 @@
...
+    show-warning --> warning-accepted
+    warning-accepted -- yes --> show-checked-accepted
+    warning-accepted -- no --> show-unchecked-accepted
+    show-checked-accepted --> flow-end
+    show-unchecked-accepted --> flow-end
+    warning-cur-page -- no --> flow-end
+```
+
+
+## Click "Volgende"
```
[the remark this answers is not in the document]
jschuurk-kr: good one, removed both the 'render's

documentatie/flowcharts/date-entry-forms.md
jschuurk-kr: Shouldn't that then go in the flowchart about rendering? I'll add it there.

jorisleker (dismissed):

jschuurk-kr:
> What about using the browser's back and forward buttons? How do we deal with that for the "save changes?" and "remain on current page" steps?

@praseodym @lkleuver Is it technically feasible to implement the same behavior as when the user clicks a navigation item? Something like this: https://stackademic.com/blog/handling-browser-back-and-forward-in-react?

jorisleker (approved):
Regarding browser back/forward button (or any other case where the user manipulates the url): the user can end up on a page that they are no longer allowed to view/edit. I.e.: after the user has finalized their input they can use the back button to navigate back to the last list, but they won't be able to edit it anymore. 

Not sure we want to address that issue in this diagram though. If we want to do that, I would generalize the back/forward/manual url entry cases into 1 item, and add a 'user allowed to edit page' check, that points to 'on furthest page' if the user is allowed, and to a 'handle not allowed' item if user is not allowed.

Approved nonetheless; we can describe this somewhere else as well

lkleuver (commented):
Click "Volgende"

Not sure if this is 100% accurate.
"Volgende" submits the current form and if no errors / warnings goes to the next form.
so "previous page" is a bit reversed. You "stay" on the page if errors.

Click navigation item
I think this misses a part, if you navigate to a page that is after the "current" form (in the form flow)
you are redirected to the current form.

lkleuver:
> > What about using the browser's back and forward buttons? How do we deal with that for the "save changes?" and "remain on current page" steps?
> 
> @praseodym @lkleuver Is it technically feasible to implement the same behavior as when the user clicks a navigation item? Something like this: https://stackademic.com/blog/handling-browser-back-and-forward-in-react?

Need to check if useBlocker works on this, it should. in theory this should already work, but browser back/forward actions are weird.

documentatie/flowcharts/data-entry-forms.md
```diff
@@ -0,0 +1,219 @@
...
+## Render navigation menu
+
+Render happens based on last received API response.
+
+```mermaid
+flowchart TD
+    %% elements
+    flow-start([start])
+
+    disabled-styling([disabled styling])
```
marjoleintamis, line 21: het duurde even voordat ik dit las als 'styling for disabled state' ipv 'styling is disabled'. Misschien iets preciezer bewoorden, zie suggestie. Dit geldt ook voor de andere states.
```suggestion
    disabled-styling([styling for disabled state])
```

documentatie/flowcharts/data-entry-forms.md
```diff
@@ -0,0 +1,219 @@
...
+
+    on-furthest-page -- no --> user-made-changes
+    user-made-changes -- no --> go-to-page
+    user-made-changes -- yes --> save-changes
+    save-changes -- no --> reset-changes
+    reset-changes --> go-to-page
+    save-changes -- yes --> call-api
+    call-api --> errors-or-warnings
+    errors-or-warnings -- no --> go-to-page
+    errors-or-warnings -- yes --> remain-on-page
```
marjoleintamis, line 217: wordt de error of warning hier ook gelijk weergegeven of is dat een aparte stap?

marjoleintamis (commented):
Een suggestie gedaan voor de bewoording van de styling-blokjes, en een vraag gesteld over wat er in de laatste flow chart gebeurt als je op de huidige pagina blijft.

documentatie/flowcharts/data-entry-forms.md
jschuurk-kr: Die worden weergegeven. Alle logica daarvoor staat beschreven in de "Render form" flowchart.

documentatie/flowcharts/data-entry-forms.md
jschuurk-kr: Ga er "styling: disabled" etc. van maken. Is korter en hopelijk ook voldoende duidelijker dan wat het was.

marjoleintamis (approved):

documentatie/flowcharts/data-entry-forms.md
```diff
@@ -18,13 +18,13 @@ flowchart TD
...
-    disabled-styling([disabled styling])
-    current-styling([current styling])
-    unsubmitted-styling([unsubmitted styling])
-    error-styling([error styling])
-    warning-styling([warning styling])
-    empty-styling([empty styling])
-    green-check-styling([valid styling])
+    styling-disabled([styling: disabled])
+    styling-current([styling: current])
+    styling-unsubmitted([styling: unsubmitted])
```
jorisleker, line 23: should this be: 'not-validated' in stead of 'unsubmitted'? Not sure how/where we are going to save/cache this data, but if we do that in the database, we do submit the data, we just don't validate it yet.

jorisleker (approved):
One small suggestion, but ok to ignore. Good enough :-)

documentatie/flowcharts/data-entry-forms.md
jschuurk-kr: Good point, let's update the flowchart when we start to save/cache the data. Because right now I don't think that state is even reachable, since we either submit + validate or we discard.

PR #240, opened 2024-08-07
Set the API to be the default run target for cargo

rnijveld:
Set the API to be the default run target in cargo (saves some typing)


cikzh (approved):

rnijveld:
I've updated the readme, and included the text for resetting and seeding the database as well.

cikzh (approved):

PR #255, opened 2024-08-12
Fixed issues on voters and votespage

Lionqueen94:
## Fixed:

- [x] up and down navigation of the fields on the Voters and Votespage
- [x] Marking the W.201 and W.202 warnings only on the field they are about and not also on the total field

### DIFF
--- a/frontend/lib/ui/InputGrid/InputGrid.module.css
+++ b/frontend/lib/ui/InputGrid/InputGrid.module.css
@@ -1,16 +1,20 @@
 .inputgrid {
   margin-top: 1rem;
   width: 100%;
-  border-spacing: 0;
   th {
     border-bottom: 1px solid var(--bg-gray-darkest);
     font-size: var(--font-size-xs);
     text-align: right;
     padding: 0.75rem 1.5rem;
     color: var(--color-help-text);
-    &:nth-of-type(3) {
+    &:nth-of-type(3),
+    &.section_title {
       text-align: left;
     }
+  }
+  th.section_title {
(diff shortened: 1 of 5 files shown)


praseodym (dismissed):

jorisleker (approved):
cycling through the inputs with up and down arrows and enter now works as expected

jorisleker:
Something did break in the styling I just noticed; all cells now seem to have a 1px padding. See screenshots (zoomed in)

Your PR:
Scherm­afbeelding 2024-08-13 om 07 45 06

Before (main):
Scherm­afbeelding 2024-08-13 om 07 46 38

jschuurk-kr:
@Lionqueen94 Can you please mention the two issues you fixed in the description of the PR?

frontend/lib/ui/InputGrid/InputGrid.module.css
```diff
@@ -1,17 +1,21 @@
 .inputgrid {
   margin-top: 1rem;
   width: 100%;
-  border-spacing: 0;
```
praseodym, line 4: Looks like removing this caused the cell padding as mentioned in https://github.com/kiesraad/abacus/pull/255#issuecomment-2285385684.

Lionqueen94: Yes I re-added it, so weird because the browser said this had no effect XD

praseodym (approved):

jorisleker (approved):

Issue #224, opened 2024-08-05
Combine input for Typst PDF generation in backend


cikzh:
Add up results of all polling stations to the total counts of the municipality.

- Assert that there are no errors in the data entry (run validation, check for no errors).
- Use `voters_recounts` instead of `voters_counts` when recounted.

PR #262, opened 2024-08-15
Add PDF download endpoint to backend

### LINKED ISSUE #223: Add PDF download functionality to backend

cikzh:
Closes #223 

Adds an endpoint that downloads a PDF. Mainly acts as a stub that will be expanded on in #224 and #225

### DIFF
--- a/backend/src/election/mod.rs
+++ b/backend/src/election/mod.rs
@@ -95,14 +95,14 @@ pub async fn election_status(
 /// Download a generated PDF
 #[utoipa::path(
     get,
-    path = "/api/elections/{election_id}/download-pdf",
+    path = "/api/elections/{election_id}/download_pdf",
     responses(
         (
             status = 200,
             description = "PDF",
-            content_type="text/pdf; charset=utf-8",
+            content_type="application/pdf",
             headers(
-                ("Content-Type", description = "text/pdf; charset=utf-8"),
+                ("Content-Type", description = "application/pdf"),
                 ("Content-Disposition", description = "attachment; filename=\"filename.pdf\"")
             )
         ),
@@ -126,7 +126,7 @@ pub async fn election_download_pdf(
     let disposition_header = format!("attachment; filename=\"{}\"", filename);
 
     let mut headers = HeaderMap::new();
-    headers.insert(header::CONTENT_TYPE, "text/pdf; charset=utf-8".parse()?);
+    headers.insert(header::CONTENT_TYPE, "application/pdf".parse()?);
     headers.insert(header::CONTENT_DISPOSITION, disposition_header.parse()?);
 
     Ok((headers, content.to_vec()))
--- a/backend/src/election/mod.rs
+++ b/backend/src/election/mod.rs
@@ -1,5 +1,6 @@
 use axum::extract::{Path, State};
 use axum::Json;
+use hyper::{header, HeaderMap};
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
@@ -70,12 +71,10 @@ pub async fn election_details(
     Ok(Json(ElectionDetailsResponse { election }))
 }
 
-/// Get election data entry status
 #[utoipa::path(
     get,
-    path = "/api/elections/{election_id}/status",
+    path = "/api/elections/{election_id}/pv",
     responses(
-        (status = 200, description = "Election", body = ElectionStatusResponse),
         (status = 404, description = "Not found", body = ErrorResponse),
         (status = 500, description = "Internal server error", body = ErrorResponse),
     ),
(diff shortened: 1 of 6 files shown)


Lionqueen94 (approved):
Only comment I have is that it would be nice if you also update the frontend openapi file when you update the backend one, to keep them in sync.

backend/src/election/mod.rs
```diff
@@ -90,3 +91,43 @@ pub async fn election_status(
...
+    let _election = elections_repo.get(id).await?;
+
+    // TODO: Replace this with the generated PDF
+    let filename = "proces-verbaal.pdf";
+    let content = &[];
+
+    let disposition_header = format!("attachment; filename=\"{}\"", filename);
+
+    let mut headers = HeaderMap::new();
+    headers.insert(header::CONTENT_TYPE, "text/pdf; charset=utf-8".parse()?);
```
praseodym, line 129: The MIME type for PDF is `application/pdf` without any charset. (See [MDN Common MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types)).

backend/src/lib.rs
```diff
@@ -40,6 +41,10 @@ pub fn router(pool: SqlitePool) -> Result<Router, Box<dyn Error>> {
     let election_routes = Router::new()
         .route("/", get(election::election_list))
         .route("/:election_id", get(election::election_details))
+        .route(
+            "/:election_id/download-pdf",
```
praseodym, line 45: We use snake case for other paths, so:
```suggestion
            "/:election_id/download_pdf",
```

praseodym (changes requested):
Also, could you give the PR a proper title?

backend/src/election/mod.rs
cikzh: [58f0f1e](https://github.com/kiesraad/abacus/commit/58f0f1eee51de154409f9e2784ca8d2e54632c57) addresses this

backend/src/lib.rs
cikzh: [58f0f1e](https://github.com/kiesraad/abacus/commit/58f0f1eee51de154409f9e2784ca8d2e54632c57) addresses this

cikzh:
> Only comment I have is that it would be nice if you also update the frontend openapi file when you update the backend one, to keep them in sync.

Nice one, I thought the commit hook would take care of that, since it does so for the backend. Would be nice to have!
Fixed it in https://github.com/kiesraad/abacus/commit/d7113fc0f81b5942a89176bebe79b5d579db1dcd

backend/src/election/mod.rs
```diff
@@ -90,3 +91,43 @@ pub async fn election_status(
...
+#[utoipa::path(
+    get,
+    path = "/api/elections/{election_id}/download_pdf",
+    responses(
+        (
+            status = 200,
+            description = "PDF",
+            content_type="application/pdf",
+            headers(
+                ("Content-Type", description = "text/pdf; charset=utf-8"),
```
praseodym, line 105: You forgot to change the content type here.

cikzh: d'oh! Fix: [38f4c30](https://github.com/kiesraad/abacus/pull/262/commits/38f4c30153929caafd44a9022ee6ae5d2aec984c)

backend/src/election/mod.rs
```diff
@@ -90,3 +91,43 @@ pub async fn election_status(
...
+#[utoipa::path(
+    get,
+    path = "/api/elections/{election_id}/download_pdf",
+    responses(
+        (
+            status = 200,
+            description = "PDF",
+            content_type="application/pdf",
+            headers(
+                ("Content-Type", description = "application/pdf; charset=utf-8"),
```
praseodym, line 105: There's no charset because PDF is a binary format:
```suggestion
                ("Content-Type", description = "application/pdf"),
```

cikzh: Sorry, I'm doing these corrections too hasty. Fixed in [f643473](https://github.com/kiesraad/abacus/commit/f643473ac5a3411692c4acd3091eafab0ded1458)

praseodym (approved):

backend/src/election/mod.rs
```diff
@@ -90,3 +91,43 @@ pub async fn election_status(
     let statuses = polling_station_repo.status(id).await?;
     Ok(Json(ElectionStatusResponse { statuses }))
 }
+
+/// Download a generated PDF
+#[utoipa::path(
+    get,
+    path = "/api/elections/{election_id}/download_pdf",
```
jschuurk-kr, line 98: I find `download_pdf` a little vague as it describes the format, not the data.

We can change it to `download_results`? ("Certified Election Results" is the translation of "Proces Verbaal" we have in our "woordenlijst". I also suppose they are not actually certified, until the documents are signed?)

Am also fine with changing the endpoint, once we generate the pdf with the results, since for the scope of this PR, `download_pdf` arguably is a better name than `download_results`.

cikzh: We haven't decided yet on the exact API. That will become clear once we develop more of the epic. I've changed anyway in [2fee568](https://github.com/kiesraad/abacus/pull/262/commits/2fee568fb06d7df540929b85831709e53a770496), since it's indeed probably better

Lionqueen94 (approved):

praseodym (approved):

PR #278, opened 2024-08-23
Add no console lint rule

lkleuver:
Add a rule that prevents committing console.log statements (warn and error are allowed)
the rule is only active when running through lefthook but not during development.

### DIFF
--- a/frontend/.eslintrc.cjs
+++ b/frontend/.eslintrc.cjs
@@ -28,6 +28,8 @@ module.exports = {
         allowBoolean: true,
       },
     ],
+    "no-console":
+      process.env.ESLINT_ENV === "production" ? ["error", { allow: ["warn", "error"] }] : "off",
   },
   parserOptions: {
     project: "**/tsconfig.json",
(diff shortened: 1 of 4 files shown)


Lionqueen94 (dismissed):

praseodym:
I don't mind enabling the rule all the time without the 'development/production' difference. 

If we do want to only enable it for 'production' we need to change the GitHub Actions check to also include the new environment variable.

lkleuver:
> I don't mind enabling the rule all the time without the 'development/production' difference.
> 
> If we do want to only enable it for 'production' we need to change the GitHub Actions check to also include the new environment variable.

I quite often use console.log during development (hence this PR) so enabling it in dev gives a constant stream of errors.

praseodym:
I changed it so that `npm run lint` uses `ESLINT_ENV=production`, that should still leave the rule disabled in your IDE (I assume it calls `eslint` directly). This way it is also enabled in GitHub Actions.

Lionqueen94 (approved):

praseodym (approved):

PR #266, opened 2024-08-16
Implement validation copy using unique codes

Lionqueen94:
Closes #136 
Closes #85 
Closes #86
Closes #185 
Closes #150 
Closes #151 
Closes #186 

- [x] Implement changed rules in backend and frontend
- [x] Implement validation copy using unique codes
- [x] Update Feedback component to be able to show multiple errors/warnings and correct action list
- [x] Update Feedback component to match design
- [x] Added W.203 in backend (6cd640099bbb3b6ed32f065f8d47312e7505b16d)
- [x] Added W.204 in backend (6cd640099bbb3b6ed32f065f8d47312e7505b16d)
- [x] Added W.205 in backend (ed9c563a950bc5a3aa40446a795c92c10085542c)
- [x] Added W.206 in backend (9586120efd757fb955c36540135c22ad624a3f98)
- [x] Added W.207 in backend (9586120efd757fb955c36540135c22ad624a3f98)
- [x] Added W.208 in backend (9586120efd757fb955c36540135c22ad624a3f98)
- [x] Added sorting of errors and warnings in backend (22af73923680f9a1fc8646173d28c00c260b8604)
- [x] Improved styling of code in Feedback component (4b39beac03f0d6efee86ec32a73f037cad42cf97)
- [x] Set number of eligible voters on PollingStation 34 to 1000 to be able to test the related warnings
- [x] Implemented changes in https://github.com/kiesraad/abacus-internal/pull/223
- [x] Added tests for W.203-207 in frontend
- [x] Improve d2d tests
- [x] Add links to flowchart doc
- [x] Implement test for F.204 or check if this has been done with the right behaviour
- [x] Duplicate errors and warnings in the frontend https://github.com/kiesraad/abacus/pull/266#pullrequestreview-2277820922 (5195bc56aae21ce38dd796e8e574c3216a181fa2)

Issue #95, opened 2024-06-24
Epic: data entry check and finalisation


praseodym:
Implement according to [Figma section "Controleren en opslaan"](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---URN?node-id=1249-53267&t=aoBOCjLid2UT1ImM-4).

- [x] #133
- [x] #102
- [x] #103
- [x] #104
- [x] #212
- [x] #236
- [x] #265

praseodym:
If #265 is too complicated we can drop it from this epic and implement it later.

PR #281, opened 2024-08-26
Update data entry forms flowchart

praseodym:
Update after discussion on Mattermost.

Includes the update from #266.

### DIFF
--- a/documentatie/flowcharts/data-entry-forms.md
+++ b/documentatie/flowcharts/data-entry-forms.md
@@ -72,16 +72,16 @@ flowchart TD
     render-submitted-data([render with submitted data])
 
     show-error([show error])
-    hide-checkbox-accepted(["hide checkbox warnings accepted"])
+    hide-checkbox-accepted(["hide checkbox \n warnings accepted"])
     show-warning([show warning])
-    show-checked-accepted(["show checked checkbox warnings accepted"])
-    show-unchecked-accepted(["show unchecked checkbox warnings accepted"])
+    show-checked-accepted(["show checked checkbox \n warnings accepted"])
+    show-unchecked-accepted(["show unchecked checkbox \n warnings accepted"])
 
     page-submitted{page submitted?}
     error-cur-page{error for current page?}
     warning-cur-page{warning for current page?}
     cached-input-available{cached input available?}
-    input-changed{"input changed since submit?"}
+    input-changed{"input changed \n since submit?"}
     warning-accepted{"warning(s) accepted?"}
 
     %% flow
@@ -120,9 +120,9 @@ flowchart TD
     go-to-prev-page([go to previous page])
     abort-input([abort input])
 
-    error-any-prev-page{"error for any previous page?"}
-    error-cur-page{"error for current page?"}
-    warning-cur-page{"warning for current page?"}
+    error-any-prev-page{"error for any \n previous page?"}
+    error-cur-page{"error for \n current page?"}
+    warning-cur-page{"warning for \n current page?"}
     warnings-accepted{"warning(s) accepted?"}
     user-addresses-error{user addresses error}
 
--- a/documentatie/flowcharts/data-entry-forms.md
+++ b/documentatie/flowcharts/data-entry-forms.md
@@ -166,20 +166,23 @@ Note that if there is a warning and the user changes the input, they should no l
 
 ## Navigate away from page
 
+Navigating away from a page can happen in several ways:
+- Clicking an item in the left navigation menu
+- Clicking a link in an error or warning message
+- Clicking a link in the top navigation bar
+- Clicking the browser back/forward button
+
+The next page can be either within the form or outside it. The action flow depends on this.
+
 ```mermaid
 flowchart TD
     %% start and end
-    flow-start([start])
-
+    flow-start([navigate action])
     go-to-page([go to selected page])
     remain-on-page([remain on current page])
 
     %% steps
-    nav-item(click navigation item)
-    browser-back("click browser back button")
-    link-in-error-or-warning("click link in error or warning")
-    browser-forward("click browser forward button")
-    navigate-outside-form("navigate outside form")
+    inside-outside{navigating inside form?}
     on-furthest-page{on furthest page?}
     user-made-changes{user made changes?}
     save-changes{save changes?}
@@ -192,19 +195,12 @@ flowchart TD
     save-or-delete{"save or delete?"}
 
     %% flow
-    flow-start --> navigate-outside-form
-    flow-start --> nav-item
-    flow-start --> browser-back
-    flow-start --> browser-forward
-    flow-start --> link-in-error-or-warning
-    nav-item --> on-furthest-page
-    link-in-error-or-warning --> on-furthest-page
-    browser-forward --> on-furthest-page
-    browser-back --> on-furthest-page
+    flow-start --> inside-outside
+    inside-outside -- outside --> save-or-delete
+    inside-outside -- inside --> on-furthest-page
     on-furthest-page -- yes --> cache-input
     cache-input --> go-to-page
     
-    navigate-outside-form --> save-or-delete
     save-or-delete -- save --> call-save-api
     save-or-delete -- delete --> call-delete-api
     call-delete-api --> go-to-page
--- a/documentatie/flowcharts/data-entry-forms.md
+++ b/documentatie/flowcharts/data-entry-forms.md
@@ -66,10 +66,10 @@ flowchart TD
 
     %% elements
     flow-start([start])
+    flow-done([done])
 
     render-empty-fields([render with empty fields])
     render-cached-data([render with cached data])
-    flow-end([end])
 
     render-submitted-data([render with submitted data])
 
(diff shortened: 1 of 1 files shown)


documentatie/flowcharts/data-entry-forms.md
```diff
@@ -61,28 +64,24 @@ flowchart TD
...
-    hide-checkbox-accepted(["hide checkbox
-        warnings accepted"])
+    hide-checkbox-accepted(["hide checkbox warnings accepted"])
     show-warning([show warning])
-    show-checked-accepted(["show checked checkbox
-        warnings accepted"])
-    show-unchecked-accepted(["show unchecked checkbox
-        warnings accepted"])
+    show-checked-accepted(["show checked checkbox warnings accepted"])
+    show-unchecked-accepted(["show unchecked checkbox warnings accepted"])
```
jschuurk-kr, line 78: These were split over two lines, because that carries over into the rendered version, making the containing box narrower and thus the chart more compact. You prefer the wider version?

documentatie/flowcharts/data-entry-forms.md
```diff
@@ -153,67 +148,72 @@ flowchart TD
...
 
 Note that if there is a warning and the user changes the input, they should no longer have the option to accept the warning. They need to click "Next" first, to validate the changed input.
 
+## Navigate away from page
+
+Navigating away from a page can happen in several ways:
+- Clicking an item in the left navigation menu
+- Clicking a link in an error or warning message
+- Clicking a link in the top navigation bar
+- Clicking the browser back/forward button
```
jschuurk-kr, line 173: good improvement!

jschuurk-kr (approved):

documentatie/flowcharts/data-entry-forms.md
praseodym: It didn't seem to make a difference yesterday, but going over it again now it does make a difference.

I've re-added the line breaks with `\n` so that my IDE doesn't complain.

jschuurk-kr (dismissed):

jorisleker (commented):
Not sure about the "outside" arrow from "navigate inside form" to "save or delete" is correct. There are various ways to navigate outside the form, and I think _only_ the `abort input` button should offer user the option to _delete_ their input. If the user uses any other method to navigate outside the form (i.e. going to the overview page or their profile), asking them wether they want to save their (latest) changes would be more friendly.

praseodym:
> Not sure about the "outside" arrow from "navigate inside form" to "save or delete" is correct. There are various ways to navigate outside the form, and I think _only_ the `abort input` button should offer user the option to _delete_ their input. If the user uses any other method to navigate outside the form (i.e. going to the overview page or their profile), asking them wether they want to save their (latest) changes would be more friendly.

If the user finds out that they chose the wrong polling station and use their browser's back button, the abort modal would be the right one to show. In other cases another modal might be more friendly indeed.

jorisleker:
> If the user finds out that they chose the wrong polling station and use their browser's back button, the abort modal would be the right one to show. In other cases another modal might be more friendly indeed.

In the case you describe, I think there would not be any input yet. So nothing to delete? Or do we create a data entry immediately after the user has selected a polling station number?

praseodym:
> > If the user finds out that they chose the wrong polling station and use their browser's back button, the abort modal would be the right one to show. In other cases another modal might be more friendly indeed.
> 
> In the case you describe, I think there would not be any input yet. So nothing to delete? Or do we create a data entry immediately after the user has selected a polling station number?

Right now we don't, but it might be a good idea in the future so that we can do some concurrency control to prevent multiple people from working on the same data entry at the same time.

jorisleker:
> > > If the user finds out that they chose the wrong polling station and use their browser's back button, the abort modal would be the right one to show. In other cases another modal might be more friendly indeed.
> > 
> > 
> > In the case you describe, I think there would not be any input yet. So nothing to delete? Or do we create a data entry immediately after the user has selected a polling station number?
> 
> Right now we don't, but it might be a good idea in the future so that we can do some concurrency control to prevent multiple people from working on the same data entry at the same time.

That's a good one. Just looked at the copy in the design, still think it won't be a good idea to show the 'abort' modal in cases where the user did not click the abort button:

Abort modal:
image

Save modal:
image

jschuurk-kr:
> > > > If the user finds out that they chose the wrong polling station and use their browser's back button, the abort modal would be the right one to show. In other cases another modal might be more friendly indeed.
> > > 
> > > 
> > > In the case you describe, I think there would not be any input yet. So nothing to delete? Or do we create a data entry immediately after the user has selected a polling station number?
> > 
> > 
> > Right now we don't, but it might be a good idea in the future so that we can do some concurrency control to prevent multiple people from working on the same data entry at the same time.
> 
> That's a good one. Just looked at the copy in the design, still think it won't be a good idea to show the 'abort' modal in cases where the user did not click the abort button:

Agree on not allowing the user to delete their input unless they click "invoer afbreken".

That would mean we have four modals in total:
1. abort input
1. navigate outside of form with changes
1. navigate within form with changes
1. fix or ignore errors or warnings after saving changes

As someone who has been clicking through these at an unrealistic speed because doing testing, let's make sure they are distinct enough and clear enough in what they do.

---

This discussion also raises some good questions about concurrency control etc:
- Why would a user navigate to the overview page of their profile in the middle of doing data entry?
- If a user starts data entry for a polling station, is that the only user that can complete the data entry, unless they (or the coordinator) delete it again? Or would it be acceptable that one user does the first half and a second user the second half?

Don't think those questions are that relevant to this discussion about the modals, though.

jorisleker:
> Why would a user navigate to the overview page of their profile in the middle of doing data entry?

That's a valid question. We could simplify everything by removing all navigation options from the top bar and just have it display static data. That would make a choice to leave the form an explicit one because "invoer afbreken" (or the browsers back button from page 1) would be only way to leave.

praseodym:
> > Why would a user navigate to the overview page of their profile in the middle of doing data entry?
> 
> That's a valid question. We could simplify everything by removing all navigation options from the top bar and just have it display static data. That would make a choice to leave the form an explicit one because "invoer afbreken" (or the browsers back button from page 1) would be only way to leave.

It doesn't really matter how many ways there are to navigate away, if there is at least one we need to think about it. Removing the top navigation won't simplify the flowchart, so I'd say we keep it for now. If we later decide that it's too confusing (e.g. from user testing) we can remove it.

Lionqueen94:
@jorisleker Is there a specific reason why it's `invoer bewaren` in one modal and `wijzigingen opslaan` in the other, while the text and the other buttons all say `bewaren` instead of `opslaan`?

jorisleker:
> @jorisleker Is there a specific reason why it's `invoer bewaren` in one modal and `wijzigingen opslaan` in the other, while the text and the other buttons all say `bewaren` instead of `opslaan`? Just something that caught my eye.

Yeah noticed that as well and put it on my todo list to look into. Will problably end up in the design debt ticket. Not broken enough to fix now.

jschuurk-kr:
> > > That would mean we have four modals in total:
> > > 
> > > 1. abort input
> > > 2. navigate outside of form with changes
> > > 3. navigate within form with changes
> > > 4. fix or ignore errors or warnings after saving changes
> > 
> > 
> > I think 2 and 3 are the same modal?
> 
> Agreed. Those are the same modal.

The reason I said they were different, was based on Joris's [earlier comment](https://github.com/kiesraad/abacus/pull/281#pullrequestreview-2262439902):

1. abort input -> save input or delete all input
2. navigate outside of form with changes -> save input (no option to delete all input)
3. navigate within form with changes -> save input or discard changes on current page

But yeah, modal 2 and 3 can be the same one. Or even should be the same one, so users don't have to manually reset their changes on the current page, if they want to navigate away from the form without saving.

jorisleker (approved):

documentatie/flowcharts/data-entry-forms.md
```diff
@@ -153,67 +148,73 @@ flowchart TD
...
-    %% if you can click the browser forward button, you by definition are not on the furthest page
-    click-browser-forward --> user-made-changes
-    click-nav-item --> on-furthest-page
+    flow-start --> inside-outside
+    inside-outside -- outside --> save-or-delete
+    inside-outside -- inside --> on-furthest-page
     on-furthest-page -- yes --> cache-input
     cache-input --> go-to-page
+    
+    save-or-delete -- save --> call-save-api
```
jschuurk-kr, line 205: This part of the flow seems too restrictive. If you want to navigate out of the data entry of a polling station without deleting all the data entry you have done so far, your only option is to save what's in the current form. Even if you want to revert to the latest input you actually submitted. And if there's an error or warning for what's in the current form, you're not allowed to navigate away unless you fix those.

We can revisit this when we pick up epic #95. For this PR I'd leave this part as-is.

documentatie/flowcharts/data-entry-forms.md
```diff
@@ -153,67 +148,73 @@ flowchart TD
...
     on-furthest-page{on furthest page?}
     user-made-changes{user made changes?}
     save-changes{save changes?}
-    call-api(call api)
+    call-save-api(call save api)
+    call-delete-api(call delete api)
     cache-input(cache input)
     reset-changes(reset changes)
     errors-or-warnings{errors or warnings?}
+    modal-fix-or-ignore{fix or ignore?}
```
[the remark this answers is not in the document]
praseodym: I've removed it!

documentatie/flowcharts/data-entry-forms.md
praseodym: This is changed now to be the same modal as navigating inside the form. Only when explicitly clicking the 'abort data entry' button we will have  the option to delete.

praseodym:
I've updated the navigation flowchart. Now the user can leave errors on the page when navigating inside the form. Maybe we need to block navigation when navigation inside the form, clicking 'save' in the modal and then have errors come back?

documentatie/flowcharts/data-entry-forms.md
```diff
@@ -122,27 +120,24 @@ flowchart TD
     go-to-prev-page([go to previous page])
     abort-input([abort input])
```
jorisleker, line 121: Do we need `abort input` in this flowchart? Not really related to clicking volgende? Right now it feels as if aborting input is one of the options presented in the dialogs, which is not the case

jorisleker (changes requested):

documentatie/flowcharts/data-entry-forms.md
praseodym: I agree that doesn't make sense, I've removed it.

jorisleker (approved):

jschuurk-kr (approved):

PR #284, opened 2024-08-27
Some design fixes

Lionqueen94:
## Fixes:
- [x] on `/input` allow polling-station-numbers starting with `0` (i.e. `033`). Strip leading zeros before doing lookup of the polling station number in the database (`002` and `02` should evaluate to `2`)
- [x] Hints on how to submit forms with keyboard have been [redesigned](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=812-20649&m=dev) to better show they are keyboard shortcuts. Implement new design.
- [x] in input-grids, when the contents of a cell wraps over two lines, the row-height increases because the text does not fit into remaining cell height. Fix: 
  - `input-grid td`'s should have `padding:.5rem 1.5 rem` (iso `1rem 1.5rem`) and `line-height:1.25rem` (iso inherited value of `1.5rem`
  - see screenshots. 1 is before fix, 2 is how it should look
Scherm­afbeelding 2024-08-22 om 12 31 39
Scherm­afbeelding 2024-08-22 om 12 31 07


jorisleker (dismissed):
Yes, ziet er goed uit!

jschuurk-kr:
@Lionqueen94 @jorisleker Small thing and up for debate: If I search for polling station "00330", I get the error message "Geen stembureau gevonden met nummer 00330". Arguably we should strip the leading zeroes in the error message, so that it shows the value we are actually looking up.

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
```diff
@@ -166,11 +166,14 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
...
-      <BottomBar type="inputgrid">
+      <BottomBar type="input-grid">
         <Button type="submit" size="lg" disabled={loading}>
           Volgende
         </Button>
-        <span className="button_hint">SHIFT + Enter</span>
+        <span className="button_hint">
+          <Shift />
+          <Enter />
+        </span>
```
praseodym, line 176: What do you think of combining all of this into a single 'keyboard keys' component?

jorisleker: Might be a good idea, I expect there will be more types of keyboard in the future

frontend/lib/ui/KeyboardKey/Enter.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { cn } from "@kiesraad/util";
+
+import cls from "./KeyboardKey.module.css";
+
+export function Enter() {
+  return (
+    <div className={cn(cls["keyboard-key"])}>
+      <span>Enter</span>
+      <IconCornerDownLeft />
+    </div>
```
praseodym, line 11: We should use the [`<kbd>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) here.

Also, did you consider using Unicode symbols instead of an icon? (e.g. <kbd>⇧ Shift</kbd> <kbd>Enter ⏎</kbd>)

Lionqueen94: image
I think having non unicode icons looks a bit better. The one with unicode has the <kbd> element around it.

praseodym: I think we can make the Unicode icons look a bit better still, but let's just stick to the SVG icons instead.

jorisleker: Agreed, let's stick to the svg. 
regarding the `<kbd>` element: semantically better. Can we override the font so it's the same as in design?

Lionqueen94: Yes I have just made the changes, please have a look

jorisleker: Yeah saw your PR. Looks good to me! 👌

Lionqueen94:
> @Lionqueen94 @jorisleker Small thing and up for debate: If I search for polling station "00330", I get the error message "Geen stembureau gevonden met nummer 00330". Arguably we should strip the leading zeroes in the error message, so that it shows the value we are actually looking up.

I think it might be a little confusing for the user to see something else than they entered. @jorisleker what do you think?

jorisleker (approved):

frontend/lib/util/strings.ts
```diff
@@ -10,5 +10,5 @@ export function ellipsis(text: string, maxLength: number = 20): string {
 // returns undefined when non-numeric characters are encountered
 export function parseIntStrict(text: string): number | undefined {
   const num = parseInt(text, 10);
-  return !isNaN(num) && num.toString() === text ? num : undefined;
```
praseodym, line 13: We should add a new function with a new name for this, instead of changing the behaviour of `parseIntStrict` to also strip zeroes.

praseodym (changes requested):

frontend/lib/util/strings.ts
```diff
@@ -10,5 +10,5 @@ export function ellipsis(text: string, maxLength: number = 20): string {
...
+  return !isNaN(num) && num.toString() === text.replace(/^0+/g, "") ? num : undefined;
```
praseodym, line 13: We should add a new function with a new name for this, instead of changing the behaviour of `parseIntStrict` to also strip leading zeroes.

praseodym (changes requested):

jorisleker:
> I think it might be a little confusing for the user to see something else than they entered. @jorisleker what do you think?

No strong opinion on this. Both your arguments are valid. Had the same reaction as @jschuurk-kr when I first saw the feedback, but I think that is because we know how the system actually parses your input. Without knowing that leading zeros are stripped, just feedbacking the initial query is perfectly logical.

One thing that could happen, is that users first try the query with a leading zero, and if that does not give results, is that they retry their query without the zero to see if that does give the desired outcome. So showing the user the actual query that was used could prevent this.

frontend/lib/util/strings.ts
Lionqueen94: @praseodym Well what it does is it already strips the zeroes, because parseInt does this. The only thing I changed was the comparison to the original string, now removing the leading zeroes in the original string. Maybe I can change the name of the function to parsePollingStationNumber, to make it specific to this usage?

praseodym: I would create a new `parsePollingStationNumber` function that does this, and leave `parseIntStrict` as is. We also use `parseIntStrict` for parsing numeric IDs in URLs (`useNumericParam`) and there we do not want to strip leading zeroes.

Lionqueen94: Updated in f834595efba02b693ee02046f91be1a6b9d7a55f

praseodym (dismissed):

jorisleker (dismissed):

praseodym (dismissed):

praseodym (approved):

jorisleker (approved):

Issue #263, opened 2024-08-15
UX/design debt 3


jorisleker:
- [x] Consistent page loading experience. The nav menu and footer should not be visible while content is still loading. → #267
  - On `/overview`, you briefly see the page footer right below the top navigation bar whilst the content of the page is loading. 
  - On `/1/input` the entire page is white whilst page is loading: this is how we want it to be. 
- [x] on `/input` allow polling-station-numbers starting with `0` (i.e. `033`). Strip leading zeros before doing lookup of the polling station number in the database (`002` and `02` should evaluate to `2`) → #284
- [x] Hints on how to submit forms with keyboard have been [redesigned](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=812-20649&m=dev) to better show they are keyboard shortcuts. Implement new design. → #284
- [x] in input-grids, when the contents of a cell wraps over two lines, the row-height increases because the text does not fit into remaining cell height. Fix: → #284
  - `input-grid td`'s should have `padding:.5rem 1.5 rem` (iso `1rem 1.5rem`) and `line-height:1.25rem` (iso inherited value of `1.5rem`
  - see screenshots. 1 is before fix, 2 is how it should look
![Scherm­afbeelding 2024-08-22 om 12 31 39](https://github.com/user-attachments/assets/fb09082b-2196-48bf-935f-830a103f31c6)
![Scherm­afbeelding 2024-08-22 om 12 31 07](https://github.com/user-attachments/assets/3f792a37-2727-4c39-8d07-6e7ae3f68993)
- [x] #346
  - See https://github.com/floriankarsten/space-grotesk/issues/44, what do we do with this?
  - → #337 

After #133 is fixed:
- [x] ~Pressing Enter in the last input row returns focus to the first row, it should move focus to the submit button.~ will be fixed in #207 
- [x] Shift+Enter doesn't work (Safari on macOS). → #295
- [x] ~With the tab-key I can highlight the "volgende" button and then submit by hitting the space bar. I can also highlight the navigation items on the left, but there hitting the space bar does not get me to the item I highlighted.~ No fix needed. Spacebar does not submit actions by default. Only `ENTER` does.
[...]

PR #292, opened 2024-09-03
Custom checkbox

lkleuver:
Change checkbox so it has custom styling to allow for error state.

part of #263


frontend/lib/ui/Checkbox/Checkbox.tsx
```diff
@@ -12,13 +15,26 @@ interface CheckboxProps {
 //TODO: you can't style a border for a checkbox, currently outline is used as a workaround but the border radius doesnt match
```
jorisleker, line 15: I think this comment can be removed?

jorisleker (changes requested):
- The border-color of the unchecked unfocused (not in error state) checkbox is too dark ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2871-9247&node-type=FRAME&m=dev))
- Please use the focus (border+glow) we also use for buttons for the focus state of checkboxes (See [components](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2906-35842&node-type=FRAME&m=dev) in Figma)
- When the warning is shown, the checkbox should have a error/600 red border. Currently it doesn't. ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3626-39567&node-type=FRAME&m=dev))
- The checkmark in design is a bit heavier then the current implementation ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3396-74006&node-type=FRAME&m=dev))
- The label element should have a top and border padding or margin of 1.5rem, so it is properly spaced to the alert and the submit button. 
- On small viewports, the checkbox's width is reduced, and the checkbox isn't square anymore. 

 Scherm­afbeelding 2024-09-03 om 13 04 22
Scherm­afbeelding 2024-09-03 om 13 13 20

frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
```diff
@@ -448,7 +450,9 @@ test.describe("navigation", () => {
         total_votes_cast_count: "100",
       };
       await votersVotesPage.fillInPageAndClickNext(voters, votes);
-      await votersVotesPage.acceptWarnings.click();
+
+      // eslint-disable-next-line playwright/no-force-option -- force option needed to click on hidden element
+      await votersVotesPage.acceptWarnings.click({ force: true, noWaitAfter: true });
```
praseodym, line 455: Maybe we should also using be `.check` here? (cc @jschuurk-kr)

praseodym (changes requested):
Could you remove the commits to `main` that are duplicated in this branch?

lkleuver:
> * The border-color of the unchecked unfocused (not in error state) checkbox is too dark ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2871-9247&node-type=FRAME&m=dev))
> * Please use the focus (border+glow) we also use for buttons for the focus state of checkboxes (See [components](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2906-35842&node-type=FRAME&m=dev) in Figma)
> * When the warning is shown, the checkbox should have a error/600 red border. Currently it doesn't. ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3626-39567&node-type=FRAME&m=dev))
> * The checkmark in design is a bit heavier then the current implementation ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3396-74006&node-type=FRAME&m=dev))
> * The label element should have a top and border padding or margin of 1.5rem, so it is properly spaced to the alert and the submit button.
> * On small viewports, the checkbox's width is reduced, and the checkbox isn't square anymore.
> 

I've think I've addressed all issues, except the not showing the red border, I can't reproduce this.
If you can, could you check if the custom checkbox div has a class "has-error" ?

frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
jschuurk-kr: That would be better. I'd also prefer to have a `checkAcceptWarnings()` method in the page object instead of having to do this everywhere we want to accept warnings.

More importantly, locally the test fails every time at this point for me. Adding a `await expect(votersVotesPage.warning).toBeVisible();` before accepting the warning solved that.

jschuurk-kr:
@lkleuver I don't get the red checkbox on the Voters and Votes page, only on the Differences page.

frontend/lib/ui/Checkbox/Checkbox.module.css
```diff
@@ -2,17 +2,55 @@
...
+    width: 1px;
+  }
+
+  > div {
+    display: inline-block;
     margin-right: 8px;
+    flex: 0 0 20px;
+    width: 20px;
+    height: 20px;
+    border-radius: 6px;
```
Lionqueen94, line 22: Please convert the larger `px` amounts to `rem` (I made a lot of these changes in https://github.com/kiesraad/abacus/pull/303 for the codebase). See also `14px` on line 42.

lkleuver: done

jorisleker (dismissed):

Lionqueen94 (dismissed):

frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
lkleuver: I've added a "checkAcceptWarnings" function to VotersVotesPgObj.ts
interestingly, there doesnt need to be a lint disable warning line here.

jschuurk-kr: That's because the page object files are not included in the eslint `overrides` for Playwright.

Addressed in 772d6e770ea0bc382e08b31bb830e0e26d012c32.

Lionqueen94 (dismissed):

jschuurk-kr (changes requested):
If I check the checkbox, then make a change to the page that also results in a warning (same or different warning) and click "Volgende", the checkbox to accept warnings is already checked.

---

(edit) I wanted to write a test for this, but ran into a problem. I want to select the checkbox with:
```
screen.getByLabelText("Ik heb de aantallen gecontroleerd met het papier en correct overgenomen.")
```
but when I check for the checkbox for not being visible after editing the input, RTL claims the element is visible. I wonder if that's because the `hidden` is on the `<section>`, but that would mean RTL doesn't see children of a hidden parent as hidden, which seems weird.

This is not an issue with Playwright.

jschuurk-kr:
> (edit) I wanted to write a test for this, but ran into a problem. I want to select the checkbox with:
> 
> ```
> screen.getByLabelText("Ik heb de aantallen gecontroleerd met het papier en correct overgenomen.")
> ```
> 
> but when I check for the checkbox for not being visible after editing the input, RTL claims the element is visible. I wonder if that's because the `hidden` is on the `<section>`, but that would mean RTL doesn't see children of a hidden parent as hidden, which seems weird.

:point_up: This does seem to be the case,though. This fails:

```
      await user.clear(screen.getByTestId("blank_votes_count"));
      await user.type(screen.getByTestId("blank_votes_count"), "100");
      await user.tab();
      expect(screen.getByTestId("blank_votes_count"), "100").toHaveValue("100")
      
      await waitFor(() => {
        expect(screen.getByRole("checkbox", {
          name: "Ik heb de aantallen gecontroleerd met het papier en correct overgenomen.",
        })).not.tobeVisible();
      })
```
because according to RTL the element is visible, while the DOM output of the test does contain `<section class="row hidden">`.

lkleuver:
> * it is present in the document
> * it does not have its css property `display` set to `none`
> * it does not have its css property `visibility` set to either `hidden` or `collapse`
> * it does not have its css property `opacity` set to `0`
> * its parent element is also visible (and so on up to the top of the DOM tree)
> * it does not have the [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute
> * if `<details />` it has the `open` attribute

To summarize, styles aren't loaded by our testing library currently, we can either load them or focus RTL tests on checking for the correct classes/attributes.

Currently I've added the hidden attribute to the section that hides the checkbox. This works and circumvents loading styles. We could also check for the presence of "hidden" on the section, but this couples it a lot to the html structure.

jorisleker:
There is still the 'checkbox remains checked' bug. 

Reproduction steps:
- On voters and votes enter some numbers that will result in a warning (ie A100, D100, E90, F10, H100)
- press volgende
- check the 'accept warnings' checkbox, don't press volgende
- change your input to something that will trigger a warning (ie change E94, F6)
- the checkbox disappears
- press volgende. 
- you get a warning, the checkbox reappears and is checked (should be unchecked)

jorisleker (changes requested):
see last comment

lkleuver:
onChange didn't fire, changed it.
Might be a controlled input in the future.

jorisleker (approved):
retested the bug I found earlier, has been resolved!

frontend/app/component/form/voters_and_votes/VotersAndVotesForm.test.tsx
```diff
@@ -368,6 +368,17 @@ describe("Test VotersAndVotesForm", () => {
...
+      await user.clear(screen.getByTestId("blank_votes_count"));
+      await user.type(screen.getByTestId("blank_votes_count"), "100");
+      await user.tab();
+      expect(screen.getByTestId("blank_votes_count"), "100").toHaveValue("100");
+
+      await waitFor(() => {
+        const checkbox = screen.getByTestId("voters_and_votes_form_ignore_warnings");
+        expect(checkbox).toBeInTheDocument();
+        expect(checkbox).not.toBeVisible();
+      });
```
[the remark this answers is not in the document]
lkleuver: I've updated the test

Lionqueen94 (approved):

praseodym (approved):

Issue #285, opened 2024-08-30
Inconsistent link colors and override of visited link color in feedback and alert components


jorisleker:
We've reviewed the colors of our link styles to make them pass accessibility requirements.

The colors should be updated as follows:
- [x] link default: #155EEF 
- [x] link hover: #004EEB
- [x] link active: #0040C1
- [x] link visited: #6927DA

Also see [definition in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3087-58933&node-type=FRAME&m=dev)

The colors apply for all links in texts (not in buttons).

There is one exception. In [feedback](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=136-2286&node-type=FRAME&m=dev) and [alert](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=985-19666&node-type=FRAME&m=dev) components (red/yellow/green or blue background) the _visited_ state of links should use the same color as the default state. Hover and active colors should override the visited color.

Acceptance criteria:
- [x] color of links adhere to scheme defined in this ticket
- [x] style of visited links in `feedback` and `alert` elements is same as non visited links
- [x] hover and active states always override visited state
- [x] style of visited links on `Controleren en opslaan` page is same as non visited links

Example of link in feedback component [in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3085-18418&node-type=FRAME&t=yrnZQIcxc8CMyLZF-11)

jorisleker:
Link colors are updated in https://github.com/kiesraad/abacus/pull/303

PR #303, opened 2024-09-05
Some code cleanup

Lionqueen94:
- [x] Removed unused `form.less` file and config
- [x] Improved imports
- [x] Added missing color variables
- [x] Replaced most px with rem
- [x] Fixed BottomBar button rendering on login and account setup page


praseodym (approved):

frontend/lib/ui/ProgressList/ProgressList.module.css
```diff
@@ -47,7 +47,7 @@
     &:global(.updates) {
       aside {
         svg {
-          fill: var(--menu-updates);
+          fill: var(--link-hover);
```
jorisleker, line 50: Not sure we should use `--link-hover` var here? Semantically, it would be better to keep the `--menu-updates` var here to control the color of this icon? Might want to rename it's name to be a bit more concise.

frontend/lib/ui/style/form.less
```diff
@@ -1,7 +1,7 @@
 form {
   input {
     border: 1px solid var(--bg-gray-darkest);
-    background-color: #fff;
+    background-color: var(--base-white);
     font-size: 16px;
```
jorisleker, line 5: Not part of your change, but do we want font-sizes to be defined in px here? Shouldn't this be `1rem`? (applies to other font-size definitions as well)

frontend/lib/ui/style/variables.css
```diff
@@ -100,7 +102,10 @@
   --blue-light-800: #065986;
   --blue-dark-100: #d1e0ff;
   --blue-dark-400: #528bff;
-  --menu-updates: #004eeb;
+  --link-default: #155eef;
+  --link-hover: #004eeb;
+  --link-active: #0040c1;
+  --link-visited: #6927da;
```
jorisleker, line 108: Great, this resolves most of https://github.com/kiesraad/abacus/issues/285 👍 I've ticked of some of the checkboxes there.

jorisleker (changes requested):

frontend/lib/ui/ProgressList/ProgressList.module.css
Lionqueen94: I can keep 2 variables, pointing to the same color, if that is what we prefer?

jorisleker: yeah think that would be wise here. Easier if we decide to change the color of the icon for some reason.

Lionqueen94: Changed in 3daa4d9e0b473944f6053ce90993cd290237f796

frontend/lib/ui/style/form.less
Lionqueen94: Replaced most px for rem in the whole codebase in 3daa4d9e0b473944f6053ce90993cd290237f796

jorisleker (approved):

praseodym (approved):

lkleuver (approved):
Unsure about using "@" imports within the lib itself to itself. but I can't really think of a reason not to do it 😅 

The UI library is already too coupled to the other libs, in a perfect world it would work standalone.

PR #307, opened 2024-09-05
Integrate pdf poc in backend and add typst template

### LINKED ISSUE #225: Integrate PDF PoC in backend

### LINKED ISSUE #270: Typst template styling
This issue tracks design ToDo's for the Typst templates, generating the PDF documents

- [ ] Number formatting: 1.000.000 instead of 1000000
- [ ] Use DM Sans as font for everything except lists of numbers
- [ ] Use Geist Mono as font for lists of numbers in tables
- [ ] Make sure left and right margins are sufficient for pages with punched holes (at least 16mm, preferably 20mm)
- [ ] Make sure top and bottom margins are sufficient for older printers (at least 10mm, bottom preferable 13mm)
- [ ] Can we replace `SHA-256-Hashcode:` by something more readable and understandable to average users? i.e. `Controlecode telresultaten:`

rnijveld:
Resolves #225 and resolves #270

I've attached the PDF that is generated using the example json input file.

[model-na-31-2.pdf](https://github.com/user-attachments/files/16887675/model-na-31-2.pdf) (outdated - see comments)


backend/src/pdf_gen/models/model_na_31_2.rs
```diff
@@ -0,0 +1,83 @@
...
+            other_explanation_count: SummaryCount::zero(),
+            no_explanation_count: SummaryCount::zero(),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct SummaryCount {
+    pub count: u32,
+    pub polling_stations: Vec<String>,
```
praseodym, line 73: What will this field be used for?

backend/src/pdf_gen/models/model_na_31_2.rs
```diff
@@ -0,0 +1,83 @@
...
+            too_few_ballots_handed_out_count: SummaryCount::zero(),
+            too_many_ballots_handed_out_count: SummaryCount::zero(),
+            other_explanation_count: SummaryCount::zero(),
+            no_explanation_count: SummaryCount::zero(),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct SummaryCount {
```
praseodym, line 71: If `SummaryCount` is the total count for all polling stations in the municipality, maybe `TotalCount` is a better name? Or `SumCount`?

backend/src/pdf_gen/models/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+        Ok(Bytes::from(data.as_bytes()))
+    }
+
+    pub fn from_name_with_input(
+        name: &str,
+        input: &str,
+    ) -> Result<PdfModel, Box<dyn std::error::Error>> {
+        match name {
+            "model-na-31-2" => Ok(Self::ModelNa31_2(serde_json::from_str(input)?)),
+            _ => Err(std::io::Error::new(std::io::ErrorKind::InvalidInput, "Unknown model").into()),
```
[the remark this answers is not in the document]
praseodym: Both `std::error::Error` and `std::io::Error` are used, so importing them as `Error` won't work. I think could import `std::io` so that we can write `io::Error` instead.

backend/templates/model-na-31-2.typ
```diff
@@ -0,0 +1,351 @@
...
+  (
+    ([Aantal geldige stempassen], [A], mono[#input.summary.voters_counts.poll_card_count]),
+    ([Aantal geldige volmachtbewijzen (schriftelijk of via ingevulde stem- of kiezerspas)], [B], mono[#input.summary.voters_counts.proxy_certificate_count]),
+    ([Aantal geldige kiezerspassen], [C], mono[#input.summary.voters_counts.voter_card_count]),
+  ),
+  added_together: [Tel op +],
+  total: ([Het aantal tot de stemming toegelaten kiezers ( A + B + C = D )], [D], mono[#input.summary.voters_counts.total_admitted_voters_count])
+)
+
+= Aantal getelde stembillettenuitgebrachte stemmen zoals vastgesteld door het #is_municipality[gemeentelijk stembureau][stembureau voor het openbaar lichaam]
```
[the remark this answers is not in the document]
rnijveld: I was wondering how much freedom we have in changing the text from these models, I would very much prefer to use the suggestions though if we have that freedom.

backend/src/pdf_gen/models/model_na_31_2.rs
rnijveld: That field is used for the list of polling stations with more/fewer ballots than voters in section 6 of the form. Although the other differences counts don't really need it, I feel it does make sense to keep the type (and thus the eventual addition logic) the same for all these fields.

backend/templates/model-na-31-2.typ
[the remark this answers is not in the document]
Lionqueen94: I got the suggestions from the current model, see https://wetten.overheid.nl/afbeelding?toestandid=BWBR0034180/2023-11-01_0&naam=271034.png&schalen=nee

backend/src/pdf_gen/models/mod.rs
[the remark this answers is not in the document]
Lionqueen94: That would be fine, and then importing InvalidInput fully at the start of the file would also help

backend/templates/model-na-31-2.typ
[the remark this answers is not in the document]
rnijveld: Interesting, I used the file from https://www.rijksoverheid.nl/documenten/publicaties/2022/11/18/model-na-31-2 as my starting point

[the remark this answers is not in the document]
Lionqueen94: What the heck, that's weird. Also this `stembillettenuitgebrachte stemmen` just makes no sense in that text. There should at least be a `/` in between :confused: @jorisleker @praseodym @jschuurk-kr Any idea what file is correct?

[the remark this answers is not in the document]
jschuurk-kr: @arjen-ba What's the best reference/source for the latest version of the "modellen"?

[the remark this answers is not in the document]
Lionqueen94: > Interesting, I used the file from https://www.rijksoverheid.nl/documenten/publicaties/2022/11/18/model-na-31-2 as my starting point

Looks like they just made an error in the PDF, because the ODT is correct compared to the wetten.overheid.nl file and the suggestion I made.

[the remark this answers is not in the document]
arjen-ba: > @arjen-ba What's the best reference/source for the latest version of the "modellen"?

https://www.rijksoverheid.nl/onderwerpen/verkiezingen/documenten/publicaties/2022/11/18/model-na-31-2

(the one dated 29-04-2024 is the correct one, so it's the same one as in this (different) link)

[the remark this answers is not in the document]
jorisleker: > I was wondering how much freedom we have in changing the text from these models, I would very much prefer to use the suggestions though if we have that freedom.

There is an ongoing project to revise these documents, so wouldn't put too much effort into changing things now. There are much bigger changes coming, so as @praseodym also mentioned: no need to be 100% accurate right now. And textual changes/improvements you make will probably become obsolete.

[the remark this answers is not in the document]
rnijveld: I've updated the title here to conform to the odt

backend/src/pdf_gen/models/mod.rs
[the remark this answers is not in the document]
rnijveld: I've used a local use statement to allow importing `std::io::Error` within the scope of this function, while importing `std::error::Error` outside the scope that makes it much more readable.

backend/src/pdf_gen/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+            let buffer: Vec<u8> = typst_pdf::pdf(&document, Smart::Auto, None);
+            println!(
+                "PDF generation took {} ms",
+                pdf_gen_start.elapsed().as_millis()
+            );
+            Ok(buffer)
+        }
+        Err(err) => {
+            eprintln!("{:?}", err);
+            // TODO: Improve this error
```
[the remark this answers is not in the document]
rnijveld: I believe so. I've propagated the errors typst gives us back up, I believe that is the best thing we can do here. At least we now get a logged message with the typst errors if something were to go wrong.

backend/templates/model-na-31-2.typ
```diff
@@ -0,0 +1,351 @@
...
+  #table(
+    columns: (80pt, 1fr, auto),
+    inset: 8pt,
+    stroke: 0.5pt + black,
+    fill: (_, y) => if y > 2 and calc.even(y) { rgb("EAF2F5") },
+    table.header(
+      table.cell(colspan: 3, grid(
+        columns: (auto, auto),
+        gutter: 12pt,
+        [*Lijstnaam*],   [#TODO],
```
rnijveld, line 291: Note that list name and candidate names are currently unavailable, which does result in a large set of TODOs right now, is there something else we could display here instead?

praseodym: Why are they unavailable?

rnijveld: I must be missing something, where would they be stored?

backend/templates/inputs/model-na-31-2.json
```diff
@@ -49,11 +64,15 @@
     "political_group_votes": [
       {
         "number": 1,
-        "total": 0,
+        "total": 30,
```
Lionqueen94, line 67:
```suggestion
        "total": 0,
```
These should be 0 I think, as all values above are also 0.

backend/templates/inputs/model-na-31-2.json
```diff
@@ -49,11 +64,15 @@
...
         "candidate_votes": [
           {
             "number": 1,
             "votes": 20
```
Lionqueen94, line 71:
```suggestion
            "votes": 0
```

rnijveld:
Updated preview PDF for latest version of code: [model-na-31-2.pdf](https://github.com/user-attachments/files/16929018/model-na-31-2.pdf)

backend/templates/inputs/model-na-31-2.json
```diff
@@ -49,11 +64,15 @@
...
+          },
+          {
+            "number": 2,
+            "votes": 10
```
Lionqueen94, line 75:
```suggestion
            "votes": 0
```

backend/templates/model-na-31-2.typ
```diff
@@ -0,0 +1,354 @@
...
+
+#show: doc => conf(input, doc)
+
+#let is_municipality = (municipal, public_body) => if input.election.category == "Municipal" [#municipal] else [#public_body]
+
+#title(
+  [Model Na 31-2],
+  [Proces-verbaal van een #is_municipality[gemeentelijk stembureau][stembureau voor het openbaar lichaam] (centrale stemopneming)],
+  [
+    De verkiezing van de leden van *#input.election.name* \ op *#format_date(input.election.election_date)* \
```
rnijveld, line 13: `election.name` doesn't really match what should be input here, is there something else we could display, or leave it as is?

backend/templates/model-na-31-2.typ
```diff
@@ -0,0 +1,354 @@
...
+    #input.location
```
rnijveld, line 14: There is no location stored in the current datastructure (it currently is just a copy of `election.name`), is there something we could put here instead?

backend/templates/model-na-31-2.typ
```diff
@@ -0,0 +1,354 @@
...
+    In opdracht van het centraal stembureau heeft het #is_municipality[gemeentelijk stembureau][stembureau voor het openbaar lichaam] de in
+    dit proces-verbaal opgenomen aantallen (opnieuw) onderzocht. Dit heeft niet geleid tot wijzigingen.
+  ]
+])
+
+= Zitting en aantal kiesgerechtigden
+
+Het betreft de openbare zitting van het #is_municipality[gemeentelijk stembureau in de gemeente][stembureau voor het openbaar lichaam] *#input.location*.
+
+Kieskring #TODO
```
rnijveld, line 64: Should not be displayed for a municipal election, right?

backend/templates/model-na-31-2.typ
```diff
@@ -0,0 +1,354 @@
...
+
+Datum en tijdstip aanvang zitting: #date_input(date: none, top_label: ("Dag", "Maand", "Jaar")) #time_input(time: none, top_label: "Tijd")
+
+Aantal kiesgerechtigden in #is_municipality[de gemeente][het openbaar lichaam] bedraagt #TODO
```
rnijveld, line 68: We currently don't have this information?

backend/templates/model-na-31-2.typ
```diff
@@ -0,0 +1,354 @@
...
+  ]
+  #block_with_checkbox(checked: false)[
+    In opdracht van het centraal stembureau heeft het #is_municipality[gemeentelijk stembureau][stembureau voor het openbaar lichaam] de in
+    dit proces-verbaal opgenomen aantallen (opnieuw) onderzocht. Dit heeft niet geleid tot wijzigingen.
+  ]
+])
+
+= Zitting en aantal kiesgerechtigden
+
+Het betreft de openbare zitting van het #is_municipality[gemeentelijk stembureau in de gemeente][stembureau voor het openbaar lichaam] *#input.location*.
```
rnijveld, line 62: Again `location` here doesn't really cover this, nor is the election name, is there something specific we could display here?

backend/templates/inputs/model-na-31-2.json
rnijveld: I've updated the voters and votes counts to reflect these, so that we don't just have all zeroes in this example.

backend/templates/model-na-31-2.typ
arjen-ba: correct, not applicable for municipal elections

backend/templates/model-na-31-2.typ
praseodym: I think this should be based on the `election.category`

backend/templates/model-na-31-2.typ
praseodym: I think we should add a new field in the election for this.

backend/templates/model-na-31-2.typ
praseodym: We need to add a new field to the election for the number of voters of the whole election. We do have this for each polling station (`polling_stations.number_of_voters`) but not yet for the election.

backend/templates/model-na-31-2.typ
praseodym: Same as https://github.com/kiesraad/abacus/pull/307#discussion_r1750076260.

backend/templates/model-na-31-2.typ
rnijveld: Would it not make sense to just add up all the number_of_voters for all polling stations then?

rnijveld:
Another updated PDF: [model-na-31-2.pdf](https://github.com/user-attachments/files/16930296/model-na-31-2.pdf)

backend/openapi.json
```diff
@@ -663,6 +663,7 @@
         "required": [
           "id",
           "name",
+          "location",
```
Lionqueen94, line 666: Please update the frontend openapi file as well and make sure all election definitions there are updated as well.

backend/templates/model-na-31-2.typ
praseodym: No, because the number of voters for each polling stations is not a required field.

Lionqueen94 (approved):

praseodym (approved):

PR #311, opened 2024-09-06
More data entry dom-to-db tests

jschuurk-kr:
changes:
- added RTL test for submitting with Shift+Enter (d2d tests should wait until full support for keyboard navigation)
- added three data entry d2d tests:
  - recount, no differences
  - no recount, difference of more ballots counted
  - recount, difference of fewer ballots counted
- updated existing d2d test with no recount, no differences to use more realistic numbers
- added some convenience methods to page objects
- made page object convenience methods to input counts accept `number`s and do the conversion to string
- explicit typing of arguments for page object method calls

### DIFF
--- a/frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
+++ b/frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
@@ -1,11 +1,20 @@
 import { expect, test } from "@playwright/test";
 import { formatNumber } from "e2e-tests/e2e-test-utils";
 import { CandidatesListPage } from "e2e-tests/page-objects/input/CandidatesListPgObj";
-import { DifferencesPage } from "e2e-tests/page-objects/input/DifferencesPgObj";
+import {
+  DifferencesPage,
+  FewerBallotsFields,
+  MoreBallotsFields,
+} from "e2e-tests/page-objects/input/DifferencesPgObj";
 import { InputPage } from "e2e-tests/page-objects/input/InputPgObj";
 import { RecountedPage } from "e2e-tests/page-objects/input/RecountedPgObj";
 import { SaveFormPage } from "e2e-tests/page-objects/input/SaveFormPgObj";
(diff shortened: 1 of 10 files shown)


frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
```diff
@@ -1,76 +1,268 @@
...
+      total_votes_cast_count: 1150,
+    };
+    await votersVotesPage.inputVotesCounts(votes);
     await votersVotesPage.next.click();
 
-    await expect(votersVotesPage.error).toBeHidden();
-    await expect(votersVotesPage.warning).toBeHidden();
+    await expect(votersVotesPage.warning).toContainText("W.203");
+    await expect(votersVotesPage.warning).toContainText(
+      "r is een onverwacht verschil tussen het aantal toegelaten kiezers (A t/m D) en het aantal uitgebrachte stemmen (E t/m H).",
```
Lionqueen94, line 165:
```suggestion
      "Er is een onverwacht verschil tussen het aantal toegelaten kiezers (A t/m D) en het aantal uitgebrachte stemmen (E t/m H).",
```

frontend/e2e-tests/page-objects/input/VotersVotesPgObj.ts
```diff
@@ -2,22 +2,30 @@ import { type Locator, type Page } from "@playwright/test";
...
+  invalid_votes_count: number;
+  total_votes_cast_count: number;
+}
+
+export interface VotersRecounts {
+  poll_card_recount: number;
+  proxy_certificate_recount: number;
+  voter_card_recount: number;
+  total_admitted_voters_recount: number;
 }
```
Lionqueen94, line 24: Can't you import these from `openapi.ts`?

jschuurk-kr: I don't think we should. If at some point the API specs change (however unlikely for these particular fields), that might then break the test code.

Lionqueen94 (approved):

praseodym (approved):

Issue #103, opened 2024-06-26
Frontend: screen to show all errors and warnings


praseodym:
Add check screen according to [Figma section "Controleren en opslaan"](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3725-15701&node-type=SECTION&t=Jo48EtFi2NgcMsKe-11).

Business logic:
- when there are no errors and no not accepted warnings:
  - first item in list is 'Alle optellingen kloppen'
  - last two item in list are 'Er zijn geen blokkerende fouten of waarschuwingen' and 'Je kan de resultaten van dit stembureau opslaan'
- when there are no errors but there are not accepted warnings:
  - First item in list is 'Alle optellingen kloppen'
  - Last item in list is 'Je kan de resultaten van dit stembureau nog niet opslaan'
  - there is no submit button.
- when there are errors:
  -  Last item in list is 'Je kan de resultaten van dit stembureau nog niet opslaan'
  - there is no submit button
 
between the list items mentioned above, we loop through all pages/sections of the form and we
- do not show items that don't have warnings or errors
- if an item contains no/all 0 input we show 'Bij `<step>` zijn geen aantallen ingevoerd'
- if an item contains accepted warnings: '`<step>` heeft geaccepteerde waarschuwingen'
- if an item contains not accepted warnings: 'Controleer waarschuwingen bij `<step>`'
- if an item contains errors: '`<step>` heeft blokkerende fouten' 

`<step>` is a link to the section that contains the error/warning. The link texts is the name of the step as used in the navigation menu, except for political groups. There we only use 'Lijst `<listnumber>`', we leave out the name of the political group

praseodym:
@jorisleker is going to re-check the design for this.

jorisleker:
Done. Updated design and description of the ticket

lkleuver:
Should we perhaps have a slight visual difference in a "list item" that is blocking "saving". Like unaccepted warnings.

jorisleker:
> Should we perhaps have a slight visual difference in a "list item" that is blocking "saving". Like unaccepted warnings.

This crossed my mind too. 'Fix' could be to differentiate between accepted and unaccepted warnings. My idea suggestion be to update _accepted_ warnings to the green checkmark, and only use the triangle+exclamation mark for unaccepted warnings. Line of thought would be that an accepted/checked warning isn't really a warning anymore.

[...]

PR #312, opened 2024-09-06
Implement data entry final save

### LINKED ISSUE #104: Frontend: implement final save
Extend the check screen from #103 to finalise the data entry using the appropriate API endpoint from #102. After successful save, update UI according to [Figma section "Controleren en opslaan"](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---URN?node-id=1249-53267&t=aoBOCjLid2UT1ImM-4).

Make sure it is not allowed to do a finalised data entry again.

Business logic:
- When no errors or not accepted warnings are present, show first bullet: 'Alle optellingen kloppen' and last bullet: 'Je kan de resultaten van dit stembureau opslaan'
- If there are errors or not accepted warnings, do not show the first bullet, and change the last bullet to 'Je kan de resultaten van dit stembureau nog niet opslaan'. And do not show the submit button.
- For each empty step. Show "Op `<step>` zijn geen aantallen ingevoerd". Replace `<step>` with link to that form section. For lists, only show 'Lijst n', not the name of the political group. 
- For each _accepted_ warnings, show '`<step>` heeft geaccepteerde waarschuwingen'. 
- For each _not accepted_ warnings: show 'Controleer waarschuwingen bij `<step>`'

praseodym:
Implement data entry final save. The logic to hide the save ("Opslaan") button is also implemented because it wasn't that complicated. The overview of errors and warnings according to the design will be implemented in #103.

Resolves #104.


frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
```diff
@@ -259,10 +259,14 @@
~    await candidatesListPage_1.fillCandidatesAndTotal([837, 253], 1090);
~    await candidatesListPage_1.next.click();
~
     const saveFormPage = new SaveFormPage(page);
     await expect(saveFormPage.heading).toBeVisible();
 
-    // TODO: extend as part of epic #95: data entry check and finalisation
+    const saveFormPage = new SaveFormPage(page);
```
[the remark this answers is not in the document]
praseodym: Caused by a badly resolved merge conflict 😓

jorisleker (dismissed):

frontend/lib/api/form/pollingstation/PollingStationFormController.tsx
```diff
@@ -446,6 +447,18 @@ export function PollingStationFormController({
     status.current = "deleted";
   };
 
+  const finaliseDataEntry = async () => {
+    status.current = "finalising";
+    const response = await client.postRequest(request_path + "/finalise");
+    status.current = "idle";
+    if (response.status !== ApiResponseStatus.Success) {
+      console.error("Failed to finalise data entry", response);
+      status.current = "idle";
```
Lionqueen94, line 456: This is not needed, since the `status.current = "idle"` is already done on line 453, or that one has to be removed and line 456 kept.

frontend/app/component/form/pollingstation_save/PollingStationSaveForm.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+    // check that the user is navigated back to the input page
+    expect(mockNavigate).toHaveBeenCalledWith("/1/input#data_entry_saved");
+  });
+
+  test("Data entry does not show finalise button with errors", async () => {
+    const formState = structuredClone(defaultFormState);
+    formState.sections.voters_votes_counts.errors = [
+      {
+        code: "F201",
+        fields: ["data.voters_counts.poll_card_count"],
```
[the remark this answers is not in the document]
praseodym: The specific fields don't really matter for this test, but I've applied your suggestions nonetheless.

frontend/lib/api/form/pollingstation/PollingStationFormController.tsx
```diff
@@ -446,6 +447,17 @@ export function PollingStationFormController({
...
 
+  const finaliseDataEntry = async () => {
+    status.current = "finalising";
+    const response = await client.postRequest(request_path + "/finalise");
+    if (response.status !== ApiResponseStatus.Success) {
+      console.error("Failed to finalise data entry", response);
+      status.current = "idle";
+      setError(response.data as ApiResponseErrorData);
+    }
+    status.current = "finalised";
```
Lionqueen94, line 458: Shouldn't this be an explicit `else`? Because the if does not have a return and the state will always be set to `finalised` now I believe.

praseodym: Good catch, this was missing a `throw` that I've added back.

lkleuver (dismissed):

frontend/app/component/form/pollingstation_save/PollingStationSaveForm.test.tsx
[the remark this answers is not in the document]
Lionqueen94: Yes true, but I think it's good to use the codes and fields as consistently as possible to prevent accidental errors where it does matter :)

frontend/app/component/form/pollingstation_save/PollingStationSaveForm.tsx
```diff
@@ -13,29 +18,69 @@ export function PollingStationSaveForm() {
...
-        </div>
-      ))}
-    </div>
+      {finalisationAllowed && (
+        <BottomBar type="input-grid">
+          <BottomBar.Row>
+            <Button type="submit" size="lg" disabled={status.current === "finalising"}>
+              Opslaan
+            </Button>
+            <KeyboardKeys keys={[KeyboardKey.Shift, KeyboardKey.Enter]} />
```
jschuurk-kr, line 80: Submitting with Shift+Enter does not work for me.

frontend/app/component/form/pollingstation_save/PollingStationSaveForm.tsx
```diff
@@ -13,29 +18,69 @@ export function PollingStationSaveForm() {
     });
   }, [registerCurrentForm]);
 
+  const finalisationAllowed = Object.values(formState.sections).every(
```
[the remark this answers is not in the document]
praseodym: It's not recommended to put information in the state that can be easily calculated from other state, see [avoid redundant state](https://react.dev/learn/choosing-the-state-structure#avoid-redundant-state). We also only need this variable in this component.

frontend/app/component/form/pollingstation_save/PollingStationSaveForm.tsx
praseodym: That should be provided by the `Form` component. If that component doesn't work maybe we can fix that in a separate issue, where we can also make sure to write test coverage for the keyboard shortcut?

jschuurk-kr: We do have test coverage for shift+enter for the four forms and those tests pass. So not sure what's different about this form.

praseodym: Ok, then I'll write a test for this form too and will try to solve out the issue.

frontend/app/component/form/pollingstation_save/PollingStationSaveForm.tsx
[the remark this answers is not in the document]
lkleuver: Perhaps the logic can be reusable and testable?
even though it's quite trivial.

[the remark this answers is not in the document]
praseodym: `PollingStationSaveForm.test.tsx` already tests whether the save button is shown only when there are no errors and no unaccepted warnings.

frontend/app/component/form/pollingstation_save/PollingStationSaveForm.tsx
praseodym: Fixed in 4e89eca221471986ddf6d25ff29c8c3033252869 by attaching the keyboard shortcut listener to the document instead of the `<form>` node. @lkleuver what do you think of that solution?

lkleuver: yeah thats fine, it sort of felt safe to have the listener only active when the form is active, but this leads to unexpected behavior if you clicked outside the form and the shortcut no longer works.

Lionqueen94 (approved):

lkleuver (approved):

PR #319, opened 2024-09-09
Test what field is marked with an icon for every rule and check implemented

### LINKED ISSUE #271: Test what field is marked with an icon for every rule and check implemented

Lionqueen94:
Closes #271 

- [x] Added `aria-invalid` to inputs that contain an error or an unaccepted warning
- [x] Added `aria-errormessage` to those inputs, pointing to the element with the feedback message
- [x] Added `aria-label` to the input error and warning icons
- [x] Added `aria-label` to remaining error, warning, success and notifiy icons
- [x] Added helper functions to check validity of fields
- [x] Tested validity and icons of all fields for all warnings and errors
- [x] Renamed `votes_candidates_counts` to `votes_candidates_count` for consistency of fieldnames


Lionqueen94:
@jorisleker Do we want the same aria-labels (bevat een fout and bevat een waarschuwing) on the error icon on the polling station selector page as well? And what about the icons on the Feedback component and on the Alert component?

jorisleker:
> @jorisleker Do we want the same aria-labels (bevat een fout and bevat een waarschuwing) on the error icon on the polling station selector page as well? And what about the icons on the Feedback component and on the Alert component?

@Lionqueen94 Yes, I think we'd want that!

frontend/app/component/form/testHelperFunctions.tsx
```diff
@@ -0,0 +1,31 @@
...
+) {
+  fields.forEach((field) => {
+    expect(field).toBeInvalid();
+    expect(field).toHaveAccessibleErrorMessage(feedbackMessage);
+  });
+}
+
+export function expectFieldsToHaveIconAndToHaveAccessibleName(fields: Array<HTMLElement>, accessibleName: string) {
+  fields.forEach((field) => {
+    expect(within(field.previousElementSibling as HTMLElement).getByRole("img")).toHaveAccessibleName(accessibleName);
```
jschuurk-kr, line 16: Am not happy about the `previousElementSibling`. Am looking into alternative solutions, might need some changes to `InputGridRow` or `FormField`.

jschuurk-kr: React Testing Library and tables are not the easiest combination. Would be nice if you could select a cell based on `rowheader` and `columnheader`, but alas.

Best solution I can come up with is to put an id on the table cell (`cell-<input field id>`), similar to the `row-<input-field-id>` we have already. If we put the `id`s in some `const`s it should be maintainable.

It's a straightforward refactor but will take some effort. So that raises the question, considering the other open PRs: should we do that in this PR or in a next one? As long as the DOM structure does not change, the current solution works. So we could merge this PR and pick up the refactor next week.

jschuurk-kr: Discussed in standup. This PR can be merged. I've created a new issue https://github.com/kiesraad/abacus/issues/329 for improving the locator, to be picked up next week.

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
```diff
@@ -146,6 +146,7 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
                 id={`candidate_votes[${candidate.number - 1}].votes`}
                 title={`${candidate.last_name}, ${candidate.initials} (${candidate.first_name})`}
                 errorsAndWarnings={isSaved ? errorsAndWarnings : undefined}
+                warningsAccepted={getIgnoreWarnings()}
```
jschuurk-kr, line 149: It's good we're consistent about the CandidatesVotesForm supporting warnings, even though the backend currently does not have any warnings for this form implemented.

However, if we support the functionality in the frontend, we should have some tests for it. The alternative would be to not have the warnings functionality for this form and throw an error if one does show up in the response from the backend.

Lionqueen94: Added in cdd913115d1429b3f1c71bbd65829f67d5013f67

jorisleker (changes requested):
A lot of renames not related to the scope of the ticket in this PR. Maybe put those in a separate PR next time? Makes it easier to review. 

Added suggestions for the aria-labels in Alerts, because realized the context of use for those icons is a bit different then in the form navigation and check and save page. They do not refer to another page, but to something that's happening on the current page.

frontend/app/component/form/differences/DifferencesForm.tsx
```diff
@@ -141,6 +141,7 @@ export function DifferencesForm() {
             id="more_ballots_count"
             title="Stembiljetten méér geteld"
             errorsAndWarnings={isSaved ? errorsAndWarnings : undefined}
+            warningsAccepted={getIgnoreWarnings()}
```
lkleuver, line 144: For the future, maybe this is a good idea?
```
const defaultProps = { warningsAccepted: getIgnoreWarnings(), errorsAndWarnings: isSaved ? errorsAndWarnings : undefined, format};

<InputGridRow {...defaultProps />
```

lkleuver (dismissed):

frontend/app/component/form/differences/DifferencesForm.tsx
Lionqueen94: Good suggestion, implemented in c7faface31a7468c5ae56be5091677ca00e44261

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
```diff
@@ -125,6 +125,13 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
   const hasValidationError = errors.length > 0;
   const hasValidationWarning = warnings.length > 0;
 
+  const defaultProps = {
+    errorsAndWarnings: isSaved ? errorsAndWarnings : undefined,
+    warningsAccepted: getIgnoreWarnings(),
+    inputProps: register(),
```
lkleuver, line 131: was doubting if this would work with register as well but if all tests run :D

lkleuver (approved):

frontend/app/component/form/candidates_votes_form/CandidatesVotesForm.tsx
Lionqueen94: Yeah I just tried it and it worked haha

praseodym (approved):

PR #324, opened 2024-09-10
Sum up all the votes and voters for the entire election for PDF

### LINKED ISSUE #224: Combine input for Typst PDF generation in backend

rnijveld:
Fixes #224 

The tests right now are relatively barebones, but at least we test some successful additions. I would suggest to at add much more significant testing as an issue for a later time (soonish)


backend/src/pdf_gen/models/model_na_31_2.rs
```diff
@@ -41,6 +45,66 @@ impl ModelNa31_2Summary {
...
+                        })
+                        .collect(),
+                });
+            }
+        }
+
+        // loop over results and add them to the running total
+        for (polling_station, result) in results {
+            // add voters and votes to the total
+            totals.voters_counts += &result.voters_counts;
```
praseodym, line 77: If recounted, we need to use the recounted voters counts instead.

praseodym (changes requested):

backend/src/pdf_gen/models/model_na_31_2.rs
jorisleker: Adding up results for the entire election is the core functionality of our software. 

Feels a bit odd to do the addition inside `/pfd_gen` and inside a specific model. Shouldn't this be in a dedicated module (that is also reused for generating eml's?

praseodym: Agreed! I don't think its necessary to move the code at this point already, but we should keep it in mind.

jorisleker: Maybe put in into a ticket, so we don't forget?

jschuurk-kr: I'd hope/expect that as soon as we do a calculation that's not for an Na31-2, moving that code is the obvious choice, without needing a ticket to remember.

jschuurk-kr:
When I run it, pdf generation prints some debugging info: the data, some durations, and the number of warnings. Do we want to keep that for now?

(note: the relevant code is not in scope of this PR, it was added in [PR #307](https://github.com/kiesraad/abacus/pull/307))

backend/src/polling_station/structs.rs
```diff
@@ -389,6 +397,15 @@ pub struct VotersCounts {
...
 
+impl AddAssign<&VotersCounts> for VotersCounts {
+    fn add_assign(&mut self, other: &Self) {
+        self.poll_card_count += other.poll_card_count;
+        self.proxy_certificate_count += other.proxy_certificate_count;
+        self.voter_card_count += other.voter_card_count;
+        self.total_admitted_voters_count += other.total_admitted_voters_count;
+    }
+}
+
```
Lionqueen94, line 408: There needs to be an implementation for adding votersRecounts to VotersCounts as well

backend/src/pdf_gen/models/model_na_31_2.rs
jorisleker: 🤞

rnijveld: Moving it to a more generic place comes naturally as more functionality is added I would say. As a part of KISS (and for the little time we currently have) I would try and prevent doing more than is necessary for the functionality we're currently adding. I'd imagine that the EML output and adding other models will have other consequences that we currently can't foresee anyway.

backend/src/pdf_gen/models/model_na_31_2.rs
```diff
@@ -41,6 +45,66 @@ impl ModelNa31_2Summary {
...
+            totals
+                .differences_counts
+                .add_polling_station_results(polling_station, &result.differences_counts);
+
+            // if this polling station was recounted, add it to the list
+            if result.recounted {
+                totals
+                    .recounted_polling_stations
+                    .push(polling_station.number);
+            }
```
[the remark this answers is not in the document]
praseodym: It is used in the report:
Screenshot 2024-09-12 at 11 40 14

jorisleker (approved):
Tested, download now works on macos in chrome, safari and firefox

backend/src/pdf_gen/models/model_na_31_2.rs
Lionqueen94: I see what you mean @rnijveld , but the section for recounts and differences is unfortunately the same, see:
Screenshot from 2024-09-12 12-27-09
Screenshot from 2024-09-12 12-27-26

We could add some more information in the comments though, what do you think @praseodym @jschuurk-kr ?

praseodym: Improved in 3ddc7e0de5c21401cb8c36991b427a1609c37e93, let me know what you think!

backend/src/pdf_gen/models/model_na_31_2.rs
```diff
@@ -81,4 +170,280 @@ impl SumCount {
...
+
+    pub fn add(&mut self, polling_station: &PollingStation, count: Count) {
+        if count > 0 {
+            self.count += count;
+            self.polling_stations.push(polling_station.number);
+        }
+    }
+}
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
praseodym: Added in 3ddc7e0de5c21401cb8c36991b427a1609c37e93.

Lionqueen94 (approved):

backend/src/pdf_gen/models/model_na_31_2.rs
rnijveld: Looks good to me! <sub>I can't give approval on the PR though because I'm the author of it</sub>

Lionqueen94 (approved):

praseodym (approved):

PR #325, opened 2024-09-11
Fix handling of API errors

### LINKED ISSUE #321: Errors from API are not shown correctly in frontend
Issues with how they are shown:
1. Both the 4xx and 5xx range are showns as "Server error".
2. The specifics of the error response are not shown on the screen, only the `:` is shown.

There are two options to solve issue 1:
- distinguish Client errors from Server errors
- call them all Errors, because the distinction is not helpful to our users

image

We should change the API client code to return a better error object (maybe something like the [Rust `Result` type but in TypeScript](https://imhoff.blog/posts/using-results-in-typescript)?). The `Feedback` component should render this correctly, preferably by passing down the error object in a new prop instead of using the current `data` prop for both validation results (errors and warnings) and API errors.

praseodym:
Fixes #321


Lionqueen94 (dismissed):

frontend/app/component/form/pollingstation_save/PollingStationSaveForm.tsx
jschuurk-kr, line 1: The "Controleren en opslaan" page does not scroll up automatically when there's an API error. This behavior is relevant if there are enough lists with zero votes that you do need to scroll down to the "Opslaan" button.

frontend/app/component/pollingstation/PollingStationFormNavigation.tsx
```diff
@@ -152,7 +152,7 @@ export function PollingStationFormNavigation({ pollingStationId, election }: Pol
         </Modal>
       )}
 
-      {error && <Feedback id="feedback-server-error" type="error" data={error} />}
+      {apiError && <Feedback id="feedback-server-error" type="error" apiError={apiError} />}
```
jschuurk-kr, line 155: It would be nice to have some whitespace below this error message.
image

jschuurk-kr, line 155: Once this kind of error is shown, it's shown on all pages. It also does not go away after a succesful submit.

Tested by making the finalise endpoint return a 500 response.

frontend/app/component/form/pollingstation_save/PollingStationSaveForm.tsx
praseodym: Fixed in 28bbd0eb6e08faade61173208947eb0a39df4344.

frontend/app/component/pollingstation/PollingStationFormNavigation.tsx
praseodym: Fixed in 0c1467b884c6e32855ef6188c5cf8a793ed93650.

jschuurk-kr (approved):

Issue #318, opened 2024-09-09
Add ability to reset database in dom-to-database tests


praseodym:
In #312 we implement finalisation of data entries, but this can only be done once for each polling station. To be able to run our dom-to-database tests for this we should be able to reset the database so that tests can run independently without being affected by data from other tests.

Playwright is able to [send API requests](https://playwright.dev/docs/api-testing). We should use [fixtures](https://playwright.dev/docs/test-fixtures) for setup/teardown.

PR #326, opened 2024-09-11
Prevent new data entry after finalisation

### LINKED ISSUE #316: Do not allow data entry save when already finalised
When the data entry for a polling stations is fully completed, we shouldn't allow new data entries to be saved for that polling station.

praseodym:
Resolves #316.


praseodym:
d2d tests now fail because they expect to be able to do conflicting saves. We need to reset the database between tests (#318) to fix this.

praseodym:
> d2d tests now fail because they expect to be able to do conflicting saves. We need to reset the database between tests (#318) to fix this.

Resolved by #327, so this PR is ready for review now.

Lionqueen94 (commented):
Can you add a test for the new functionality?

praseodym:
> Can you add a test for the new functionality?

Added in f1d02ff067fbfdd230b875b0af666f855db4a481.

Lionqueen94 (approved):

rnijveld (approved):

Issue #207, opened 2024-08-02
Improve keyboard navigation on list data entry page


jorisleker:
Right now we offer some keyboard navigation on the list page: 
- down arrow, enter and tab navigates to next field
- up arrow and shift-tab navigate to previous field
- shift + enter submits the form

Right now, you either have to use your mouse or repeatedly use the down/enter/tab key to get where you want to be. This is sometimes counter intuitive: ie to skip 4 rows, you need 4+1 keypresses.

Also the enter key always moves you to the next interactive element, unless you are at the bottom of the form, then it moves you back to the top.

We can improve keyboard navigation and reduce keyboard/mouse changes by adding some keyboard shortcuts:
- [x] `SHIFT` + `DOWN` arrow: navigate to last input on page (the totals field) 
- [x] `SHIFT` + `UP` arrow: navigate to first input on page (the first candidate)
- [x] `ENTER` from last input field moves focus to submit button. (closed duplicate #306)

## To do:
- [x] implement keyboard shortcuts on list data-entry pages
- [x] update design of bottombar on list data-entry pages:
  - [x] add text describing shift + down new keyboard-shortcut, use these texts:
    - on lists: 'Snel naar totaal van de lijst:'
    - on other forms ~'Snel naar het onderste invoerveld:'~ no hint

[Design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3728-23145&t=nFxASayzDUpV6xHl-11)

related: #342

PR #331, opened 2024-09-13
add shortcut to move to last input

lkleuver:
A tiny experiment introducing shift+arrow-down to move to the last input field (often "total")


jorisleker:
related: https://github.com/kiesraad/abacus/issues/207

jorisleker (approved):
Nice @lkleuver. I had similar ideas, see #207, which adds some more keyboard interactions and some instructions to make them easier to discover.

praseodym (approved):

PR #345, opened 2024-09-16
Renamed routes and renamed some files and directories accordingly

### LINKED ISSUE #184: Adjust routes in frontend to be consistent with the backend and adjust folder structure accordingly
Adjust routes in frontend to be consistent with the backend and adjust folder structure and component names accordingly.

See discussion in: https://github.com/kiesraad/abacus/issues/133

Proposed changes:
- `/1/input/1/numbers` -> `/1/data-entry/1/voters-and-votes`
- `:electionId/input` -> `elections/:electionId/data-entry`

Lionqueen94:
Closes #184


praseodym:
Could you rename 'input' to 'data entry'? (When 'input' does not relate to HTML `<input>` fields)

Lionqueen94:
> Could you rename 'input' to 'data entry'? (When 'input' does not relate to HTML `<input>` fields)

Do we want to rename this in the URL as well?

praseodym:
> > Could you rename 'input' to 'data entry'? (When 'input' does not relate to HTML `<input>` fields)
> 
> @praseodym Do we want to rename this in the URL as well?

Yes please! That would match the related API requests.

Lionqueen94:
> > > Could you rename 'input' to 'data entry'? (When 'input' does not relate to HTML `<input>` fields)
> > 
> > 
> > @praseodym Do we want to rename this in the URL as well?
> 
> Yes please! That would match the related API requests.

@praseodym Done! 1 question, I now used dashes in the urls because that is what the issue stated, but we use underscores in the API urls, shall I change it to underscores in the frontend urls too? So `data-entry` -> `data_entry` and `voters-and-votes` -> `voters_and_votes`?

jorisleker:
> @praseodym Done! 1 question, I now used dashes in the urls because that is what the issue stated, but we use underscores in the API urls, shall I change it to underscores in the frontend urls too? So `data-entry` -> `data_entry` and `voters-and-votes` -> `voters_and_votes`?

For user facing urls, it's [best practice to use dashes instead of underscores](https://developers.google.com/search/docs/crawling-indexing/url-structure). Better readability (and better for SEO, but that's not an issue here ;-). Still, if we can, would prefer `voters-and-votes` over `voters_and_votes`.

Lionqueen94:
> > @praseodym Done! 1 question, I now used dashes in the urls because that is what the issue stated, but we use underscores in the API urls, shall I change it to underscores in the frontend urls too? So `data-entry` -> `data_entry` and `voters-and-votes` -> `voters_and_votes`?
> 
> For user facing urls, it's [best practice to use dashes instead of underscores](https://developers.google.com/search/docs/crawling-indexing/url-structure). Better readability (and better for SEO, but that's not an issue here ;-)). Still, if we can, would prefer `voters-and-votes` over `voters_and_votes`.

Then I suggest renaming `/elections/${election.id}/data-entry#data_entry_saved` to `/elections/${election.id}/data-entry#data-entry-saved` as well, then there are no underscores in the user facing urls. Agreed?

frontend/e2e-tests/page-objects/data_entry/DataEntryPgObj.ts
jschuurk-kr, line 1: I think we should rename the file and the class to `PollingStationsPgObj` (or something similar) to avoid confusion about the relation between `DataEntryBasePgObj` and `DataEntryPgObj`.

frontend/e2e-tests/page-objects/data_entry/DataEntryBasePgObj.ts
jschuurk-kr, line 1: Class should also be renamed to `DataEntryBasePage`.

Lionqueen94: Done, thanks for spotting!

frontend/e2e-tests/page-objects/data_entry/DataEntryPgObj.ts
Lionqueen94: Done, also renamed everything that was not in line with the renaming done in this PR

jschuurk-kr (approved):

jschuurk-kr (approved):

Issue #243, opened 2024-08-07
Implement polling station status behavior


cikzh:
When the first input of a polling station is finalized, you shouldn't be able to input again

- Throw an error when the user tries to access the input page
- Don't allow the user to select the polling station in the polling station choice page

Issue #336, opened 2024-09-16
Finalise button ("Invoerfase afronden") only appears after reload


praseodym:
After completing the data entry for the last polling station, the finalise button ("Invoerfase afronden") is not shown right away. It only appears after a reload of the polling station choice page.

Same is happening with the status badges of the polling stations (see #347).

PR #347, opened 2024-09-17
Added status badge

cikzh:
Added the status badges to the frontend where applicable and added a status field to the Polling Station types.
Also added (and renamed) the different status types so it matches the Figma design and the frontend Badge component. For now I've refrained from implementing the determination of the statuses we don't yet use, to keep the PR small.

Frontend behavior of the statuses will be implemented in #243


Lionqueen94 (approved):

backend/src/polling_station/structs.rs
```diff
@@ -50,6 +50,7 @@ pub struct PollingStation {
     pub house_number_addition: Option<String>,
     pub postal_code: String,
     pub locality: String,
+    pub status: PollingStationStatus,
```
praseodym, line 53: Why extend this api endpoint with a status field? We already have a separate endpoint to return the status for each polling station, and that endpoint is more suitable for refreshing often. This endpoint currently returns very static data and I don't see a good reason to add the relatively dynamic status field to that.

cikzh: It made the frontend implementation a lot easier. I hadn't considered the staticness of the struct as it was.

backend/src/polling_station/structs.rs
```diff
@@ -370,10 +370,16 @@ pub struct PollingStationStatusEntry {
...
 pub enum PollingStationStatus {
-    Incomplete,
-    Complete,
+    Correction,
+    Definitive,
+    Difference,
+    ExtraEntry,
+    FirstEntry,
+    Objections,
+    SecondEntry,
```
jschuurk-kr, line 382: I'd prefer limiting the statuses to the ones that are relavant to / reachable by what we've built so far.

praseodym (approved):

backend/src/polling_station/structs.rs
praseodym: Thanks for removing it! I think the frontend implementation wasn't too complicated after all? And this way, reloading the statuses (#336) should be a bit more lightweight.

Lionqueen94 (approved):

jschuurk-kr:
FYI I ran into issue #336 while testing this, i.e. you need to refresh the page to see the completed status of the polling station you just filled in. Will leave a comment on that issue, so we make sure it's fixed alongside showing the finalise button immediately.

backend/src/polling_station/structs.rs
cikzh: Why? They were already present in the frontend as labels for `<Badge>`. Now we have it concisely typed throughout

praseodym:
> FYI I ran into issue #336 while testing this, i.e. you need to refresh the page to see the completed status of the polling station you just filled in. Will leave a comment on that issue, so we make sure it's fixed alongside showing the finalise button immediately.

Yes that's the same issue, the status currently only gets loaded once and never gets reloaded. That causes the finalise button issue but also stale status badges.

backend/src/polling_station/structs.rs
jschuurk-kr: Agree that it's better to have it concisely typed throughout, but 5 out of those 7 statuses are dead code. And our API specs are now incorrect, since they list those 5 statuses.

This (including that `<Badge>` existed at all) ties back into the bigger and older discussion of how we built UI placeholders. That's not a discussion to drag into this PR, though.

My prefered solution would be to clean up the statuses throughout the whole stack, reducing them to the ones supported by the functionality, so `FirstEntry` and `Definitive`.

praseodym (dismissed):

praseodym (approved):

PR #349, opened 2024-09-17
Re-fetch statuses on opening data entry home page

### LINKED ISSUE #336: Finalise button ("Invoerfase afronden") only appears after reload

praseodym:
Fixes #336

### DIFF
--- a/frontend/lib/api/election/ElectionStatusProvider.tsx
+++ b/frontend/lib/api/election/ElectionStatusProvider.tsx
@@ -4,6 +4,7 @@ import { PollingStationStatusEntry, useElectionStatusRequest } from "@kiesraad/a
 
 export interface iElectionStatusProviderContext {
   statuses: Required<PollingStationStatusEntry[]>;
+  refetch: () => void;
 }
 
 export const ElectionStatusProviderContext = React.createContext<iElectionStatusProviderContext | undefined>(undefined);
@@ -14,20 +15,20 @@ export interface ElectionStatusProviderProps {
 }
 
 export function ElectionStatusProvider({ children, electionId }: ElectionStatusProviderProps) {
-  const { data, loading, error } = useElectionStatusRequest({
+  const { data, error, refetch } = useElectionStatusRequest({
     election_id: electionId,
   });
 
-  if (loading) {
+  if (data === null) {
     return null;
   }
 
-  if (!data || error) {
-    throw new Error();
+  if (error) {
+    throw new Error("Could not fetch election statuses");
   }
 
   return (
-    <ElectionStatusProviderContext.Provider value={{ statuses: data.statuses }}>
+    <ElectionStatusProviderContext.Provider value={{ statuses: data.statuses, refetch }}>
       {children}
     </ElectionStatusProviderContext.Provider>
   );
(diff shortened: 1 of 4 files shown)


jschuurk-kr:
When I go to `/overview` and click on one of the elections, the call to get the status of the election I selected (e.g. `/elections/2/status`) is made twice.

praseodym:
> When I go to `/overview` and click on one of the elections, the call to get the status of the election I selected (e.g. `/elections/2/status`) is made twice.

Indeed, but unfortunately it's not that easy to fix. The `DataEntryHomePage` component only gets rendered once the initial statuses request completes, and then immediately calls `refetch()` to update the statuses with the second statuses request. 

If we don't block the `DataEntryHomePage` render we can let that do the initial load and prevent a second request, but it will cause the footer to be already rendered before the request completes. I think that's not trivial to fix: it would require something like [`<Suspense>`](https://react.dev/reference/react/Suspense) which is only usable with a query framework (such as [TanStack Query](https://tanstack.com/query/latest/docs/framework/react/guides/suspense)) because there is no official API yet (see ['Note' block in Suspense docs](https://react.dev/reference/react/Suspense#displaying-a-fallback-while-content-is-loading)).

jschuurk-kr:
> > When I go to `/overview` and click on one of the elections, the call to get the status of the election I selected (e.g. `/elections/2/status`) is made twice.
> 
> Indeed, but unfortunately it's not that easy to fix. [\<snip\>](https://github.com/kiesraad/abacus/pull/349#issuecomment-2357769113)

Fixing it sounds worse than leaving as it is. So works for me to keep the two calls.

jschuurk-kr (dismissed):

Lionqueen94:
I see I forgot to rename renderInputHomePage, @praseodym can you do this for me? 😇

praseodym:
> I see I forgot to rename renderInputHomePage, @praseodym can you do this for me? 😇

Done: be66bb13d58cb97eca6f953dbc87753cd51a9d73

jschuurk-kr (approved):

Lionqueen94 (approved):

frontend/lib/api/election/ElectionStatusProvider.tsx
```diff
@@ -14,20 +15,20 @@ export interface ElectionStatusProviderProps {
...
   });
 
-  if (loading) {
+  if (data === null) {
     return null;
   }
 
-  if (!data || error) {
-    throw new Error();
+  if (error) {
```
lkleuver, line 26: shouldn't this be defined above `if (data === null) return null;` ? (just in case)

lkleuver (approved):

frontend/lib/api/election/ElectionStatusProvider.tsx
praseodym: I'll create a PR to fix that!

praseodym: #354

PR #352, opened 2024-09-18
Add Git commit and branch details to footer

### LINKED ISSUE #328: Show branch name and commit hash in frontend
Would be nice if the frontend shows which version of Abacus you are interacting with. 

Could we add branch name and commit hash to the footer? See screenshot for how/where to put this

- branch name, bottom left. Branch names can be rather long, should be truncated with `&hellip;` if longer then say 32 chars.
- commit hash, bottom right. I've put it after the version num now, but we could also consider replacing the version num, since that is hardcoded now. If the commit is tagged (i.e. official release), we could show the tag instead of the commit hash, but think that would be something we could implement later.

image

praseodym:
Add Git commit and branch details to footer, but only for Vite builds. During Vite serve (e.g. `npm run dev`) the Git tree might change without a chance to reload the variables, and we do not want to show stale details.

The second commit cleans up up Vite `define` usage to use global variables as used in the Vite docs:
https://vitejs.dev/config/shared-options.html#define

It also fixes the `host` field in `ApiClient` being unused.

Closes #328.

### DIFF
--- a/frontend/app/component/footer/Footer.tsx
+++ b/frontend/app/component/footer/Footer.tsx
@@ -1,10 +1,36 @@
 export function Footer() {
+  const gitBranch = import.meta.env.GIT_BRANCH as string | undefined;
+  let gitBranchShort = gitBranch;
+  if (gitBranch && gitBranch.length > 32) {
+    gitBranchShort = gitBranch.substring(0, 32) + "…";
+  }
+  const gitCommit = import.meta.env.GIT_COMMIT as string | undefined;
+  const gitDirty = import.meta.env.GIT_DIRTY as boolean;
+  let mode = (import.meta.env.MODE as string | undefined) || "unknown";
+  mode = (mode[0]?.toUpperCase() ?? "") + mode.slice(1);
+
   return (
     <footer>
-      <section>Abacus</section>
       <section>
-        <strong>Server</strong> {process.env.MSW ? "Mocked" : "Live"} &nbsp;&nbsp;
-        <strong>Versie</strong> v{process.env.VERSION}
+        Abacus{" "}
+        {gitBranch && (
+          <>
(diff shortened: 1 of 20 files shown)


jorisleker:
Don't have local instance on this machine, so testing with the deployment on github pages. Branch name looks ok, but commit hash I am seeing on GitHub does not match the hash shown in the footer? 

Scherm­afbeelding 2024-09-19 om 07 42 27
Scherm­afbeelding 2024-09-19 om 07 42 42

praseodym:
That's because GitHub Actions is configured to build the PR branch as if it was merged with `main`. If you click the commit id you'll see that and can also see the commit id from the branch.

jschuurk-kr (commented):
Tried out the preview and running locally, inc. local commit and dirty state.

Found one thing: if you have an un-pushed commit, the GitHub link to the commit 404s (for obvious reasons). `git log @{u}.. --oneline` shows unpushed commits, so we could use that to do something similar to the dirty state for this case.
Or we leave this as is, since it should be clear enough why it 404s. If we do fix the 404, we should also account for unpushed commits plus dirty state. And before you know it, we're implementing a full git prompt into our app.

praseodym:
> Found one thing: if you have an un-pushed commit, the GitHub link to the commit 404s (for obvious reasons). `git log @{u}.. --oneline` shows unpushed commits, so we could use that to do something similar to the dirty state for this case. Or we leave this as is, since it should be clear enough why it 404s. If we do fix the 404, we should also account for unpushed commits plus dirty state. And before you know it, we're implementing a full git prompt into our app.

I think that's going to make it a bit too complicated. I'm also not sure if there are edge cases that would cause this command to wrongly report commits as not pushed (e.g. in GitHub Actions?). And a commit can always be pushed to GitHub a little later, which would make the link work from then on.

jschuurk-kr (approved):

Lionqueen94 (approved):

PR #353, opened 2024-09-19
Add NumberInput component replacing usePositiveNumberInputMask

### LINKED ISSUE #301: Cursor moves to last char when editing numbers >= 1000
### What's the problem
Reformatting numbers after each keystroke leads to unexpected behavior when editing anything other then the last digit in a number:

#### Reproduction steps (case 1)
- enter a number > 999 into an numeric input, ie 1200
- the number is automatically formatted to 1.200
- your caret is at the last position, so every new number you enter appears after the 0 (1200**I**)
- imagine you have made an error: you should have entered 12500
- move your caret to position after the 2 (1.2**I**00)
- typ 5
- the number is formatted as 12.500
- **observed**: the caret is moved to the last position again: 12.500**I**
- **expected**: caret remains after the number you entered, so 12.5**I**00

#### Reproduction steps (case 2)
- enter 1500, which is automatically formatted as 1.500
- imagine you had to type 2500 instead.
- move your caret to the position between the . and the 5: 1.**|**500
- press backspace, intending to remove the 1
- **observed:** you remove the '.', the number is immediately reformated and the caret is moved to the last postition: observed state is : 1.500**|** (so effectively, nothing happens). 
- **expected** keyboard entries ignore the '.', backspace removes the 1 and caret remains at same position. So expected state is: **|**500

### Potential solutions:
- Nice, but potentially complex: Ignore formatting when moving the caret (this can be rather complicated, we should add logic that performs some magic when using keys that have an effect on the caret position (left/right arrows, backspace)
- Fastest: Remove formatting while editing. Add formatting on blur, remove it on focus.

### LINKED ISSUE #341: Refactor `usePostitiveNumberInputMask`
Move the logic to live alongside the input.
Dismissal of the hoverpopup is easier as well as processing the generated warning from the input mask.
It will also simplify the form components.

lkleuver:
This is an unstyled component that accepts standard <input /> props but masks number input by formatting it with decimal points.

Also adds an implementation to restore caret position after edit.

Future todo:
Add tooltip support and paste warning implementation

closes #341 
closes #301


jorisleker:
@lkleuver nice, almost there!
I think we can do a little bit better with backspace support.

Two 'bugs':
- if your caret is right _after_ a thousands separator, **backspace** does nothing (it removes the dot, which is immediately re-added).
- if your caret is right _before_ a thousands separator, **delete** does nothing (it removes the dot, which is immediately re-added).

Other issues have been solved.

jorisleker (changes requested):
See previous comment

lkleuver:
I've pushed a fix for delete and backspace behavior. 
This component feels extra scary though since Abacus is all about numbers. 
Please help me try and break it 😅

frontend/lib/ui/NumberInput/NumberInput.test.tsx
```diff
@@ -38,4 +38,28 @@ describe("UI Component: number input", () => {
...
+  test("should delete the right character", async () => {
+    render(<NumberInput id="test" defaultValue={12345} />);
+    const input: HTMLInputElement = screen.getByTestId("test");
+    const user = userEvent.setup();
+
+    input.focus();
+    //value should be 12.345
+    await user.keyboard("{arrowleft}{arrowleft}{arrowleft}{arrowLeft}");
+    await user.keyboard("{delete}");
+    expect(input).toHaveValue("1.245");
```
Lionqueen94, line 63: Shouldn't the 2 be deleted instead of the 3 when you move the caret 4 times to the left? Or we count the dot as a character I guess?

jorisleker: Good suggestion @Lionqueen94, not sure if we could implement/build that. But since you can't really type the `.` yourself, we might try and ignore it altogether when moving the caret. Just like we do now when using backspace or delete. Would mean that the cursor would jump over it when moving the caret.

Lionqueen94: Exactly, that makes more sense to me since it's just a styling dot, that you cannot add or remove yourself.

jorisleker:
> Please help me try and break it 😅
tried a lot, but can't get it to break yet 🎉

frontend/lib/ui/NumberInput/NumberInput.test.tsx
lkleuver: I think this is very doable.

lkleuver: I left the other behavior in as wel since you can still "reach" the dot by mouse or by deleting a char.
it looks really weird if you click on "pos x" and the cursor jumps.

jorisleker: @lkleuver (looks like the caret only 'jumps' now when you are moving with the left arrow, and not with the right arrow. Shouldn't this be both ways?

jorisleker: mmm, jumping the cursor breaks the current selection range. 
repro:
- input 1234
- move cursor to last position
- press and hold shift
- use left arrow to expand selection 
- observed: after you pass the `.`, a new selection range is set (containing the `1.`)
- expected: selection range contains `1.234` after passing the dot.

jschuurk-kr: Related, using shift+arrow:

1. enter "12.345"
2. use arrows to navigate to just after the first character: `1|2.345`
3. press shift + arrow right

This results in the separator symbol being selected instead of the "2".

jschuurk-kr:
Am not able to reproduce, so just something to keep an eye out for: I was able to get my cursor into field D, but moving my mouse over the input in field A, selected the numbers in that field.

frontend/lib/ui/NumberInput/NumberInput.test.tsx
jschuurk-kr: Another one related to selection:

1. enter `123.456.789`
2. put your cursor before the `3`: `12|3.456.789`
3. select up to the `5`, so `.45` are selected
4. hit the delete key

This results in `12.356.789`, so the `5` is still there, even though it was selected and should have been deleted.

praseodym:
#301 also mentions an easier solution:

> Fastest: Remove formatting while editing. Add formatting on blur, remove it on focus.

This is also how e.g. Microsoft Excel works, and I personally prefer this from a UX standpoint too.

Maybe we should reconsider and go for that solution?

jorisleker:
Agree with @praseodym. Think we are just scratching the surface of everything that can go wrong when manipulating cursor positions and selection ranges. Hard to build, test and maintain.

frontend/lib/ui/NumberInput/NumberInput.test.tsx
lkleuver: I've changed is so formatting is removed on focus and returned on blur, also added an onKeyDown handler to make it behave as input type="number"

jorisleker (changes requested):
Noticed selection-range isn't applied correctly when formatting is removed:

Compare case A:
- Enter 123 in an input
- tab to next input
- shift tab back to previous input
- 123 is selected

to case B:
- Enter 1234 in an input
- tab to next input (1234 is reformatted as 1.234)
- shift tab back to previous input (1.234 formatting is removed to 1234)
- nothing is selected
- expected: 1234 to be selected

jorisleker (approved):

Lionqueen94:
> Noticed selection-range isn't applied correctly when formatting is removed:
> 
> Compare case A:
> 
>     * Enter 123 in an input
> 
>     * tab to next input
> 
>     * shift tab back to previous input
> 
>     * 123 is selected
> 
> 
> to case B:
> 
>     * Enter 1234 in an input
> 
>     * tab to next input (1234 is reformatted as 1.234)
> 
>     * shift tab back to previous input (1.234 formatting is removed to 1234)
> 
>     * nothing is selected
> 
>     * expected: 1234 to be selected

@lkleuver How does it work when unselecting now, it seems like the value is always selected when moving back with shift + tab. Is that the desired behaviour? @jorisleker

praseodym (changes requested):
Caret positioning when clicking into the field does not take into account the removed dots. If I have number `123.456` and place the caret by clicking at `|` in `123|.456`, formatting is removed and the caret appears as `12|3456` instead of as `123|456`.

jorisleker:
> How does it work when unselecting now, it seems like the value is always selected when moving back with shift + tab. Is that the desired behaviour? 

@Lionqueen94 yes this is desired behaviour. Default for inputs. When focussing them by keyboard, contents are fully selected.

> Caret positioning when clicking into the field does not take into account the removed dots. If I have number 123.456 and place the caret by clicking at | in 123|.456, formatting is removed and the caret appears as 12|3456 instead of as 123|456.

Might be harder to fix this? Looks like the mouse event that focusses the input is fired _after_ the reformatting has taken place. So this is just the 'click' registering on the unformatted string, whereas the user intended to click on a position in the formatted string.

lkleuver:
> > How does it work when unselecting now, it seems like the value is always selected when moving back with shift + tab. Is that the desired behaviour?
> 
> @Lionqueen94 yes this is desired behaviour. Default for inputs. When focussing them by keyboard, contents are fully selected.
> 
> > Caret positioning when clicking into the field does not take into account the removed dots. If I have number 123.456 and place the caret by clicking at | in 123|.456, formatting is removed and the caret appears as 12|3456 instead of as 123|456.
> 
> @praseodym @lkleuver Might be harder to fix this? Looks like the mouse event that focusses the input is fired _after_ the reformatting has taken place. So this is just the 'click' registering on the unformatted string, whereas the user intended to click on a position in the formatted string.

Yes, you are kind of going back to the initial approach with caret control this way.
Can probably detect this and move the caret if neccesarry

lkleuver:
> Caret positioning when clicking into the field does not take into account the removed dots. If I have number `123.456` and place the caret by clicking at `|` in `123|.456`, formatting is removed and the caret appears as `12|3456` instead of as `123|456`.

I've added something to make it more natural. Does this work for you @praseodym ?

praseodym:
> > Caret positioning when clicking into the field does not take into account the removed dots. If I have number `123.456` and place the caret by clicking at `|` in `123|.456`, formatting is removed and the caret appears as `12|3456` instead of as `123|456`.
> 
> I've added something to make it more natural. Does this work for you @praseodym ?

I don't see any difference, the behaviour is still the same.

praseodym:
We've decided not to try to compensate the caret position after all. This is hacky and both Microsoft Excel and Apple Numbers also don't do anything like this when working with formatted cells.

praseodym (approved):

jorisleker (approved):

PR #233, opened 2024-08-07
Update data entry navigation

lkleuver:
central storage of values and server responses for pollingstation form sections

controlled navigation through form sections
- go to the next section to fill in
- restrict navigation when the user forgot to save
- block sections that are later in the flow

[flowcharts describing the rendering and navigation logic](https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/data-entry-forms.md)

# TODO while in draft:
- [x] fix current tests to check for navigation through the form
- [x] merge/rebase main
- [x] fix showing errors too early (differences form)
- [x] fix political group votes errors not showing at all
- [x] fix showing votes on candidates incorrect total error too early (voters and votes form) #149 
- [x] scroll up when there is an error on a page
- [x] improve aria-labels for icons
  - aria-labels should describe meaning of icon ("je bent hier") instead of appearance of icon ("pijl naar rechts")

# Testing findings 1

(by @jschuurk-kr)

## General
- [x] 1. The `console.log()`s  (e.g. "Checking changes!!!" and "new form state" with data) should be removed or be more informative → Fixed in 632aaf5ea51c6445a654e5d385cef31945f063d2
- [x] 2. Shift+Enter does not seem to work on any of the pages. (Ubuntu + Firefox) → Part of #263 
- [x] 3. Proposed url change: `/1/input/1/numbers` -> `/1/input/1/voters-and-votes` → Part of #184 
- [x] 4. Proposed url change: `:electionId/input` -> `elections/:electionId/input` → Part of #184
- [x] 5. Should we sort the errors and warnings, i.e. always show F201 above F202, etc? Currently they aren't. → Part of #266 
- [x] 6. If I have a warning and change the input so I get an error, the warning checkbox is very briefly rendered.
- [x] 7. Clicking "Volgende" seems to bring me to the furthest page visited, not necessarily the next page.

## Warnings checkbox
- [x] 1. The checkbox disappears even when not changing fields, e.g. when pressing ctrl, esc, alt, tab, arrow keys. → Will fix
- [x] 2. The checkbox disappears when replacing the value in a field with the value it contained. → Won't fix
- [x] 3. The checkbox often does not appear on the Differences page.
- [x] 4. If I uncheck the checkbox and click next, I don't get any feedback from the UI about what to do next.
[...]

PR #367, opened 2024-09-23
Polling station status behavior

cikzh:
- Adds an alert when the users enters a valid polling station number, but it's in the "Definitief" state
- Filters polling stations from the list with a "Definitief" state
- Throws an error and send the user to the "Er is iets mis" page, in case the user tries to reach a finalised polling station through the URL bar or browser history

@jorisleker When the user tries to select a finalised polling station, it will get an alert, but at the same time the polling station message will appear green. Perhaps we want the polling station message to be a different color? Yellow or red? What do you think?

### DIFF
--- a/frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
+++ b/frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
@@ -49,9 +49,9 @@ export function PollingStationChoiceForm({ anotherEntry }: PollingStationChoiceF
 
     const parsedStationNumber = parsePollingStationNumber(pollingStationNumber);
     const pollingStation = pollingStations.find((pollingStation) => pollingStation.number === parsedStationNumber);
-    const pollingStationStatus = electionStatuses.statuses.find((status) => status.id === pollingStation?.id);
+    const pollingStationStatusEntry = electionStatuses.statuses.find((status) => status.id === pollingStation?.id);
 
-    if (pollingStationStatus?.status === "definitive") {
+    if (pollingStationStatusEntry?.status === "definitive") {
       setAlert(DEFINITIVE_POLLING_STATION_ALERT);
       setLoading(false);
       return;
--- a/frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
+++ b/frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
@@ -1,7 +1,7 @@
 import { useMemo, useState } from "react";
 import { useNavigate } from "react-router-dom";
 
-import { PollingStation, useElection } from "@kiesraad/api";
+import { PollingStation, useElection, useElectionStatus } from "@kiesraad/api";
 import { IconError } from "@kiesraad/icon";
 import { Alert, BottomBar, Button, Icon, KeyboardKey, KeyboardKeys } from "@kiesraad/ui";
 import { cn, parsePollingStationNumber, useDebouncedCallback } from "@kiesraad/util";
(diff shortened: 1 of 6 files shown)


frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
```diff
@@ -38,17 +43,24 @@ export function PollingStationChoiceForm({ anotherEntry }: PollingStationChoiceF
...
-      setShowAlert(true);
+      setAlert(INVALID_POLLING_STATION_ALERT);
       return;
     }
 
     const parsedStationNumber = parsePollingStationNumber(pollingStationNumber);
     const pollingStation = pollingStations.find((pollingStation) => pollingStation.number === parsedStationNumber);
+    const pollingStationStatus = electionStatuses.statuses.find((status) => status.id === pollingStation?.id);
+
+    if (pollingStationStatus?.status === "definitive") {
```
Lionqueen94, line 54: It looks a bit weird to use pollingStationStatus.status, maybe we can rename the const?

frontend/app/routes.test.tsx
```diff
@@ -90,4 +90,22 @@ describe("routes", () => {
...
+    await router.navigate("/elections/1/data-entry/2");
+    expect(router.state.location.pathname).toEqual("/elections/1/data-entry/2");
+
+    // NOTE: We're not using the wrapped render function here,
+    // since we want control over our own memory router.
+    rtlRender(
+      <ApiProvider>
+        <RouterProvider router={router} />
+      </ApiProvider>,
+    );
```
praseodym, line 107: We can simplify this after my changes in https://github.com/kiesraad/abacus/pull/233/commits/cc2afed8e602cdca510e1f48ed850295d440f806:
```typescript
rtlRender(<Providers router={router} />);
```
Or to replace all invocations in this file:
```typescript
// top of file
const render = () => rtlRender(<Providers router={router} />);

// in the test
render();
```
and replace all duplicate code blocks.

praseodym (commented):
Two minor comments, otherwise it looks good!

Also, maybe not for this PR/issue, I think the design of the polling station choice page can be improved for when all inputs have been completed. See this screenshot:
localhost_3000_elections_1_data-entry (2)
The green alert isn't even meant for the 'invoerder' role, so we'll be removing this in a later iteration. @jorisleker any ideas on what we should show in that case?

frontend/app/routes.test.tsx
cikzh: [72a87fe](https://github.com/kiesraad/abacus/pull/367/commits/72a87fee88145db3182e255116d1662337f4a3f0)

frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
cikzh: [a932acc](https://github.com/kiesraad/abacus/pull/367/commits/a932accef1b712d26fa3d2a86a502af6202dc2fd)

frontend/app/component/form/data_entry/polling_station_choice/PollingStationsList.tsx
```diff
@@ -28,6 +28,10 @@ export function PollingStationsList({ pollingStations }: PollingStationsListProp
       <tbody>
         {pollingStations.map((pollingStation: PollingStation) => {
           const pollingStationStatus = electionStatus.statuses.find((status) => status.id === pollingStation.id);
+          if (pollingStationStatus?.status === "definitive") {
```
Lionqueen94, line 31:
```suggestion
          const pollingStationStatusEntry = electionStatus.statuses.find((status) => status.id === pollingStation.id);
          if (pollingStationStatusEntry?.status === "definitive") {
```

praseodym (dismissed):

Lionqueen94 (approved):

praseodym:
> WDYT? Shall I create a new ticket to implement these?

Yes please!

jorisleker:
✅ https://github.com/kiesraad/abacus/issues/388

PR #377, opened 2024-09-24
Fixed navigating form sections and focus state and Alert styling

Lionqueen94:
## TODO
- [x] Fix focusing on Feedback header in case of error F.204, now the focus is on the form title
- [x] Fix rendering of unsaved changes modal when you get error F.204 and try to navigate away from the page after making no changes

## Description
- [x] Removed autofocus from input fields
- [x] Added focus to form/feedback/modal heading where needed
- [x] Made sure navigation outside of form works with keyboard
- [x] Updated Alert styling to adhere to design
- [x] Focus on previously active element after closing modal
- [x] Changed modal from using `createPortal` to a `<dialog>` element to increase accessibility
- [x] Added workaround for running tests because of jsdom lacking support for HTMLDialogElement (see https://github.com/jsdom/jsdom/issues/3294#issuecomment-1268330372)
- [x] Modal close button does not focus with keyboard ([focus state for close button](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2630-29542&node-type=frame&t=hR2XbCLIZaEi4pNS-11))
- [x] Add focus state to Alert close button ([design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=985-19666&node-type=frame&t=hR2XbCLIZaEi4pNS-11))

Close buttons focus states have a 2px inside border with same color (variant 400)

## What to test
- [x] Alert styling (the error one for accept warnings in particular)
- [x] Focus of close buttons on Alert and Modal
- [x] Title on each form page should be focused and with <kbd>Tab</kbd> you can go to the first input
- [x] Radio button behaviour on Recounted page with only keyboard (not sure if this is accessibility compliant)
- [x] When opening the modal, the title should be focused, with <kbd>Tab</kbd> you can go to the buttons and with <kbd>Shift+Tab</kbd> you can go to the close button
- [x] When closing modal the last focused element should be focused again (for example `Invoer afbreken` button)
- [x] <kbd>Enter</kbd> and <kbd>Shift+Enter</kbd> behaviour is as designed on each form page


frontend/lib/ui/Form/Form.tsx
```diff
@@ -109,6 +112,11 @@ export const Form = React.forwardRef<HTMLFormElement, FormProps>(({ children, sk
     submitButton.current = innerRef.current?.querySelector("button[type=submit]") as HTMLButtonElement | null;
   }, [children, skip]);
 
+  // TODO: Need to perform this also when modal closes
```
Lionqueen94, line 115: @lkleuver Do you have any suggestion how to focus the form title again after a modal closes (onClose/onCancel)?

frontend/lib/ui/Modal/Modal.tsx
```diff
@@ -7,17 +7,23 @@ import { IconButton } from "@kiesraad/ui";
...
 
+  React.useEffect(() => {
+    document.getElementById(id)?.focus();
+  }, [id]);
+
   return createPortal(
     <div className={cls.modal} role="dialog">
       <div className={cls["modal-container"]}>
         {onClose && (
+          // TODO: How to make sure thus button can focus with keyboard?
```
Lionqueen94, line 26: @lkleuver Do you know how I can make sure this close button can focus with the keyboard?

frontend/lib/ui/Form/Form.tsx
jorisleker: @Lionqueen94 shouldn't we return the focus to where it was when opening the modal? (no clue whether this will make implementation harder or easier)

frontend/lib/ui/Modal/Modal.stories.tsx
```diff
@@ -5,24 +5,22 @@ import { Button } from "@kiesraad/ui";
...
-    </Modal>
-  </>
+  <Modal id="modal-title" onClose={() => {}}>
+    <h2 id="modal-title" tabIndex={-1}>
+      Wat wil je doen met je invoer?
+    </h2>
+    <p>
+      Ga je op een later moment verder met het invoeren van dit stembureau? Dan kan je de invoer die je al hebt gedaan
+      bewaren.
+      <br />
```
jorisleker, line 15: `</p><p>`

frontend/lib/ui/Modal/Modal.tsx
```diff
@@ -1,28 +1,49 @@
...
+            onClick={() => {
+              if (dialogRef.current) {
+                dialogRef.current.close();
+                dialogRef.current = null;
+                lastActiveElement.current?.focus();
+              }
+              onClose();
+            }}
+            icon={<IconCross />}
+            title="Annuleren"
```
Lionqueen94, line 40: @jorisleker is it ok to call the close button "Annuleren" instead of "Melding sluiten"?

jorisleker: Yes that's ok / better 👍

jorisleker:
@Lionqueen94 LGTM. Just noticed #404 was merged with a small focus-'bug'. The close button in the alert doesn't have a focus state there. Will this PR fix that?

jorisleker (approved):

jorisleker:
And one question remark regarding "Focus on previously active element after closing modal". This works, but I think the behavior is not really what we want it to be. When using the keyboard for navigation, this is fine: focus is returned to the element that was focussed before triggering the modal. But: when using the mouse, the result is a bit weird. 

ie: I have my focus on an input field. I click the 'invoer afbreken' button, and then hit cancel. Now my focus is on the 'invoer afbreken' button. Would be more natural if in this case, focus would still be on the input. Not a very big deal, happy to accept current implementation.

Lionqueen94:
> @Lionqueen94 LGTM. Just noticed #404 was merged with a small focus-'bug'. The close button in the alert doesn't have a focus state there. Will this PR fix that?

Yes it uses a regular `<Alert>` component so it should have a focus state now

Lionqueen94:
> And one question remark regarding "Focus on previously active element after closing modal". This works, but I think the behavior is not really what we want it to be. When using the keyboard for navigation, this is fine: focus is returned to the element that was focussed before triggering the modal. But: when using the mouse, the result is a bit weird.
> 
> ie: I have my focus on an input field. I click the 'invoer afbreken' button, and then hit cancel. Now my focus is on the 'invoer afbreken' button. Would be more natural if in this case, focus would still be on the input. Not a very big deal, happy to accept current implementation.

I think it is better to add this as a separate issue, because I'm going to have to do some saving of the last input field to make this work and make a distinction between keyboard and mouse users and that's quite a bit more effort.

jorisleker:
> I think it is better to add this as a separate issue, because I'm going to have to do some saving of the last input field to make this work and make a distinction between keyboard and mouse users and that's quite a bit more effort.

Yeah, same thought. Will postpone adding it as an issue until we notice it is really an issue (ie from a11y audit)

praseodym:
Two findings, I'm not sure how severe they are:
- On the recounted page, when navigating with <kbd>Tab</kbd> I can't get the radio buttons checked using <kbd>Spacebar</kbd> or <kbd>Enter</kbd>. Only <kbd>↓</kbd> and <kbd>↑</kbd> work to check the radio buttons.
- When navigating to a new section using the next button, focus gets set on the header. If I press <kbd>Tab</kbd> the first input field gets focus, but if I instead press <kbd>Enter</kbd> the second input fields gets focus.

Lionqueen94:
@jorisleker Mark noticed a few bugs and we worked on fixing them. We think the radio button behaviour is now correct (spacebar to select the radio works now for example) and the focusing on the input fields now does not skip one field. Could you review again and let us know if you notice anything wrong?

frontend/app/component/form/data_entry/voters_and_votes/VotersAndVotesForm.tsx
```diff
@@ -146,20 +144,24 @@ export function VotersAndVotesForm() {
   React.useEffect(() => {
     if (isSaved) {
       window.scrollTo(0, 0);
+      document.getElementById("feedback-title-0")?.focus();
```
lkleuver, line 147: couldn't we use refs for this as well? 
```
const returnFocusRef = React.useRef<HTMLElement>(null);
...
returnFocusRef.current?.focus();
```

frontend/lib/ui/Feedback/Feedback.tsx
```diff
@@ -30,7 +30,9 @@ export function Feedback({ id, type, data, apiError, children }: FeedbackProps)
         <div className="feedback-item">
           <header>
             {renderIconForType(type)}
-            <h3>Sorry, er ging iets mis</h3>
+            <h3 id="feedback-title-0" tabIndex={-1}>
```
lkleuver, line 33: you can't put fixed ID's in ui elements, they are not guaranteed to be used only once on a page.
so it has to be a prop.

frontend/lib/ui/Form/Form.tsx
```diff
@@ -101,13 +89,12 @@ export const Form = React.forwardRef<HTMLFormElement, FormProps>(({ children, sk
...
-    //filter out inputs we should skip
-    if (skip && skip.length) {
-      inputList.current = inputList.current.filter((input) => !skip.includes(input.id));
-    }
     submitButton.current = innerRef.current?.querySelector("button[type=submit]") as HTMLButtonElement | null;
-  }, [children, skip]);
+  }, [children]);
+
+  React.useEffect(() => {
+    document.getElementById("form-title")?.focus();
```
lkleuver, line 96: instead of id, maybe find the first header element?
or find by className

lkleuver (changes requested):

jorisleker (approved):
Retested the changes regarding radio-button interaction and using enter when opening a new section. Both approved!

frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
```diff
@@ -123,15 +122,16 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
...
   const defaultProps = {
     errorsAndWarnings: isSaved ? errorsAndWarnings : undefined,
     warningsAccepted: getAcceptWarnings(),
   };
 
   return (
-    <Form onSubmit={handleSubmit} ref={formRef} id={`candidates_form_${group.number}`} skip={[_ACCEPT_WARNINGS_ID]}>
-      <h2>
+    <Form onSubmit={handleSubmit} ref={formRef} id={`candidates_form_${group.number}`}>
+      <h2 id="form-title" tabIndex={-1}>
```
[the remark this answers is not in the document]
Lionqueen94: Changed in d1bf0ed431654f0ec14291f58a46a924fc3242c0

frontend/app/component/form/data_entry/voters_and_votes/VotersAndVotesForm.tsx
Lionqueen94: Changed in d1bf0ed431654f0ec14291f58a46a924fc3242c0

frontend/lib/ui/Feedback/Feedback.tsx
Lionqueen94: Very good point! Changed in d1bf0ed431654f0ec14291f58a46a924fc3242c0

frontend/lib/ui/Form/Form.tsx
Lionqueen94: Changed in d1bf0ed431654f0ec14291f58a46a924fc3242c0

jorisleker (approved):

praseodym (approved):

praseodym:
d2d test fails:
```
  1) [chrome] › data-entry.e2e.ts:293:3 › full data entry flow › submit with accepted warning on voters and votes page 

    Error: locator.check: Element is outside of the viewport
    Call log:
      - waiting for getByLabel('Ik heb de aantallen gecontroleerd met het papier en correct overgenomen.')
      -   locator resolved to <input type="checkbox" id="voters_and_votes_form_accept_warnings" name="voters_and_votes_form_accept_warnings"/>
      - attempting click action
      -   scrolling into view if needed
      -   done scrolling


       at ../page-objects/data_entry/VotersAndVotesPgObj.ts:113

      111 |     await this.acceptWarnings.waitFor();
      112 |     // eslint-disable-next-line playwright/no-force-option -- force option needed to click on hidden element
    > 113 |     await this.acceptWarnings.check({ force: true, noWaitAfter: true });
          |                               ^
      114 |   }
      115 | }
      116 |

        at VotersAndVotesPage.checkAcceptWarnings (/home/runner/work/abacus/abacus/frontend/e2e-tests/page-objects/data_entry/VotersAndVotesPgObj.ts:113:31)
        at /home/runner/work/abacus/abacus/frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts:322:5
```

praseodym (dismissed):

jorisleker (approved):

frontend/lib/ui/Modal/Modal.stories.tsx
```diff
@@ -5,22 +5,20 @@ import { Button } from "@kiesraad/ui";
...
-      </p>
-      <nav>
-        <Button size="lg">Invoer bewaren</Button>
-        <Button size="lg" variant="secondary">
-          Niet bewaren
-        </Button>
-      </nav>
-    </Modal>
-  </>
+  <Modal id="modal-title" onClose={() => {}}>
```
oliver3, line 8: I think that the Modal prop name 'id' is a bit confusing, at first I thought that there are now two identical id's on the page. Maybe rename it with a more descriptive name like `focusId` ?

frontend/app/module/data_entry/PollingStation/PollingStation.test.tsx
```diff
@@ -20,10 +20,15 @@ const startPollingStationInput = async () => {
   expect(router.state.location.pathname).toEqual("/elections/1/data-entry/1");
 };
 
-const expectRecountedForm = async () => {
+const expectRecountedForm = async (headingShouldNotHaveFocus?: boolean) => {
```
oliver3, line 23: I personally have something against negative boolean names, especially when used in an inverse way :)
Maybe we can do `async (headingShouldHaveFocus = true) {` instead and flip all usages?

frontend/lib/api/form/pollingstation/pollingStationUtils.ts
```diff
@@ -476,4 +476,11 @@ export function updateFormStateAfterSubmit(
       section.warnings = section.warnings.filter((err) => !isGlobalValidationResult(err));
     });
   }
+
+  // Reset acceptWarnings when a page gets an error or has no warnings anymore
+  Object.values(formState.sections).forEach((section) => {
+    if (section.acceptWarnings && (section.errors.length > 0 || section.warnings.length === 0)) {
```
oliver3, line 482: Could there be a scenario that there is one section without warnings and one with warnings? Is the expected behaviour that the `acceptWarnings` will be reset in this case?

Lionqueen94: Yes, that is possible, so that's why we only reset the acceptWarnings for the section it is applicable for. We save this variable for every section individually.

frontend/app/module/data_entry/PollingStation/PollingStation.test.tsx
Lionqueen94: Changed in 8dc28d65183b3e99fae58abe59646b139e4ae842

frontend/lib/ui/Modal/Modal.stories.tsx
Lionqueen94: Added the title as a prop to modal, to be in line with the form and to not have to add this id as a prop in 8dc28d65183b3e99fae58abe59646b139e4ae842

oliver3: Ah even better!

frontend/lib/api/form/pollingstation/pollingStationUtils.ts
oliver3: Ah this is for one section, got it, thanks!

oliver3 (approved):

praseodym (approved):

jorisleker:
@Lionqueen94 So nearly there 😅 
The close button on the modal moves up+left by (I think) 2px when it receives focus. It should not :-)
Scherm­afbeelding 2024-10-10 om 17 50 46

All other behavior as expected, including the keyboard-nav for the new radio-button/group.

Lionqueen94:
> @Lionqueen94 So nearly there 😅 The close button on the modal moves up+to the side by (I think) 2px when it receives focus. It should not :-) Scherm­afbeelding 2024-10-10 om 17 50 46
> 
> All other behavior as expected, including the keyboard-nav for the new radio-button/group.

Oeh well spotted Joris! There was a var that was forgotten to be renamed causing the border to not exist at all. It's fixed in 7aed6536ae78a762dba1bda959c38418282841d8

frontend/lib/ui/Feedback/Feedback.tsx
```diff
@@ -1,5 +1,4 @@
-import * as React from "react";
-import { ReactNode } from "react";
+import { useEffect, useRef } from "react";
```
praseodym, line 1: Use `React.` instead?

jorisleker (approved):
🎉

praseodym (approved):

PR #378, opened 2024-09-24
Election polling station progress bars

### LINKED ISSUE #242: Add polling station progress bars
Create under 'Voortgang' on `/1/input`.

See design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=1195-25602&m=dev

lkleuver:
Adds progress bars on the data-entry home page showing percentages of first entry and completed polling stations

closes #242

### DIFF
--- a/frontend/lib/ui/ProgressBar/ProgressBar.tsx
+++ b/frontend/lib/ui/ProgressBar/ProgressBar.tsx
@@ -1,0 +1,25 @@
+import cls from "./ProgressBar.module.css";
+
+export interface ProgressBarProps {
+  percent: number;
+  title: string;
+}
+export function ProgressBar({ percent, title }: ProgressBarProps) {
+  return (
+    <div className={cls["progressbar-container"]}>
+      <label>{title}</label>
+      <section>
+        <div
+          className={cls["progressbar-outer"]}
+          role="progressbar"
+          aria-valuenow={percent}
+          aria-valuemin={0}
+          aria-valuemax={100}
+        >
+          <div className={cls["progressbar-inner"]} />
+        </div>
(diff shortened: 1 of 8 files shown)


jorisleker (changes requested):
✅ design looks good.
✅ counting of 'alles samen' works as expected (counts percentage of polling stations with results
❌ counting of 'eerste invoer' (and probably 'tweede invoer') does not work as designed.

- 'Eerste invoer' should count the percentage with a *completed* first data entry as a percentage of the total number of polling stations. So this number should start at '0' and go to '100' when more polling station data comes in. As polling station data entries are deleted when the polling station is finalized, a polling station's first data entry should be counted as completed as follows: 
  - there is a record in `polling_station_data_entries` with (data.finalized=1 and entry_number=1) 
  - OR there is a record in `polling_station_results`

frontend/lib/ui/ProgressBar/ProgressBar.tsx
praseodym, line 1: Any reason not to use the [`<progress>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress) element?

lkleuver:
As discussed on vidconf, as long as there is only first_entry we are only showing the definitive progress bar for now.

jorisleker (approved):

frontend/lib/ui/ProgressBar/ProgressBar.tsx
lkleuver: I considered it but was worried about consistent styling.

praseodym (approved):

PR #387, opened 2024-09-25
Fixed showing the icon on the progress list for check and save page

### LINKED ISSUE #382: Progress list: Controleren en opslaan does not show current / furthest page icon
In progress list, we do not show an icon on the 'controleren en opslaan' step. We should.

These 2 cases are missing:
- [x] When user is on the "Controleren en opslaan" step, it should show the current page icon (black arrow)
- [x] When user is not on the "Controleren en opslaan" step, but is allowed to navigate there (because it is the furthest page), it should show the blue pencil icon.

Screenshot 1: User is on "Controleren en opslaan", but black arrow icon is missing:
Scherm­afbeelding 2024-09-25 om 10 59 27

Screenshot 2: User is allowed to navigate to "Controleren en opslaan" (see hover effect), but blue pencil icon is missing:
Scherm­afbeelding 2024-09-25 om 10 59 33

Lionqueen94:
Closes #382


jorisleker (approved):
👌 LGTM

praseodym (approved):

PR #401, opened 2024-09-30
Disable autocomplete for inputs

### LINKED ISSUE #394: Make sure we never show autocomplete on inputs
On the laptops we use for demo-ing (ubuntu/firefox), browsers are in their default config, which means autocomplete for input's is switched on. 

We should make sure we never show autocomplete/auto populate on the inputs in our application. 
https://www.w3schools.com/howto/howto_html_autocomplete_off.asp

lkleuver:
On the laptops we use for demo-ing (ubuntu/firefox), browsers are in their default config, which means autocomplete for input's is switched on.

We should make sure we never show autocomplete/auto populate on the inputs in our application.

closes #394

### DIFF
--- a/frontend/lib/ui/InputField/InputField.tsx
+++ b/frontend/lib/ui/InputField/InputField.tsx
@@ -39,6 +39,7 @@ export function InputField({
           <textarea
             name={name}
             value={value}
+            autoComplete="off"
             aria-invalid={error ? "true" : "false"}
             aria-errormessage={error ? `${name}-hint_or_error` : undefined}
             rows={7}
@@ -48,6 +49,7 @@ export function InputField({
             name={name}
             value={value}
             type={type}
+            autoComplete="off"
             aria-invalid={error ? "true" : "false"}
             aria-errormessage={error ? `${name}-hint_or_error` : undefined}
             {...InputFieldProps}
--- a/frontend/lib/ui/NumberInput/NumberInput.tsx
+++ b/frontend/lib/ui/NumberInput/NumberInput.tsx
@@ -10,6 +10,7 @@ export interface NumberInputProps
 export function NumberInput({ id, ...inputProps }: NumberInputProps) {
   const props = {
     maxLength: 9,
+    autocomplete: "off",
     ...inputProps,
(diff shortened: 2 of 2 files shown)


praseodym (commented):
Looks like this doesn't show up in the DOM, or am I overlooking something?

Example from the poll card count field:
```html
<div class="_form-field_j0ude_1"><aside></aside><input maxlength="9" name="poll_card_count" aria-invalid="false" type="text" id="poll_card_count" value=""></div>
```

lkleuver:
> Looks like this doesn't show up in the DOM, or am I overlooking something?
> 
> Example from the poll card count field:
> 
> ```
> <div class="_form-field_j0ude_1"><aside></aside><input maxlength="9" name="poll_card_count" aria-invalid="false" type="text" id="poll_card_count" value=""></div>
> ```

Hmm was after my latest commit? What browser?
This is off my local machine
```
<input maxlength="9" autocomplete="off" name="poll_card_count" aria-invalid="false" type="text" id="poll_card_count" value="">
```

praseodym:
> Hmm was after my latest commit? What browser? This is off my local machine [..]

My bad, looks like I was looking at the wrong branch or something. I can see the attribute after all.

jorisleker:
@praseodym could you test this branch on the demo laptops to see if the problem is solved there? I can see the attribute is added to fields, but the autocomplete was off already in all my browsers

Lionqueen94:
> @praseodym could you test this branch on the demo laptops to see if the problem is solved there? I can see the attribute is added to fields, but the autocomplete was off already in all my browsers

I have autocomplete on in my browser and it doesn't autocomplete anymore :)

jorisleker:
> I have autocomplete on in my browser and it doesn't autocomplete anymore :)

Double check: if you test another branch, do you see autocomplete/suggestions on inputs? So does this PR actually change the behavior?

Lionqueen94:
> > I have autocomplete on in my browser and it doesn't autocomplete anymore :)
> 
> Double check: if you test another branch, do you see autocomplete/suggestions on inputs? So does this PR actually change the behavior?

Yes I do get suggestions on another branch, so this PR prevents it

Lionqueen94 (approved):

jorisleker (approved):

PR #406, opened 2024-10-01
Add nullable types to OpenAPI generator

praseodym:
- Add nullable types to OpenAPI generator 
- Mark `skip_serializing_if` fields as non-nullable
  This prevents us from requiring checks for both `undefined` and `null`, even though the API will only return `undefined`.


frontend/lib/api/gen/openapi.ts
```diff
@@ -158,12 +158,12 @@ export interface PoliticalGroupVotes {
 export interface PollingStation {
   election_id: number;
   house_number: string;
-  house_number_addition?: string;
+  house_number_addition?: string | null;
```
cikzh, line 161: Can we lose the question mark here? If I understand correctly the API will return either `string` or `null`, since the Rust type doesn't have a `skip_serializing_if` macro, it will always return the field
```suggestion
  house_number_addition: string | null;
```

frontend/lib/api/gen/openapi.ts
```diff
@@ -158,12 +158,12 @@ export interface PoliticalGroupVotes {
...
   id: number;
   locality: string;
   name: string;
   number: number;
-  number_of_voters?: number;
+  number_of_voters?: number | null;
```
cikzh, line 166: Same here
```suggestion
  number_of_voters: number | null;
```

cikzh (changes requested):
Nice addition!

lkleuver (approved):

frontend/lib/api/gen/openapi.ts
praseodym: Yes you're right. We can do that by annotating it with `#[schema(required = true)]` in the Rust code. 

The only thing is that the deserialisation code will still accept both `undefined` and `null`. https://github.com/juhaku/utoipa/issues/638 has some more discussion about this.

praseodym: Alternatively we can mark it as `#[serde(skip_serializing_if = "Option::is_none")]` so it can be `undefined` but never `null`. What do we prefer?

cikzh: Personally I prefer `| null` types over the question mark operator, so type information is conveyed in the same place as much as possible.

I guess it's not that big of a deal that deserialisation still accepts both? I think the "profit" of this change lies mainly in the generated TS type, so we'll get the best help of the TS compiler. 

Is `#[serde(deny_unknown_fields)] ` any help to us here?

oliver3: I would personally prefer question mark for optional fields in TypeScript 😅 but this is also an API design question I think. What would you expect the frontend to send in e.g. POST requests. Set all the optional fields to null or leave them out?

praseodym: I decided to type them as `undefined` and never `null`, that then we don't have any `null` values in the API spec.

Lionqueen94 (approved):

frontend/lib/api/gen/openapi.ts
praseodym: See https://github.com/kiesraad/abacus/pull/406#discussion_r1793108126.

oliver3 (approved):

cikzh (approved):

Issue #402, opened 2024-09-30
Refactor PollingstationFormController getValues


praseodym:
Refactor forms and controller to store "draft" values in a ref instead of using getValues provider by setCurrentForm.

optionally: refactor controller to move logic into a class with a triggered rerender instead of state based rerenders
```ts
const controller = React.useMemo(() => return new Controller(rerenderFunc, props), [electionId, ....]);
```

jschuurk-kr:
will be covered in #850

PR #417, opened 2024-10-03
Save and resume data entry with form state

### LINKED ISSUE #381: Save and resume data entry with form state
Save user data from the form state (accepted warnings, current page, last page) to the backend and restore it on resume.

praseodym:
* Add client state to data entry API
* Resume form progress with client state: save client state on each save to backend, and resume by loading client state. To achieve resuming using the state, some of the controller functionality is refactored.

Resolves #381 

What to test:
* Does saving or aborting data entry and resuming later always get back to the same point?
* Is the current behaviour with regards to showing errors and warnings when resuming ok? These are shown when coming back after clicking next ("Volgende"), but not when coming back after aborting ("Invoer afbreken") or navigating away.

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -757,9 +757,13 @@
         "description": "Response structure for getting data entry of polling station results",
         "required": [
           "data",
+          "client_state",
           "validation_results"
         ],
         "properties": {
+          "client_state": {
+            "type": "object"
+          },
           "data": {
             "$ref": "#/components/schemas/PollingStationResults"
           },
@@ -956,9 +960,14 @@
         "type": "object",
         "description": "Request structure for saving data entry of polling station results",
         "required": [
-          "data"
+          "data",
+          "client_state"
         ],
         "properties": {
+          "client_state": {
+            "type": "object",
+            "description": "Client state for the data entry (arbitrary JSON)"
+          },
           "data": {
             "$ref": "#/components/schemas/PollingStationResults"
           }
--- a/frontend/app/routes.tsx
+++ b/frontend/app/routes.tsx
@@ -36,7 +36,7 @@ export const routes = createRoutesFromElements(
       <Route path="data-entry" element={<DataEntryLayout />}>
         <Route index element={<DataEntryHomePage />} />
         <Route path=":pollingStationId" element={<PollingStationLayout />}>
-          <Route index element={<Navigate to="./recounted" replace />} />
+          <Route index element={null} />
           <Route path="recounted" element={<RecountedPage />} />
           <Route path="voters-and-votes" element={<VotersAndVotesPage />} />
           <Route path="differences" element={<DifferencesPage />} />
--- a/frontend/lib/api/form/pollingstation/PollingStationFormController.tsx
+++ b/frontend/lib/api/form/pollingstation/PollingStationFormController.tsx
@@ -1,24 +1,27 @@
 import * as React from "react";
+import { useNavigate } from "react-router-dom";
+
+import { getUrlForFormSectionID } from "app/component/pollingstation/utils";
 
 import {
-  addValidationResultToFormState,
   ApiError,
   ApiResponseStatus,
   Election,
+  getClientState,
   GetDataEntryResponse,
   getInitialFormState,
   getInitialValues,
-  isGlobalValidationResult,
+  getNextSectionID,
   POLLING_STATION_DATA_ENTRY_SAVE_REQUEST_PATH,
   PollingStationResults,
+  SaveDataEntryRequest,
   SaveDataEntryResponse,
+  updateFormStateAfterSubmit,
   useApi,
   useApiGetRequest,
   ValidationResult,
   VotersRecounts,
 } from "@kiesraad/api";
-
-import { formSectionComplete, getNextSection, resetFormSectionState } from "./pollingStationUtils";
 
 export interface PollingStationValues extends Omit<PollingStationResults, "recounted"> {
   recounted: boolean | undefined;
@@ -38,7 +41,7 @@ export interface FormReference<T> {
   type: string;
   id: FormSectionID;
   getValues: () => T;
-  getIgnoreWarnings?: () => boolean;
+  getAcceptWarnings?: () => boolean;
 }
 
 export interface FormReferenceRecounted extends FormReference<Pick<PollingStationValues, "recounted">> {
@@ -71,16 +74,21 @@ export type AnyFormReference =
   | FormReferencePoliticalGroupVotes
   | FormReferenceSave;
 
+interface SubmitCurrentFormOptions {
+  acceptWarnings?: boolean;
+  aborting?: boolean;
+  continueToNextSection?: boolean;
+}
+
 export interface iPollingStationControllerContext {
   status: React.RefObject<Status>;
   apiError: ApiError | null;
   formState: FormState;
-  targetFormSection: FormSectionID | null;
   values: PollingStationValues;
   setTemporaryCache: (cache: TemporaryCache | null) => boolean;
   cache: TemporaryCache | null;
   currentForm: AnyFormReference | null;
-  submitCurrentForm: (acceptWarnings?: boolean, aborting?: boolean) => Promise<void>;
+  submitCurrentForm: (params?: SubmitCurrentFormOptions) => Promise<void>;
   registerCurrentForm: (form: AnyFormReference) => void;
   deleteDataEntry: () => Promise<void>;
   finaliseDataEntry: () => Promise<void>;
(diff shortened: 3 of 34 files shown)


cikzh (dismissed):
Looks good! Mostly did a code review of the backend and some of the frontend, so it would be wise if someone took an extra look on the frontend, especially the navigation refactor. Maybe @lkleuver and/or @Lionqueen94?

Lionqueen94:
Regarding `- If the furthest page was edited and not explicitly saved, these changes are lost. Is this ok?`
To me it makes sense if you navigate away from the form that the cached input is sent to the backend to be saved. Since you explicitly click `Invoer bewaren` (even if you are not on that specific page). @jorisleker what do you think?

jorisleker:
Agree with @Lionqueen94. Input on furthest page should be saved if you choose invoer bewaren.

praseodym:
> Regarding `- If the furthest page was edited and not explicitly saved, these changes are lost. Is this ok?` To me it makes sense if you navigate away from the form that the cached input is sent to the backend to be saved. Since you explicitly click `Invoer bewaren` (even if you are not on that specific page). @jorisleker what do you think?

> Agree with @Lionqueen94. Input on furthest page should be saved if you choose invoer bewaren.

I've created #421 to figure ou the behaviour, clarify documentation and implement any fixes.

backend/openapi.json
```diff
@@ -757,9 +757,13 @@
         "description": "Response structure for getting data entry of polling station results",
         "required": [
           "data",
+          "client_state",
           "validation_results"
         ],
         "properties": {
+          "client_state": {
+            "type": "object"
```
lkleuver, line 765: I guess fine for now but it would be nice if we could type this.

frontend/app/routes.tsx
```diff
@@ -36,7 +36,7 @@ export const routes = createRoutesFromElements(
       <Route path="data-entry" element={<DataEntryLayout />}>
         <Route index element={<DataEntryHomePage />} />
         <Route path=":pollingStationId" element={<PollingStationLayout />}>
-          <Route index element={<Navigate to="./recounted" replace />} />
+          <Route index element={null} />
```
lkleuver, line 39: From just browsing "routes.tsx" perspective this is quite weird. how do you find the component responsible for rendering the index?

frontend/lib/api/form/pollingstation/PollingStationFormController.tsx
```diff
@@ -1,25 +1,28 @@
...
   POLLING_STATION_DATA_ENTRY_SAVE_REQUEST_PATH,
   PollingStationResults,
+  SaveDataEntryRequest,
   SaveDataEntryResponse,
+  updateFormStateAfterSubmit,
   useApi,
   useApiGetRequest,
   ValidationResult,
   VotersRecounts,
 } from "@kiesraad/api";
```
lkleuver, line 24: just noticed randomly, should libs use the @kiesraad alias to themselves? 
I guess only a bit annoying if you ever plan to make it standalone.

frontend/lib/api/form/pollingstation/PollingStationFormController.tsx
```diff
@@ -71,16 +74,21 @@ export type AnyFormReference =
   | FormReferencePoliticalGroupVotes
   | FormReferenceSave;
 
+interface SubmitCurrentFormOptions {
+  acceptWarnings?: boolean;
```
lkleuver, line 78: This is not part of values yet?
would make storing clientState easier?

lkleuver (approved):
only some small insignificant things.
Lots of name changes, makes it tricky but this is more logica.

lgtm

jorisleker (changes requested):
**Something weird is happening in the following case:**

Filled in polling station 33:
- recounted: no
- A:100, B:2, D:102, E 100, G:2, H:102
- differences: empty
- Lijst 1: 1:90, 2:10, total: 100
- do not submit list 1
- click invoer afbreken from list 1, and 'invoer bewaren'
- open polling station 33 from the blue alert
- you are on list 1, with cached input saved (expected)
- navigate to 'aantal kiezers en stemmen', don't make any changes
- click lijst 1
- **observed**: modal: 'Wat wil je doen met je invoer' is shown
- **expected**: modal not shown, as I did not make any changes.

It looks like the cache/saved state isn't fully reloaded in the frontend. @lkleuver has created something to monitor wether you made changes compared to the last submitted state. Is this 'last saved state' not recreated?

**And another bug:**
- fill in polling station 33 as above, but...
- after filling in list 1 (without submitting it), go to step 'aantal kiezers en stemmen'.
- click 'invoer afbreken' and 'invoer bewaren'
- open polling station 33 from the blue alert
- you are on "Aantal kiezers en stemmen"
- navigate to list 1
- **observed**: no icon before menu-item "aantal kiezers en stemmen"
-  **expected**: green check icon before menu-item "aantal kiezers en stemmen"
- (also list 1 is not saved in this case, whilst it should (see #421)

praseodym:
> Filled in polling station 33:
> [..]
> * **observed**: modal: 'Wat wil je doen met je invoer' is shown
> * **expected**: modal not shown, as I did not make any changes.
> 
> It looks like the cache/saved state isn't fully reloaded in the frontend. @lkleuver has created something to monitor wether you made changes compared to the last submitted state. Is this 'last saved state' not recreated?

This I can reproduce, I've written a regression test with Playwright and will implement a fix.

> **And another bug:**
> [..]
> * navigate to list 1
> * **observed**: no icon before menu-item "aantal kiezers en stemmen"
> * **expected**: green check icon before menu-item "aantal kiezers en stemmen"

When clicking list 1 to navigate to it, I get the "Let op: niet opgeslagen wijzigingen" modal. Only when I click "Niet bewaren", no icon is shown. Is that what's happening for you as well, or is there no modal in this case?

praseodym:
> > Filled in polling station 33:
> > [..]
> > 
> > * **observed**: modal: 'Wat wil je doen met je invoer' is shown
> > * **expected**: modal not shown, as I did not make any changes.
> > 
> > It looks like the cache/saved state isn't fully reloaded in the frontend. @lkleuver has created something to monitor wether you made changes compared to the last submitted state. Is this 'last saved state' not recreated?
> 
> This I can reproduce, I've written a regression test with Playwright and will implement a fix.

This is fixed in 3dfaf38f2ddd99528578adace6d7e3503dd58066.

> > **And another bug:**
> > [..]
> > 
> > * navigate to list 1
> > * **observed**: no icon before menu-item "aantal kiezers en stemmen"
> > * **expected**: green check icon before menu-item "aantal kiezers en stemmen"
> 
> When clicking list 1 to navigate to it, I get the "Let op: niet opgeslagen wijzigingen" modal. Only when I click "Niet bewaren", no icon is shown. Is that what's happening for you as well, or is there no modal in this case?

This is fixed in 4e70ede631e700ce0f864b7c836d5e33887b49b7.

This commit slightly changes the behaviour: when resuming, now all sections up to the furthest submitted section are mark as saved. Validation results (errors and warnings) are shown if a section is saved. This change means that any validation results are shown right away of if the user went back to a previous section, made some changes, and clicked 'Invoer afbreken' and then 'Invoer opslaan'.

frontend/app/routes.tsx
praseodym: I've added a comment in 3a6fd373c8c4648004a4989e1ea6870ebe9ae8b5, does that make it more clear?

backend/openapi.json
praseodym: @cikzh also mentioned this on Mattermost. For now it's fine because the backend doesn't need the client state, and the type in the frontend is more specific than what  we have in OpenAPI (using `FormSectionID` instead of just `string`).

frontend/lib/api/form/pollingstation/PollingStationFormController.tsx
praseodym: This happens in more places. We should decide what to do and set up linting for this if possible, what do you think?

frontend/lib/api/form/pollingstation/PollingStationFormController.tsx
praseodym: The backend doesn't use this, so that's why it's part of the client state.

We could return it from the form component in a different way as part of #402.

Lionqueen94 (dismissed):

jorisleker:
Great! Issues previously found have been solved. 
One new one
- fill in the forms
- make sure you trigger a warning in step "Aantal kiezers en stemmen"
- accept the warning, fill in the next steps
- go back to 'Aantal kiezers en stemmen'
- click 'invoer afbreken' and 'invoer bewaren'
- resume your entry
- **observed** the "accept warnings" checkbox is unchecked
- **expected** the "accept warnings" checkbox is still checked

nb: the "accept warnings" checkbox remains checked if I save my input from any other step then the step that contains the warning.

praseodym:
> nb: the "accept warnings" checkbox remains checked if I save my input from any other step then the step that contains the warning.

Good find! Should be fixed by 2150b3be12f493d2b3dd0498e687fb5d1b90a809, can you retest?

jorisleker (approved):
Retested, works as expected now!

Lionqueen94 (approved):

jorisleker (changes requested):
~Woops. Approved to soon.~

~the checkbox remains checked now, but the 'volgende' button does not navigate to the next step anymore~

edit: checked: this behavior is present in main as well, will create a separate issue

jorisleker (approved):

PR #427, opened 2024-10-07
Use `Attachment` from `axum-extra` for PDF results

### LINKED ISSUE #372: Axum: use new `Attachment` type
Use the new `Attachment` type in our PDF API endpoint. See https://github.com/tokio-rs/axum/pull/2789.

praseodym:
Resolves #372

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -190,6 +190,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "axum-extra"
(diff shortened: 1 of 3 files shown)


cikzh (approved):
Normally I'd say the few extra lines of code isn't worth the dependency, but as this doesn't pull in any new sub-dependencies, I think it's a nice improvement :+1:

Lionqueen94 (approved):

PR #428, opened 2024-10-07
Fix submit-navigate after navigating back one page

### LINKED ISSUE #426: Submitting step with accepted warnings does not always navigate to next step
Reproduction:
- fill in a polling station and make sure to trigger a warning in step 'voters and votes' (ie: by using same numbers for A - D and E - H)
- click volgende, accept warnings, and click volgende again. 
- now, use the navigation to go back to step 'voters and votes'
- click 'volgende'
- **expected**: since the checkbox is checked: navigate to next step (differences)
- **observed**: you stay on voters and votes.

Compare to:
- Now: click the 'recounted' menu item'
- click volgende to navigate to voters and votes
- click volgende
- **observed** you navigate to differences (as **expected**)

So, this only happens when clicking volgende after opening the step 'voters and votes' via navigation, not when opening the page via volgende'

praseodym:
This issue occurred whenever navigating back by one page. In that case, the 'next section after submit' did not change so the form did not navigate to the expected next section. The 'accept warnings' checkbox does not seem to matter.

Fixes #426.

### DIFF
--- a/frontend/app/module/data_entry/PollingStation/PollingStation.test.tsx
+++ b/frontend/app/module/data_entry/PollingStation/PollingStation.test.tsx
@@ -350,7 +350,24 @@ describe("Polling Station data entry integration tests", () => {
       }
     });
 
-    test("Navigate to next page after navigating back and submitting", async () => {
+    test("Navigate to next page after navigating back one page submitting", async () => {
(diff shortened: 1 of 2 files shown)


jorisleker (approved):
Retested, now the 'volgende' button works reliably

Issue #305, opened 2024-09-05
Navigating form sections and focus state


jorisleker:
- When navigating between form sections, we currently set the focus to the first input on the next page. 
- When there are unaccepted warnings or errors that prevent navigation, we scroll up the page, but we don't change the focus state. Focus remains were the user was before submitting the form section.

Both are not [recommended practice for single page applications](https://johnsweetaccessibility.com/2020/05/accessibility-in-spas-part-2/). 

What needs to change:
- [x] When **successfully** navigating to another form section, set focus to the h2 (this is the header that describes the content of the page, and corresponds to the menu-item clicked). 
- [x] When an error or unaccepted warning prevents a navigation attempt, set focus to the first feedback header 

nb: To be able to set focus to a non interactive element like an `h2` or `header`, add a `tabindex="-1"` attribute it (this will make it possible to focus the element, whilst ignoring it when tabbing through the page.
 
We might need something to indicate to the user that focus is on the first h2/feedback element. Will look into this after this ticket has been merged.

Something else that needs fixing:
- [x] The `ENTER` to move to another form field logic we build, prevents using `ENTER` from being used to confirm your focus on other elements. i.e.: when I focus an item in the progress list or a link in the header and press `ENTER` I expect to navigate to that step/page. Instead, the focus is moved to the 'next' form field. This should only happen when the focus is on another form field, not when it is somewhere else on the page.

praseodym:
If we don't focus the first input field, users will first need to press the tab key to start data entry?

jorisleker:
> If we don't focus the first input field, users will first need to press the tab key to start data entry?

Yes that's right. This is the way it should be done if we want to pass accessibility audits. Am open to discussion about this though, because we might unintentionally force more users to use their mouse to focus the first input on the page if we implement this. My goal would be to have as much users as possible use their keyboard only when entering numbers, because that's the more efficient way to do it. A strong focus indicator on the focussed heading might help here. 

Main issue that I want to tackle is outlined in the second bullet. If you submit a form section, and there are warnings/errors, something funny happens with the focus: it remains where it was (when using mouse to click 'volgende') or it is moved to the next input (if you use `SHIFT` + `ENTER` to submit). Both are confusing.

PR #435, opened 2024-10-08
Implement custom radiobuttons + choicelist component

Lionqueen94:
## Description
- [x] Refactored current styling of custom checkbox to make it easier to use (no manual div toggling needed) and to make it suitable to use with radio buttons as well. Thanks @oliver3 for the help!
- [x] Added radio button and intermediate checkbox styling. 
- [x] Added ChoiceList component (fieldset) to group multiple inputs.

## What to test
- [ ] Is the behaviour of the checkbox and radio inputs as desired
- [x] Is the styling of the radio inputs and checkbox (still) as designed
- [ ] Is it ok that you sometimes have to click a radio button twice to check it (is this normal?) @jorisleker


jorisleker (changes requested):
- ✅ styling of radio's looks okay (all states)
- ✅ styling and behavior of checkbox is still ok. (can't retest a group of checkboxes though, because we don't have that yet)
- ❌ when hovering, the label has a hand/pointer cursor, the radio has a default cursor. Could we use pointer for both?
- ❌ padding/margin left of radio's should be removed; radio's should left align with elements above and below it
- ❌ when using tab to navigate through the page, it is now impossible to continu to the submit button from the choice group. Please compare to vanilla html form, behavior should be same. 
- ❌ having to click a radio twice is not expected behavior. Don't know exactly when/how this happens, but it should not. Clicking a radiobutton or it's label should immediately change it's state to checked.

Lionqueen94:
> Tested on MacOS (Chrome, Safari, Firefox):
> 
>     * ✅ styling of radio's looks okay (all states)
> 
>     * ✅ styling and behavior of checkbox is still ok. (can't retest a group of checkboxes though, because we don't have that yet)
> 
>     * ❌ when hovering, the label has a hand/pointer cursor, the radio has a default cursor. Could we use pointer for both?
> 
>     * ❌ padding/margin left of radio's should be removed; radio's should left align with elements above and below it
> 
>     * ❌ when using tab to navigate through the page, it is now impossible to continu to the submit button from the choice group. Please compare to vanilla html form, behavior should be same. (might be fixed in [Navigating form sections and focus state #305](https://github.com/kiesraad/abacus/issues/305)?)
> 
>     * ❌ having to click a radio twice is not expected behavior. Don't know exactly when/how this happens, but it should not. Clicking a radiobutton or it's label should immediately change it's state to checked.

Fixed the cursor, padding and margin in ee2a9bb68e47997ca8e6eba521efddb11ca656e6
The navigate issue should indeed be fixed by merging my other pr
We will investigate the sometimes needed double click and hopefully figure out how to fix it

jorisleker (approved):

jorisleker:
LGTM now @Lionqueen94! Let's make sure to check the tab/navigation issue after merging this to main

oliver3 (approved):

Issue #433, opened 2024-10-08
Create UI Table component


oliver3:
- [x] Create basic table component - https://github.com/kiesraad/abacus/pull/521
- [x] Use component instead of html tables throughout the application - https://github.com/kiesraad/abacus/pull/563

#542 and #506 will be implemented at a later stage

Notes:
- Number columns have proper spacing between integers (e.g. polling station numbers, see [this comment](https://github.com/kiesraad/abacus/pull/453#discussion_r1804763599))

jorisleker:
re: Number columns have proper spacing between integers

We should only do this for number columns where comparing/adding numbers is expected. Ie: columns with counts or other statistics. In situations where the presented number is an **identifier** (ie polling station number, candidate number, list number), I prefer using the default font + 'normal' letterspacing that is optimized for legibility.

Issue #431, opened 2024-10-08
Create polling station


oliver3:
- Add button on list
- Plus icon from Figma 
- Create backend endpoint
- Create frontend form
  - nummer: length 6
- Validation on number unique + show error
- Validation on required
- Create frontend page
- Show confirmation alert success

Issue #277, opened 2024-08-22
Improve error pages design


praseodym:
We're working on an improved error page design ([see Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=3933-59177&t=PubTloYfoXHhAha9-4)).

Once the design is finished this should be implemented in the frontend.
- The 404 Not Found page should be the handler for invalid frontend URLs 
- The 500 Internal Server Error page should be rendered for errors from the backend that we cannot easily recover from

PR #453, opened 2024-10-14
Add Polling Station List

### LINKED ISSUE #430: Polling station list
- [x] Restore Polling Station list endpoint
- [x] Create frontend page, use simple html table with existing class for now. Sorting (and separate component) will be added in another sub-issue.

No export functionality for now

oliver3:
Closes #430 

- Add polling station list API endpoint and tests
- Add polling station list frontend page and test
- Add mock endpoint handler

Page can be accessed through `/dev` page and should list polling stations with backend and with mock server


cikzh (changes requested):
Looks good! Only one minor suggestion.

Congrats on the first substantial PR :)!

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
```diff
@@ -0,0 +1,55 @@
+import { PollingStationType, usePollingStationList } from "@kiesraad/api";
+import { IconChevronRight } from "@kiesraad/icon";
+
+function pollingStationType(pollingStationType: PollingStationType): string {
```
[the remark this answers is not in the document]
oliver3: Hi, thanks for your review! 

This function already does an exhaustiveness check: because of the return type string it is not allowed to return undefined when you did not handle all cases (`TS2366: Function lacks ending return statement and return type does not include undefined`). I was myself looking at a `default` with something like `never` but that seemed not to be needed in this case.

Would you still prefer separating out a lookup object?

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <>
+      <header>
+        <section>
+          <h1>Stembureaus</h1>
+        </section>
+      </header>
+      <main>
+        <article>
+          <table id="polling_stations" className="polling_stations_table">
```
lkleuver, line 27: Suggestion for later: Create Table component with css module

frontend/lib/api/polling_station/PollingStationListProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+  undefined,
+);
+
+export interface PollingStationListProviderProps {
+  children: React.ReactNode;
+  electionId: number;
+}
+
+export function PollingStationListProvider({ children, electionId }: PollingStationListProviderProps) {
+  const { data, loading } = usePollingStationListRequest({ election_id: electionId });
```
lkleuver, line 19: Does this need to be in the global context / via a provider, instead of using 
usePollingStationListRequest() directly in the PollingStationListPage ?

lkleuver (dismissed):

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
Lionqueen94: @lkleuver We have already created an issue for this: https://github.com/kiesraad/abacus/issues/433 :)

frontend/lib/api/polling_station/PollingStationListProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  if (loading) {
+    return null;
+  }
+
+  if (!data) {
+    return <div>Error no polling stations data</div>;
+  }
```
Lionqueen94, line 27: Maybe it's good to make this a little more friendly and in Dutch. Something like: `<div>Er zijn nog geen stembureaus aangemaakt voor deze verkiezing.</div>`

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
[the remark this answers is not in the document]
cikzh: > This function already does an exhaustiveness check: because of the return type string it is not allowed to return undefined when you did not handle all cases (`TS2366: Function lacks ending return statement and return type does not include undefined`). I was myself looking at a `default` with something like `never` but that seemed not to be needed in this case.

I didn't know TS would catch this, fair! The error message is a little more cryptic then when (for example) you leave out a variant in the typed object:
```
typescript [2741]: Property 'Bijzonder' is missing in type '{ Mobiel: string; VasteLocatie: string; }' but required in type '{ VasteLocatie: string; Bijzonder: string; Mobiel: string; }'.
```

> Would you still prefer separating out a lookup object?

I was thinking of creating the typed object instead of the function and just perform the lookup instead of doing a function call. To me it's a little more readable, but I guess it's a matter of taste.

frontend/app/module/polling_stations/PollingStationsLayout.tsx
```diff
@@ -0,0 +1,23 @@
...
+export function PollingStationsLayout() {
+  const electionId = useNumericParam("electionId");
+
+  return (
+    <div className="app-layout">
+      <PageTitle title="Overzicht verkiezingen - Abacus" />
+      <PollingStationListProvider electionId={electionId}>
+        <NavBar />
+        <Outlet />
+        <Footer />
```
jschuurk-kr, line 19: There's a double footer on the page, see https://f174b725.kiesraad-abacus.pages.dev/elections/1/polling-stations

Might be because of the `<Footer />` in the `OverviewLayout`?

frontend/app/module/election/page/ElectionHomePage.tsx
jschuurk-kr, line 1: Arguably not in scope of this PR, but the nav bar of this page and its sub-pages is rather empty:
- /elections/1#coordinator only has the role in the top right
- /elections/1/polling-stations#coordinator only has the role in the top right
- /elections/1/status#coordinator has "municipal election" in the top left and the role in the top right

The [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=5442-50666&t=ST8OH49dNClph5bd-1) has "GSB 045 Juinen — Gemeenteraadsverkiezingen 2026". For now I think it would be good to have the name of the election in the nav bar, i.e. what the `ElectionStatusPage` does.

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
```diff
@@ -0,0 +1,55 @@
...
+            <thead>
+              <tr>
+                <th>Nummer</th>
```
jschuurk-kr, line 30: Should we consider a monospaced font for the numbers? Our `0` is significantly wider than the other integers. 
cc @jorisleker 

image

frontend/lib/api/polling_station/PollingStationListProvider.tsx
jschuurk-kr: Current solution is consistent with `ElectionListProvider`. If we want to make the error more friendly, we should also have more on the page than just the error message. And I think that's in scope of #277 Improve error pages design.

backend/tests/polling_station_integration_test.rs
```diff
@@ -0,0 +1,39 @@
...
+use sqlx::SqlitePool;
+
+use backend::polling_station::PollingStationListResponse;
+
+use crate::utils::serve_api;
+
+mod utils;
+
+#[sqlx::test(fixtures(path = "../fixtures", scripts("elections", "polling_stations")))]
+async fn test_polling_station_listing_works(pool: SqlitePool) {
```
jschuurk-kr, line 13: Am not a big fan of "works" in a test name, because it's not very specific. On the other hand, just noticed we do the same in `election_integration_test.rs`.

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
oliver3: This is actually the use case for [`font-variant-numeric: tabular-nums`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric) but I don't know if our font supports that.

frontend/app/module/polling_stations/page/PollingStationListPage.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import { PollingStationListPage } from "app/module/polling_stations";
+import { render } from "app/test/unit";
+
+import { PollingStationListProvider } from "@kiesraad/api";
+
+describe("PollingStationListPage", () => {
+  test("Show polling stations", async () => {
+    render(
+      <PollingStationListProvider electionId={1}>
```
jschuurk-kr, line 12: It doesn't matter which `electionId` you provide here, since the mock will always return data.

Should we have the mock return a 404 for `electionId` `404`? Or solve this in a different way? Because then we can also build an RTL test for the "Error no polling stations data" message.

frontend/lib/api/polling_station/PollingStationListProvider.tsx
oliver3: Was doubting about that but the Provider also handles the loading and error state (although both are not yet really implemented), so that the `PollingStationListPage` does not have to deal with that. Is that ok with you?

backend/tests/polling_station_integration_test.rs
oliver3: That was indeed my blueprint for this PR. Not sure if I should change it here, or on both locations, or as a separate PR throughout the code?

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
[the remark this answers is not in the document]
oliver3: > I was thinking of creating the typed object instead of the function ...
Ah yes that turns out to be a better solution, thanks! Implemented in https://github.com/kiesraad/abacus/pull/453/commits/b96547431de16c6db2e0051f082c694b5515d2f1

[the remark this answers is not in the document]
oliver3: > Shouldn't the PollingStationType be english? looking for instance at:

Good suggestion! Updated the backend enum and all that followed in https://github.com/kiesraad/abacus/commit/b96547431de16c6db2e0051f082c694b5515d2f1

frontend/app/module/election/page/ElectionHomePage.tsx
oliver3: Will do in a separate PR

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
oliver3: Please discuss further in # https://github.com/kiesraad/abacus/issues/433 Create UI Table component

frontend/lib/ui/style/table.css
```diff
@@ -59,6 +60,13 @@ table.overview_table {
   }
 }
 
+table.polling_stations_table {
+  td.number,
+  th.number {
+    width: 10rem;
+  }
+}
```
Lionqueen94, line 68: On the polling station list on the polling station selector page the width of the number column is set to 6.5rem, which is long enough for 6 digit polling station numbers. Should we standardize this everywhere to the same width? The 6.5rem is set with an inline width, which should be set in css when the separate Table component is created I think.

frontend/lib/api/polling_station/PollingStationListProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+export function PollingStationListProvider({ children, electionId }: PollingStationListProviderProps) {
+  const { data, loading } = usePollingStationListRequest({ election_id: electionId });
+
+  if (loading) {
+    return null;
+  }
+
+  if (!data || !data.polling_stations.length) {
+    return <div id="no-polling-station-data">Er zijn nog geen stembureaus aangemaakt voor deze verkiezing.</div>;
+  }
```
Lionqueen94, line 27: I think that when there are no polling stations created yet for the election, there should not be this sentence but just a text in place of the table and the create button should be shown (added in the next issue). Currently there will be an empty age with this sentence shown.

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+                <th>Naam</th>
+                <th>Soort</th>
+                <th></th>
+              </tr>
+            </thead>
+            <tbody>
+              {pollingStations.map((station) => (
+                <tr key={station.id}>
+                  <td className={"number"}>{station.number}</td>
```
Lionqueen94, line 34:
```suggestion
                  <td className="number">{station.number}</td>
```

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <section>
+          <h1>Stembureaus</h1>
+        </section>
+      </header>
+      <main>
+        <article>
+          <table id="polling_stations" className="polling_stations_table">
+            <thead>
+              <tr>
+                <th>Nummer</th>
```
Lionqueen94, line 25:
```suggestion
                <th className="number">Nummer</th>
```

jorisleker (changes requested):
Not sure which element is causing this, but the `Nummer` column is wider then it needs to be (and stays too wide when making the viewport very narrow)

Scherm­afbeelding 2024-10-18 om 11 01 42

frontend/lib/ui/style/table.css
jorisleker: Expected max length for polling station numbers is 3 digits, so I would propose using the length of the 'Nummer' labels as the entity that dictates the width of this column. Anything wider is not needed; space should be used by the other columns. See screenshot. 
Scherm­afbeelding 2024-10-18 om 11 01 42

frontend/lib/api/polling_station/PollingStationListProvider.tsx
lkleuver: I just worry about the pattern, will we add providers for all lists for example?

frontend/lib/api/gen/openapi.ts
```diff
@@ -192,7 +195,7 @@ Bijlage 2: uitkomsten per stembureau" from the
 export interface PollingStationResults {
   differences_counts: DifferencesCounts;
   political_group_votes: PoliticalGroupVotes[];
-  recounted?: boolean;
+  recounted?: boolean | null;
```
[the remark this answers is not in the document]
oliver3: Resolved in https://github.com/kiesraad/abacus/pull/475

frontend/lib/ui/style/table.css
oliver3: Set to 6.5rem for now in https://github.com/kiesraad/abacus/pull/453/commits/26acb7b4e4300c1ec30d77b80e48b55669a0660d

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
oliver3: Fixed in https://github.com/kiesraad/abacus/pull/453/commits/26acb7b4e4300c1ec30d77b80e48b55669a0660d

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
oliver3: Added in https://github.com/kiesraad/abacus/pull/453/commits/26acb7b4e4300c1ec30d77b80e48b55669a0660d

frontend/app/module/DevHomePage.tsx
```diff
@@ -43,6 +43,9 @@ export function DevHomePage() {
               <li>
                 <Link to={`/elections/1#coordinator`}>Verkiezing 1</Link>
               </li>
+              <li>
+                <Link to={`/elections/1/polling-stations#coordinator`}>Verkiezing 1 stembureaus</Link>
```
marjoleintamis, line 47: De woordvolgorde ziet er hier vreemd uit, zou 'Stembureaus Verkiezing 1' moeten zijn. Ik kan hier geen suggestie doen (knopje ontbreekt).

documentatie/referentie/woordenlijst-EN-NL.md
```diff
@@ -12,6 +12,9 @@ Voor de vertaling van begrippen uit de wet naar variabelen en functienamen in on
 | Kiesraad | KR | Electoral Council | |
 | Stemlokaal | | Polling Place | Deze zouden we in de code niet moeten gebruiken |
 | Stembureau | SB | Polling Station | |
+| Vaste Locatie | | Fixed Location | Stembureau type |
```
marjoleintamis, line 15:
```suggestion
| Vaste Locatie | | Fixed Location | Stembureautype |
```

documentatie/referentie/woordenlijst-EN-NL.md
```diff
@@ -12,6 +12,9 @@ Voor de vertaling van begrippen uit de wet naar variabelen en functienamen in on
...
+| Mobiel | | Mobile | Stembureau type |
```
marjoleintamis, line 16:
```suggestion
| Mobiel | | Mobile | Stembureautype |
```

documentatie/referentie/woordenlijst-EN-NL.md
```diff
@@ -12,6 +12,9 @@ Voor de vertaling van begrippen uit de wet naar variabelen en functienamen in on
...
+| Bijzonder | | Special | Stembureau type |
```
marjoleintamis, line 17:
```suggestion
| Bijzonder | | Special | Stembureautype |
```

frontend/app/module/polling_stations/page/PollingStationListPage.test.tsx
jschuurk-kr: Discussed. Test for empty list has been added. Test for 404 not because handled in different component.

jorisleker:
@oliver3 reviewed again, number column looks good now, so will approve this PR.

Do you want to add the [empty state](https://github.com/kiesraad/abacus/pull/453#discussion_r1810075896) as part of this PR, or merge this first and add the empty state in a separate PR?

jorisleker (approved):

frontend/app/module/polling_stations/PollingStationsLayout.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { PollingStationListProvider } from "@kiesraad/api";
+import { PageTitle } from "@kiesraad/ui";
+import { useNumericParam } from "@kiesraad/util";
+
+export function PollingStationsLayout() {
+  const electionId = useNumericParam("electionId");
+
+  return (
+    <div className="app-layout">
+      <PageTitle title="Overzicht verkiezingen - Abacus" />
```
Lionqueen94, line 14: The PageTitle here needs to be moved to PollingStationsListPage and the title renamed

frontend/lib/api/polling_station/PollingStationListProvider.tsx
oliver3: Now that the empty state is specified, in the Page, there is really no excuse anymore for the Provider to exist :) I've removed it.

frontend/lib/api/polling_station/PollingStationListProvider.tsx
oliver3: I've added the empty state, without any buttons for now. The Handmatig invullen will be added for issue https://github.com/kiesraad/abacus/issues/431

frontend/app/module/polling_stations/PollingStationsLayout.tsx
oliver3: Moved to page in https://github.com/kiesraad/abacus/pull/453/commits/acb94c041a5dd7c7d9a4c24414812b98760791ec

lkleuver (dismissed):

lkleuver (dismissed):

oliver3:
> Do you want to add the [empty state](https://github.com/kiesraad/abacus/pull/453#discussion_r1810075896) as part of this PR, or merge this first and add the empty state in a separate PR?

I've added it in this PR, and it can be seen in https://c81e4c30.kiesraad-abacus.pages.dev/elections/3/polling-stations#coordinator

frontend/lib/api/gen/openapi.ts
```diff
@@ -182,7 +195,7 @@ Bijlage 2: uitkomsten per stembureau" from the
...
```
Lionqueen94, line 198:
```suggestion
  recounted?: boolean;
```

backend/openapi.json
jschuurk-kr, line 1: When I preview this file with my VS Codium plugin, I get 6 resolver errors (number of errors depends on which part of the spec I fold out). For the `polling_stations` endpoint the Example Value says "string" and the Schema `$ref#/components/schemas/PollingStationListResponse`.

On `main` branch the file is rendered by the plugin without issues.

jorisleker (approved):
UX is as designed 👌

backend/openapi.json
jschuurk-kr: When I have https://validator.swagger.io/ check the file by giving it a link to the raw file on GitHub, it parses and renders the file just fine. :man_shrugging:

jschuurk-kr: Example resolver error on an unchanged part of the file:
```
Resolver error at responses.200.content.application/json.schema.$ref
Could not resolve reference: Evaluation failed on URI: "vscode-webview://0t4dlamu8hgmrbe3p3g63dfg1jmp3nvqe9oqp08mb38idouacrm3/index.html?id=79361920-6651-4bc0-9887-4fd7d1d39fc0&origin=6c4f5b9c-0061-4766-afd4-9655c3d03df8&swVersion=4&extensionId=42Crunch.vscode-openapi&platform=electron&vscode-resource-base-authority=vscode-resource.vscode-cdn.net&parentOrigin=vscode-file%3A%2F%2Fvscode-app#/components/schemas/ElectionListResponse"
```

There are two issues about resolver errors reported on the plugin. Does require a circular reference in the spec:
- https://github.com/42Crunch/vscode-openapi/issues/147
- https://github.com/42Crunch/vscode-openapi/issues/286

frontend/lib/api/gen/openapi.ts
Lionqueen94: Because you have to manually re-run the gen command, unfortunately that's not built into commit/push yet

backend/openapi.json
jschuurk-kr: The plugin also supports a preview using ReDoc. That one works fine. So it's only the preview using Swagger that gives issues.

jschuurk-kr: Cause is the update to version 3.1.0 in the openapi spec: https://github.com/kiesraad/abacus/pull/476/files#diff-0cabbcfedc2cff27c017284dc9284b72172e9dd89cd501f5ffbceb7af5d7e48a

Since it's specific to one type of preview of a VS Code plugin, let's not roll that back. :-)

jorisleker (approved):

PR #459, opened 2024-10-15
i18n MVP

lkleuver:
Exploring the approach of implementing our own i18n library, reducing dependencies and security risks. But in a way that would make it "easy" to replace it with i18next


frontend/lib/i18n/i18n.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return Object.entries(vars).reduce(
+      (acc, [key, value]) => (typeof value === "string" ? acc.replace(`$${key}`, value) : acc),
+      translate(k),
+    );
+  }
+
+  return translate(k);
+}
+
+export function tx(k: keyof Translation, vars?: Record<string, RenderCallback | string>): Array<string | ReactElement> {
```
oliver3, line 24: Can you add some JSDoc to this function? The parameter names `k` and `vars` are not really speaking to me.

oliver3, line 24: Please add some specific (unit) tests for this rather complex code

oliver3 (changes requested):

marlonbaeten:
Note: this adds translation functions and contains some examples on how to use them. We need this to start translating all interfaces.

jorisleker:
> Note: this adds translation functions and contains some examples on how to use them. We need this to start translating all interfaces.

Note: we agreed that we currently only need dutch localization. We are adding i18n now because adding it later will be a lot more work. But adding english (or other) translations is out of scope for now.

jorisleker (dismissed):

marlonbaeten:
> > Note: this adds translation functions and contains some examples on how to use them. We need this to start translating all interfaces.
> 
> Note: we agreed that we currently only need dutch localization. We are adding i18n now because adding it later will be a lot more work. But adding english (or other) translations is out of scope for now. So no need to 'start translating' all interfaces. What needs to happen is move all content that needs translation out of our templates and into i18n files.

Correct! What I meant is that we need to move plain text from templates into the language file - only for Dutch for now. I'll remove the English language in to make this more clear.

cikzh (approved):
Nice work @lkleuver and @marlonbaeten !

jorisleker (approved):

frontend/app/module/election/page/ElectionStatusPage.tsx
```diff
@@ -28,18 +29,15 @@ export function ElectionStatusPage() {
...
-          <h2>Alle stembureaus zijn twee keer ingevoerd</h2>
-          <p>
-            De resultaten van alle stembureaus in jouw gemeente zijn correct ingevoerd. Je kunt de uitslag nu definitief
-            maken en het proces verbaal opmaken. Doe dit alleen als er vandaag niks meer herteld hoeft te worden.
-          </p>
+          <h2>{t("election_status.difinitive.title")}</h2>
+          <p>{t("election_status.difititive.message")}</p>
           <Button onClick={finishInput} size="md">
-            Invoerfase afronden
+            {t("election_status.difititive.finish_button")}
```
oliver3, line 40: If we're not translating to English for now, should we have English or Dutch keys? 

Finding good English keys is almost like translating already.. (triggered by this misspelling of "definitive")

jorisleker: Would suggest using english, because this will make it a lot easier to spot missing translations. Unless the translation engine does something fancy to mark missing strings.

PR #471, opened 2024-10-17
Introduce error boundary and basic error page styling

### LINKED ISSUE #277: Improve error pages design

marlonbaeten:
Fixes #277, #472 and #474


marlonbaeten:
Added #472 #473 and #474 to improve error handling further and introduce the error modal.

frontend/app/component/error/Error.tsx
```diff
@@ -0,0 +1,42 @@
+import { useNavigate } from "react-router-dom";
+
+import { IconArrowLeft } from "@kiesraad/icon";
+
+import { Button } from "../../../lib/ui/Button/Button";
```
cikzh, line 5:
```suggestion
import { Button } from "@kiesraad/ui
```

cikzh (changes requested):

jorisleker (changes requested):
Requesting `elections/1/data-entry/123123`
Expected to get a 404 (not found), got a 500 (server error)

Scherm­afbeelding 2024-10-18 om 11 19 20

Some design feedback:
- Button 'back to previous page' should have an arrow point left in stead of right
- Heading 'abacus is stuk' is the only heading on the page, thus should be `h1`
- font size should be `3rem`, now 3.75/60px
- line-height should be 125%

marlonbaeten:
> * font size should be `3rem`, now 3.75/60px

There does not seem to be a variable for this size:

```css
    --font-size-body: 1.125rem;
    --font-size-xs: 0.75rem;
    --font-size-sm: 0.875rem;
    --font-size-md: 1rem;
    --font-size-lg: 1.8rem;
    --font-size-xl: 3.75rem;
```

Should I add one?

marlonbaeten:
> * Button 'back to previous page' should have an arrow point left in stead of right

I used the icon `<IconArrowLeft />` should I change this icon to point the "right" direction?

jorisleker:
> > * font size should be `3rem`, now 3.75/60px
> 
> There does not seem to be a variable for this size:
> 
> ```css
>     --font-size-body: 1.125rem;
>     --font-size-xs: 0.75rem;
>     --font-size-sm: 0.875rem;
>     --font-size-md: 1rem;
>     --font-size-lg: 1.8rem;
>     --font-size-xl: 3.75rem;
> ```
> 
> Should I add one?

The style for the heading on this page is unique within Abacus, so I think it has not been defined yet. On the other hand, I don't recoginize these font-sizes / they don't match the font sizes defined in [Figma (page: fundamentals)](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2723-64563&t=EyzKwX91IZxGpFoB-4)

@Lionqueen94 can you help out?

marlonbaeten:
> > * Button 'back to previous page' should have an arrow point left in stead of right
> 
> I used the icon `<IconArrowLeft />` should I change this icon to point the "right" direction?

This seems to be an error (there is no usage of this icon elsewhere) - I'll correct it.

marlonbaeten:
> The style for the heading on this page is unique within Abacus, so I think it has not been defined yet. On the other hand, I don't recoginize these font-sizes / they don't match the font sizes defined in [Figma (page: fundamentals)](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2723-64563&t=EyzKwX91IZxGpFoB-4)
> 
> @Lionqueen94 can you help out?

I will change the larges size to `3rem` for now - since it is not used anywhere else.

Lionqueen94:
> > > * font size should be `3rem`, now 3.75/60px
> > 
> > 
> > There does not seem to be a variable for this size:
> > ```css
> >     --font-size-body: 1.125rem;
> >     --font-size-xs: 0.75rem;
> >     --font-size-sm: 0.875rem;
> >     --font-size-md: 1rem;
> >     --font-size-lg: 1.8rem;
> >     --font-size-xl: 3.75rem;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should I add one?
> 
> The style for the heading on this page is unique within Abacus, so I think it has not been defined yet. On the other hand, I don't recoginize these font-sizes / they don't match the font sizes defined in [Figma (page: fundamentals)](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2723-64563&t=EyzKwX91IZxGpFoB-4)
> 
> @Lionqueen94 can you help out?

I think this is still a remnant from the POC for the data entry page made before I started, it can and should definitely be corrected to the current design.

Lionqueen94:
> > * Button 'back to previous page' should have an arrow point left in stead of right
> 
> I used the icon `<IconArrowLeft />` should I change this icon to point the "right" direction?

Oh lol indeed, this one has the wrong name, and there should be one pointing left added indeed.

jorisleker:
> > The style for the heading on this page is unique within Abacus, so I think it has not been defined yet. On the other hand, I don't recoginize these font-sizes / they don't match the font sizes defined in [Figma (page: fundamentals)](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2723-64563&t=EyzKwX91IZxGpFoB-4)
> > @Lionqueen94 can you help out?
> 
> I will change the larges size to `3rem` for now - since it is not used anywhere else.

- Heading font should be DM Sans here 
- Arrow should have a stroke width comparable to other icons (I think it defaults to 1px now, should be a bit wider)
- Can we add a max-width to the `<p>` element? (think we have that defined on other places as well, please use value used in other places like alerts)

marlonbaeten:
> * Heading font should be DM Sans here
> * Arrow should have a stroke width comparable to other icons (I think it defaults to 1px now, should be a bit wider)
> * Can we add a max-width to the `<p>` element? (think we have that defined on other places as well, please use value used in other places like alerts)

Fixed :+1:

frontend/app/component/error/Error.module.css
```diff
@@ -0,0 +1,38 @@
+.errorContainer {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+}
+
+.errorContainer h1 {
+  font-size: var(--font-size-xl);
+  font-family: var(--font-text);
+  font-weight: 700;
```
Lionqueen94, line 10:
```suggestion
  font-weight: bold;
```
we always use `bold` instead of 700 and `normal` instead of 400 to make it more readable :)

frontend/app/component/error/Error.module.css
```diff
@@ -0,0 +1,38 @@
...
+.errorStack {
+  flex: 1 1 auto;
+  padding: 2.5rem 5rem;
+  border-bottom: 1rem solid var(--bg-gray);
+}
+
+.errorStack h2 {
+  font-size: var(--font-size-lg);
+  font-family: var(--font-text);
+  font-weight: 700;
```
Lionqueen94, line 32:
```suggestion
  font-weight: bold;
```

frontend/lib/api/election/ElectionStatusProvider.tsx
```diff
@@ -20,7 +22,7 @@ export function ElectionStatusProvider({ children, electionId }: ElectionStatusP
   });
 
   if (error) {
-    throw new Error("Could not fetch election statuses");
+    throw new NotFoundError("Could not fetch election statuses");
```
Lionqueen94, line 25: Shall we make this error message in Dutch as well?

frontend/app/component/error/ErrorBoundry.tsx
```diff
@@ -0,0 +1,22 @@
+import { useRouteError } from "react-router-dom";
```
Lionqueen94, line 1: Can you correct the filename to `ErrorBoundary.tsx`?

frontend/lib/icon/generated.tsx
```diff
@@ -7,12 +7,8 @@ export const IconArrowBlockUp = (props: React.SVGAttributes<SVGElement>) => (
...
 export const IconArrowLeft = (props: React.SVGAttributes<SVGElement>) => (
-  <svg {...props} x="0px" y="0px" viewBox="0 0 24 24" role="img">
-    <path
-      d="M12,20c-0.3,0-0.5-0.1-0.7-0.3c-0.4-0.4-0.4-1,0-1.4l5.3-5.3H5c-0.6,0-1-0.4-1-1s0.4-1,1-1h11.6l-5.3-5.3
-	c-0.4-0.4-0.4-1,0-1.4s1-0.4,1.4,0l7,7c0.1,0.1,0.2,0.2,0.2,0.3c0,0.1,0.1,0.2,0.1,0.4l0,0c0,0,0,0,0,0l0,0c0,0.1,0,0.3-0.1,0.4
-	c0,0.1-0.1,0.2-0.2,0.3l-7,7C12.5,19.9,12.3,20,12,20z"
-    />
+  <svg {...props} viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" role="img">
+    <path d="M15 8H1M1 8L8 15M1 8L8 1" stroke-width="2" />
   </svg>
```
Lionqueen94, line 12: The icons are generated from the `lib/icon/svg` list, please update the icon there and then run `npm run gen:icons` to update this file.

marlonbaeten: Thanks - good to know :+1:

jorisleker (dismissed):

frontend/app/component/error/Error.module.css
marlonbaeten: Fixed in 53ffbfba4f703f64c491a1eb853a092e4c62b807

frontend/lib/api/election/ElectionStatusProvider.tsx
marlonbaeten: I was planning to properly translate with the i18n PR - to prevent duplicate work.

cikzh (dismissed):

jorisleker:
@marlonbaeten 
- How can I test the modal? 
- Same for stack trace? 
- How do I switch between dev and production mode?

If the answer to the question involves 'break the code somewhere': can you provide screenshots of your implementation?

marlonbaeten:
> * How can I test the modal?

Run the frontend with `npm run dev:server` and the backend with `cargo run` go to `http://localhost:3000/elections/1/data-entry/1/voters-and-votes` then quit the backend and click "Volgende".

> * Same for stack trace?

Trigger a error somewhere `throw new Error("Paniek!");` for instance in `/home/marlon/tweedegolf/abacus/frontend/app/module/RootLayout.tsx`.

> * How do I switch between dev and production mode?

Toggle the boolean in /home/marlon/tweedegolf/abacus/frontend/lib/util/env.ts;

Lionqueen94:
image
There is now an empty Feedback visible when you kill the backend and press Volgende and get the 500 error modal. See in the background of the above screenshot.

frontend/app/component/error/ErrorModal.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <div className="content">
+        <p>
+          <strong>
+            Foutcode: <code>{error.code}</code>
+          </strong>
+        </p>
+        <p>{error.error}</p>
+        <Button variant="default" onClick={() => {}}>
+          Melding sluiten
+        </Button>
```
Lionqueen94, line 20: This button does not do anything

frontend/app/component/error/Error.tsx
```diff
@@ -0,0 +1,56 @@
...
+            {children}
+            {action === ErrorAction.Back && (
+              <Button
+                size="lg"
+                variant="secondary"
+                onClick={() => {
+                  navigate(-1);
+                }}
+              >
+                <Icon icon={<IconArrowLeft />} size="sm" color="primary" />
```
Lionqueen94, line 36:
```suggestion
                <Icon icon={<IconArrowLeft />} size="md" color="primary" />
```
In the design the icon has a size of 1.5rem

Lionqueen94:
The error page looks good and works well! Only a few small comments, mostly about the error modal. Nice work!

frontend/app/component/error/ErrorModal.tsx
marlonbaeten: Fixed in 076879b4da23dbabdb80c84c7b3452896459e691

frontend/app/component/error/Error.tsx
marlonbaeten: Fixed in 076879b4da23dbabdb80c84c7b3452896459e691

frontend/lib/ui/Feedback/Feedback.stories.tsx
```diff
@@ -33,10 +35,8 @@ export const MultipleWarnings: Story = () => {
 
 export const SingleServerError: Story = () => {
```
Lionqueen94, line 36: Can you move this test to a new file next to ErrorModal.tsx since it doesn't test the Feedback component anymore.

marlonbaeten: Like this? https://github.com/kiesraad/abacus/pull/471/commits/92e23d3400883762662829565dede231c23dee8f

Lionqueen94 (dismissed):

frontend/app/component/error/Error.module.css
```diff
@@ -0,0 +1,38 @@
+.errorContainer {
```
lkleuver, line 1: I think the consensus will be to use kebab/dash case for classnames. 
anyway, nothing case sensitive :)

frontend/app/component/error/Error.types.ts
```diff
@@ -0,0 +1,13 @@
+export enum ErrorAction {
+  None = "none",
+  Back = "back",
+}
+
+export class NotFoundError extends Error {
```
lkleuver, line 6: i have no suggestion, but this is not a type 😬

frontend/app/component/error/ErrorBoundary.tsx
lkleuver, line 1: With the title of this PR I'm expecting something "root level ish" like Reacts ErrorBoundary (Or Brian Vaughn [implementation](https://github.com/bvaughn/react-error-boundary))

Wouldnt we need a "root catcher" somewhere? Or does react-router handle this?

frontend/lib/util/env.ts
```diff
@@ -0,0 +1 @@
+export const isDevelopment = process.env.NODE_ENV === "development";
```
lkleuver, line 1: there are more process.env vars used throughout the app I think, could you add those as well? (i like this)

lkleuver (dismissed):

frontend/app/component/error/ErrorBoundary.tsx
marlonbaeten: I've first tried adding a ErrorBoundary on the root level - but indeed the router catches everything. I've registered a error handler on the root element of the router- that works as expected.

jorisleker (dismissed):
Almost there @marlonbaeten !

- Vertical space between the paragraph and the button should be 2rem
- Copy 'foutcode: 500' in modal does not match design. Text 'Er is een technisch probleem ...' is missing
- Vertical spacing between paragraph of text and button in modal should be 1.5rem;
- when tabbing through the modal, things move around a little when focus is placed on the 'melding sluiten' button. 
- the dev mode version of the error page is missing some copy on how to report a bug: 
Scherm­afbeelding 2024-10-22 om 10 15 39

frontend/app/component/error/Error.module.css
marlonbaeten: A name with a dash is not usable as a property of the css module import without using string quotes, but I can change this to an single name since it is scoped to this component :+1:

frontend/app/component/error/Error.types.ts
marlonbaeten: Although its not a pure Typescript type, a TypeScript class definition creates a type from itself - so maybe at least half the correct place 😅

frontend/lib/util/env.ts
marlonbaeten: I have only found them in configuration files - moving those to the codebase might be problematic. So for now this is the only one used at compile/runtime.

frontend/lib/api/ApiClient.ts
```diff
@@ -12,27 +12,44 @@ export interface ApiResponse<T> {
   data: T;
 }
 
-export type ApiError = {
+export class ApiError extends Error {
```
oliver3, line 15: Nice that you use an actual `Error`! 👍 

You can use https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties to clean this up a bit

```ts
export class ApiError extends Error {
  constructor(
    public status: ApiResponseStatus.ClientError | ApiResponseStatus.ServerError,
    public code: number,
    public error?: string,
  ) {
    super(error);
  }

  withContext(message: string): this {
    this.error = message;

    return this;
  }
}
```

frontend/lib/api/ApiClient.ts
```diff
@@ -12,27 +12,44 @@ export interface ApiResponse<T> {
...
+
+  constructor(status: ApiResponseStatus.ClientError | ApiResponseStatus.ServerError, code: number, error?: string) {
+    super(error);
+    this.status = status;
+    this.code = code;
+    this.error = error;
+  }
+
+  withContext(message: string): this {
+    this.error = message;
```
oliver3, line 28: maybe set `this.message` as well to not let them diverge? Ideally we would use message only and not have a shadow `error` field..

oliver3 (changes requested):

frontend/lib/api/ApiClient.ts
marlonbaeten: Thanks :+1: I like it clean :-)

frontend/lib/api/ApiClient.ts
marlonbaeten: Check -I've merged error and message field.

frontend/lib/api/ApiClient.ts
```diff
@@ -12,54 +12,95 @@ export interface ApiResponse<T> {
...
-  code: number;
-  error?: string;
-};
+export class ApiError extends Error {
+  constructor(
+    public status: ApiResponseStatus.ClientError | ApiResponseStatus.ServerError,
+    public code: number,
+    public message: string,
+  ) {
+    super(message || "Unknown error");
```
oliver3, line 21: Message is not an optional field, and I do not think you mean to override an empty string? If you do want to make it optional, you can also use default parameter: `public message = "Unknown error"`

marlonbaeten: Check - previously JSON responses from the API were directly cast to `ApiError` that's why there was a error field (and this fallback). But I removed all casting. I'll add your suggestion - thanks!

Lionqueen94 (approved):

oliver3 (approved):

marlonbaeten:
FYI @jorisleker gave me a GO to merge on a side channel.

PR #475, opened 2024-10-17
Add 'skip serialization' and 'nullable false' attributes to optional recounted field

oliver3:
The field was made optional in https://github.com/kiesraad/abacus/pull/446 but I believe the attributes as discussed in https://github.com/kiesraad/abacus/pull/406#discussion_r1793108126 were accidentally not added.

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -896,8 +896,7 @@
           },
           "recounted": {
             "type": "boolean",
-            "description": "Recounted (\"Is er herteld? - See form for official long description of the checkbox\")",
-            "nullable": true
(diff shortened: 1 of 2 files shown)


Lionqueen94:
Is it an idea to fix the PDF generation in this PR as well, since it was related to the same PR?

cikzh (approved):

cikzh:
> Is it an idea to fix the PDF generation in this PR as well, since it was related to the same PR?

I'd say we create a different PR, it's kind of related, but no reason to keep this one unmerged

oliver3:
I would prefer a separate PR as well for the PDF.

Lionqueen94 (approved):

PR #485, opened 2024-10-22
Add some additional tests for addition and pdf rendering

### LINKED ISSUE #335: Expand tests for PDF generation and result summation
> > The tests right now are relatively barebones, but at least we test some successful additions. I would suggest to at add much more significant testing as an issue for a later time (soonish)
> 
> Fine with me to do it in a later issue, as long as that issue is part of the same epic as this PR/issue.

_Originally posted by @jschuurk-kr in https://github.com/kiesraad/abacus/issues/324#issuecomment-2342993720_
            
# TODO

- Expand d2d test with PDF generation
- Expand unit tests
  - using the *"0, 1, many, lots, oops"* heuristic (*many*: more than one; *lots*: very very many; *oops*: errors), can be applied both to the counts, as to the number of polling stations
  - Cover the case @ring-ring-ring  ran into during a demo: `"dictionary does not contain key \"house_number_addition\""`

rnijveld:
Fixes #335 

This also fixes a bug that caused a crash when no house number addition was given because the JSON serialization behavior was changed. I've added a test that checks for this as well.

One additional change that was made is that the release profile now always includes integer overflow checks. I don't think we need the performance for this, but I do think we need the correctness here. While I don't expect any overflow or underflows to occur with any regular input (and even if there are large enough numbers entered they should normally be captured by validation rules), if one does occur, we should really have the software crash instead of it continuing as if nothing bad happened.

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -42,3 +42,6 @@ unsafe_code = "forbid"
 
 [lints.clippy]
 unwrap_used = "warn"
+
+[profile.release]
+overflow-checks = true
--- a/backend/src/data_entry/structs.rs
+++ b/backend/src/data_entry/structs.rs
@@ -703,6 +703,21 @@ pub struct DifferencesCounts {
     pub no_explanation_count: Count,
 }
 
+#[cfg(test)]
+impl DifferencesCounts {
+    pub fn zero() -> DifferencesCounts {
+        DifferencesCounts {
+            more_ballots_count: 0,
+            fewer_ballots_count: 0,
+            unreturned_ballots_count: 0,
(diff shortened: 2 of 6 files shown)


cikzh (approved):

Lionqueen94 (approved):

backend/Cargo.toml
```diff
@@ -42,3 +42,6 @@ unsafe_code = "forbid"
 
 [lints.clippy]
 unwrap_used = "warn"
+
+[profile.release]
+overflow-checks = true
```
praseodym, line 47: Nice, this was still on my todo list.

PR #487, opened 2024-10-22
No migration files in Dockerfile

cikzh:
Because we use the `migrate!()` macro, the migrations files are embedded in the binary. Copying them into the Docker container is therefore not needed.

See https://docs.rs/sqlx/latest/sqlx/macro.migrate.html for more information

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,10 +6,9 @@ RUN groupadd -r abacus --gid=999
 RUN useradd --no-log-init -r -g abacus --uid=999 abacus
 USER 999
 
-# Copy the binary, database and migrations
+# Copy the binary and database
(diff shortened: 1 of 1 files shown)


rnijveld (approved):
I'm a little bit surprised that we still need the database file. I'd assume that it always auto-migrates to the latest database version if at all possible, that would make the most sense to me. But other than that this looks good to me!

cikzh:
@rnijveld and I found a way to exclude also the database file from the Dockerfile, fixed in [4bd5275](https://github.com/kiesraad/abacus/pull/487/commits/4bd5275c7fcd704b5ce0263bf9de2e6b93ff07f1)

Lionqueen94 (dismissed):

Lionqueen94 (approved):

rnijveld (approved):

PR #492, opened 2024-10-25
Added npm install command to build and run script

cikzh:
To make sure `npm` runs build on the correct dependencies, `npm install` is run first.

Also turned on command printing and immediate exiting on error.

### DIFF
--- a/build-and-run
+++ b/build-and-run
@@ -14,7 +14,7 @@ set -x
 echo "Building frontend..."
 
 pushd "frontend" > /dev/null 2>&1
-npm install
+npm ci
 npm run build
 popd > /dev/null 2>&1
 
--- a/build-and-run
+++ b/build-and-run
@@ -5,9 +5,18 @@ then
   git checkout "$1"
 fi
 
+# Exit on errors
+set -e
+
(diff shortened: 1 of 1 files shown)


Lionqueen94 (dismissed):

build-and-run
```diff
@@ -5,9 +5,18 @@ then
...
+set -e
+
+# Print the commands as they are executed
+set -x
+
 echo "Building frontend..."
 
-npm run build --prefix frontend
+pushd "frontend" > /dev/null 2>&1
+npm install
```
Lionqueen94, line 17: It might be better to use `npm ci` instead, so it really installs only what's in the `package-lock.json`

cikzh: That does significantly increase startup time, but I think if it's used in situations like today, that's a fine tradeoff :+1:

cikzh: Fixed in [71a1b39](https://github.com/kiesraad/abacus/pull/492/commits/71a1b39b8c5e5e35073cfb902f2744809efd0517)

Lionqueen94 (approved):

jschuurk-kr (approved):

PR #521, opened 2024-10-29
First iteration of table component

oliver3:
Implements the component/visual part of https://github.com/kiesraad/abacus/issues/433

This `Table` component and friends will render and style an HTML table, inspired by `InputGrid`.

The `Table.LinkRow` component handles a click event and navigates to the specified url. Keyboard support will be added later.

Behaviour such as sorting will be done in a separate hook, that will play nicely with this `Table`.

Please have a look at `Table.stories.tsx` to see how it can be used, and let me know:
- if this setup will work for us
- if we indeed want `Table.Row` instead of `TableRow`
- other implementation comments

### DIFF
--- a/frontend/lib/ui/Table/Table.tsx
+++ b/frontend/lib/ui/Table/Table.tsx
@@ -1,0 +1,52 @@
+import * as React from "react";
+import { useNavigate } from "react-router-dom";
+
+import cls from "./Table.module.css";
+
+export interface TableProps {
+  id?: string;
+  children?: React.ReactNode;
+}
+
+export function Table({ id, children }: TableProps) {
+  return (
+    <table id={id} className={cls["table"]}>
+      {children}
+    </table>
+  );
+}
+
+Table.Header = ({ children }: { children: React.ReactNode[] }) => (
+  <thead>
+    <tr>{children}</tr>
+  </thead>
+);
+
+Table.Column = ({ children, number, width }: { children: React.ReactNode; number?: boolean; width?: string }) => (
+  <th className={number ? cls["number"] : undefined} style={width ? { width } : undefined}>
+    {children}
+  </th>
+);
+
+Table.Body = ({ children }: { children: React.ReactNode[] }) => <tbody>{children}</tbody>;
+
+Table.Row = ({ children }: { children: React.ReactNode[] }) => <tr>{children}</tr>;
+
+Table.LinkRow = ({ children, to }: { children: React.ReactNode[]; to: string }) => {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const navigate = useNavigate();
+
+  function handleClick() {
+    navigate(to);
+  }
+
+  return (
+    <tr className={cls["row-link"]} onClick={handleClick}>
+      {children}
+    </tr>
+  );
+};
+
+Table.Cell = ({ children, number }: { children?: React.ReactNode; number?: boolean }) => (
+  <td className={number ? cls["number"] : undefined}>{children}</td>
+);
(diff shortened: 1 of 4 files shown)


frontend/lib/ui/Table/Table.tsx
```diff
@@ -0,0 +1,52 @@
...
+    {children}
+  </th>
+);
+
+Table.Body = ({ children }: { children: React.ReactNode[] }) => <tbody>{children}</tbody>;
+
+Table.Row = ({ children }: { children: React.ReactNode[] }) => <tr>{children}</tr>;
+
+Table.LinkRow = ({ children, to }: { children: React.ReactNode[]; to: string }) => {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
oliver3, line 36: Seems that ESLint does not recognise this `Table.LinkRow` as a React component

jorisleker:
@oliver3 How can I test this? Do all tables in Abacus use this new component now? If so: looks good!

jschuurk-kr:
@jorisleker Heb het aan @oliver3 gevraagd: is alleen het component, dus zit nog niet in de applicatie. Testen kan dus alleen met 'npm run ladle'.

jorisleker (dismissed):

frontend/lib/ui/Table/Table.tsx
praseodym: Apparently this pattern of component declarations is not recommended, but there is a workaround for the lint error: https://github.com/facebook/react/issues/20700#issuecomment-1919251632

lkleuver: Do you mean the <Table.Cell> part?
I've been slowly extracting the separate components for some (like ProgressList)
The reason to choose these type of exports is to have some namespacing and reduce the amount of imports required, with the goal of improving readability for "new" coders.

I have no feelings towards removing this "code style" but curious about the reasons.

lkleuver (dismissed):

frontend/lib/ui/Table/Table.tsx
oliver3: I've updated it to have the function definition separated, to make it more clear (also to the linter) that they are React components, thanks for the suggestion!

And I prefer this dot notation as well for discoverability in the IDE (autocomplete).

frontend/lib/ui/Table/Table.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+function LinkRow({ children, to }: { children: React.ReactNode[]; to: string }) {
+  const navigate = useNavigate();
+
+  function handleClick() {
+    navigate(to);
+  }
+
+  return (
+    <tr className={cls["row-link"]} onClick={handleClick}>
```
marlonbaeten, line 58: We might want to consider using a different element for this - for example:

```jsx
<a role="row" className={cls["row-link"]} onClick={handleClick}>
```

And then:

```css
.row-link {
  display: table-row;
}
```

This will still be rendered correctly as a table, but the clickable element is better recognized as such by browsers.

marlonbaeten (approved):

frontend/lib/ui/Table/Table.tsx
oliver3: I believe (and checked with mdn and validator.w3.org) that the only content allowed in a `tbody` is zero or more `tr` elements. For now I propose to keep it this way, and later find another more accessible way that will be either not using `table` at all, or a link within a cell and some css trickery to make it cover the whole row.

lkleuver (approved):

PR #527, opened 2024-10-30
Show error when list-total is left empty by mistake

### LINKED ISSUE #500: Show error when list-total is left empty by mistake
Because this often happens by mistake. We show the error at the bottom of the page; 
[design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=6128-28260&node-type=frame&t=VOfeUhME521tHpCy-11)

We show the error when:
- The user submits the form using 'Volgende', and
- there are votes on candidates, and
- the "Totaal lijst x" field is empty

Acceptance criteria
- We only show the error message in the situation described above
- When we show this error, we do _not_ show error F.401
- Scroll position remains the same; do not scroll up
- a11y: the error is noticeable for users with a screen reader (ie by adding `aria-invalid=true` and `aria-errormessage=`-attributes ([documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage))
- The "Totaal lijst x" field has an error-icon
- On blur: do not remove the error icon from the field (as we don't do this for other errors; validation is performed on the backend, we should keep that behavior consistent)

note: the error in Figma uses a new color-palette that has not been implemented yet. Please use the colors as currently defined in css for this type of alerts.

Lionqueen94:
Closes #500

### DIFF
--- a/frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
+++ b/frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
@@ -105,13 +105,14 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
         document.getElementById("total")?.focus();
       } else if (errors.length === 0 && warnings.length > 0) {
         const acceptWarnings = acceptWarningsRef.current?.checked || false;
-
+        setMissingTotalError(false);
         if (!hasChanges && !acceptWarnings) {
           setWarningsWarning(true);
         } else {
           await submit({ acceptWarnings });
         }
       } else {
+        setMissingTotalError(false);
         await submit();
       }
     })(event);
--- a/frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
+++ b/frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
@@ -92,13 +92,17 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
     }
   }, [isSaved, errors, warnings]);
 
+  const [missingTotalError, setMissingTotalError] = React.useState(false);
   const [warningsWarning, setWarningsWarning] = React.useState(false);
 
   const handleSubmit = (event: React.FormEvent<CandidatesVotesFormElement>) =>
     void (async (event: React.FormEvent<CandidatesVotesFormElement>) => {
       event.preventDefault();
 
-      if (errors.length === 0 && warnings.length > 0) {
+      const values = getValues();
+      if (values.candidate_votes.some((candidate) => candidate.votes > 0) && values.total === 0) {
+        setMissingTotalError(true);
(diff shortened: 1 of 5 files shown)


jorisleker (changes requested):
@Lionqueen94 noticed the 'don't change scroll position acceptance criterium' does work out as intended. It was intended for situations where the user was already scrolled down, and should prevent us from scrolling up as we do with other errors. 

Now consider the following scenario (with a list that's longer then the viewport)
- fill in x votes at candidate 1.
- submit the form with `shift+enter` (so don't scroll down to the 'volgende' button)
- you won't notice the error. 

Instead of 'don't change the scroll position', we should scroll to the bottom of the list, if the error is outside of the viewport. Can you update the PR to do this?

PR meets all other acceptance criteria ✅

jorisleker (changes requested):
Yes, works now!

Noticed something else (sorry for not spotting it previously).
The focus should be set to the totals field, but it's not.

jorisleker (approved):
👍

frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
```diff
@@ -92,21 +92,27 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
...
 
   const handleSubmit = (event: React.FormEvent<CandidatesVotesFormElement>) =>
     void (async (event: React.FormEvent<CandidatesVotesFormElement>) => {
       event.preventDefault();
 
-      if (errors.length === 0 && warnings.length > 0) {
+      const values = getValues();
+      if (values.candidate_votes.some((candidate) => candidate.votes > 0) && values.total === 0) {
+        setMissingTotalError(true);
+        document.getElementById("total")?.focus();
```
lkleuver, line 105: I think a ref would be nicer here.

lkleuver: oh hmm, nvm, that would require implementing a forwardRef which is kinda iffy.

lkleuver (approved):
Feels a bit "hacky" now with errorsAndWarnings and an extra error field.
fine for now imho, but might want to note a refactor to separate errors and warnings props, so you could just append to the errors array instead of introducing a new prop.

Lionqueen94:
> Feels a bit "hacky" now with errorsAndWarnings and an extra error field. fine for now imho, but might want to note a refactor to separate errors and warnings props, so you could just append to the errors array instead of introducing a new prop.

@lkleuver I agree, we should definitely refactor this later on, not having the field process the errorsAndWarnings but just tell the field if there is an error or warning or not

frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
Lionqueen94: Yeah I thought the same

Issue #503, opened 2024-10-28
Focus first input when navigating to next step


jorisleker:
Right now, we focus the header of section. Usertesting has shown this is confusing / leads people to use the mouse instead of the keyboard to navigate between inputs.

- [ ] Add `aria-live="polite"` attribute to the h2 that contains the name of the political group, so screen readers announce the element's content. 
- [ ] Remove `tabindex="-1"` from the h2 element
- [ ] Focus on first input on the page, unless there is an error / warning. In that case, retain current behavior
- [ ] On the 'recounted' page, focus on the fieldset containing the radio's. Make sure they are labeled in an accessible way.
- [ ] On the 'controleren en opslaan' step, don't set a focus (or set on the page body), so screenreaders will read out all relevant content, even when there is no 'opslaan' button because of blocking errors.

in scope: all steps in the data-entry form.
out of scope: modals, warning/errors (and rest of the application)

oliver3:
It seems that the consensus on the internet is that for a SPA with routing, you need to change the title and announce the change. Announcing can be done either by moving focus to the new content or header (as we do now), or use an [ARIA live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) if you cannot do that, e.g. when you want the first input to have the focus (as we would like to).

Also summarized on https://hidde.blog/accessible-page-titles-in-a-single-page-app/

Note that [role status](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role) also turns something into a live region 

Possible implementations:
- Maybe it is okay to turn our heading into a status, but it will loose role heading?
- Add a [visually hidden](https://www.a11yproject.com/posts/how-to-hide-content/) element with role status on every page, maybe using out `PageTitle` component
- Implement an application wide aria live announcement system, possibly coupled to our `PageTitle` component, similar to https://github.com/Heydon/on-demand-live-region or https://github.com/adobe/react-spectrum/blob/main/packages/@react-aria/live-announcer/src/LiveAnnouncer.tsx

PR #529, opened 2024-10-30
Focus first input when navigating to next step

oliver3:
Fixes https://github.com/kiesraad/abacus/issues/503 where the focus should not be on the data entry headers but the first input element when appropriate.
- Auto focus first input for data entry pages, instead of focus on section header
- Change section header to `fieldset` with `legend` for screen reader context
- Remove role table from data entry table so that it does not interfere for the screenreader
- Use field (left) plus title (right) as label for data entry inputs
- Adjust unit tests and playwright tests accordingly and add accessibility name assertions
- Disable "jsx-a11y/no-autofocus"

Keyboard focus might still be not ideal in e.g. scenarios with a modal, but this is acceptable for now.


lkleuver:
What about adding the first focus thing to the `<Form />` Component?

oliver3:
> What about adding the first focus thing to the `<Form />` Component?

I was worried that you then have to be able to tell Form _not_ to do that again in certain cases (warnings, ..?)

frontend/.eslintrc.cjs
```diff
@@ -20,6 +20,7 @@ module.exports = {
   parser: "@typescript-eslint/parser",
   plugins: ["react-refresh", "jsx-a11y", "prettier", "@typescript-eslint"],
   rules: {
+    "jsx-a11y/no-autofocus": "off",
```
oliver3, line 23: We do want to manage focus in our application, so I suggest to disable this rule. Doing it with JS instead of autofocus has the same issues and is not a solution for this.

frontend/e2e-tests/dom-to-db-tests/resume-data-entry.e2e.ts
```diff
@@ -75,11 +75,6 @@ test.describe("resume data entry flow", () => {
       await expect(votersAndVotesPage.blankVotesCount).toBeEmpty();
       await expect(votersAndVotesPage.invalidVotesCount).toBeEmpty();
       await expect(votersAndVotesPage.totalVotesCastCount).toHaveValue("100");
-
-      await votersAndVotesPage.abortInput.click();
-      await expect(abortInputModal.heading).toBeFocused();
-      await abortInputModal.close.click();
-      await expect(votersAndVotesPage.abortInput).toBeFocused();
```
Lionqueen94, line 82: The modal should still get a focused heading right? @jorisleker  I think you cannot tab through it otherwise

jorisleker: Yes, and it the h2 still gets focussed. So see no reason to remove this test.

oliver3: The test was removed here because the subject of the test is "resume after saving" and tests should not do too much (ideally only one thing). Modal will still get a focused heading and this is also already checked in tests on six other locations.

The last line was the trigger actually, because the abortInput will not receive focus again after closing the modal in Safari only.

Lionqueen94:
Shall we autoFocus the input on PollingStationChoiceForm too?

jorisleker (changes requested):
Focussing first input LGTM!

Something else broke in this PR: 
- the alert isn't full width anymore
- the progress bar doesn't show progress anymore (blue part of bar is missing)
Scherm­afbeelding 2024-11-11 om 16 14 32

Lionqueen94:
> Focussing first input LGTM!
> 
> Something else broke in this PR:
> 
>     * the alert isn't full width anymore
> 
>     * the progress bar doesn't show progress anymore (blue part of bar is missing)
> 
> 
> Scherm­afbeelding 2024-11-11 om 16 14 32

image

This has been fixed in the current version of this PR. After I fixed the Alert in another PR and Joep and Oliver fixed the progress bar color. Please pull and check the most recent version :+1:

Lionqueen94 (approved):

lkleuver (approved):

jorisleker (approved):

PR #540, opened 2024-11-04
Add datetime of last submit to polling_station_data_entries and polling_station_results

### LINKED ISSUE #462: Add datetime of last submit to polling_station_data_entries and polling_station_results
These are used in the tables, see https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2232-31003&node-type=frame&m=dev
Timezone aware datetime needed?

Add datetime to `/api/elections/{election_id}/status` endpoint.

Lionqueen94:
Closes #462 

## Question
- Any particular test I need to add for the backend changes?


marlonbaeten (approved):
I see `CURRENT_TIMESTAMP` was nog the function we were looking for ;-)

backend/migrations/3_polling_station_data_entries.sql
```diff
@@ -4,6 +4,7 @@ CREATE TABLE polling_station_data_entries
     entry_number       INTEGER NOT NULL,
     data               BLOB,
     client_state       BLOB,
+    timestamp          INTEGER NOT NULL DEFAULT (unixepoch()),
```
jschuurk-kr, line 7: I have concerns about using unix epoch as timestamp. We know the clocks of the machines on which Abacus will run might be off. Hopefully that's just a small amount of time drift and not an issue with timezone settings or daylight saving time settings.

But still, by storing timestamp as unix epoch we store seconds from 1970-01-01 00:00:00 UTC according to the machine, while losing the information of which timezone the machine thinks it's in.

This also raises the question how we would display timestamps when combining data from the BES-islands with that from the European Netherlands. Do we infer the timezone based on location?

jorisleker: To consider when deciding upon implementation: the times we store here a primarily used in the UI (coordinator status screens: ie to sort data entries based on the time they were completed), and to measure how long data entry took (telemetry). They are not meant as absolute truths and I don't think they should end up in eml's or other exports. Not sure we will ever run into the situation that 1 installation will be used to enter data in a BES timezone and a European one?

Apart from the formatting: a more descriptive name for this field might be good: what does the timestamp represent? Is it the moment data entry was `created`, `updated`, `finished`?

jschuurk-kr: > They are not meant as absolute truths and I don't think they should end up in eml's or other exports.

I'd prefer storing all dates and times in the same way and not trying to anticipate which dates/times will be used how.

> Not sure we will ever run into the situation that 1 installation will be used to enter data in a BES timezone and a European one?

Agreed, was thinking about the CSB. And if a single installation will be used to enter data both from BES and European Netherlands, timestamps should be in local time of the machine Abacus is running on.

praseodym: > I have concerns about using unix epoch as timestamp. We know the clocks of the machines on which Abacus will run might be off. Hopefully that's just a small amount of time drift and not an issue with timezone settings or daylight saving time settings.
> 
> But still, by storing timestamp as unix epoch we store seconds from 1970-01-01 00:00:00 UTC according to the machine, while losing the information of which timezone the machine thinks it's in.

We're using these timestamps in a local database, not in exported data. The server knows its own timezone, so no information is lost by saving a unix epoch without an explicit timezone.

For exports I'd propose using ISO 8601 date time with offset (e.g. 2024-11-05T10:07:12+01:00), but that's for a future epic (e.g. #546).

praseodym: > Apart from the formatting: a more descriptive name for this field might be good: what does the timestamp represent? Is it the moment data entry was `created`, `updated`, `finished`?

Agreed that `updated` is a better description for the data entry rows, and maybe `created` for the results rows.

jschuurk-kr: > We're using these timestamps in a local database, not in exported data. The server knows its own timezone, so no information is lost by saving a unix epoch without an explicit timezone.
> 
> For exports I'd propose using ISO 8601 date time with offset (e.g. 2024-11-05T10:07:12+01:00), but that's for a future epic (e.g. #546).

That leaves some exotic scenarios in which the timezone of the server is changed between data entry and export, but I think accounting for that is more strict thant we're aiming for.

praseodym: > That leaves some exotic scenarios in which the timezone of the server is changed between data entry and export, but I think accounting for that is more strict thant we're aiming for.

That's also not an issue because the unix timestamp is UTC, so it is not affected by timezone changes on the server. 

Changes to the server clock (in UTC) will have implications, e.g. if the clock is changed backwards the data entries look like they were created in the future. However, we can't do much about this.

jschuurk-kr: The default in Windows is to set the hardware clock to local time. (There's a registry setting to make it use UTC instead.)

But agreed that it all comes down to the fact that we can't do much about a server admin making changes to date, time, or timezone settings. And if we'd want to detect and log someone making such changes, there has to be a better way than storing a timezone with every timestamp.

oliver3: I believe storing unix epoch in UTC is actually the best option, and when showing it on the screen we have to make sure that it is in local time. 

And I agree with more a more descriptive name than timestamp.

Lionqueen94: I changed `timestamp` to `updated_at` for the data_entries and to `created_at` for the results. In the polling station status endpoint the field is called `finished_at` since it's regarding a certain state, meaning the previous entry number is finished at a certain moment (or the field is empty when the polling station data entry has not been started yet). The field will only be used in the `First entry done` and `First and second entry done` tables on the coordinator election status overview.

oliver3 (approved):

praseodym (approved):

jschuurk-kr (approved):

jorisleker (approved):

PR #551, opened 2024-11-05
Update openapi generate script with proper prettier and generate all request paths

### LINKED ISSUE #544: Verify generated `openapi.ts` is up to date
In the backend we have a [cargo test](https://github.com/kiesraad/abacus/blob/3498e3aace5fedcd4fb00036da16d2df676187df/backend/src/bin/gen-openapi.rs#L23-L28) to check whether the generated `openapi.json` is up to date with the Rust code.

We should also do this in the frontend, to have CI check whether the generated `openapi.ts` is up to date with the `openapi.json` source file. Because the file is currently not generated with the correct formatting (maybe because `gen_openapi_types.ts` does not use the correct Prettier settings), this should be fixed first to avoid always having formatting changes.

Additionally, `gen_openapi_types.ts` should be added to the Lefthook pre-commit hooks so that we don't forget to run it.

### LINKED ISSUE #548: `gen_openapi_types.ts` overwrites paramters with paths that have more than one method
When generating the TypeScript types for PR https://github.com/kiesraad/abacus/pull/538, I notice that somes types for listing a `PollingStation` gets removed, while they are still present in `openapi.json`, resulting in this example diff:

```diff
-export type POLLING_STATION_LIST_REQUEST_PATH = `/api/elections/${number}/polling_stations`;
+export type POLLING_STATION_CREATE_REQUEST_PATH =
+  `/api/elections/${number}/polling_stations`;
```

Expected behavior would be that both types would be generated. It looks like the latter type overrides the first.
I suspect `gen_open_api_types.ts` should loop over all methods in the `addPath()` function, right now it seems to just check if it's a `get` or `post`.

lkleuver:
Add local prettier config to generator formatter so that it aligns with the linter/lefthook
Add test to check local openapi.ts against openapi.json
Generate all requests in a path

closes #544 
closes #548


Lionqueen94 (approved):

frontend/scripts/gen_openapi_types.ts
```diff
@@ -1,172 +1,28 @@
...
+import path from "path";
 
-import { OpenAPIV3, OperationObject, PathsObject, ReferenceObject, SchemaObject } from "./openapi";
+import { generate } from "./openapi/generator";
 
-const TARGET_PATH = "./lib/api/gen";
+const PROJECT_ROOT = path.join(__dirname, "..");
+const TARGET_PATH = `${PROJECT_ROOT}/lib/api/gen`;
 const FILE_NAME = "openapi.ts";
+//const PRETTIER_CONFIG = path.join(PROJECT_ROOT, "/.prettierrc.cjs");
```
oliver3, line 9: Could be removed?

oliver3 (dismissed):

Lionqueen94 (approved):

frontend/lib/api/openapi.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const OPEN_API_JSON_PATH = `${BACKEND_ROOT}/openapi.json`;
+const CURRENT_SPEC_PATH = `${FRONTEND_ROOT}/lib/api/gen/openapi.ts`;
+
+describe("openapi spec", () => {
+  test("openapi spec is up to date", async () => {
+    const openAPIJSONString = fs.readFileSync(OPEN_API_JSON_PATH, "utf8");
+
+    const latestSpecString = await generate(openAPIJSONString);
+    const currentSpecString = fs.readFileSync(CURRENT_SPEC_PATH, "utf8");
```
praseodym, line 18: You could use the async `fs.readFile` function instead.

lkleuver: my typescript thinks its not promisified yet.

lkleuver: updated, but it's a bit iffy using a different import. works here but makes the gen_openapi_types file more complex

praseodym (approved):

frontend/lib/api/openapi.test.ts
praseodym: Let's keep it to only changing the test then!

oliver3 (approved):

PR #554, opened 2024-11-05
Use SQLx offline mode for CI

praseodym:
This shaves off 1min30s of sqlx-cli install time from the CI run time.

For this to work, the SQLx query metadata is expanded to all features and targets with two extra arguments to `cargo sqlx prepare`.

Relates to #488.

PR #561, opened 2024-11-06
Use new `force-embed` feature for `memory-serve

### LINKED ISSUE #488: Backend CI runs are slow
CI build times for the backend action have significantly increased. It seems there's a problem with caching

praseodym:
With this update, we can use faster debug builds instead of slow release builds in CI and the `build-and-run` script.

Thanks to @marlonbaeten for adding the feature in `memory-serve`!

Resolves #488


praseodym:
This shaves off about 4 minutes of build time, so together with #554 this will cut the backend CI run time almost in half.

jschuurk-kr (approved):
`build-and-run` works on my machine.
Downloading the build artifact works too (tried with and without `--seed-data`).

Issue #473, opened 2024-10-17
Escalate some errors in `PollingStationFormController.tsx` 


marlonbaeten:

praseodym:
I noticed in #566 when the API request after clicking the 'Invoer opslaan' button in the 'Invoer afbreken' modal returns an error (e.g. status code 422), no error is shown to the user. I think that should be in scope to be fixed with this issue.

marlonbaeten:
> I noticed in [#566](https://github.com/kiesraad/abacus/pull/566) when the API request after clicking the 'Invoer opslaan' button in the 'Invoer afbreken' modal returns an error (e.g. status code 422), no error is shown to the user. I think that should be in scope to be fixed with this issue.

I think this is resolved with: https://github.com/kiesraad/abacus/pull/577/commits/4a7da4bafba5896a18f6bb6398462e1e30766aa9

PR #566, opened 2024-11-06
Save cached input when clicking save in modal

### LINKED ISSUE #421: Save cached input after clicking save in modal
> Regarding "- If the furthest page was edited and not explicitly saved, these changes are lost. Is this ok?"
> To me it makes sense if you navigate away from the form that the cached input is sent to the backend to be saved. Since you explicitly click `Invoer bewaren` (even if you are not on that specific page). @jorisleker what do you think?

_Originally posted by @Lionqueen94 in https://github.com/kiesraad/abacus/issues/417#issuecomment-2391838912_

> Agree with @Lionqueen94. Input on furthest page should be saved if you choose invoer bewaren.

_Originally posted by @jorisleker in https://github.com/kiesraad/abacus/issues/417#issuecomment-2392146990_

The flow is described in the [data entry flowchart](https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/data-entry-forms.md). This covers the 'cached data' (input on the furthest page that was not yet saved by clicking next, while the user navigated to a different form page in the meantime), but not whether this data should be saved when navigating away or aborting the data entry using the button.

Let's figure out what behaviour we should have, clarify in our documentation (the flowchart) and implement any changes in the code if needed.

praseodym:
Resolves #421

### DIFF
--- a/documentatie/flowcharts/data-entry-forms.md
+++ b/documentatie/flowcharts/data-entry-forms.md
@@ -223,7 +223,7 @@ flowchart TD
 
     %% steps
     modal{"modal: <br/> save or delete?"}
-    call-save-api(call save api)
+    call-save-api(call save api, <br/> include cached data)
     call-delete-api(call delete api)
 
     %% flow
--- a/documentatie/flowcharts/data-entry-forms.md
+++ b/documentatie/flowcharts/data-entry-forms.md
@@ -2,10 +2,11 @@
 
 This page describes the navigation and rendering logic of the data entry forms through the following flow charts:
 
-- __Render navigation menu__: styling of the different items in the left-hand navigation menu
-- __Render form__: rendering of the different forms for data entry
-- __Click "Volgende"__: what should happen when the user clicks the "Volgende" ("Next") button
-- __Navigate away from page__: what should happen when the user navigates away from the page in any other way than clicking the "Volgende" button
+- [__Render navigation menu__](#render-navigation-menu): styling of the different items in the left-hand navigation menu
+- [__Render form__](#render-form): rendering of the different forms for data entry
+- [__Click "Volgende"__](#click-volgende): what should happen when the user clicks the "Volgende" ("Next") button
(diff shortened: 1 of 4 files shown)


jschuurk-kr:
It's possible to trigger a 422 Invalid Data response. The user does not get feedback about this and the latest state is not saved.

Steps to reproduce:
1. Set recounted to yes.
2. Fill in at least one field in "Toegelaten kiezers na hertelling door gemeentelijk stembureau" (A.2 - D2.)
3. Return to the recounted page.
4. Set recounted to no.
5. Click "Invoer afbreken" and then "Invoer opslaan".

This does not happen when instead of clicking "Invoer afbreken" you click "Volgende", we remove `voters_recounts` from the request.

jschuurk-kr (changes requested):
It's possible to trigger a 422 Invalid Data response. The user does not get feedback about this and the latest state is not saved.

Steps to reproduce:
1. Set recounted to yes.
2. Fill in at least one field in "Toegelaten kiezers na hertelling door gemeentelijk stembureau" (A.2 - D2.)
3. Return to the recounted page.
4. Set recounted to no.
5. Click "Invoer afbreken" and then "Invoer opslaan".
6. The request to the server now contains both `recounted: false` and `voters_recounts` fields, triggereing the 422.

This does not happen when instead of clicking "Invoer afbreken" you click "Volgende", we remove `voters_recounts` from the request.

praseodym:
> It's possible to trigger a 422 Invalid Data response. The user does not get feedback about this and the latest state is not saved.

I think fixing the error handling for this is part of #473.

I've fixed the issue itself by making sure `voters_recounts` is always set/unset depending on the value of `recounted`, regardless of the current form page (82d2f6f). This also uncovered a test which relied on broken behaviour, which is also fixed (9e95418).

Lionqueen94 (approved):

jschuurk-kr (commented):
Tested the fix in https://github.com/kiesraad/abacus/commit/82d2f6febdf762d3650bc6416262b6affd35ca2e. LGTM!

Did not review the code of this PR.

marlonbaeten (approved):

PR #570, opened 2024-11-07
CSS guidelines

oliver3:
This is the result from a conversation with some team members about CSS guidelines and improvements:
- Beginning of `documentatie/style-guide/css.md`
- Vite CSS config to allow kebab-case CSS classes to be imported as camelCase in our components, with examples
- Some starter utility classes in `util.css` that should be used instead of many different CSS rules that do this one thing

This is just a starting point from which we can extend and improve


praseodym (dismissed):
lgtm, thanks for creating the style guide!

Lionqueen94 (approved):

praseodym (approved):

PR #577, opened 2024-11-11
Correct API error escalation and translations

### LINKED ISSUE #473: Escalate some errors in `PollingStationFormController.tsx`

marlonbaeten:
Fixes #473 and #486

- [Do not navigate from AbortDataEntryModal when the api returns a error](https://github.com/kiesraad/abacus/pull/577/commits/4a7da4bafba5896a18f6bb6398462e1e30766aa9)
- Add translations for all possible backend error references.
- Create NotFoundError in ApiClient based on HTTP response code
- Add translations for texts on error and not found pages


frontend/lib/api/form/pollingstation/PollingStationFormController.tsx
```diff
@@ -255,7 +261,7 @@ export function PollingStationFormController({
     acceptWarnings = false,
     aborting = false,
     continueToNextSection = true,
-  }: SubmitCurrentFormOptions = {}) => {
+  }: SubmitCurrentFormOptions = {}): Promise<ApiResult<SaveDataEntryResponse>> => {
```
oliver3, line 264:
```suggestion
  }: SubmitCurrentFormOptions = {}): Promise<ApiResult<SaveDataEntryResponse, ApiError>> => {
```
to enforce that only `ApiError` errors can be returned

frontend/lib/api/ApiError.ts
```diff
@@ -1,23 +1,44 @@
...
 
     return this;
   }
 }
 
-export class NetworkError extends Error {
-  constructor(public message: string) {
-    super(message);
-  }
+export function isFatalError(error: ApiResult<unknown>): error is ApiError | NetworkError | NotFoundError {
```
oliver3, line 42: The type predicate is not correct here, causing my IDE to complain about unreachable code in `deleteDataEntry()`.

If error is an `ApiError` with `fatal=false` it will return false, but TypeScript cannot see that this scenario exists.

frontend/lib/api/ApiClient.ts
```diff
@@ -120,8 +141,14 @@ export class ApiClient {
 
       return await this.handleEmptyBody(response);
     } catch (e: unknown) {
-      const message = `Network error: ${(e as Error).message}` || "Network error";
-      return new ApiError(ApiResponseStatus.ServerError, 500, message);
+      if (e === "Component unmounted") {
```
oliver3, line 144: Can we make this string a constant somewhere?

oliver3 (changes requested):

frontend/app/component/error/Error.types.ts
```diff
@@ -2,12 +2,3 @@ export enum ErrorAction {
   None = "none",
   Back = "back",
 }
```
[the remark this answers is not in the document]
marlonbaeten: Ok - we can always add it when necessary.

frontend/app/module/NotFound.tsx
```diff
@@ -2,24 +2,21 @@ import { Error, ErrorAction } from "app/component/error";
...
 import { AppLayout } from "@kiesraad/ui";
 
 export interface NotFoundProps {
-  message?: string;
-  path?: string;
+  message: TranslationPath;
+  path: string;
 }
 
 export function NotFound({ message, path }: NotFoundProps) {
```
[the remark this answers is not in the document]
marlonbaeten: `PageNotFound` noes not always cover what is happening - an URL / path might be valid and found, but a part of the fetched data might be missing. I will postfix both the fatal error and the not found page with the word `Page`, since they both render a full page message.

frontend/app/component/error/ErrorBoundary.tsx
```diff
@@ -17,8 +15,12 @@ export function ErrorBoundary() {
     return <NotFound message={error.message} path={error.path} />;
   }
 
+  if (error instanceof NetworkError) {
+    return <FatalError message={error.message} />;
+  }
+
   if (error instanceof ApiError) {
-    return <FatalError message={error.message} code={error.code} />;
+    return <FatalError message={error.message} reference={error.reference} code={error.code} />;
```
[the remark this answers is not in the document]
marlonbaeten: The `reference` references a message that that should be shown to the user. That might not qwlays overlap with the cause or reason for the error. But I will add some documentation to clear this up.

frontend/lib/api/ApiError.ts
marlonbaeten: Yeah, maybe I need to split ApiError and FatalApiError in two different types - what do you think?

marlonbaeten:
TODO: add flag to `useApi` to throw errors and only return success result

frontend/lib/api/ApiError.ts
oliver3: I think that would make things more straightforward indeed

marlonbaeten:
@oliver3 could you take a look at my latest additions, and see if your concerns are met?

frontend/app/module/FatalErrorPage.tsx
lkleuver, line 1: should this page use components that potentially can throw fatal errors?

lkleuver (commented):
lgtm besides the tests that were failing during review.

does feel a bit, off-ish to use translation strings for the notFoundMessage
not sure, just feels like the api shouldnt be aware of i18n.

oliver3 (approved):

frontend/app/module/FatalErrorPage.tsx
marlonbaeten: I don't think there are any components that can throw an error on this page. Do you see any? The translations are linted - so should never give an error at runtime.

oliver3 (approved):

PR #582, opened 2024-11-12
Added ElectionStatus to backend and frontend

Lionqueen94:
Backend
- add `status` field to elections table and corresponding endpoints, that represents the election status (only "DataEntryInProgress" or "DataEntryFinished" for now)

Frontend
- show status with icon and text on election report page and in header of election status page

Scripts
- add `data-icon` with the icon name to generated SVGs.


frontend/app/component/election/ElectionStatusWithIcon.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { Icon } from "@kiesraad/ui";
+
+function getIconForElectionStatus(status: ElectionStatus) {
+  switch (status) {
+    case "DataEntryInProgress":
+      return <Icon size="md" color="accept" icon={<IconCheckHeart />} />;
+    case "FinishDataEntry":
+      return <Icon size="md" color="error" icon={<IconLock />} />;
+    case "DataEntryFinished":
+      return <Icon size="md" color="default" icon={<IconCheckVerified />} />;
```
jschuurk-kr, line 13: For this `case` I also have the button "Invoerfase afronden" on the screen. That seems inconsistent.

image

frontend/app/component/election/ElectionStatusWithIcon.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+import { ElectionStatusWithIcon } from "./ElectionStatusWithIcon";
+
+describe("ElectionStatusWithIcon", () => {
+  test("renders the correct header coordinator status for DataEntryInProgress", () => {
+    const { getByText, getByRole } = render(ElectionStatusWithIcon("DataEntryInProgress", true, true));
+
+    expect(getByText("Steminvoer bezig")).toBeVisible();
+    expect(getByText("(eerste zitting)")).toBeVisible();
+    expect(getByRole("img")).toBeVisible();
```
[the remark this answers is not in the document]
jschuurk-kr: We'll have to do something with either `title` or `alt text`, so we can use `getByTitle` or `getByAltText`.

Or we can use the `alt` attribute in combination with `toHaveAccessibleName()`. For some other icons we use `toHaveAccessibleName()` and `aria-label`, but there we need the `aria-label` to convey information that's otherwise not available. Here we have text plus icon, so the icon is 'just' an extra.

@jorisleker What can we do with this icon, while avoiding that a screen reader first reads the text and then also 'reads' the icon?
image

frontend/app/component/election/ElectionStatusWithIcon.test.tsx
```diff
@@ -0,0 +1,102 @@
+import { render } from "@testing-library/react";
+import { describe, expect, test } from "vitest";
+
+import { ElectionStatusWithIcon } from "./ElectionStatusWithIcon";
+
+describe("ElectionStatusWithIcon", () => {
+  test("renders the correct header coordinator status for DataEntryInProgress", () => {
+    const { getByText, getByRole } = render(ElectionStatusWithIcon("DataEntryInProgress", true, true));
```
jschuurk-kr, line 8: It's recommended to use `screen` instead of destructuring the `render`: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen.

frontend/app/component/election/ElectionStatusWithIcon.test.tsx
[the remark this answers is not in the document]
jschuurk-kr: I can't find anything about React Testing Library being able to check if the icon is in front or behind the text. I'm not even sure if RTL renders the DOM in a way that it could know.

Playwright doesn't seem to be able either, see this [bug](https://github.com/microsoft/playwright/issues/9923). The closest you can get is clicking with `{ trial: true }` on all four corners of the image (using x/y coordinates), because the click will fail if something else is on top of the image at those points.

So I don't think it makes sense to assert on this.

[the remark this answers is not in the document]
jorisleker: @jschuurk-kr @oliver3 What if we do this?

```html
<img src="icon.svg" alt="icon description" aria-hidden="true" />
```

`Aria-hidden` should instruct screenreaders that the image (and it's alt text) should be ignored. And I don't think testing wether the icon is before or behind the text makes much sense. Bit too much detail?

frontend/app/component/election/ElectionStatusWithIcon.test.tsx
Lionqueen94: Updated in b514939d3a57324737f2b1e833db436db664ad6d

frontend/app/component/election/ElectionStatusWithIcon.tsx
Lionqueen94: Fixed in b514939d3a57324737f2b1e833db436db664ad6d

frontend/app/component/election/ElectionStatusWithIcon.test.tsx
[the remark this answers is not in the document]
oliver3: Talking to @jschuurk-kr we came with the following suggestion
```typescript
  test("renders the correct header coordinator status for DataEntryInProgress", () => {
    const { container } = render(ElectionStatusWithIcon("DataEntryInProgress", true, true));

    expect(container.children.length).toBe(2);
    expect(container.children[0]).toHaveTextContent("Steminvoer bezig (eerste zitting)");
    const icon = within(container.children[1] as HTMLElement).getByRole("img");
    expect(icon).toHaveAttribute("data-icon", "IconCheckHeart");
}
```

frontend/lib/i18n/locales/nl/election_status.json
```diff
@@ -2,10 +2,18 @@
...
   },
-  "finish_first_session_data_entry_status": "<strong>Steminvoer afronden</strong> (eerste zitting)"
+  "coordinator": {
+    "data_entry_in_progress": "Steminvoer bezig",
+    "data_entry_finished": "Steminvoer afgerond"
+  },
+  "typist": {
+    "data_entry_in_progress": "Invoer gestart",
+    "data_entry_finished": "Steminvoer voltooid"
+  }
```
jschuurk-kr, line 18: We should use the same words for the same status, not different ones depening on role.

Case in point: the translation file uses `"coordinator"`, while the `ElectionStatusWithIcon` wants to know if `isAdministrator`.

backend/src/election/structs.rs
```diff
@@ -27,6 +28,13 @@ pub enum ElectionCategory {
     Municipal,
 }
 
+/// Election status (limited for now)
+#[derive(Serialize, Deserialize, ToSchema, Clone, Debug, PartialEq, Eq, Hash, Type)]
+pub enum ElectionStatus {
+    DataEntryInProgress,
+    DataEntryFinished,
```
jschuurk-kr, line 35: We don't seem to have functionality to get an election from `DataEntryInProgress` to `DataEntryFinished`.

I can click "Invoerfase afronden" and then download the *proces-verbaal*, but that doesn't change the status of the election.

Lionqueen94: Yes we have not implemented this flow yet (see Figma). I set one of the elections in the fixtures and mock to finished so you can test how it looks. I can set the election to finished upon downloading the report, but this is not the moment this should happen so I was hesitant to do this.

jschuurk-kr: Good point. Implementing the status change on a button that's only temporarily on that page would not really be better.

Arguably we should have kept the statuses limited to the single `DataEntryInProgress` for now, but let's not revert all the work you've done.

frontend/lib/i18n/locales/nl/election_status.json
Lionqueen94: @jschuurk-kr `isAdministrator` will be removed once we have users with roles, so for now this is just used to show what a coordinator/admistrator person would see. I can make a separate `isCoordinator` flag for those specific views, for now I bundled those under isAdministrator (with the #coordinator or #administrator url flag)

jorisleker (approved):

frontend/lib/i18n/locales/nl/election_status.json
jorisleker: @jschuurk-kr For these two statuses I agree, these labels are very similar and could be the same. 

In other cases (not yet implemented) the label for coordinator and typist will differ, because of their different perspective. ie. If an election is fully configured, but the coordinator hasn't opened it for data-entry yet, we will show 'Klaar voor invoer' to the coordinator. To the typist we will communicate 'Wacht op coördinator'. 

So, for the sake of consistency, we could harmonize the labels here, but we should not remove the logic to differentiate the label based on the user's role.

jschuurk-kr: Thanks, @jorisleker! I'd still be in favor of removing the logic and add it again when we actually need it. But that's @oliver3's decision, since he took over this PR from Ellen.

oliver3: I've kept the logic but did some other refactor instead.

frontend/lib/api-mocks/ElectionMockData.ts
```diff
@@ -365,6 +367,27 @@ export const electionListMockResponse: ElectionListResponse = {
...
+    },
+    {
+      id: 5,
+      name: "Gemeenteraadsverkiezingen afgerond",
+      location: "Heemdamseburg",
+      number_of_voters: 100,
+      category: "Municipal",
+      election_date: "2020-01-30",
+      nomination_date: "2020-01-01",
+      status: "DataEntryFinished",
```
praseodym, line 390: Why don't I see these new elections in the MSW build?

rnijveld (dismissed):
Backend code LGTM! I've looked at the frontend as well, which looks ok to me, but someone else should probably do a better review of that part.

backend/fixtures/elections.sql
```diff
@@ -827,7 +827,54 @@ VALUES (1, 'Municipal Election', 'Heemdamseburg', 100, 'Municipal', '2024-11-30'
...
+                "initials": "B.",
+                "first_name": "Bar",
+                "last_name": "Foo",
+                "locality": "Rotterdam",
+                "gender": null
+              }
+            ]
+          }
+        ]'),
+       (5, 'Finished Election', 'Heemdamseburg', 100, 'Municipal', '2024-11-30', '2024-11-1', 'DataEntryFinished',
```
jschuurk-kr, line 877: This election (id 5) has no polling stations and no polling station results, even though it has as status `DataEntryFinished`. I don't think we shoud have incomplete test data like this.

I discovered this by going to by going to the status page of election 5, where it currently says "Steminvoer klaar (eerste zitting)" but also "Er zijn nog geen stembureaus toegevoegd voor deze verkiezing.".

jschuurk-kr: pair-fixed

rnijveld (approved):

PR #583, opened 2024-11-12
Nested translations & split up translation files

### LINKED ISSUE #576: Split up translation files
- Scoped translations
- Multiple files

marlonbaeten:
Fixes #576

- Translations can now be nested -> you do not have to repeat prefixes in the keys of translations :tada: 
- Translations can de split up in files, by their root key
- Use `npm run gen:po` to generate a single PO translation file per locale
- Use `npm run gen:translations-json` to convert the `.po` files to translation json files again.


frontend/README.md
```diff
@@ -121,3 +121,23 @@ npm run gen:icons
...
+
+#### gen_translation-json
+
+Generate json translation files from `.po` translation files in `/frontend/translations/<locale>.po`:
+
+```sh
+npm run gen:translation-json
+```
+
+This results in `/frontend/lib/i18n/loclales/<locale>/*.json`.
```
jschuurk-kr, line 143: Why do we need both? It's not the case that we edit the `json` files and generate the `.po` files for use by the application?

lkleuver (dismissed):
I think translators will really appreciate this.
Personally I like recursive types :D

frontend/README.md
marlonbaeten: - We write json ourselves (this gives us all kind of checks during development, like whether the key and/or the translation exists).
- Translation software expects some standard, `.po` files seems to be the most common one.
- To update translations , or to add a new language, we convert the json to `.po` files, which we then load into translation software.
- When a translation is added or updated the `.po` files are converted to json again.

jschuurk-kr: Would be a good idea then to mention in the README that the `.po` files are for the translation software and the `.json` ones for the application.

marlonbaeten: We do not commit any `.po` files (they are explicitly ignored) since that would create a duplicate "source of truth".
I could however describe the process of adding / updating translations, but before doing that it would make sense to first choose the software we are going to use. This is nice-to-have for now, since Dutch is the only locale we are going to support for now.

cikzh (dismissed):

cikzh (approved):

frontend/README.md
marlonbaeten: See https://github.com/kiesraad/abacus/pull/583/commits/903b8216558be12fcfaf8ac92129feb2db8776ec

lkleuver (approved):

PR #606, opened 2024-11-18
speed up Playwright tests in pipeline

jschuurk-kr:
Scope:
- run the Playwright lib-ui and app tests in their own job, parallel to the frontend and backend jobs
- run the Playwright e2e tests across three shards

**About sharding**
Because we tests against three browsers and use three shards, each shard now covers one browser. We could not use shards (yet) and create a separate job for each browser. At some later point we could then shard each of the browser jobs. I didn't see an advantage to doing it that way. And the current way makes it really easy to add a shard, i.e. add a number to `shard: [1, 2, 3]`.

**About caching**
I tried installing and chaching node modules and browser engines in a job. There's several issues with this:
- The app tests fail.
- Webkit caching does not work, because Playwright installs dependencies outside of the cached folder `~/.cache/ms-playwright`. I didn't try to find out how many additional folders would need to be cached to get it to work.
- The difference between installing Webkit and installing all three browser engines (Chrome, Firefox, Webkit) is quite small. So caching node modules / browser engines and the installing Webkit on every runner that needs it, does not result in a significant speed increase.

So caching node modules and browser engines is not worth it.

---

Main branch:
image

This branch:
image


jschuurk-kr:
Options for the missing WebKit dependencies:
- re-install them for every Webkit job
- cache more things, e.g. `/usr/lib/x86_64-linux-gnu/libwoff2dec.so.1.0.2`

.github/workflows/build-lint-test.yml
```diff
@@ -101,14 +101,119 @@ jobs:
...
+            frontend/node_modules
+          key: modules-${{ hashFiles('frontend/package-lock.json') }}
+
+      - name: Cache Playwright binaries
+        uses: actions/cache@v4
+        id: cache-playwright
+        with:
+          path: |
+            ~/.cache/ms-playwright
+          key: playwright-${{ hashFiles('frontend/package-lock.json') }}
```
praseodym, line 119: Did you check if this improves performance? The cache also needs to be downloaded and unpacked, so the performance improvement might be negligible or even non-existent.

jschuurk-kr: It results in a significant improvement. From my fork:
image

10s to download and unpack the cache. 30s/37s to run the UI tests on Firefox/Chrome. 1m10s to install the Webkit dependencies and run the tests on it.

(I intend to put all the UI tests back in a single job, because that probably still takes less time than the d2d tests job on Webkit.)

jschuurk-kr: Because caching does not work for WebKit, we need to install it on each runner. And doing that is barely faster than installing all three browser engines. And using cached node modules vs running `npm ci` also takes about the same time.

So until WebKit chaching works, there's no value in caching node modules and browser engines for Playwright tests. (Or until we care about a 10-ish seconds in difference.)

jschuurk-kr: The Playwright `app` tests also magically start passing again if you don't use cache. :man_shrugging:

praseodym (approved):

rnijveld (approved):

PR #607, opened 2024-11-19
Renamed binary to Abacus

cikzh:
The current binary name is "api", but since the binary will be the thing that will be run by end users, I think "abacus" is a more appropriate name.

While reviewing, please check that I didn't miss any renames in the scripts/CI pipelines/etc.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -60,7 +60,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           name: backend-build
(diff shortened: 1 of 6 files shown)


praseodym (approved):

praseodym:
I think you missed a name change in  https://github.com/kiesraad/abacus/blob/03884a471d9091373f8b4ce9c95ece6d0520d247/frontend/playwright.d2d.config.ts#L5-L18

praseodym (approved):

rnijveld (approved):

Issue #588, opened 2024-11-12
Implement second data entry in frontend


cikzh:
Change frontend to allow second data entry after first data entry

PR #481, opened 2024-10-21
Create election status page progress section

Lionqueen94:
Closes #460

PR #612, opened 2024-11-19
Add second data entry statuses in backend

cikzh:
Adds:

- The actual statuses
- Logic for updating the status
- Logic for determining the current status
- Tests


praseodym:
I'm not sure what happened here, but I've linked #587 (backend) instead of #588 (frontend) again.
Screenshot 2024-11-20 at 19 19 32

backend/src/data_entry/mod.rs
```diff
@@ -65,20 +65,25 @@ pub async fn polling_station_data_entry_save(
...
-        return Err(APIError::NotFound(
-            "Only the first data entry is supported".to_string(),
-            ErrorReference::EntryNumberNotSupported,
-        ));
-    }
-
-    if polling_station_data_entries.exists_finalised(id).await? {
-        return Err(APIError::Conflict(
-            "Cannot save data entry for a polling station that has already been finalised"
+    // Check if it is valid to save the data entry
```
praseodym, line 68: This code accepts `entry_number`s greater than 2 now, which should not be the case.

backend/src/data_entry/mod.rs
```diff
@@ -252,7 +256,20 @@ pub async fn polling_station_data_entry_finalise(
...
     }
 
-    polling_station_data_entries.finalise(&mut tx, id).await?;
+    match entry_number {
+        1 => {
+            polling_station_data_entries
+                .finalise_first_entry(&mut tx, id)
+                .await?
+        }
+        2 => polling_station_data_entries.finalise(&mut tx, id).await?,
```
praseodym, line 265: This needs to be the same for both the first and second data entry.

The creation of the data entry result needs to be triggered after the second data entry is finalised. In a next issue, we need to add a condition to only do this when the first and second data entry are equal. For this issue, we can skip that check.

backend/src/error.rs
```diff
@@ -16,12 +16,15 @@ use utoipa::ToSchema;
 pub enum ErrorReference {
     EntryNumberNotSupported,
     EntryNotFound,
-    PollingStationAlreadyFinalized,
+    PollingStationAlreadyFinalised,
+    PollingStationFirstEntryAlreadyFinished,
+    PollingStationFirstEntryNotFinished,
```
praseodym, line 21: Should we use `Finalised` instead of `Finished` here?

backend/src/polling_station/repository.rs
```diff
@@ -182,25 +182,52 @@ impl PollingStations {
...
-LEFT JOIN polling_station_results AS r ON r.polling_station_id = p.id
 LEFT JOIN polling_station_data_entries AS de ON de.polling_station_id = p.id
+LEFT JOIN polling_station_results AS r ON r.polling_station_id = p.id
 WHERE election_id = $1
+  AND de.polling_station_id IS NULL OR de.entry_number IN
+    (SELECT MAX(entry_number)
+     FROM polling_station_data_entries
+     WHERE polling_station_id = p.id
+     GROUP BY polling_station_id);
 "#,
```
praseodym, line 226: I feel like we're trying to cram way too much logic in the SQL layer here, and as I mentioned in https://github.com/kiesraad/abacus/pull/481#discussion_r1850822737 I don't like the `json_extract` use at all. I think we should try to come up with something more maintainable.

(Also, the doc comment is out of date now)

cikzh:
> I'm not sure what happened here

It is the result of double misclick :/

backend/src/error.rs
cikzh: Yeah, I agree that's probably more clear. Fixed in [deeaf03](https://github.com/kiesraad/abacus/pull/612/commits/deeaf0313f9e9d9c5e2889df3384e603a0c90068)

backend/src/data_entry/mod.rs
```diff
@@ -377,27 +398,37 @@ mod tests {
...
+        // and that a second entry was created
+        // TODO: We should test the "finalised_at" field and entry numbers
+        let row = query!("SELECT * FROM polling_station_data_entries")
             .fetch_one(&pool)
             .await
             .unwrap();
-        assert_eq!(row_count.count, 0);
-        // ...and added to polling_station_results
+        assert!(row.finalised_at.is_some());
+        // ...and is _not_ added to polling_station_results
```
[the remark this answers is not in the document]
cikzh: In [babb9f2](https://github.com/kiesraad/abacus/pull/612/commits/babb9f21e4af4f4df9a513223a12922115a31524), I've added a second row count check. The first is to test that nothing yet is inserted into the results table. Since that is how it used to work, it seems like a nice test to keep. The second tests that there is at least something added.

backend/src/polling_station/repository.rs
cikzh: As just discussed in person, we'll keep this for now and rewrite it in a different PR, based on this PR

cikzh:
> Support for `entry_number == 2` should also be added to `polling_station_data_entry_delete`.

Done in [253806c](https://github.com/kiesraad/abacus/pull/612/commits/253806c1ced894ccb1f4f4f6a0b0021505827ca4)

backend/src/data_entry/mod.rs
cikzh: Fixed in [ab34fce](https://github.com/kiesraad/abacus/pull/612/commits/ab34fced8a4d57f0effbf4ffc9f5e0b8a861c392).

I verified that the frontend still works as-is by clicking through an election.

cikzh:
This PR is ready for review again. First entries are now immediately finalised again, so the frontend stays usable. I've ignored the tests that already test for second entry logic.

praseodym (approved):

marlonbaeten (approved):

PR #625, opened 2024-11-20
Add tests for InputGridRow and tweak parameters

oliver3:
- Added tests for InputGridRow
- Made `warningsAccepted` optional
- Rename `error` prop to `errorMessageId`

These tests do not use a ladle story to test it in more isolation, and also because we feel that this should move to `app/component` because it is tightly coupled to the application.


marlonbaeten (approved):
LGTM!

Issue #432, opened 2024-10-08
Update polling station


oliver3:
- Create backend endpoint 
- Create frontend page
- Reuse frontend form
  - Rename submit button  
  - Cancel button
- Same validations
- Different confirmation alert success

Issue #542, opened 2024-11-04
Table component keyboard interaction


oliver3:
Add keyboard interactions for Table component, implemented in https://github.com/kiesraad/abacus/issues/433
- Show row focus
- Navigate to link on Enter

Issue #489, opened 2024-10-23
Discussion: what TypeScript type should a Rust enum have in our openapi.ts


oliver3:
## Current implementation
In Abacus, we generate an OpenAPI specification from the Rust types, and an `openapi.ts` file with TypeScript types from that OpenAPI spec.

Currently, Rust enum types, such as 

```rs
#[serde(rename_all = "snake_case")]
pub enum PollingStationStatus {
    FirstEntry,
    FirstEntryInProgress,
    Definitive,
}
```

are represented in OpenAPI as

```json
"PollingStationStatus": {
  "type": "string",
  "enum": [
    "first_entry",
    "first_entry_in_progress",
    "definitive"
  ]
},
```
and the generated TypeScript type is a union type of string literals

```typescript
export type PollingStationStatus = "first_entry" | "first_entry_in_progress" | "definitive";
```

## Problem
With this solution, it seems to that there is no way to iterate over the possible values of this type, which would have been useful in a recent PR. 

The generation of `openapi.ts` is done in `gen_openapi_types.ts`, and could be changed [here](https://github.com/kiesraad/abacus/blob/92f75e40bfbef103fcfb0b30f96c0b6eb3b110fc/frontend/scripts/gen_openapi_types.ts#L125) to generate something else for these enums.

## Possible other TypeScript types

A quick investigation leads to two other ways to represent this enum in the frontend, whose values can be iterated over:

[TypeScript string enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums)
```typescript
export enum PollingStationStatus {
  first_entry = "first_entry",
  first_entry_in_progress = "first_entry_in_progress",
  definitive = "definitive",
}
``` 

String array as const, and a type that is generated from its values

```typescript
export const pollingStationStatusValues = ["first_entry", "first_entry_in_progress", "definitive"] as const;
export type PollingStationStatus = (typeof pollingStationStatusValues)[number];
```

Do we see the value of being able to iterate over the values, and which solution do we prefer?

lkleuver:
another option: (least amount of refactor I think)
```
export const pollingStationStatusList: PollingStationStatus[] = [...];
```

But probably typescript enum is the most "proper", only reason string literal type is "nicer' is you don't need an extra import / less code.

jschuurk-kr:
> which would have been useful in a recent PR

[...]

PR #630, opened 2024-11-20
Polling station CRUD: create and update 

lkleuver:
Part of #431 and #432

Polling station create and update frontend with partial MSW support.

Out of scope:
Validation
Delete

For validation side of things:
Which fields are shown, which are required, which are dropped.
Figma 'mismatch' with backend currently.

UI Todo:
1. Table misses a "highlight" option
2. Alert's close button doesnt render perfectly. 

Brittle:
Changes to InputField and NumberInput touching a large portion of the UI


oliver3:
>     1. Table misses a "highlight" option

This is part of issue https://github.com/kiesraad/abacus/issues/542

jorisleker:
Took a quick look @lkleuver 
- No validation for now is ok, but can we make `Nummer` and `Aantal kiesgerechtigden` numeric inputs? They can only contain numbers. 
- In your branch, something funny is happening on the PollingStationChoice page:
Scherm­afbeelding 2024-11-21 om 15 11 23

jorisleker:
Great, pollingStationChoice page is fixed now I see, and 'polling station number' and 'aantal kiesgerechtigden' are numeric inputs 👌 

- Can we remove the stepper (up/down arrows) from the polling station number field?
Scherm­afbeelding 2024-11-22 om 11 19 36

- Is it possible to make the label of the 'Aantal kiesgerechtigden' field wider? (see design 👇 )
Scherm­afbeelding 2024-11-22 om 11 18 12

jorisleker (changes requested):
Issues found previously have been fixed 👌 

Did some more reviewing:
- PollingStation overview page is missing a footer
- PollingStation add/edit page is missing a footer
- When editing a PollingStation, the submit button label should be "Wijzigingen opslaan" (currently: "Opslaan en toevoegen" which is the label for the 'add' page)
- When updating a PollingStation to a non unique number, the validation message in the alert:
  - is missing a title
  - has not been translated
  - does not have a close button
  - has a large padding on the right side, making a horizontal scrollbar appear on the page
  - also: the 'invalid' state on the PollingStation number field is missing
- After saving changes to an existing stembureau, no feedback/alert is shown on the polling stations overview page

marlonbaeten (approved):
LGTM

As discussed: I will create a PR to allow passing paths to the `useCrud` hook, which would make `usePollingStationMutation.ts` more trivial.

Using route parameters to render  the save / update notifications feels a bit ad-hoc in this PR - maybe we could generalize this in a next PR?

lkleuver:
> Using route parameters to render the save / update notifications feels a bit ad-hoc in this PR - maybe we could generalize this in a next PR?

I agree, although this seems ok-ish for now.
This would require some form of global app state or api client history?

lkleuver:
> Issues found previously have been fixed 👌
> 
> Did some more reviewing:
> 
> * PollingStation overview page is missing a footer
> * PollingStation add/edit page is missing a footer
> * When editing a PollingStation, the submit button label should be "Wijzigingen opslaan" (currently: "Opslaan en toevoegen" which is the label for the 'add' page)
Done

> * When updating a PollingStation to a non unique number, the validation message in the alert:
>   
>   * is missing a title
>   * has not been translated
>   * does not have a close button
>   * has a large padding on the right side, making a horizontal scrollbar appear on the page
>   * also: the 'invalid' state on the PollingStation number field is missing
I would like to add this to the validation step, also the alert rendering might require a larger refactor to the way we are rendering the layout.

> * After saving changes to an existing PollingStation, no feedback/alert is shown on the polling stations overview page
I can't reproduce this.

praseodym:
> Using route parameters to render the save / update notifications feels a bit ad-hoc in this PR - maybe we could generalize this in a next PR?

We're using URL hashes in other places for this, e.g. https://kiesraad-abacus.pages.dev/elections/1/data-entry#data-entry-saved

frontend/lib/api/polling_station/usePollingStationMutation.ts
```diff
@@ -0,0 +1,46 @@
...
+export type UsePollingStationMutationReturn = {
+  create: PollingStationSubmitCreate;
+  update: PollingStationSubmitUpdate;
+  requestState: LimitedApiRequestState<PollingStation>;
+};
+
+export function usePollingStationMutation(): UsePollingStationMutationReturn {
+  const { requestState, create: crudCreate, update: crudUpdate } = useCrud<PollingStation>(true);
+
+  const create: PollingStationSubmitCreate = React.useCallback(
```
oliver3, line 27: I would make these functions async, and use the void operator when using them in the component, i.e. `void create(electionId, requestObj);` in `PollingStationForm`. Then it is more obvious in the handler that this is an async operation.
The `callAsync` util can then be removed again.

frontend/app/module/polling_stations/page/PollingStationUpdatePage.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useNumericParam } from "@kiesraad/util";
+
+export function PollingStationUpdatePage() {
+  const electionId = useNumericParam("electionId");
+  const pollingStationId = useNumericParam("pollingStationId");
+  const navigate = useNavigate();
+
+  const { requestState } = usePollingStationGet(pollingStationId);
+
+  const onSave = (ps: PollingStation) => {
```
oliver3, line 16:
```suggestion
  const handleSaved = (ps: PollingStation) => {
```

frontend/app/test/unit/test-utils.ts
```diff
@@ -43,12 +43,14 @@ export function getUrlMethodAndBody(call: [input: string | URL | Request, init?:
...
-export async function userTypeInputs(user: UserEvent, inputs: { [key: string]: string | number }) {
+export async function userTypeInputs(user: UserEvent, inputs: { [key: string]: string | number }, skipCheck?: boolean) {
   for (const [key, value] of Object.entries(inputs)) {
     const input = await screen.findByTestId(key);
     await user.clear(input);
     await user.type(input, value.toString());
-    expect(input).toHaveValue(value.toString());
+    if (!skipCheck) {
+      expect(input).toHaveValue(value.toString());
+    }
```
jschuurk-kr, line 53: The `skipCheck` is not a good solution and not what was suggsted in Mattermost.

More importantly, we should use locators based on roles, not on ids when possible. That wasn't possible for data entry (it is now I just learned), hence the usage of ids there.

So for this PR the locators for polling stations need to be changed. I can pick up the data entry changes in a new PR.

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    // mainly to stop eslint from complaining about unused variables
+    expect(id).toBe(1);
+    expect(election_id).toBe(1);
+
+    await userTypeInputs(user, inputs, true);
+
+    const pollingStationType = screen.getByTestId(`polling_station_type-${polling_station_type}`);
+    await userEvent.click(pollingStationType);
+
+    await userEvent.click(screen.getByRole("button", { name: t("polling_station.form.save_create") }));
```
jschuurk-kr, line 43: Am not a fan of using the translation functionality like this in a test. It makes the test harder to read. And it means the test no longer checks for the correct text, only for the correct translation key.

frontend/lib/api/polling_station/usePollingStationMutation.ts
lkleuver: It's specifically in the form component I would prefer not to use async code since this is not the flow we are using but are using state changes. using "both" to me seems confusing.
I see "create / update" as an "action" that changes state.

I'm fine either way, just prefer to reach a consensus first.

frontend/app/module/polling_stations/page/PollingStationUpdatePage.tsx
lkleuver: Hope I'm not being difficult, any specific reason for this? (If this is the convention we use thats good enough for me)
onSave is intended as a listener, handleSaved feels like it does something to the saved item.

frontend/app/test/unit/test-utils.ts
lkleuver: yes, i made this change before your suggestion, I will update the PR with a the suggested workaround.

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
lkleuver: I chose this because now changing a translation doesn't require 2 (or more) files to be edited.
But not going to argue with the expert :) will change in next update.

frontend/app/component/form/polling_station/PollingStationForm.tsx
jschuurk-kr, line 1: Adding a polling station with a really long name is not handled well by the polling station overview page:
image

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+                label={t("polling_station.locality")}
+                defaultValue={pollingStation?.locality}
+              />
+            </FormLayout.Row>
+          </FormLayout.Section>
+
+          <FormLayout.Controls>
+            <Button type="submit" name="submit">
+              {pollingStation ? t("polling_station.form.save_update") : t("polling_station.form.save_create")}
+            </Button>
```
jschuurk-kr, line 144: I'm missing a cancel button for when I update a polling station and change my mind.

frontend/app/module/polling_stations/page/PollingStationUpdatePage.tsx
oliver3: It is mostly about the save vs. saved, because this is not actually saving but will be called after it has been saved, if I understand this correctly. 
The `handle` vs. `on` is what I thought a convention indeed.

frontend/lib/api/labels/index.ts
```diff
@@ -0,0 +1,21 @@
+import { PollingStationType } from "@kiesraad/api";
+import { t } from "@kiesraad/i18n";
+
+export const labelForPollingStationType = withEntries<{ [K in PollingStationType]: string }>({
+  FixedLocation: t("polling_station.type.FixedLocation"),
```
oliver3, line 5: I think that this object is now a bit redundant and the labels can be used directly in the `PollingStationForm`. I'd also prefer not to use translations in the API "layer" of the frontend.

```tsx
{pollingStationTypeValues.map((pollingStationType) => (
  <ChoiceList.Radio
    key={pollingStationType}
    id={`polling_station_type-${pollingStationType}`}
    name={"polling_station_type"}
    defaultValue={pollingStationType}
    defaultChecked={pollingStation?.polling_station_type === pollingStationType}
    label={t(`polling_station.type.${pollingStationType}`)}
  />
))}

```
Ideally the `pollingStationTypeValues` comes from the openapi.ts as suggested in https://github.com/kiesraad/abacus/issues/489

frontend/lib/api/labels/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  Special: t("polling_station.type.Special"),
+  Mobile: t("polling_station.type.Mobile"),
+});
+
+function withEntries<T extends { [K in keyof T]: T[K] }>(
```
oliver3, line 10: I would prefer using the standard `Object.entries()` directly (less surprises reading the code), but I think that this might not be needed at all anymore.

oliver3:
@jorisleker 
>     * No validation for now is ok, but can we make `Nummer` and `Aantal kiesgerechtigden` numeric inputs? They can only contain numbers.

Numeric inputs (`input type=number`) will still not prevent typing numbers and will add complexity in removing the spinner buttons, and for our current tests. Is it okay to also leave this out for when we do validation in another PR?

frontend/lib/api/polling_station/usePollingStationMutation.ts
lkleuver: i've used `void useCrud`.

frontend/app/module/polling_stations/page/PollingStationUpdatePage.tsx
lkleuver: handleSaved

frontend/app/test/unit/test-utils.ts
lkleuver: stopped using the util and used roleBased selectors

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
lkleuver: removed programmatic translations from test files

frontend/app/component/form/polling_station/PollingStationForm.tsx
lkleuver: I think this is a nice design debt issue, but i've added a simple word-break rule to the css.

frontend/app/component/form/polling_station/PollingStationForm.tsx
lkleuver: I had discussed this with @jorisleker, I think the conclusion then was that the browser back button works.

When working on this I do miss it, specifically when you open de page via a direct (deep) link.

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+        </FormLayout.Alert>
+      )}
+
+      <Form ref={formRef} onSubmit={handleSubmit} id="polling-station-form">
+        <FormLayout disabled={requestState.status === "loading"}>
+          <FormLayout.Section title={t("general_details")}>
+            <input type="hidden" id="election_id" name="election_id" defaultValue={electionId} />
+            <input type="hidden" id="id" name="id" defaultValue={pollingStation?.id} />
+
+            {/* props that are not in design but are in the pollingstation model */}
```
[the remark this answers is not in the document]
lkleuver: I added house number but left out addition.

frontend/lib/api/labels/index.ts
lkleuver: removed, we do need to share this somewhere I think. It's used in 3 places currently.

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  return (
+    <div>
+      {requestState.status === "api-error" && (
+        <FormLayout.Alert>
+          <Alert type="error">{requestState.error.message}</Alert>
+        </FormLayout.Alert>
+      )}
+
+      <Form ref={formRef} onSubmit={handleSubmit} id="polling-station-form">
```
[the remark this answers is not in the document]
lkleuver: Form is generic enough to handle this use-case too I believe, lets create an issue?

jorisleker (changes requested):
- the 'Annuleren' secondairy button is missing from the 'edit' page
- the breadcrumb for the add/edit page misses the 'stembureaus' node
- the alert shown when polling station number is not unique still has padding on the right side, should be full width. (I think adding the actual validation message is part of the validation issue?)
- The copy of the success message when adding a stembureau is not according to design. Font-weight in design is bold, and message contains both polling station number and name (see [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=5445-52093&node-type=frame&t=Lm70TAtSuFNoppcO-11)).
- No feedback is shown after updating an existing pollingStation. The design asks for a  'changes saved' success alert ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=5452-54955&node-type=frame&t=Lm70TAtSuFNoppcO-11)).

Scherm­afbeelding 2024-11-28 om 08 39 48

frontend/app/module/polling_stations/page/PollingStationCreatePage.tsx
```diff
@@ -9,7 +9,7 @@ import { useNumericParam } from "@kiesraad/util";
 export function PollingStationCreatePage() {
   const electionId = useNumericParam("electionId");
   const navigate = useNavigate();
-  const onSave = (ps: PollingStation) => {
+  const handleSaved = (ps: PollingStation) => {
     navigate(`../?created=${ps.id}`);
```
oliver3, line 13: At this moment, we don't have an ID, but let's implement this in the validation issue https://github.com/kiesraad/abacus/issues/638

jorisleker:
> > * the 'Annuleren' secondairy button is missing from the 'edit' page
> 
> Added the button, doesnt this make sense for create as well?

Nope. Too risky button. Best practice is to not add cancel/destructive buttons to forms where every field is newly entered data.  

> > * the breadcrumb for the add/edit page misses the 'stembureaus' node ([Figma]
> 
> I think this should be a separate issue, breadcrumbs, it has no code or proper design yet.

Agreed. Will add issue in this epic.

> > * No feedback is shown after updating an existing pollingStation. The design asks for a  'changes saved' success alert
> 
> Still can't reproduce this, is the `?updated=<id>` in the url after you've saved?

I get `1/polling-stations/?updated=undefined` in the url. Tested on https://630.abacus-test.nl/elections/1/polling-stations/update/1

frontend/app/module/polling_stations/page/PollingStationCreatePage.tsx
oliver3: This is not working for me, same issue that @jorisleker is mentioning

praseodym: Same here in Safari

jorisleker (approved):

frontend/app/module/polling_stations/page/PollingStationListPage.tsx
```diff
@@ -22,14 +22,48 @@ export function PollingStationListPage() {
...
+      {createdPollingStation && (
+        <Alert type="success" onClose={closeAlert}>
+          <strong>
+            {t("polling_station.title.singular")} {createdPollingStation.number} ({createdPollingStation.name}){" "}
+            {t("added")}
+          </strong>
+        </Alert>
+      )}
       <main>
         {!data.polling_stations.length ? (
```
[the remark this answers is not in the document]
jschuurk-kr: design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=5995-47193&node-type=frame&t=ou68qrmfEBGlsvau-0

frontend/app/component/form/polling_station/PollingStationForm.tsx
jschuurk-kr: Word-break rule is working, but I now get "Bijzonder" split over two lines:
image

Is word-break etc still as much of a drama across browsers as it was several years ago? Am testing with Firefox.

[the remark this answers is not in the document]
jschuurk-kr: @marlonbaeten What are your thoughts on this? :point_up_2:

frontend/app/component/form/polling_station/PollingStationForm.tsx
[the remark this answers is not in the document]
jschuurk-kr: The keyboard features of Form are not working well with this page:
- up/down moves through the radio buttons in the order 2nd, 1st, 3rd and vice-versa
- enter takes you through Nummber - Naam - Soort stembureau - Opslaan en toevoegen, skipping the fields in between
- shift + down moves the cursor to Plaats
- shift + up does not work
- down does not let you to reach the "Annuleren" button on the edit-page

Shift+enter lets you submit. That's fine, but not needed and als not communicated by the page.

frontend/lib/i18n/locales/nl/generic.json
```diff
@@ -2,8 +2,10 @@
   "": "",
   "test": "test",
   "account_configured": "Je account is ingesteld",
+  "added": "Toegevoed",
```
jschuurk-kr, line 5:
```suggestion
  "added": "Toegevoegd",
```

frontend/lib/api-mocks/RequestHandlers.ts
```diff
@@ -225,6 +229,83 @@ export const PollingStationDataEntryFinaliseHandler = http.post<
...
+      const updatedPollingStation: PollingStation = {
+        ...ps,
+        ...json,
+      };
+
+      Database.pollingStations = Database.pollingStations.map((ps) =>
+        ps.id === pollingStationId ? updatedPollingStation : ps,
+      );
+
+      return HttpResponse.json(updatedPollingStation, { status: 200 });
```
jschuurk-kr, line 281: As you mentioned, body of 200 response from backend is empty.

frontend/app/component/form/polling_station/PollingStationForm.tsx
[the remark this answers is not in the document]
praseodym: > The current design was based on the assumption that there was one address field, but our backend and database model indeed has separate fields. An older design has all the fields, so let's do that until we change it everywhere.

Is there an issue to track that change?

frontend/app/module/election/page/ElectionStatusPage.tsx
jschuurk-kr, line 1: There's a TODO in this file that I think can be removed:
`{/* TODO: Add button onClick to Create Polling Station page */}`

marlonbaeten (dismissed):

frontend/app/component/form/polling_station/PollingStationForm.tsx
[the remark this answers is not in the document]
marlonbaeten: Our generic error handling splits errors in two categories: fatal and non-fatal. By default fatal errors will be thrown and caught by our router in in which case an applicable error page is rendered. Non-fatal errors will be returned to the caller, in this case "requestState" wil be in an error state. Each error returned from the backend should have a "reference" that can de translated into an error message. I'm not sure what causes the "hang" in this case, but I can improve error handling by rendering a generic overlay for non-fatal errors that I'm working on in #645

frontend/app/component/form/polling_station/PollingStationForm.tsx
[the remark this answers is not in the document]
lkleuver: created a separate issue: 
https://github.com/kiesraad/abacus/issues/672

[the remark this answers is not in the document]
jorisleker: Up/down arrows should only work for the data entry forms. These have input fields that are - like in a spreadsheet - above each other. 

Editing a polling station is much more a classic web form (and used by a different user group). Keyboardnav should be limited to default browser behavior, without the add-ons we made for the results forms.

frontend/app/module/polling_stations/page/PollingStationCreatePage.tsx
jschuurk-kr: @lkleuver What's the status of this thread? What @oliver3 said, i.e. will be done in issue #638?

frontend/app/component/form/polling_station/PollingStationForm.tsx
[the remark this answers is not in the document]
jschuurk-kr: Resolving because will be picked up in #672 as part of this epic.

frontend/app/component/form/polling_station/PollingStationForm.tsx
jschuurk-kr: @lkleuver With long polling stations, the "Soort" column becomes too narrow. Any updates on this?

frontend/app/module/polling_stations/page/PollingStationCreatePage.tsx
lkleuver: the undefined came from the assumption of the backend returning the updated polling station, I've changed this to use the param from react router in PollingStationUpdatePage so the undefined is now defined.

frontend/app/component/form/polling_station/PollingStationForm.tsx
lkleuver: We've added an issue for the Table to fix this separately

oliver3 (approved):

marlonbaeten (approved):

frontend/app/component/form/polling_station/PollingStationForm.tsx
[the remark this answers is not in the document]
marlonbaeten: This will be fixed with #670

frontend/app/component/form/polling_station/PollingStationForm.tsx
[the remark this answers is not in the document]
lkleuver: created issue #684

Issue #621, opened 2024-11-20
Update react-router-dom to 7.x


oliver3:
Update react-router-dom to version 6.28.0 which needs several flags to be set to prevent warnings:

https://reactrouter.com/en/6.28.0/upgrading/future#update-to-latest-v6x

This caused issues in the automatic navigation from `data-entry/1` to `data-entry/1/recounted` in the tests 

Branch https://github.com/kiesraad/abacus/tree/react-router-dom-issues

praseodym:
react-router-dom version 7.0.1 has been released a few days ago, there is an upgrade guide available: https://reactrouter.com/upgrading/v6

praseodym:
Useful addition in react-router-dom 7: [Typesafety improvements](https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#typesafety-improvements) / [Route Module Type Safety](https://reactrouter.com/dev/how-to/route-module-type-safety)

PR #641, opened 2024-11-25
Dependabot should ignore react-router-dom versions 6.28 and 7

oliver3:
Updating react-router-dom needs code changes, see also #621 

This PR is to let dependabot ignore minor and major updates for react-router-dom


praseodym:
I think we need to ignore both major and minor versions so we don't get the 6.28 version again?

oliver3:
> I think we need to ignore both major and minor versions so we don't get the 6.28 version again?

Indeed, not sure why I thought that would be covered by major..

praseodym (approved):

jschuurk-kr (approved):

Issue #623, opened 2024-11-20
Unfinalised data entry not shown on polling station choice page


jorisleker:
Tested on main (e2ff56f7)

**Reproduction steps**
- start a data entry
- abort the data entry, and choose to save it
- you are redirected to the polling station choice page, which contains a blue box with a link to the unfinished data entry. So far so good.
- open the data entry again
- don't make any changes
- Click on one of the links in the breadcrumb navigation (ie. the election name)
- Expected: the unfinished data entry is shown on polling station choice page (as in 3rd bullet)
- Observed: the polling station choice page does not show your unfinished data entry

**Note**: 
On the election status page (for coordinators) there is difference in how the data entry is shown:
- if it is shown on polling station choice page, it is displayed in the 'invoer afgebroken' table
- If it not shown on the polling station choice page, it is displayed in the 'invoer bezig' table

**What needs to be fixed**:
- For the polling station choice page, the difference between 'aborted' and 'busy' is irrelevant. It should be shown in both cases. 
- When navigating away from a data-entry without changes, the 'do you want to save' dialog is not shown. But the 'aborted' status should be set, as you are no longer working on that data-entry.

praseodym:
Please also fix the `json_extract` nightmare introduced in #481:

              Impressive SQL work here, but architecturally this is a nightmare: we're now using the untyped _client state_ (only meant to let the client resume where it left off!) in the backend by letting SQLite parse JSON for each row.

I think it would be good to reconsider the approach when fixing #623, e.g. by adding a column for this state.

_Originally posted by @praseodym in https://github.com/kiesraad/abacus/pull/481#discussion_r1850822737_

oliver3:
Decided together with @jorisleker to leave the 'aborted' state for what it is, for now, and only fix that data entry with both statuses should show up as 'openstaande invoer' on the polling station choice page.

praseodym:
> Decided together with [@jorisleker](https://github.com/jorisleker) to leave the 'aborted' state for what it is, for now, and only fix that data entry with both statuses should show up as 'openstaande invoer' on the polling station choice page.

If we leave that part as-is, are we still going to find a way not to use the client state in the backend?

PR #643, opened 2024-11-26
Also show first_entry_in_progress as "openstaande invoer"

oliver3:
Fixes issue https://github.com/kiesraad/abacus/issues/623 by showing "openstaande invoer" for both `first_entry_unfinished` and `first_entry_in_progress` polling stations.

Added testcase in `PollingStationChoiceForm.test.tsx`

The changing of the status when navigating away from a resumed data entry has not been done.

Please refer to the original issue for how to test it.


praseodym (dismissed):
lgtm, but see https://github.com/kiesraad/abacus/issues/623#issuecomment-2501965472 for the scope of this issue

jorisleker (changes requested):
✅ The in-progress data entry is correctly shown now. 

Found another bug (that was probably already present, but became more obvious now because we show the 'in progress' box on the polling station choice page more often):

- start a data entry for 1 election
- don't finished it, abort it (do save)
- navigate to another election
- observed: the 'you have open data entries' alert is shown, but without any polling stations
- expected: the alert is not shown for this election.

jschuurk-kr:
While reproducing the issue described by Joris, I found something related to the save/discard modal shown when navigating away from the data entry pages.

Steps to reproduce:
1. Fill in a polling station until you're on the "Toegelaten kiezers en uitgebrachte stemmen" page.
2. Click in the top bar on the election name.
3. The save/discard modal appears. Save your input.
4. You're on the polling station selection page.
5. Select the same polling station again.
6. Fill in any of the fields.
7. Click in the top bar on the election name.
8. The save/discard modal does not appear. The input is also not cached.

Based on our [Navigate away from page flowchart](https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/data-entry-forms.md#navigate-away-from-page) I'd expect the opposite behavior:
- I did not make changes on the page at step 3, so I shouldn't see the modal.
- I did make changes on the page at step 7, so I should see the modal.
- Input caching (saving in frontend without submittin to backend) should happen in neither scenario.

jschuurk-kr:
The [bug](https://github.com/kiesraad/abacus/pull/643#pullrequestreview-2464337403) mentioned by @jorisleker above als occurs on the coordinator's status page. The "Invoer bezig (1)" table is shown but with zero entries.

So it seems that somewhere we don't properly filter polling stations on election id?

praseodym:
> The [bug](https://github.com/kiesraad/abacus/pull/643#pullrequestreview-2464337403) mentioned by @jorisleker above als occurs on the coordinator's status page. The "Invoer bezig (1)" table is shown but with zero entries.
> 
> So it seems that somewhere we don't properly filter polling stations on election id?
> 
> This is a backend issue:
> 
> * `http://localhost:3000/api/elections/3` returns three polling stations. (ids 4,5,6)
> * `http://localhost:3000/api/elections/3/status` returns four polling stations. (ids 3,4,5,6)

I suspect the bug is in this code:

https://github.com/kiesraad/abacus/blob/08d5a5d8923c8846d554fcbf58db92540e8b97cd/backend/src/polling_station/repository.rs#L226-L231

In SQL, `AND` takes precedence over `OR`, which means that the `WHERE` clause is evaluated like this:
```sql
(election_id = $1 AND de.polling_station_id IS NULL)
OR
 (de.entry_number IN
    (SELECT MAX(entry_number)
     FROM polling_station_data_entries
     WHERE polling_station_id = p.id
     GROUP BY polling_station_id))
```
So the second part of the `WHERE` clause does not filter by election id.

@cikzh is already working on refactoring this query into a state machine, so let's make sure this bug gets fixed in his branch.

oliver3:
Updated the sql query and added a specific backend integration test

jorisleker (approved):
Retested. 
- open and aborted data entry correctly shown on pollingStationChoice page and status page
- open/aborted data entry not shown on pollingStationChoice page and status page for other elections. 

✅ approved

jschuurk-kr (approved):

PR #646, opened 2024-11-26
Improve data entry locators in RTL and Playwright tests

jschuurk-kr:
Now that our data entry tables have `aria-labelledby` attributes, we can replace the `...ByTestId` locators with `...ByRole` locators. This applies both to React Testing Library (RTL) and Playwright tests. So I replaced the ones that were easy enough to replace.

All the changes were straightforward, except the one in `CandidatesVotesForm.test.tsx` where I added the helper function `candidateName()` to compose the value for the `name` parameter of the `...ByRole` locator.

Locators I didn't replace in this PR:
1. The ones used by `"CandidatesVotesForm renders correctly"` in `CandidatesVotesForm.test.tsx` (RTL), because we want to assert on the value of the "Kandidaat" field, not locate the textbox field.
2. The one used by `userTypeInputs()` in `frontend/app/test/unit/test-utils.ts` (RTL), because significant refactor and we have enough other tests using `...ByRole` locators for the same fields.
3. Any `...ByTestId` locator used for something else than a textbox field (RTL), because not in scope.
4. The ones in `CandidatesListPgObj.ts` (Playwright), because bigger refactor with limited value since some of the RTL tests  use `...ByRole` locators.

---

I also discovered we didn't have RTL tests for the recounted fields. So I extended the test that fills in all the fields and added a test for checking the API request sent by the form.


praseodym (approved):

frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.test.tsx
```diff
@@ -116,7 +123,9 @@ describe("Test CandidatesVotesForm", () => {
 
       renderForm({ recounted: false });
 
-      const candidate1 = await screen.findByTestId("candidate_votes[0].votes");
+      const candidate1 = await screen.findByRole("textbox", {
+        name: `1 ${candidateName(politicalGroupMockData.candidates[0] as Candidate)}`,
+      });
```
jschuurk-kr, line 128: Discussed with @oliver3: change into a `map()` to create an array of candidate names, then use that array in the `findbyRole()`s.

jschuurk-kr: done in https://github.com/kiesraad/abacus/pull/646/commits/70f72e696d1a60a07ae3ebcce16e46b532748ee7

frontend/app/test/unit/test-utils.ts
```diff
@@ -52,18 +70,11 @@ export async function userTypeInputs(user: UserEvent, inputs: { [key: string]: s
...
-export const setupTestRouter = () => {
-  return createMemoryRouter(routes, {
-    future: {
-      v7_normalizeFormMethod: true,
-    },
+export function getCandidateFullNamesFromMockData(politicalGroupMockData: PoliticalGroup): string[] {
+  const candidateNames = politicalGroupMockData.candidates.map((candidate) => {
+    return candidate.first_name
+      ? `${candidate.last_name}, ${candidate.initials} (${candidate.first_name})`
+      : `${candidate.last_name}, ${candidate.initials}`;
```
lkleuver, line 77: might want a formal name rendering utility for this.

lkleuver (approved):
LGTM

When should you use getByRole vs await findByRole?
Couldnt quite catch the method of choosing which from the code in this PR

frontend/app/test/unit/test-utils.ts
jschuurk-kr: Not sure what you mean by "fornal name rendering utility"? Goal of this function is to concatenate the relevant parts of the mock data into the string shown on the screen. The simpler we keep that function the better.

jschuurk-kr:
> When should you use getByRole vs await findByRole? Couldnt quite catch the method of choosing which from the code in this PR

@lkleuver Good catch, I did not keep what was there. Will fix.

As to when to use what:
- getBy: queries without retries/waiting
- findBy; queries with retries/waiting (default timeout is 1000 ms)
- queryBy: [only if](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence) you want to check `.toBeInTheDocument()` or `not.toBeInTheDocument()`

lkleuver (approved):

frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.test.tsx
```diff
@@ -123,9 +117,9 @@ describe("Test CandidatesVotesForm", () => {
 
       renderForm({ recounted: false });
 
-      const candidate1 = await screen.findByRole("textbox", {
-        name: `1 ${candidateName(politicalGroupMockData.candidates[0] as Candidate)}`,
-      });
+      const candidateNames = getCandidateFullNamesFromMockData(politicalGroupMockData);
+
+      const candidate1 = await screen.findByRole("textbox", { name: `1 ${candidateNames[0]}` });
```
praseodym, line 122: Why not add the index prefix to the return value of `getCandidateFullNamesFromMockData`?

jschuurk-kr: Because that would result in a single function that goes from mock data to an aria-label that's the concatenation of two columns in the UI. That seems to me like too many things (and something too specific) for a single function.

oliver3 (approved):

PR #655, opened 2024-11-28
Button tertiary

oliver3:
Added tertiary button variant for delete polling station.

Button can be seen in the (hopefully) now deployed preview ladle pages at `/ladle`.

It should have all the proper states and behave as expected.

### DIFF
--- a/frontend/lib/ui/Button/Button.module.css
+++ b/frontend/lib/ui/Button/Button.module.css
@@ -78,6 +78,46 @@
     }
   }
 
+  &.tertiary {
+    --button-border-width: 0px;
+    border: none;
+    background: transparent;
+    color: var(--color-error-darkest);
+
+    &:not(:hover, :active) {
+      box-shadow: none;
+    }
+
+    &:hover {
+      background: #fef9fa;
+    }
+
+    &:focus,
+    &:focus-visible {
+      box-shadow: none;
+      outline: none;
+      --button-border-width: 2px;
+      border: var(--button-border-width) solid var(--border-color-focus);
+    }
+
+    &:active {
+      background-color: var(--color-error-bg);
+    }
+
+    &:disabled {
+      box-shadow: none;
+      background: var(--button-color-bg-disabled);
+      border-color: var(--button-color-bg-disabled);
+      color: var(--button-color-disabled);
+    }
+  }
+
+  &:disabled {
+    background: var(--button-color-bg-disabled);
+    border-color: var(--button-color-bg-disabled);
+    color: var(--button-color-disabled);
+  }
+
   &.alert {
     border-color: var(--color-error);
     background: var(--color-error);
--- a/frontend/lib/ui/Button/Button.stories.tsx
+++ b/frontend/lib/ui/Button/Button.stories.tsx
@@ -1,65 +1,44 @@
-import type { Story } from "@ladle/react";
+import type { Story, StoryDefault } from "@ladle/react";
 
-import { Size, Variant } from "@kiesraad/ui";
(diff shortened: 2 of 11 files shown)


jorisleker:
Some questions:
- do we expect all tertiary buttons to be red? I don't think so. Think the red font is a separate attribute, something like 'destructive' and the default state of the button should be same as 
- same for the 'alert' button. Buttons in alerts are not red by default. This is a 'destructive' variant of our primary button. Our secundairy button could have a destructive state as well. 

(I you agree, I will update Figma accordingly)

oliver3:
>     * do we expect all tertiary buttons to be red? I don't think so. Think the red font is a separate attribute, something like 'destructive' and the default state of the button should be same as
> 
>     * same for the 'alert' button. Buttons in alerts are not red by default. This is a 'destructive' variant of our primary button. Our secundairy button could have a destructive state as well.

It is more than just the red font, also background for different states. I think I would prefer it as a separate variant and not create a matrix of variant x attribute, or just apply a different font color to a button variant, because buttons have many states and it is hard to keep them consistent.

We could create variants:
- primary (current default) and primary-destructive (current alert), and
- tertiary (current ghost?) and tertiary-destructive (current tertiary)

frontend/lib/ui/Button/Button.stories.tsx
```diff
@@ -1,65 +1,44 @@
...
     text: "Invoer",
   },
   argTypes: {
-    variant: {
-      options: ["default", "secondary", "ghost", "alert"],
-      control: { type: "radio" },
-    },
     size: {
-      options: ["xs", "sm", "md", "lg"],
+      options: ["xs", "sm", "md", "lg", "xl"],
```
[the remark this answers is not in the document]
jschuurk-kr: Discussed with @oliver3. `xl` button being smaller than the `lg` one is the way it is. Addressing this is not in scope of the current PR, which goal is to add a tertiary button.

frontend/lib/ui/Button/Button.module.css
```diff
@@ -78,6 +78,40 @@
...
+    border: none;
+    background: transparent;
+    color: var(--color-error-darkest);
+
+    &:not(:hover, :active) {
+      box-shadow: none;
+    }
+
+    &:hover {
+      background: #fef9fa;
```
lkleuver, line 92: any reason this isn't a var?

frontend/lib/ui/Button/Button.module.css
```diff
@@ -78,6 +78,40 @@
...
+    }
+
+    &:focus,
+    &:focus-visible {
+      box-shadow: none;
+      outline: none;
+      --button-border-width: 2px;
```
lkleuver, line 99: what does this do?

lkleuver (dismissed):

frontend/lib/ui/Button/Button.module.css
oliver3: This CSS var is used to accommodate for different border sizes for different button states, without making the button size change. This was done in https://github.com/kiesraad/abacus/pull/591/files

frontend/lib/ui/Button/Button.module.css
oliver3: This color was not yet in our variables and I'm not sure what old and new colors are. I'll have a call with @jorisleker about this soon.

jorisleker (approved):

frontend/lib/ui/Button/Button.module.css
oliver3: Added the Figma colors as css variables and used those in the new button variant.

lkleuver (approved):

PR #659, opened 2024-11-29
add integration tests for deleting polling station with data entry or with result

jschuurk-kr:
If you try to delete a polling station that has either a data entry or a result associated with it, the API call fails with a 422: "error":"Invalid data","fatal":false,"reference":"InvalidData"}. The error originates from the `ForeignKeyViolation` in `backend/src/error.rs` and the foreign keys defined for the data entries table and the results table.

We did not have tests for these two scenarios. While it may be unlikely that we break this functionality by removing the foreign keys, these tests also serve as documentation of this important behavior.

To be able to add the tests, I had to add a new fixture for the data entries table. So polling station 2 "Testplek" of election 1 "Municipal Election" now has an entry in that table. I generated the data by opening the recounted page for that polling station and immediately click "afbreken" en then "opslaan". Luckily that did not break any of the other tests.

Questions:
- should the new fixture be an API call instead?
- if it remains a fixture, should it be loaded when seeding fixtures?


praseodym:
> * should the new fixture be an API call instead?

I think API calls are a bit more maintainable because they are type checked. We can make them into reusable 'fixtures' by extracting them into separate helper functions.

> * if it remains a fixture, should it be loaded when seeding fixtures?

No, this way is fine to me. I think we should aim for less variety of different elections for the seeded fixtures, because as a user the current situation gets very confusing. I would prefer having only one test election (e.g. Juinen), in Dutch, to make the application with seeded fixtures easier to understand. That's something for a new issue, though.

praseodym (approved):

oliver3 (approved):

PR #668, opened 2024-12-02
Add jobs to upload binaries to abacus-test.nl for test and release

marlonbaeten:


.github/workflows/release.yml
```diff
@@ -143,3 +143,28 @@ jobs:
           files: release/*
           token: ${{ secrets.GITHUB_TOKEN }}
           name: Version ${{ env.TAG }}
+
+
+  deploy:
+    name: Deploy to abacus-test.nl
```
jschuurk-kr, line 149: With build-lint-test doing this for the `main` branch, what's the reason for also including it here?

marlonbaeten: The release build is a bit different, its build in release mode, the binary is smaller (for linux) and compiled against musl. We want to test this binary with abacus-test.nl

jschuurk-kr: Oh of course, download step is the same, but the build job that does the upload is different between the two workflows.

marlonbaeten:
Note: the commit hash used for the deploy is currently based on the merge-commit: I'll change this to the latest commit hash, for simplicity.

praseodym:
> Note: the commit hash used for the deploy is currently based on the merge-commit, when the action was triggered by a PR.

That's fine, because that's the actual commit that was built. It could change when the branch is re-built after an update the PR target branch.

.github/workflows/build-lint-test.yml
```diff
@@ -183,7 +183,7 @@ jobs:
       - backend
     environment:
       name: test
-      url: https://${{ github.sha }}.abacus-test.nl
+      url: https://${{ github.event.pull_request.head.sha || github.sha }}.abacus-test.nl
```
praseodym, line 186: I think we should always use `github.sha` because that's what we built. If we use `github.event.pull_request.head.sha`, we can have two builds with the same 'hash' that have been merged with different versions of `main` with no way to tell the difference.

marlonbaeten: That is true - but I have no way of linking that virtual merge commit to a PR. GitHub does not expose this commit hash in the (graphql) API as far as I can see. For the GitHub state I prefer to have a single source of truth, thus "linking" these commits by an upload call would not be the nicest option.

marlonbaeten:
Note: I discussed with @praseodym that we need a binding between the commit that triggered the build and the build commit in order to show the information correctly on abacus-test. I will update this PR once abacus-test has the binding feature. ETA for the feature is tomorrow.

praseodym (approved):

.github/workflows/build-lint-test.yml
```diff
@@ -174,3 +176,26 @@ jobs:
...
+  deploy:
+    name: Deploy to abacus-test.nl
+    needs:
+      - backend
+    environment:
+      name: test
+      url: https://${{ github.sha }}.abacus-test.nl
+    permissions:
+      contents: read
+      deployments: write
```
oliver3, line 189: Why does this need deployment write access?

marlonbaeten: It creates a new deployment ->  see the block below this comment "This branch was successfully deployed
2 active deployments"

rnijveld (approved):

praseodym (approved):

PR #671, opened 2024-12-02
Unite voters counts and voters recounts

rnijveld:
This changes voters counts and voters recounts to be the same struct, this also requires some frontend renames of variables, but the user interface should stay exactly the same. This is change is needed to simplify working with PollingStationResults greatly, especially for the XML output (but already demonstrated for the summary).

Aside from the simplification, I've also added a `FieldPath` struct that removes al the `format!` macro calls that were used previously for constructing the path to a field. This FieldPath is used by the combined VotersCounts struct to determine which error to construct (the error number for the recount is different). An alternative I considered was to move the code up a level, but I felt that the validation would become too centralized if I did that (validate method for PollingStationResults is already quite large).

I've tested entering the data for an election that uses recounts and I could successfully complete entering valid data, but this should probably be tested by someone else as well, just to confirm nothing broke.

In order to let the VotersAndVotesForm work with this new setup, in which there are repeating fields, I had to add some (for my taste) very hacky code in 590ea7e1f74cdc3a68e0145e0a3042fbd0f15b4d. If there is any good alternative without having to do a massive rewrite here, I'd love to change it. And it probably needs some testing as well.


praseodym (approved):

lkleuver (approved):

rnijveld:
I've added issue #678 to make sure we don't forget to improve the conversion of error paths to field ids.

PR #685, opened 2024-12-03
Add API endpoint to download EML_NL XML

rnijveld:
- This adds a new API endpoint for downloading/viewing the EML_NL 510b XML.
- ~~I've added a new property in Election for the ElectionSubcategory, which is used in the XML~~
- I've added a new property in Election for the number of seats, in order to fill the ElectionSubcategory in the EML
- There are some additional properties that will need to be added to our election datastructure as well, but most of them will be entered via the EML_NL election definition, so I added todos for these properties to be filled in at a later time.
- All of the frontend changes are just addding the new subcategory property to elections, so this mostly needs backend reviews
- I've made some elements in the XML structs non-optional because those requirements were EML_NL specific and added the `UncountedVotes`


chrismostert:
Minor point, but the hardcoded `CSB` value: https://github.com/kiesraad/abacus/blob/3bfa7563bc3a5aec62dae3d8b475ed59c42a6358/backend/src/eml/mod.rs#L36
results in invalid ReportingUnitIdentifiers (`CSB::{\d+}`):
https://github.com/kiesraad/abacus/blob/3bfa7563bc3a5aec62dae3d8b475ed59c42a6358/backend/src/eml/mod.rs#L292

This is because the EML_NL standard expects reporting unit identifiers to follow the following pattern:
`(HSB\d+)|((HSB\d+::)?\d{4})|(((HSB\d+::)?\d{4}::)?SB\d+)|(HSB\d+::SB\d+)`

praseodym:
> Minor point, but the hardcoded `CSB` value:
> 
> https://github.com/kiesraad/abacus/blob/3bfa7563bc3a5aec62dae3d8b475ed59c42a6358/backend/src/eml/mod.rs#L36
> 
> 
> results in invalid ReportingUnitIdentifiers (`CSB::{\d+}`):
> https://github.com/kiesraad/abacus/blob/3bfa7563bc3a5aec62dae3d8b475ed59c42a6358/backend/src/eml/mod.rs#L292
> 
> This is because the EML_NL standard expects reporting unit identifiers to follow the following pattern, so the polling station number should also include the prefix 'SB': `(HSB\d+)|((HSB\d+::)?\d{4})|(((HSB\d+::)?\d{4}::)?SB\d+)|(HSB\d+::SB\d+)`

Maybe we can replace `authority_id` with a hardcoded `0000` for now, then it should validate. This number can be replaced with the [CBS code](https://en.wikipedia.org/wiki/Municipalities_of_the_Netherlands) of the municipality (from the election definition EML_NL XML) in the future.

backend/migrations/1_elections.sql
```diff
@@ -5,6 +5,7 @@ CREATE TABLE elections
     location         TEXT                              NOT NULL,
     number_of_voters INTEGER                           NOT NULL,
     category         TEXT                              NOT NULL,
+    subcategory      TEXT                              NOT NULL,
```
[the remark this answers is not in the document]
rnijveld: Replaced by `number_of_seats` in 99f03bfd5c1c4b9815d5a24835d490bee6969015

backend/src/eml/mod.rs
```diff
@@ -87,6 +87,13 @@ impl EML510 {
 
 impl base::EMLDocument for EML510 {}
 
+fn election_subcategory(election: &crate::election::Election) -> Option<String> {
+    match (&election.category, election.number_of_seats) {
+        (crate::election::ElectionCategory::Municipal, ..19) => Some("GR1".into()),
```
jschuurk-kr, line 92: I'm getting `error[E0658]: exclusive range pattern syntax is experimental` on my local `cargo run`. Our backend build jobs are green, though. So should I change a local setting or should our pipeline also fail on experimental features?

rnijveld: This was actually stabilized in Rust 1.80, so I think you probably need to update your Rust version (i.e. `rustup update` if installed that way.

rnijveld:
> > Minor point, but the hardcoded `CSB` value:
> > https://github.com/kiesraad/abacus/blob/3bfa7563bc3a5aec62dae3d8b475ed59c42a6358/backend/src/eml/mod.rs#L36
> > 
> > results in invalid ReportingUnitIdentifiers (`CSB::{\d+}`):
> > https://github.com/kiesraad/abacus/blob/3bfa7563bc3a5aec62dae3d8b475ed59c42a6358/backend/src/eml/mod.rs#L292
> > 
> > This is because the EML_NL standard expects reporting unit identifiers to follow the following pattern, so the polling station number should also include the prefix 'SB': `(HSB\d+)|((HSB\d+::)?\d{4})|(((HSB\d+::)?\d{4}::)?SB\d+)|(HSB\d+::SB\d+)`
> 
> Maybe we can replace `authority_id` with a hardcoded `0000` for now, then it should validate. This number can be replaced with the [CBS code](https://en.wikipedia.org/wiki/Municipalities_of_the_Netherlands) of the municipality (from the election definition EML_NL XML) in the future.

In ccb6dafade3c117e459732f2d59aa8866c7ac0f9 I've both replaced the placeholder authority id with `0000` for now, but also added the `SB` prefix as well. Looking at a couple of examples this is what was done previously as well, and I don't think we have any particular reason for doing things differently here.

jschuurk-kr:
[eml2sql](https://github.com/kiesraad/eml2sql) is not happy about our xml:
```
~/workspace/eml2sql (main %=) $ ./emlconv eml-510b.xml
'eml-510b.xml' is EML form 510b
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi
Aborted (core dumped)
```

I made one change to the generated xml, i.e. change `<TransactionId>` from `TODO` to `1`.

---

So I downloaded [Verkiezingsuitslagen Gemeenteraad 2022](https://data.overheid.nl/dataset/verkiezingsuitslagen-gemeenteraad-2022#downloads) and processed those files. That seems two work. `emlconv` exited without error after printing a lot of stuff to the terminal, including an occassional "Unknown item in Results: ContestIdentifier".

I then proceeded with the next step of the README and that failed:
```
~/workspace/eml2sql (main %=) $ sqlite3 eml.sqlite < useful-views
Parse error near line 111: no such column: shortcode
  CREATE INDEX candvtcidx on candvotecounts(shortcode);
                              error here ---^
Parse error near line 117: no such column: shortcode
  INDEX formidx on candvotecounts(electionId,formid,shortcode);
                                      error here ---^
```

---

I also followed the steps from the README, which uses the TK2023 election results. Both commands from the README `./emlconv tk2023` and `sqlite3 eml.sqlite < useful-views` work on that data set.

An important difference between the GR2022 and the TK2023 data sets is that they contain several kinds of files: election definition, candidate lists, counts. So I wonder if eml2sql needs a full election data set and not just a single xml file. On the other hand, even then I'd expect a different error message than `'std::invalid_argument'`.

backend/src/eml/mod.rs
jschuurk-kr: Thanks! Was still on 1.78. :grimacing:

rnijveld:
> [eml2sql](https://github.com/kiesraad/eml2sql) is not happy about our xml:
> 
> ```
> ~/workspace/eml2sql (main %=) $ ./emlconv eml-510b.xml
> 'eml-510b.xml' is EML form 510b
> terminate called after throwing an instance of 'std::invalid_argument'
>   what():  stoi
> Aborted (core dumped)
> ```
> 
> I made one change to the generated xml, i.e. change `<TransactionId>` from `TODO` to `1`.
I don't think that technically the TransactionId needs to be an integer according to the spec, but given that they will be integers anyway with all the election specifications I've seen, I've put in the placeholder value of `1` for now.

jschuurk-kr:
> I don't think that technically the TransactionId needs to be an integer according to the spec, but given that they will be integers anyway with all the election specifications I've seen, I've put in the placeholder value of `1` for now. After that change, I don't get any complaints at all when running `emlconv`

With the latest commit and the latest Rust version, `emlconv` is happy with the xml on my machine too.

jschuurk-kr (changes requested):
I compared our xml file with the "Telling_*" files of three municipalities of GR2022. Found some differences of very varying severity. So I've split them into three categories.

__To fix or to discuss__

Our `<Contests><Contest><TotalVotes>` is missing `<UncountedVotes>`.

Our xml is missing `<ds:CanonicalizationMethod Algorithm="http://www.w3.org/TR/2001/REC-xml-c14n-20010315#WithComments"></ds:CanonicalizationMethod>`.

Their xml has the postal code of the polling station in the `<ReportingUnitIdentifier>`: `<ReportingUnitIdentifier Id="0513::SB101">Stembureau Stembureau de Agnietenkapel (postcode: 2801 GR)</ReportingUnitIdentifier>`. We don't: `<ReportingUnitIdentifier Id="0000::SB1">Studio The Rules</ReportingUnitIdentifier>`.

__We have a TODO for it__

We still need to add the currently missing `<ElectionIdentifier><kr:ElectionDomain Id="0589">Oudewater</kr:ElectionDomain>`.

__To leave as is__

We have two additional namespaces: `xmlns:rg="http://www.kiesraad.nl/reportgenerator"` and `xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"`. We also have `Id="510b" SchemaVersion="5"` as the first two attributes. They have it as the last two.

We have `<EventIdentifier/>`. They have `<EventIdentifier></EventIdentifier>`.

Our xml has the optional `<Contests><Contest><ContestIdentifier><ContestName>`.

Theirs does not have `<?xml version="1.0" encoding="UTF-8"?>`.

jschuurk-kr:
With the differences I found in xml that's valid according to several tools, should we look into approval/snapshot testing with [insta](https://docs.rs/insta/latest/insta/)?

rnijveld:
> 
> Our `<Contests><Contest><TotalVotes>` is missing `<UncountedVotes>`.

Good one, definitely should have been there. Added.

> 
> Our xml is missing `<ds:CanonicalizationMethod Algorithm="http://www.w3.org/TR/2001/REC-xml-c14n-20010315#WithComments"></ds:CanonicalizationMethod>`.

This is only used when using XML Signatures, which we won't be using in Abacus.

> 
> Their xml has the postal code of the polling station in the `<ReportingUnitIdentifier>`: `<ReportingUnitIdentifier Id="0513::SB101">Stembureau Stembureau de Agnietenkapel (postcode: 2801 GR)</ReportingUnitIdentifier>`. We don't: `<ReportingUnitIdentifier Id="0000::SB1">Studio The Rules</ReportingUnitIdentifier>`.

Hmm, I've added the postcode as well, I don't think it really adds much, but it might help quickly identify a polling station when looking at the XML if your naming scheme for polling stations isn't great.

> 
> **We have a TODO for it**
> 
> We still need to add the currently missing `<ElectionIdentifier><kr:ElectionDomain Id="0589">Oudewater</kr:ElectionDomain>`.

Right now we don't have this information yet. But it should be added once we properly import the election from its EML_NL definition.

praseodym:
> > Their xml has the postal code of the polling station in the `<ReportingUnitIdentifier>`: `<ReportingUnitIdentifier Id="0513::SB101">Stembureau Stembureau de Agnietenkapel (postcode: 2801 GR)</ReportingUnitIdentifier>`. We don't: `<ReportingUnitIdentifier Id="0000::SB1">Studio The Rules</ReportingUnitIdentifier>`.
> 
> Hmm, I've added the postcode as well, I don't think it really adds much, but it might help quickly identify a polling station when looking at the XML if your naming scheme for polling stations isn't great.

This is also not part of the standard, but tools like eml2sql do rely on it:

https://github.com/kiesraad/eml2sql/blob/4f033ad5a709e08300a7d286e77d7d2ba708a0e3/emlconv.cc#L648-L663

@chrismostert maybe we should find a way to get this data in a proper XML element in this 510b EML?

jschuurk-kr:
> > **We have a TODO for it**
> > We still need to add the currently missing `<ElectionIdentifier><kr:ElectionDomain Id="0589">Oudewater</kr:ElectionDomain>`.
> 
> Right now we don't have this information yet. But it should be added once we properly import the election from its EML_NL definition.

Sorry, I should have been clearer. I meant it indeed as: it's good for this PR, because we have a TODO to add it once we have the information.

jschuurk-kr:
Tested <UncountedVotes> in <TotalVotes> and the postal code in <ReportingUnitIdentifier> and confirmed both issues have been resolved.

jschuurk-kr (approved):

chrismostert:
> > > Their xml has the postal code of the polling station in the `<ReportingUnitIdentifier>`: `<ReportingUnitIdentifier Id="0513::SB101">Stembureau Stembureau de Agnietenkapel (postcode: 2801 GR)</ReportingUnitIdentifier>`. We don't: `<ReportingUnitIdentifier Id="0000::SB1">Studio The Rules</ReportingUnitIdentifier>`.
> > 
> > 
> > Hmm, I've added the postcode as well, I don't think it really adds much, but it might help quickly identify a polling station when looking at the XML if your naming scheme for polling stations isn't great.
> 
> This is also not part of the standard, but tools like eml2sql do rely on it: https://github.com/kiesraad/eml2sql/blob/4f033ad5a709e08300a7d286e77d7d2ba708a0e3/emlconv.cc#L648-L663
> 
> @chrismostert maybe we should find a way to get this data in a proper XML element in this 510b EML?

Currently some data checks do rely on the zip-code being present in the name of the polling station since there is no other field to put it in the `510[bcd]` EML-files, so it's good to see that (for now) it is added in Abacus as well.

I'll suggest adding the zip code as a separate attribute in the EML_NL standard as well :+1:

backend/src/election/mod.rs
```diff
@@ -148,3 +150,38 @@ pub async fn election_download_results(
...
+    get,
+    path = "/api/elections/{election_id}/download_xml_results",
+    responses(
+        (
+            status = 200,
+            description = "XML",
+            content_type="text/xml",
+            headers(
+                ("Content-Type", description = "text/xml"),
+            )
```
praseodym, line 165: nitpick: does the content type need to be specified twice?

praseodym (approved):

backend/src/election/mod.rs
rnijveld: For now I think we can just keep this as is. These two occurrences do end up on two different places in the openapi json file.

PR #689, opened 2024-12-04
Delete polling station

oliver3:
Added a "Stembureau verwijderen" button on the polling station update page that deletes the polling station after a confirmation modal has been presented.
- Deletion should go back to the list page and show a message that the polling station was deleted
- Deletion should not be possible when there is already data entry for that polling station. Instead an alert will be shown on the update page. See also the note below about showing/hiding the button.
- Deletion should be cancelled if the modal was closed with either the cross or the "Annuleren" button.

In the Figma design, there is the requirement to not show the button when data entry has started. For now, because we do not have a "data entry started" state, we will only prevent deleting a polling station after the fact. Once we will implement 'data entry started'. We can show/hide this button accordingly.

@jorisleker : is it okay to have the 'deletion failed' situation covered like this for now? And are the texts of the alert and modal correct?


jorisleker (approved):

frontend/app/module/polling_stations/page/PollingStationUpdatePage.tsx
```diff
@@ -21,6 +37,23 @@ export function PollingStationUpdatePage() {
...
+
+  React.useEffect(() => {
+    if (error) {
+      window.scrollTo(0, 0);
+    }
+  }, [error]);
+
   return (
     <>
       <PageTitle title="Stembureaus - Abacus" />
```
Lionqueen94, line 59: Please add translation here too

frontend/app/module/polling_stations/page/PollingStationUpdatePage.tsx
```diff
@@ -29,17 +62,44 @@ export function PollingStationUpdatePage() {
           <h1>Stembureau wijzigen</h1>
```
Lionqueen94, line 62: Please add translation here too

frontend/app/module/polling_stations/page/PollingStationUpdatePage.tsx
```diff
@@ -1,9 +1,14 @@
+import * as React from "react";
+import { useState } from "react";
```
praseodym, line 2: This import is already covered by the one before it

praseodym (approved):

jorisleker (changes requested):
Looks like the background-color of the primary destructive button is a bit off. Should be Rood/600

praseodym (dismissed):

jorisleker (dismissed):

praseodym (approved):

frontend/lib/i18n/locales/nl/polling_station.json
```diff
@@ -8,8 +8,11 @@
   },
   "message": {
     "no_polling_stations": "Er zijn nog geen stembureaus ingevoerd voor deze verkiezing. Kies hoe je stembureaus gaat toevoegen.",
-    "polling_station_updated": "Wijzigingen stembureau {number} ({name}) opgeslagen.",
-    "polling_station_created": "Stembureau {number} ({name}) toegevoegd."
+    "polling_station_updated": "Wijzigingen stembureau {number} ({name}) opgeslagen",
+    "polling_station_created": "Stembureau {number} ({name}) toegevoegd",
+    "polling_station_deleted": "Stembureau {pollingStation} verwijderd",
```
Lionqueen94, line 13: Is pollingStation here a number like in the sentences above or something else?

oliver3: It is the `{number} ({name})` combined, because for deleted, the polling station cannot be looked up in the list of polling stations anymore so the whole string is supplied to show in the message.

jorisleker (approved):
Tested:
- destructive button background looks good now
- deleting polling station without input works
- deleting polling station with input is blocked

PR #692, opened 2024-12-05
Add XML hashcode to PDF

rnijveld:
- ~~Note: #685 should probably be merged first~~
- Adds XML Hashcode to the PDF
- Changes indentation of XML to 2 spaces
- ~~Example output: [model-na-31-2.pdf](https://github.com/user-attachments/files/18021786/model-na-31-2.pdf)~~
- Example output: [model-na-31-2-22.pdf](https://github.com/user-attachments/files/18024332/model-na-31-2-22.pdf)

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -37,6 +37,8 @@ tracing = "0.1.41"
 typst = "0.12.0"
 typst-pdf = "0.12.0"
 quick-xml = { version = "0.37.1", features = ["serialize"] }
+sha2 = "0.10.8"
+const-hex = "1.14.0"
 
 [dev-dependencies]
 reqwest = { version = "0.12", default-features = false, features = ["json"] }
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -38,7 +38,7 @@ typst = "0.12.0"
 typst-pdf = "0.12.0"
 quick-xml = { version = "0.37.1", features = ["serialize"] }
 sha2 = "0.10.8"
-const-hex = "1.14.0"
+hex = "0.4.3"
 
 [dev-dependencies]
 reqwest = { version = "0.12", default-features = false, features = ["json"] }
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -38,7 +38,6 @@ typst = "0.12.0"
 typst-pdf = "0.12.0"
 quick-xml = { version = "0.37.1", features = ["serialize"] }
 sha2 = "0.10.8"
-hex = "0.4.3"
 
 [dev-dependencies]
 reqwest = { version = "0.12", default-features = false, features = ["json"] }
--- a/backend/src/eml/base.rs
+++ b/backend/src/eml/base.rs
@@ -88,26 +88,18 @@ pub trait EMLDocument: Sized + DeserializeOwned + Serialize {
 
 pub fn eml_document_hash(input: &str, chunked: bool) -> String {
     use sha2::Digest;
-    let mut bytes = vec![0u8; sha2::Sha256::output_size() * 2];
-    let digest = sha2::Sha256::digest(input.as_bytes());
+    let digest = sha2::Sha256::digest(input.as_bytes())
+        .into_iter()
+        .map(|b| format!("{:02x}", b));
 
-    // This can never fail unless `Sha256::output_size()` would give a wrong value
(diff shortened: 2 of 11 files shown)


backend/src/eml/base.rs
```diff
@@ -79,13 +79,38 @@ pub trait EMLDocument: Sized + DeserializeOwned + Serialize {
...
                 .expect("Constructing a new XML Serializer cannot fail");
-            s.indent(' ', 4);
+            s.indent(' ', 2);
             s
         };
         Serialize::serialize(&self, serializer)
     }
 }
 
+pub fn eml_document_hash(input: &str, chunked: bool) -> String {
```
praseodym, line 89: I think it's good to test this function, e.g.
```rust
    #[test]
    fn test_eml_document_hash() {
        let input = "test";
        let hash = eml_document_hash(input, false);
        assert_eq!(
            hash,
            "9F86D081884C7D659A2FEAA0C55AD015A3BF4F1B2B0B822CD15D6C15B0F00A08"
        );
        let hash = eml_document_hash(input, true);
        assert_eq!(
            hash,
            "9F86 D081 884C 7D65 9A2F EAA0 C55A D015 A3BF 4F1B 2B0B 822C D15D 6C15 B0F0 0A08"
        );
    }
```

backend/Cargo.toml
```diff
@@ -37,6 +37,8 @@ tracing = "0.1.41"
 typst = "0.12.0"
 typst-pdf = "0.12.0"
 quick-xml = { version = "0.37.1", features = ["serialize"] }
+sha2 = "0.10.8"
+const-hex = "1.14.0"
```
praseodym, line 41: Can you elaborate why you chose `const-hex`? It looks like it relies on some [unsafe Rust](https://github.com/search?q=repo%3ADaniPopes%2Fconst-hex%20unsafe&type=code) for performance, but we don't need it to be fast. 

The [hex crate](https://crates.io/crates/hex) is already in our dependency tree and it doesn't look like that one uses unsafe Rust.

backend/src/eml/base.rs
rnijveld: I copied your suggestion into a1ab3de98bb3c9e900a1e9af1e0ed65bb74a4f33 :)

backend/Cargo.toml
rnijveld: My main reason for not using hex is that it is no longer maintained and appears to be as good as abandoned. Alternatively there is also [faster-hex](https://crates.io/crates/faster-hex), but that one also has some unsafe usages.

rnijveld: Hmm I just noticed some new commits on the repository, maybe it will come back to life again?

rnijveld: As discussed on mattermost: switched to `hex` in cf5ada3fd988d9e4a4b5e0e57d9cb0f316752845

backend/templates/inputs/model-na-31-2.json
```diff
@@ -640,6 +640,6 @@
       "locality": "Heemdamseburg"
     }
   ],
-  "hash": "ED36 60EB 017A 0D3A D3EF 72B1 6865 F991 A36A 9F92 72D9 1516 39CD 422B 4756 D161",
-  "creation_date_time": "04-12-2024 12:08:55"
+  "hash": "ed36 60eb 017a 0d3a d3ef 72b1 6865 f991 a36a 9f92 72d9 1516 39cd 422b 4756 d161",
+  "creation_date_time": "04-12-2024 12:08"
```
praseodym, line 644: Why the change to lowercase and no seconds in the timestamp?

jorisleker: Lowercase on my request for better readability (and to save some horizontal space)

rnijveld: With seconds in the timestamp the footer format is very cramped, especially as the number of pages goes over 100. The full timestamp (up to the nanosecond if we want) is always in the EML anyway.

praseodym (approved):

marlonbaeten (dismissed):
Maybe we can remove the hex dependency since Rust already has format methods for this - see for instance: https://github.com/tweedegolf/etes/blob/e5b583ac122d0d2221c3e0aa31a1ba9b49b1f974/src/util.rs#L37

praseodym (dismissed):

praseodym (approved):

oliver3 (approved):

Issue #675, opened 2024-12-03
Fix table overflow


oliver3:
When the polling station name is very long, the table rendering is off, e.g. polling station type gets wrapped

praseodym:
We can also consider adding a limit to the length of the polling station name.

PR #693, opened 2024-12-05
Ladle vite config should convert CSS module dashes into camelCase too  

oliver3:
In Ladle, the dashes to camelCase was not working anymore because of a separate vite config that did not include the `localsConventions: dashes`, which resulted in e.g. the Table Link rows to not have the correct class anymore with on hover styling and a chevron.

Related PR's
- Implemented the possibility to use dashes in CSS modules and camelCase in components in https://github.com/kiesraad/abacus/pull/570
- Implemented a separate vite config for Ladle to make build and deploy possible in https://github.com/kiesraad/abacus/pull/655

PR #694, opened 2024-12-05
Replace several Table props with className

oliver3:
- [x] ~~https://github.com/kiesraad/abacus/pull/693 has to be merged beforehand~~

This refactor also fixes https://github.com/kiesraad/abacus/issues/675

Replace all specific Table props with className, to be able to add extra styling to the table component, for instance font size, header background, and word-break.

All tables, on the following pages, should still have the same styling:
- `elections` (OverviewPage)
- `elections/1/status#coordinator` (ElectionStatusPage)
- `elections/1/polling-stations#coordinator` (PollingStationListPage)
- `elections/1/data-entry` "Kies het stembureau" (PollingStationList)

The word-wrap should now be fixed for the PollingStationListPage:

image


frontend/lib/ui/Table/Table.module.css
```diff
@@ -60,3 +53,10 @@
     }
   }
 }
+
+:global(.table-number) {
+  text-align: right;
+  font-size: var(--font-size-body);
+  font-weight: bold;
+  width: 6.5rem;
+}
```
Lionqueen94, line 62:
```suggestion
:global(.table-number) {
  text-align: right;
  font-size: var(--font-size-body);
  font-weight: bold;
  width: 6.5rem;
  line-height: 1.75rem;
}
```

Lionqueen94 (approved):

praseodym (dismissed):

oliver3:
It seemed that the "Analyze JavaScript/TypeScript" check was stuck on "Waiting for status to be reported"..

frontend/lib/ui/Table/Table.tsx
```diff
@@ -61,14 +57,6 @@ function LinkRow({ children, to }: { children: React.ReactNode[]; to: To }) {
...
-  number,
-  fontSizeClass,
-}: {
-  children?: React.ReactNode;
-  number?: boolean;
-  fontSizeClass: string;
-}) {
-  return <td className={`${number ? `${cls.number} ` : ""}${fontSizeClass}`}>{children}</td>;
+function Cell({ children, className }: { children?: React.ReactNode; className?: string }) {
+  return <td className={className}>{children}</td>;
```
lkleuver, line 61: With this change, doesnt it make more sense to just use `<td>` instead of `<Table.Cell>` if not using defined properties. 
Cell props could also extend React.HTMLAttributes perhaps? className would be included.

My main problem with this is how do you know what classes are available?

frontend/lib/ui/Table/Table.stories.tsx
```diff
@@ -12,24 +12,51 @@ const data: [number, string, string][] = [
...
+
+    <div className="mb-lg">
+      <Table id="styled_table">
+        <Table.Header className="bg-gray">
+          <Table.Column>Nummer</Table.Column>
+          <Table.Column>Stembureau</Table.Column>
+        </Table.Header>
+        <Table.Body className="fs-sm">
+          <Table.Row>
+            <Table.Cell className="table-number">33</Table.Cell>
```
lkleuver, line 43: personal preference:
`<Table.NumberCell>` or `<Table.Cell isNumber>`

lkleuver (approved):
only added some personal opinions we could discuss later
lgtm.

praseodym:
> It seemed that the "Analyze JavaScript/TypeScript" check was stuck on "Waiting for status to be reported"..

"Waiting for status to be reported" means that the check did not run, which is because we only run that workflow for PRs to `main`. And when the PR base gets changed, it does not automatically trigger the workflow to run.

marlonbaeten (commented):
I think passing the classes directly might not be the nicest solution - what if you make a typo in the class name for instance? You could convert a set of properties to a list of classes using something like [`classnames`](https://www.npmjs.com/package/classnames) - we already have this as a dependency. Maybe you could use classnames in `Table.tsx` to covert properties (like flags or enums) to class names.

frontend/lib/ui/Table/Table.stories.tsx
oliver3: I went too far indeed with the classes, the `<Table.NumberCell>` is really a different kind of cell with a lot of other styling.

oliver3:
> I think passing the classes directly might not be the nicest solution ...

I went too far indeed and introduced the `<Table.NumberCell>` for a cell variant specific for e.g. polling station numbers. 

I think that the rest of the classNames together with our utility classes are an 'agile' way to handle the one-off design differences that we have at the moment, instead of creating props for each case.

marlonbaeten (approved):
I think for now this is fine. At a certain point we want to move away from classes en use properties.

marlonbaeten (approved):

cikzh (approved):

PR #695, opened 2024-12-05
Fixed a few tiny issues I saw when catching up with the work done in the past weeks

Lionqueen94:


frontend/lib/ui/NumberInput/NumberInput.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import type { Story } from "@ladle/react";
 
-import { NumberInput } from "./NumberInput";
+import { NumberInput } from "@kiesraad/ui";
```
praseodym, line 3: Here we import from `@kiesraad/ui` from within that same module..

frontend/lib/ui/Toolbar/Toolbar.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import { Story } from "@ladle/react";
 
 import { IconPlus } from "@kiesraad/icon";
+import { Button } from "@kiesraad/ui";
 
-import { Button } from "../Button/Button";
 import { Toolbar, ToolbarSection } from "./Toolbar";
```
praseodym, line 6: But here we use a relative import within the module.

What do we prefer? I have a small preference for keeping imports relative within the same module.

Lionqueen94: Jetbrains always recommends shortening it to importing from `"@kiesraad/ui"`. Is there any way we can lint this so it's consistent everywhere? I personally like the shorter imports instead of relative.

jorisleker (approved):

frontend/lib/ui/Toolbar/Toolbar.stories.tsx
praseodym: I'll do some research, for now we can leave this as is.

praseodym (approved):

Issue #624, opened 2024-11-20
Block down-arrow from looping back to top of form


jorisleker:
We allow users to use the down arrow to navigate through data entry forms. The current implementation focusses the first input of the form section when the down arrow is used while focus is on the last element on the page. This is confusing and time confusing. Users often use the down arrow in a repetitive manner to get to the bottom of the form, and overshoot the 'volgende' button. They are unaware they can navigate back to the bottom of the form by hitting the up-arrow

Requested change:
- don't "loop" through the form with up- and down-arrow:
  - when hitting down-arrow on last field, don't move focus to first field
  - when pressing up-arrow on first field, don't move focus to last field

PR #696, opened 2024-12-05
Separate useKeyboard from Form component

oliver3:
- Extracted all keyboard logic from `Form` to `useKeyboard()`, to be able to use Form outside of the data entry screens.
- Added separate story for form with the useKeyboard : /ladle/?story=form--use-keyboard
- Separated tests and added few tests to try and cover all keyboard interaction
- Resolved https://github.com/kiesraad/abacus/issues/624 
- Added useKeyboard() to data entry screens

It might be easier to code-review per commit

Data entry screens should still have keyboard navigation.
Polling station form should not have custom keyboard navigation anymore.

**Questions**
The implementation before fixing 624 had the behaviour that if there was no input or submit button selected, moving down selected the first input. Was this intended and do we want to have this behaviour back?

Keyboard usage with radio buttons now has different strange behaviour than before. Not sure if this is worse or better, but what do we expect that the behaviour is?


praseodym:
Now that you're refactoring this code, maybe fix #624 too?

frontend/lib/ui/Form/useKeyboard.ts
```diff
@@ -0,0 +1,98 @@
+import * as React from "react";
+
+type Dir = "up" | "down" | "first" | "last";
+
+export function useKeyboard(innerRef: React.MutableRefObject<HTMLFormElement | null>) {
```
lkleuver, line 5: I dont expect a hook named "useKeyboard" to only accept a specific "HTMLFormElement" ref.
(from the imports perspective you dont get the context of "form" eg. @kiesraad/ui)

This seems to be very specific form keyboard code.

lkleuver (approved):

frontend/lib/ui/Form/useKeyboard.ts
oliver3: Absolutely, I've renamed it to `useFormKeyboardNavigation()` to better reflect what it does in https://github.com/kiesraad/abacus/pull/696/commits/cd14c54715f44f73ed93e8d43c76d4975b50cc5d

oliver3:
>     * One thing I would like to see changed: If the 'accept warnings' checkbox is visible, using the down arrow from the last input, should focus the checkbox. And hitting the up-arrow when focus is on the submit button should do the same. Right now, it is not possible to reach the checkbox using the arrows.

This was a caching bug that was introduced in this PR. I've removed the caching because it is not needed performance wise.

praseodym (approved):

lkleuver (approved):

jschuurk-kr:
@jorisleker @oliver3 If there's a warning on "Toegelaten kiezers en uitgebrachte stemmen" and I do shift+arrow down, I end up on the accept warnings checkbox. Is that how it should be?

jorisleker:
> @jorisleker @oliver3 If there's a warning on "Toegelaten kiezers en uitgebrachte stemmen" and I do shift+arrow down, I end up on the accept warnings checkbox. Is that how it should be?

Yes, that's ok!

jorisleker (approved):

jschuurk-kr:
> > @jorisleker @oliver3 If there's a warning on "Toegelaten kiezers en uitgebrachte stemmen" and I do shift+arrow down, I end up on the accept warnings checkbox. Is that how it should be?
> 
> Yes, that's ok!

Good! We don't have a test for that scenario. I'll add it, I don't think @oliver3 will mind.

jschuurk-kr:
> > > @jorisleker @oliver3 If there's a warning on "Toegelaten kiezers en uitgebrachte stemmen" and I do shift+arrow down, I end up on the accept warnings checkbox. Is that how it should be?
> > 
> > 
> > Yes, that's ok!
> 
> Good! We don't have a test for that scenario. I'll add it, I don't think @oliver3 will mind.

With three approvals on this PR, let me do that in a separate PR. :-D

PR #701, opened 2024-12-06
Implement polling station breadcrumb navigation and changed polling station update path ordering

### LINKED ISSUE #651: Implement polling station breadcrumb navigation
The breadcrumb for the add/edit page misses the 'stembureaus' node (see [design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=5442-50751&node-type=frame&t=Lm70TAtSuFNoppcO-11))

Lionqueen94:
Closes #651 

Changed polling station update path ordering


praseodym (approved):
Maybe we can look into creating a new component for the `<NavBar>` breadcrumb to avoid duplicating code between e.g. `PollingStationCreatePage.tsx`, `PollingStationListPage.tsx` and `PollingStationUpdatePage.tsx`?

Lionqueen94:
> Maybe we can look into creating a new component for the `<NavBar>` breadcrumb to avoid duplicating code between e.g. `PollingStationCreatePage.tsx`, `PollingStationListPage.tsx` and `PollingStationUpdatePage.tsx`?

What would help is the option to extend the current content in the NavBar, then you can add the 'base' in the layout and add an extra crumb in detail pages for example. Something to think about for a future refactor I guess.

praseodym:
> > Maybe we can look into creating a new component for the `<NavBar>` breadcrumb to avoid duplicating code between e.g. `PollingStationCreatePage.tsx`, `PollingStationListPage.tsx` and `PollingStationUpdatePage.tsx`?
> 
> What would help is the option to extend the current content in the NavBar, then you can add the 'base' in the layout and add an extra crumb in detail pages for example. Something to think about for a future refactor I guess.

Extending it from within child component is not trivial. But we could make the `NavBar` component create the whole breadcrumb from the current path, so that we don't have to define or extend it on every page.

jorisleker (approved):
I am still missing the [hamburger menu](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=1295-62561&node-type=frame&t=lW96C56PqFLLp3jy-11) before the breadcrumb nav, but guess that's out of scope for this ticket. Do we have it on the backlog to implement this as well?

Lionqueen94:
> I am still missing the [hamburger menu](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=1295-62561&node-type=frame&t=lW96C56PqFLLp3jy-11) before the breadcrumb nav, but guess that's out of scope for this ticket. Do we have it on the backlog to implement this as well?

There is no issue for this yet, can you create one with links to the specific design for each user role?

Issue #684, opened 2024-12-03
Polling station address: Align backend with design


lkleuver:
Align fields in the database with EML design and frontend.

- address
- house_number
- house_number_addition

The backend currently still has separate fields that should be condensed into a single field.

jorisleker:
Not wanting to open this discussion again, but my assumption was that the EML_NL standard only had an address line and not separate fields. Also: do we know the format of the polling station files municipalities currently use to import their data into OSV2020? (those files are created by software used by the Burgerzaken department. 

If those files don't have housenumber and housenumber addition in separate fields, we will run into issues importing that data. Please remember that for most municipalities this is the primary way polling stations will be added to Abacus. 

@praseodym your thoughts

praseodym:
> (might just be confused by the description of this issue btw, as it isn't really clear on what needs to happen. If aligning backend with design means consolidating `address`, `house_number` and `house_number_addition` into a single address-field field, then please disregard what I said before).

That's what this issue is about, the backend currently still has separate fields that should be condensed into a single field.

PR #670, opened 2024-12-02
API client improvements

marlonbaeten:
See #645 

PR #630 should first be merged!

Fixes bug causing API error being thrown from promise and therefore not caught by the router error boundary.

+ some minor improvements in the API clients and error handling.

PR #702, opened 2024-12-06
Polling station create/update validation

### LINKED ISSUE #638: Validation in create/update form
- Frontend validates and backend enforces:
  - required fields (backend should check empty string too): number, name
  - number format: number, number_of_voters

- Backend validates and frontend shows validation error
  - unique number
  - election not in state "DataEntryFinished"

- Show succes message on list only when successful

lkleuver:
This PR adds validation to the create/update pollingstation form.
Adding clientside validation and rendering backend errors.

Added generic form validation hook and types in ui/form/validation.
Adding just whitespace to a mandatory field is not considered valid.

closes #638 

Fixed in separate issue #706: ChoiceList has no UI for rendering errors.


frontend/app/component/form/polling_station/PollingStationForm.test.tsx
```diff
@@ -34,32 +34,121 @@ async function fillForm(user: UserEvent, testPollingStation: PollingStation | Om
...
+
+      await user.clear(numberInput);
+      await user.type(numberInput, testObj.number.toString());
+
+      await userEvent.click(screen.getByRole("button", { name: "Opslaan en toevoegen" }));
+
+      await waitFor(() => {
+        expect(onSaved).toHaveBeenCalledWith(expect.objectContaining(testObj));
+      });
     });
```
praseodym, line 149:
```suggestion
    test("Validation", async () => {
      const testObj: Omit<PollingStation, "id"> = {
        election_id: 1,
        number: 42,
        name: "test",
        street: "test",
        postal_code: "1234",
        locality: "test",
        polling_station_type: "FixedLocation",
        number_of_voters: 1,
        house_number: "test",
      };

      const onSaved = vi.fn();
      render(<PollingStationForm electionId={1} onSaved={onSaved} />);

      const user = userEvent.setup();
      await user.click(screen.getByRole("button", { name: "Opslaan en toevoegen" }));

      const numberInput = screen.getByRole("textbox", { name: "Nummer" });
      const nameInput = screen.getByRole("textbox", { name: "Naam" });
      const numberOfVotersInput = screen.getByRole("textbox", { name: "Aantal kiesgerechtigden Optioneel" });
      const streetInput = screen.getByRole("textbox", { name: "Straatnaam" });
      const houseNumberInput = screen.getByRole("textbox", { name: "Huisnummer" });
      const houseNumberAdditionInput = screen.getByRole("textbox", { name: "Toevoeging" });
      const postalCodeInput = screen.getByRole("textbox", { name: "Postcode" });
      const localityInput = screen.getByRole("textbox", { name: "Plaats" });
      const typeInputOption = screen.getByRole("radio", { name: "Vaste locatie" });

      await user.click(typeInputOption);

      await waitFor(() => {
        expect(numberInput).toHaveAttribute("aria-invalid", "true");
        expect(nameInput).toHaveAttribute("aria-invalid", "true");
        expect(streetInput).toHaveAttribute("aria-invalid", "true");
        expect(houseNumberInput).toHaveAttribute("aria-invalid", "true");
        expect(postalCodeInput).toHaveAttribute("aria-invalid", "true");
        expect(localityInput).toHaveAttribute("aria-invalid", "true");
        expect(numberOfVotersInput).toHaveAttribute("aria-invalid", "false");
        expect(houseNumberAdditionInput).toHaveAttribute("aria-invalid", "false");
      });

      await user.type(nameInput, testObj.name);
      await user.type(streetInput, testObj.street);
      await user.type(houseNumberInput, testObj.house_number);
      await user.type(postalCodeInput, testObj.postal_code);
      await user.type(localityInput, testObj.locality);
      await user.type(numberOfVotersInput, testObj.number_of_voters?.toString() || "");

      await user.click(screen.getByRole("button", { name: "Opslaan en toevoegen" }));

      await waitFor(() => {
        expect(numberInput).toHaveAttribute("aria-invalid", "true");
        expect(nameInput).toHaveAttribute("aria-invalid", "false");
        expect(streetInput).toHaveAttribute("aria-invalid", "false");
        expect(houseNumberInput).toHaveAttribute("aria-invalid", "false");
        expect(postalCodeInput).toHaveAttribute("aria-invalid", "false");
        expect(localityInput).toHaveAttribute("aria-invalid", "false");
        expect(numberOfVotersInput).toHaveAttribute("aria-invalid", "false");
        expect(houseNumberAdditionInput).toHaveAttribute("aria-invalid", "false");
      });

      overrideOnce("post", `/api/elections/1/polling_stations`, 409, {
        error: "Polling station already exists",
        fatal: false,
        reference: "EntryNotUnique",
      } satisfies ErrorResponse);

      await user.type(numberInput, "1");
      await user.click(screen.getByRole("button", { name: "Opslaan en toevoegen" }));

      await waitFor(() => {
        expect(screen.getByRole("alert")).toHaveTextContent(
          ["Er bestaat al een stembureau met nummer 1.", "Het nummer van het stembureau moet uniek zijn."].join(""),
        );
      });

      await user.clear(numberInput);
      await user.type(numberInput, testObj.number.toString());

      await user.click(screen.getByRole("button", { name: "Opslaan en toevoegen" }));

      await waitFor(() => {
        expect(onSaved).toHaveBeenCalledWith(expect.objectContaining(testObj));
      });
    });
```

jorisleker:
Scherm­afbeelding 2024-12-09 om 13 40 31

I assume this is default browser behavior. Can image this is the easiest way to implement this validation, but wanted to flag that it's [not according to design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=832-34493&node-type=frame&t=lW96C56PqFLLp3jy-11). 

Realized we don't have copy for this message yet, can we update the label to 'Dit is geen getal. Voer een getal in'

For the other validation messages:
- #638 only mentions Number + name as required fields. Right now Street, housenumber, postal code and city are also requited. Bit more then needed?
- The error-icon should be red, not black
- When a required field is left empty, the validation message 'Dit is een verplicht veld' should be shown below the field.
- Choice: when I put a space (` `) in the Name field, it is considered not empty and passes validation. I think we should trim the user input before checking it is empty. 

Scherm­afbeelding 2024-12-09 om 13 45 50

jorisleker (changes requested):
See previous comment

marlonbaeten (approved):

marlonbaeten (commented):
It would be nice to move some of the logic in `PollingStationForm.tsx` to a separate hook, which makes the component more readable.

There are some changes in #670 that also touch this code - I will update that PR if this one gets merged first.

lkleuver:
> It would be nice to move some of the logic in `PollingStationForm.tsx` to a separate hook, which makes the component more readable.
> 
> There are some changes in #670 that also touch this code - I will update that PR if this one gets merged first.

I wanted to make sure this is the way we want to go with form validation, a hook are util is definitely the way to go though instead of "inline"

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
```diff
@@ -34,32 +34,122 @@ async function fillForm(user: UserEvent, testPollingStation: PollingStation | Om
...
+      await userEvent.click(screen.getByRole("button", { name: "Opslaan en toevoegen" }));
+
+      await waitFor(() => {
+        expect(onSaved).toHaveBeenCalledWith(expect.objectContaining(testPollingStation));
+      });
+    });
 
-    await waitFor(() => {
-      expect(onSaved).toHaveBeenCalledWith(expect.objectContaining(testPollingStation));
+    test("Validation", async () => {
```
jschuurk-kr, line 65: This test needs to be split up in three separate tests.

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
```diff
@@ -34,32 +34,122 @@ async function fillForm(user: UserEvent, testPollingStation: PollingStation | Om
...
+      const houseNumberInput = screen.getByRole("textbox", { name: "Huisnummer" });
+      const houseNumberAdditionInput = screen.getByRole("textbox", { name: "Toevoeging" });
+      const postalCodeInput = screen.getByRole("textbox", { name: "Postcode" });
+      const localityInput = screen.getByRole("textbox", { name: "Plaats" });
+      const typeInputOption = screen.getByRole("radio", { name: "Vaste locatie" });
+
+      await user.click(typeInputOption);
+
+      await waitFor(() => {
+        expect(numberInput).toHaveAttribute("aria-invalid", "true");
```
jschuurk-kr, line 97: There's [`toBeInvalid()`](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tobeinvalid) for this.

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -73,10 +114,21 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
...
+                error={numberError}
+                hideErrorMessage={true}
+              />
+              <InputField
+                id="name"
+                name="name"
+                label={t("name")}
+                defaultValue={pollingStation?.name}
+                error={validationResult.name ? t(`form.errors.${validationResult.name}`) : undefined}
+                hideErrorMessage={validationResult.name === "FORM_VALIDATION_RESULT_REQUIRED"}
```
oliver3, line 126: Why hide the required message?

frontend/lib/ui/InputField/InputField.tsx
```diff
@@ -69,7 +71,7 @@ export function InputField({
         </span>
         {disabled ? <div className={`${fieldSize} disabled_input`}>{value}</div> : inputEl}
       </label>
-      {(error || hint) && (
+      {!hideErrorMessage && (error || hint) && (
```
oliver3, line 74: I would prefer `showErrorMessage` instead of this double negation (easier to reason about)

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -170,3 +239,77 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
...
+};
+
+type FieldRuleString = FieldRuleBase & {
+  type: "string";
+  minLength?: number;
+  maxLength?: number;
+};
+
+type FieldRuleNumber = FieldRuleBase & {
+  type: "number";
```
oliver3, line 264: This is actually about integer right?

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -170,3 +239,77 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
...
+            result[prop] = "FORM_VALIDATION_RESULT_MAX_LENGTH";
+          }
+        } else {
+          result[prop] = "FORM_VALIDATION_RESULT_INVALID_TYPE";
+        }
+        break;
+      case "number":
+        if (typeof inputValue === "number") {
+          if (rule.min && inputValue < rule.min) {
+            result[prop] = "FORM_VALIDATION_RESULT_MIN_LENGTH";
```
oliver3, line 301:
```suggestion
            result[prop] = "FORM_VALIDATION_RESULT_MIN";
```
etc

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -170,3 +239,77 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
...
+type AnyFieldRule = FieldRuleString | FieldRuleNumber;
+type ValidationRules<T> = Record<keyof T, AnyFieldRule>;
+type ValidationResult<T> = Record<keyof T, ValidationError | undefined>;
+
+function validate<T>(rules: ValidationRules<T>, obj: T): [boolean, ValidationResult<T>] {
+  const result: ValidationResult<T> = {} as ValidationResult<T>;
+  Object.entries(rules).forEach(([key, value]) => {
+    const prop = key as keyof T;
+    const inputValue = obj[prop];
+    const rule = value as AnyFieldRule;
```
oliver3, line 278: `Object.entries().forEach()` unfortunately does not play very nicely with TypeScript, you could also do
```suggestion
  for (const field in rules) {
    const inputValue = obj[field];
    const rule = rules[field];
```

and then use `continue` instead of `return`

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
```diff
@@ -34,32 +34,116 @@ async function fillForm(user: UserEvent, testPollingStation: PollingStation | Om
...
+        expect(nameInput).toHaveAttribute("aria-invalid", "false");
+        expect(streetInput).toHaveAttribute("aria-invalid", "false");
+        expect(houseNumberInput).toHaveAttribute("aria-invalid", "false");
+        expect(postalCodeInput).toHaveAttribute("aria-invalid", "false");
+        expect(localityInput).toHaveAttribute("aria-invalid", "false");
+        expect(numberOfVotersInput).toHaveAttribute("aria-invalid", "false");
+        expect(houseNumberAdditionInput).toHaveAttribute("aria-invalid", "false");
```
jschuurk-kr, line 118: There's [`toBeInvalid()`](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tobeinvalid) for this.

jorisleker (changes requested):
Almost there, but still some small issues:

- icon in the fields is still black, should be red
- when I leave 'aantal kiesgerechtigden' empty (which is allowed), I get an error 'niet het juiste type'. Expected: no error 
Scherm­afbeelding 2024-12-13 om 10 07 00
- If I enter a non-numeric value into the 'nummer' field, I get a default browser validation message, on top of our validation message.
Scherm­afbeelding 2024-12-13 om 10 08 39

- "Niet het juiste type" --> Can we change this to: 'Dit is geen getal. Voer een getal in'

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -78,10 +95,20 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
...
+                name="name"
+                label={t("name")}
+                defaultValue={pollingStation?.name}
+                error={validationResult.name ? t(`form.errors.${validationResult.name}`) : undefined}
               />
-              <InputField id="name" name="name" label={t("name")} defaultValue={pollingStation?.name} />
             </FormLayout.Row>
 
+            {/* TODO: Choicelist (required) error handling */}
+            {/* error={validationResult.polling_station_type ? t(`form.errors.${validationResult.polling_station_type}`) : undefined} */}
```
praseodym, line 111: #706 is resolved so this should be fixed in this PR too.

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
```diff
@@ -62,85 +81,82 @@ describe("PollingStationForm", () => {
...
-      const streetInput = screen.getByRole("textbox", { name: "Straatnaam" });
-      const houseNumberInput = screen.getByRole("textbox", { name: "Huisnummer" });
-      const houseNumberAdditionInput = screen.getByRole("textbox", { name: "Toevoeging" });
-      const postalCodeInput = screen.getByRole("textbox", { name: "Postcode" });
-      const localityInput = screen.getByRole("textbox", { name: "Plaats" });
-      const typeInputOption = screen.getByRole("radio", { name: "Vaste locatie" });
-
-      await user.click(typeInputOption);
-
+      // test choice list wait for merge.
```
jschuurk-kr, line 92: I don't understand this comment.

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
```diff
@@ -7,29 +7,48 @@ import { ErrorResponse, PollingStation } from "@kiesraad/api";
...
+
+  switch (testPollingStation.polling_station_type) {
+    case "FixedLocation":
+      await userEvent.click(inputs.typeOptionFixedLocation);
+      break;
+    case "Special":
+      await userEvent.click(inputs.typeOptionSpecial);
+      break;
+    case "Mobile":
+      await userEvent.click(inputs.typeOptionMobile);
```
jschuurk-kr, line 49: What's the reason for using `userEvent` here instead of `user`?

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
```diff
@@ -7,29 +7,48 @@ import { ErrorResponse, PollingStation } from "@kiesraad/api";
...
+    locality: screen.getByRole("textbox", { name: "Plaats" }),
+    typeOptionFixedLocation: screen.getByRole("radio", { name: "Vaste locatie" }),
+    typeOptionSpecial: screen.getByRole("radio", { name: "Bijzonder" }),
+    typeOptionMobile: screen.getByRole("radio", { name: "Mobiel" }),
+  };
+
+  return result;
+}
+
 async function fillForm(user: UserEvent, testPollingStation: PollingStation | Omit<PollingStation, "id">) {
```
jschuurk-kr, line 28: This function clicks the "Opslaan" button. Both the test "PollingStationForm create" and "Validation backend errors" click that button in the test as well, while I don't think they should.

jschuurk-kr: `fillFormAndSave()` might be a better name in that sense.

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
lkleuver: i've rewritten this to separate tests

frontend/app/component/form/polling_station/PollingStationForm.tsx
lkleuver: this has been refactored

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
lkleuver: changed

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
lkleuver: removed

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
lkleuver: changed to user

lkleuver:
@jorisleker I've not resolved "icon should be red"
Created issue #765

frontend/app/component/form/polling_station/PollingStationForm.test.tsx
```diff
@@ -1,65 +1,162 @@
...
+      await user.click(screen.getByRole("button", { name: "Opslaan en toevoegen" }));
+
+      await waitFor(() => {
+        expect(screen.getByRole("alert")).toHaveTextContent(
+          ["Er bestaat al een stembureau met nummer 42.", "Het nummer van het stembureau moet uniek zijn."].join(""),
+        );
+      });
+      //TODO:
+      //this is being called, handleSubmit is called twice. but why?
+      //expect(onSaved).not.toHaveBeenCalled();
```
Lionqueen94, line 156: Does this still have to be solved?

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -60,10 +67,21 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
     <div>
       {requestState.status === "api-error" && (
         <FormLayout.Alert>
-          <Alert type="error">{requestState.error.message}</Alert>
+          {requestState.error.reference === "EntryNotUnique" ? (
+            <Alert type="error">
+              <h2 id="testt">
```
Lionqueen94, line 72:
```suggestion
              <h2 id="testt">
```
This id needs to be updated to something non `testt`

frontend/lib/ui/Form/validation/form.util.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const numberInput = document.createElement("input");
+    numberInput.name = "number";
+    numberInput.value = "17";
+
+    const { isValid, requestObject } = processForm<RequestObject>(fields, { name: nameInput, number: numberInput });
+    expect(isValid).toBe(true);
+    expect(requestObject).toEqual({ name: "Polling station A", number: 17 });
+  });
+  test("Process form, validation errors", () => {
```
Lionqueen94, line 42:
```suggestion
  });
  
  test("Process form, validation errors", () => {
```

frontend/lib/ui/Form/validation/form.util.ts
```diff
@@ -0,0 +1,113 @@
+//TODO: Refactor ValidationRules to FormProperties
```
Lionqueen94, line 1: Has this TODO been done?

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -77,13 +95,24 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
                 fieldWidth="narrow"
                 defaultValue={pollingStation?.number}
                 pattern="[0-9]*"
```
Lionqueen94, line 97: To prevent the default tooltip warning related to the pattern @jorisleker noticed in his comment, you have to set the `title` attribute on fields with a pattern with a custom text. See https://www.w3schools.com/TAGS/att_input_pattern.asp

praseodym: I think that will still cause the browser tooltip to appear, just with a custom text.

If we want to disable the browser validation completely we can add the [`novalidate` attribute to the `<form>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#novalidate).

Lionqueen94: The novalidate is indeed added to the form, but that means that you can type anything in the field with the pattern for numbers only. If I put `1a` in the field, it will save successfully and remove the `a`, instead of say it's not a number. I think that's not the behaviour we want right?

jorisleker: How do we do this in the data entry forms? There, you can enter only numbers, and we don't show the default browser tooltip

jorisleker:
> @jorisleker I've not resolved "icon should be red" Created issue #765

👍 #765 has been resolved and merged to main

frontend/app/component/form/polling_station/PollingStationForm.tsx
jorisleker: Ignore previous comment, I see this issue has been fixed. I don't see the tooltip anymore, and the correct error message is shown below the input:
Scherm­afbeelding 2024-12-20 om 09 23 23

jorisleker:
If I try to create a polling station with only number + name, I get a "Abacus is stuk 422 De JSON is niet geldig" error page. I can only create a polling station if I fill out all the fields.

frontend/app/component/form/polling_station/PollingStationForm.tsx
lkleuver: data entry form use NumberInput which has custom key press handlers.
We could do that here as well, just the same but without formatting. 

Personally I would prefer to use `<input type="number" />` but this is up for debate.

lkleuver:
> If I try to create a polling station with only number + name, I get a "Abacus is stuk 422 De JSON is niet geldig" error page. I can only create a polling station if I fill out all the fields.

This is me being ahead of the backend, this PR will need to wait for #684

frontend/app/component/form/polling_station/PollingStationForm.tsx
Lionqueen94: The `<input type="number" />` is not an option design wise because you cannot turn off the up and down arrows. Meaning you can accidentally change the number when you press arrows on your keyboard, which is not desired.

Lionqueen94: I removed the `pattern` and `novalidate` because together they do nothing.

Lionqueen94:
@jorisleker I just made sure the address field is the same width as the postal code + locality fields together (see commit b0e0da1d3592092a7835b3eee10315cfd4426575). The current page layout is not exactly the same as the design on Figma. Would you like it to be updated to the design or are you happy with how it looks now?

praseodym (changes requested):
Leaving the address, postal code or locality fields empty does not show a validation error but does return an error an error from the backend because these fields are not included in the request body (i.e. they are `undefined`). This leads to the following full-screen error page in the frontend:

> 422 De JSON is niet geldig
>
> Failed to deserialize the JSON body into the target type: missing field `address` at line 1 column 24

Lionqueen94:
> Leaving the address, postal code or locality fields empty does not show a validation error but does return an error an error from the backend because these fields are not included in the request body (i.e. they are `undefined`). This leads to the following full-screen error page in the frontend:
> 
> > 422 De JSON is niet geldig
> > Failed to deserialize the JSON body into the target type: missing field `address` at line 1 column 24

This has now been fixed in collaboration with @praseodym . 

@jorisleker Can you have a look at the current setup for polling station type, it was needed to change it from the design because you can otherwise not 'deselect' a type:
image

jorisleker (changes requested):
Design looks good. 

One thing that should be fixed:
- When there is a validation error, the page should scroll back to top after clicking submit.

Lionqueen94:
> Design looks good.
> 
> One thing that should be fixed:
> 
>     * When there is a validation error, the page should scroll back to top after clicking submit.

Added a scroll to top of page in case of validation errors in 88a75898be597c024ada824c4fdcccb9ef3c7748

praseodym (approved):

Lionqueen94 (approved):

praseodym:
@lkleuver you enabled auto-merge but there is still an open conversation that needs to be answered.

praseodym (changes requested):
On a second look, I think the files in `lib/ui/Form/validation` should be part of `lib/util` instead of `lib/ui`.

jorisleker:
> > Design looks good.
> > One thing that should be fixed:
> > ```
> > * When there is a validation error, the page should scroll back to top after clicking submit.
> > ```
> 
> Added a scroll to top of page in case of validation errors in [88a7589](https://github.com/kiesraad/abacus/commit/88a75898be597c024ada824c4fdcccb9ef3c7748)

This works for validation errors on fields (ie 'name cannot be empty'), but when I trigger the 'number not unique' error, the page does not scroll to top, whilst it should.

jorisleker (changes requested):
see previous comment

Lionqueen94:
> > > Design looks good.
> > > One thing that should be fixed:
> > > ```
> > > * When there is a validation error, the page should scroll back to top after clicking submit.
> > > ```
> > 
> > 
> > Added a scroll to top of page in case of validation errors in [88a7589](https://github.com/kiesraad/abacus/commit/88a75898be597c024ada824c4fdcccb9ef3c7748)
> 
> This works for validation errors on fields (ie 'name cannot be empty'), but when I trigger the 'number not unique' error, the page does not scroll to top, whilst it should.

Very good point, I fixed this, thanks for testing!

jorisleker (approved):

Lionqueen94:
> On a second look, I think the files in `lib/ui/Form/validation` should be part of `lib/util` instead of `lib/ui`.

I moved them into the `util` folder, this makes sense to me too.

frontend/app/component/form/polling_station/PollingStationForm.tsx
Lionqueen94: Yes it is, discussed it with @praseodym and we will keep it like this for now and potentially refactor in the future (including `NumericParam` as well) if we find it needed.

praseodym (approved):

PR #703, opened 2024-12-06
Closes #388

### LINKED ISSUE #388: Feedback when polling station in progress or completed
When searching for a polling station, you can run into a couple of scenario's that we don't have proper feedback for yet.

1. Polling station is found, but already finalized (first + second input complete, regardless whether there a differences, etc):
image

2. Polling station is found, but assigned to another user (regardless whether it's first or second data entry):
image

3. Polling station is found, has entry data and is assigned to current user (status batch for first and second entry has pencil icon): 
image

In scenario 1 and 2, if the user clicks the submit button, we show the feedback that is already implemented, no need to change.

The design of the 'search by polling station number' component can be found [on the components page in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=137-5073&t=fwYzqCKFITZhwsl2-4)

**Note**: Right now we have not implemented users/roles yet. That means scenario 2 can't be addressed yet. Please create a new ticket to implement that behavior after users and roles have been implemented.

Lionqueen94:
Closes #388 

Also adds final translations for Polling Station Choice Page

### DIFF
--- a/frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
+++ b/frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
@@ -18,9 +18,8 @@ export interface PollingStationChoiceFormProps {
   anotherEntry?: boolean;
 }
 
-const INVALID_POLLING_STATION_ALERT: string = "Voer een geldig nummer van een stembureau in om te beginnen";
-const DEFINITIVE_POLLING_STATION_ALERT: string =
-  "Het stembureau dat je geselecteerd hebt kan niet meer ingevoerd worden";
+const INVALID_POLLING_STATION_ALERT: string = t("polling_station_choice.enter_a_valid_number_to_start");
+const DEFINITIVE_POLLING_STATION_ALERT: string = t("polling_station_choice.polling_station_entry_not_possible");
 
 export function PollingStationChoiceForm({ anotherEntry }: PollingStationChoiceFormProps) {
   const navigate = useNavigate();
--- a/frontend/app/component/form/data_entry/polling_station_choice/PollingStationSelector.module.css
+++ b/frontend/app/component/form/data_entry/polling_station_choice/PollingStationSelector.module.css
@@ -21,7 +21,7 @@
   align-content: center;
   font-size: var(--font-size-body);
   display: inline-flex;
-  gap: 1rem;
+  gap: 0.5rem;
   align-items: center;
   width: 100%;
 
@@ -39,5 +39,6 @@
 
   .icon {
     padding-top: 0.375rem;
+    margin-right: 0.5rem;
   }
 }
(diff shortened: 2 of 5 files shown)


frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.test.tsx
```diff
@@ -145,9 +145,14 @@ describe("Test PollingStationChoiceForm", () => {
...
       await user.type(pollingStation, "34");
 
-      // Click submit again and see that the alert appeared again
+      // Click submit and see that the alert appears
       await user.click(submitButton);
 
+      // Test if the warning message is shown correctly
+      expect(screen.getByTestId("pollingStationSelectorFeedback").textContent).toBe(
+        "Stembureau 34 (Testplek) is al twee keer ingevoerd",
+      );
```
Lionqueen94, line 154: @jschuurk-kr Could you have a look at this test, it works in the browser, but somehow in the test the text shows `"Geen stembureau gevonden met nummer 34"`

praseodym: Wrapping it with `waitFor` works a bit better:
```
      await waitFor(() => {
        expect(screen.getByTestId("pollingStationSelectorFeedback").textContent).toBe(
          "Stembureau 34 (Testplek) is al twee keer ingevoerd",
        );
      });
```
But I find it strange that "Geen stembureau gevonden met nummer 34" would be rendered initially.

Lionqueen94: Hmm yeah no clue why that is, like I said it does not happen when I just test in the browser. Just goes from aan het zoeken... to the expected message 🤔

jorisleker (approved):
Tested: 
✅ I get the 'is al twee keer ingevoerd' feedback when a record is present in pollingStationResults. Right now, thats already the case after the first data entry is completed, but I think this will be fixed once we fully implement second data entry. 

Don't think this PR should close #388, because we are still lacking some of the other feedback messages requested in that issue.

Lionqueen94:
> Tested: ✅ I get the 'is al twee keer ingevoerd' feedback when a record is present in pollingStationResults. Right now, thats already the case after the first data entry is completed, but I think this will be fixed once we fully implement second data entry.
> 
> Don't think this PR should close #388, because we are still lacking some of the other feedback messages requested in that issue.

That's why I created a new issue for the remaining scenario's. It will take a while until we have user accounts so a separate issue for those made sense to me, see: https://github.com/kiesraad/abacus/issues/388#issuecomment-2523531943

lkleuver (approved):

jorisleker (approved):
Looks like only changes since my last review are changes to test. Maybe @jschuurk-kr should review as well?

PR #713, opened 2024-12-09
Configure npm strict-peer-deps

praseodym:
https://docs.npmjs.com/cli/v9/commands/npm-ci#strict-peer-deps

See https://github.com/kiesraad/abacus/pull/707#issuecomment-2527800376 for the reason for this PR.

### DIFF
--- a/frontend/.npmrc
+++ b/frontend/.npmrc
@@ -1,0 +1,1 @@
+strict-peer-deps=true


rnijveld (approved):

oliver3 (approved):
Also works with `--prefix` 👍

PR #714, opened 2024-12-09
Add users and sessions to the backend

### LINKED ISSUE #709: Implement user fixtures and session store
- User fixtures
- User repository
- Session repository
- Login and logout endpoints

marlonbaeten:
Fixes #709

- Adds user and session tables
- Login and logout endpoints
- These can be tested by calling the API - the the tests!

TODO:
- [x] `utoipa` for handlers
- [x] Test axum handlers (using tower one-shot)


backend/src/lib.rs
```diff
@@ -71,7 +72,12 @@ pub fn router(pool: SqlitePool) -> Result<Router, Box<dyn Error>> {
         )
         .route("/:election_id/status", get(election::election_status));
 
+    let user_router = Router::new()
+        .route("/login", post(authentication::login))
+        .route("/logout", post(authentication::logout));
+
     let app = Router::new()
+        .nest("/ali/user", user_router)
```
praseodym, line 80: What about users that are not named Ali?

backend/fixtures/users.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO users (id, username, password_hash)
+VALUES (1, 'user', '$argon2id$v=19$m=19456,t=2,p=1$frZGxFIhMHEsBJS4/VZr1A$zVIGEmiTFGy9jEy1Bphdq1ZO0lUngom8qu9PLsN6mZY');
```
praseodym, line 2: Maybe add a SQL comment with the plaintext password

backend/src/authentication/user.rs
```diff
@@ -0,0 +1,274 @@
...
+#[cfg(test)]
+mod tests {
+    use sqlx::SqlitePool;
+    use std::time::Duration;
+
+    use crate::authentication::{
+        error::AuthenticationError, session::Sessions, user::Users, util::get_current_time,
+    };
+
+    #[sqlx::test(fixtures("../../fixtures/users.sql"))]
```
praseodym, line 195: I don't think this fixture is used here?

backend/src/authentication/user.rs
```diff
@@ -0,0 +1,274 @@
...
+        let fetched_user = users.get_by_id(user.id).await.unwrap().unwrap();
+
+        assert_eq!(user, fetched_user);
+
+        let fetched_user = users.get_by_username("test_user").await.unwrap().unwrap();
+
+        assert_eq!(user, fetched_user);
+    }
+
+    #[sqlx::test(fixtures("../../fixtures/users.sql"))]
```
praseodym, line 213: I don't think this fixture is used here?

backend/src/authentication/user.rs
```diff
@@ -0,0 +1,274 @@
...
+            .await
+            .unwrap_err();
+
+        assert!(matches!(
+            authenticated_user,
+            AuthenticationError::UserNotFound
+        ));
+    }
+
+    #[sqlx::test(fixtures("../../fixtures/users.sql"))]
```
praseodym, line 244: I don't think this fixture is used here?

backend/src/authentication/handlers.rs
```diff
@@ -0,0 +1,261 @@
...
+    use sqlx::SqlitePool;
+    use tower::ServiceExt;
+
+    use super::{login, logout};
+    use crate::{
+        authentication::{Credentials, LoginResponse},
+        AppState,
+    };
+
+    fn create_app(pool: SqlitePool) -> Router {
```
praseodym, line 128: These tests look a lot like integration tests, maybe they should be moved to the `backend/tests/` folder

backend/README.md
```diff
@@ -45,6 +45,8 @@ Use `cargo test` to run the tests. The tests are also run in a GitHub Actions wo
 The following dependencies (crates) are used:
 
 - `axum`: web application framework that focuses on ergonomics and modularity.
+- `axum_extra`: handling for attachements and cookies in axum.
```
Lionqueen94, line 48:
```suggestion
- `axum_extra`: handling for attachments and cookies in axum.
```

backend/migrations/6_sessions.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE sessions
+(
+    session_key         TEXT                 PRIMARY KEY NOT NULL,
```
Lionqueen94, line 3:
```suggestion
    session_key        TEXT                 PRIMARY KEY NOT NULL,
```

backend/src/authentication/handlers.rs
```diff
@@ -0,0 +1,261 @@
...
+        return Ok((jar, StatusCode::OK));
+    };
+
+    // Remove session form the database
+    let Some(session) = sessions.get_by_key(cookie.value()).await? else {
+        // no session found, return OK
+        return Ok((jar, StatusCode::OK));
+    };
+
+    sessions.delete(session.session_key()).await?;
```
praseodym, line 105: I think we can delete the session from the database without checking if it exists first

backend/src/authentication/handlers.rs
```diff
@@ -0,0 +1,261 @@
...
+pub async fn logout(
+    State(sessions): State<Sessions>,
+    jar: CookieJar,
+) -> Result<impl IntoResponse, APIError> {
+    let Some(mut cookie) = jar.get(SESSION_COOKIE_NAME).cloned() else {
+        // no cookie found, return OK
+        return Ok((jar, StatusCode::OK));
+    };
+
+    // Remove session form the database
```
Lionqueen94, line 99:
```suggestion
    // Remove session from the database
```

backend/migrations/6_sessions.sql
praseodym, line 3: formatting issue:
```suggestion
    session_key        TEXT                 PRIMARY KEY NOT NULL,
```

backend/src/authentication/handlers.rs
```diff
@@ -0,0 +1,261 @@
...
+    let Some(session) = sessions.get_by_key(cookie.value()).await? else {
+        // no session found, return OK
+        return Ok((jar, StatusCode::OK));
+    };
+
+    sessions.delete(session.session_key()).await?;
+
+    // Set cookie parameters, these are not be retrieved by the request, and have to match in order to clear the cookie
```
Lionqueen94, line 107:
```suggestion
    // Set cookie parameters, these are not to be retrieved by the request, and have to match in order to clear the cookie
```

praseodym (changes requested):
Nice work! Added some minor comments but looks good overall.

backend/src/authentication/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+mod password;
+mod session;
+mod user;
+mod util;
+
+pub use error::AuthenticationError;
+pub use handlers::{login, logout, Credentials, LoginResponse};
+
+/// Session life time, for both cookie and database
+pub const SESSION_LIFE_TIME: Duration = Duration::from_secs(60 * 60 * 2);
```
Lionqueen94, line 14:
```suggestion
pub const SESSION_LIFE_TIME: Duration = Duration::from_secs(60 * 60 * 2);  /// 2 hours
```

backend/src/authentication/session.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub async fn delete(&self, session_key: &str) -> Result<(), AuthenticationError> {
+        sqlx::query!("DELETE FROM sessions WHERE session_key = ?", session_key)
+            .execute(&self.0)
+            .await?;
+
+        Ok(())
+    }
+
+    /// Delete all sessions that have expired
+    pub async fn delete_old_sessions(&self) -> Result<(), AuthenticationError> {
```
Lionqueen94, line 130:
```suggestion
    pub async fn delete_expired_sessions(&self) -> Result<(), AuthenticationError> {
```
Old session is a bit ambiguous in my opinion, because it can also be the previous session of a user that might still be valid. So using the word expired is more specific.

backend/src/authentication/user.rs
```diff
@@ -0,0 +1,274 @@
...
+
+pub struct Users(SqlitePool);
+
+impl Users {
+    #[cfg(test)]
+    pub fn new(pool: SqlitePool) -> Self {
+        Self(pool)
+    }
+
+    /// Authenticate a user by their username and password, returns a user instace on success or an error
```
Lionqueen94, line 71:
```suggestion
    /// Authenticate a user by their username and password, returns a user instance on success or an error
```

backend/src/authentication/user.rs
```diff
@@ -0,0 +1,274 @@
...
+        // fetch the user from the database
+        let Some(user) = self.get_by_id(session.user_id()).await? else {
+            return Err(AuthenticationError::UserNotFound);
+        };
+
+        Ok(user)
+    }
+
+    /// Create a new user, save a Argon2id v19 hash of the password
+    #[allow(unused)]
```
Lionqueen94, line 106: What does `#[allow(unused)]` mean?

backend/src/authentication/util.rs
```diff
@@ -0,0 +1,63 @@
+use std::time::{Duration, SystemTime};
+
+use rand::{distributions::Alphanumeric, Rng};
+
+use super::error::AuthenticationError;
+
+/// Create a new session key, a secure random alphanumeric string of 24 characters
+/// Which correspons to ~ 142 bits of entropy
```
Lionqueen94, line 8:
```suggestion
/// Which corresponds to ~ 142 bits of entropy
```

backend/migrations/6_sessions.sql
Lionqueen94: Duplicate

backend/src/authentication/util.rs
Lionqueen94: Duplicate

praseodym: You can also delete comments :)

backend/src/authentication/user.rs
marlonbaeten: Cargo should not warn that this code is currently not called, since we need it in the future.

backend/src/authentication/handlers.rs
praseodym: I see, in other modules we don't build the `Router` and just use the handlers directly. Not sure what's better though.

backend/src/lib.rs
```diff
@@ -116,6 +122,8 @@ pub fn create_openapi() -> utoipa::openapi::OpenApi {
     #[derive(OpenApi)]
     #[openapi(
         paths(
+            authentication::handlers::login,
+            authentication::handlers::logout,
```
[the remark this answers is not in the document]
marlonbaeten: I moved the handlers to `mod.rs`

backend/src/authentication/handlers.rs
marlonbaeten: I think using the tower oneshot method is nice since you could also test any middleware, but more importantly, also tests the request extractors.

praseodym: > I think using the tower oneshot method is nice since you could also test any middleware, but more importantly, also tests the request extractors.

Agreed, I think we can use this method for more tests.

praseodym (dismissed):

backend/src/authentication/session.rs
```diff
@@ -43,6 +43,7 @@ impl Session {
     }
 
     /// Get the session key
+    #[cfg(test)]
```
Lionqueen94, line 46: This is not a test right?

marlonbaeten: `#[cfg(test)]` does not mean its a test, the code should only below `#[cfg(test)]` should only be included when running `cargo test`. In this case this method is not used anywhere, except for in tests. In the future that might change.

Lionqueen94: Aaahh I had no idea, thanks for enlightening me :smile:

cikzh (dismissed):

backend/src/authentication/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+/// Logout endpoint, deletes the session cookie
+#[utoipa::path(
+    post,
+    path = "/api/user/logout",
+    responses(
+        (status = 200, description = "Successful logout, or user was already logged out"),
+        (status = 500, description = "Internal server error", body = ErrorResponse),
+    ),
```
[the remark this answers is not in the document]
marlonbaeten: The delete call cannot trigger a 404 - only a 500 when there are other database errors.

cikzh (approved):

Lionqueen94 (approved):

PR #658, opened 2024-11-28
Add support for second data entry to frontend

praseodym:
Implement second data entry in frontend and how that is handled by the backend, in addition to what was already done in #587. The frontend URLs for data entry have been changed to add the data entry number (1 or 2). The coordinator status screen has been updated to include a section for the second data entry. Resolves #588.

RTL and Playwright tests for second data entry are added but are relatively limited. I'm open to suggestions on what other cases we should test to improve coverage.

The Playwright test uses a combination of an API request to create the first data entry, and an extracted function to share the code of doing the data entry with the test for the first data entry.

Note that:
- ~~There is still a backend integration test failure.~~ Fixed in 86484e0.
- ~~The coordinator screen has not been updated yet.~~ Added in d945f1e.
- Second data entry in MSW is missing, but can be added if we decide it is useful. This relates to #650.
- Currently the first data entry is used as the definitive result regardless of any differences between the first and second data entry. Detecting differences for finalisation will be implemented in #633 (current epic).
- Feedback for differences between first and second data entry during data entry is not yet implemented. This will be part of #130.

PR #715, opened 2024-12-09
EntryNumber enum

cikzh:
Adds an enum for `entry_number`s. The goal is to parse the entry number once at serialization, so you when you use the `EntryNumber` type, you'll be guaranteed it represents a valid entry number.
It mitigates the need for a lot of checks throughout the API code and it offloads logic from the State Machine that will be introduced in a future PR.

- I chose an enum instead of a newtype around a `u8`, because it is harder to (accidentally) circumvent the type system.
- I haven't updated the `utoipa` attributes, as I'm not sure it makes sense to do so in this case
- I've updated the test code to use `try_from()` + `unwrap()` instead of enum variants directly, so we don't skip on testing the validation

### DIFF
--- a/backend/src/data_entry/mod.rs
+++ b/backend/src/data_entry/mod.rs
@@ -2,6 +2,7 @@ use axum::extract::{FromRequest, Path, State};
 use axum::http::StatusCode;
 use axum::response::{IntoResponse, Response};
 use axum::Json;
+use entry_number::EntryNumber;
 use serde::{Deserialize, Serialize};
 use sqlx::SqlitePool;
 use utoipa::ToSchema;
@@ -15,6 +16,7 @@ use crate::polling_station::repository::PollingStations;
 use crate::polling_station::structs::PollingStation;
 use crate::validation::ValidationResults;
 
+mod entry_number;
 pub mod repository;
 pub mod structs;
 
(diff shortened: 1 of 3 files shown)


rnijveld (dismissed):

praseodym:
@cikzh can you resolve the conflicts caused by #658 being merged?

backend/src/data_entry/mod.rs
```diff
@@ -65,12 +67,12 @@ pub async fn polling_station_data_entry_save(
     State(polling_station_data_entries): State<PollingStationDataEntries>,
     State(polling_stations_repo): State<PollingStations>,
     State(elections): State<Elections>,
-    Path((id, entry_number)): Path<(u32, u8)>,
+    Path((id, entry_number)): Path<(u32, EntryNumber)>,
     data_entry_request: SaveDataEntryRequest,
 ) -> Result<SaveDataEntryResponse, APIError> {
     // Check if it is valid to save the data entry
     match entry_number {
```
jschuurk-kr, line 74: It doesn't look like we have tests for this?

cikzh: True and they already should have been there. Only the happy path is tested. 
This `match` will be refactored in the state machine PR that is in progress now, so I'm inclined to add the tests when the refactor is done.

jschuurk-kr: Works for me. Am guessing it does not make sense to add some tests for the new enum itself?

cikzh: I'd say it does makes sense and we have them by calling `EntryNumber::try_from()` in the tests. Serialization and Encoding paths are tested via integration- and DOM2DB tests. But again, we currently only test happy paths. I'll make a note in the issue to make sure we test error cases.

cikzh:
> @cikzh can you resolve the conflicts caused by #658 being merged?

@praseodym This is now done

backend/src/data_entry/mod.rs
cikzh: Added a sub issue: https://github.com/kiesraad/abacus/issues/728

rnijveld (approved):

praseodym (approved):

PR #727, opened 2024-12-10
Fix a little performance nitpick in formatting hash

rnijveld:
I just couldn't let it go ;)

### DIFF
--- a/backend/src/eml/base.rs
+++ b/backend/src/eml/base.rs
@@ -88,6 +88,7 @@ pub trait EMLDocument: Sized + DeserializeOwned + Serialize {
 
 pub fn eml_document_hash(input: &str, chunked: bool) -> String {
     use sha2::Digest;
+    use std::fmt::Write;
     let digest = sha2::Sha256::digest(input.as_bytes());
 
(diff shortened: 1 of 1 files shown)


praseodym:
Related because I saw you using `unwrap`: we should have linting configured to warn about `unwrap` usage, but I think it doesn't work. Any ideas about what's up there?

https://github.com/kiesraad/abacus/blob/1f12f233c794d8c160cfa9b78884f6aae3146f86/backend/Cargo.toml#L48-L49
https://github.com/kiesraad/abacus/blob/2cd316552fdf1a5a4ef0935c89f289a7d1aad47f/backend/clippy.toml#L1

praseodym (dismissed):

praseodym:
> Related because I saw you using `unwrap`: we should have linting configured to warn about `unwrap` usage, but I think it doesn't work. Any ideas about what's up there?

We think this is caused by https://github.com/rust-lang/rust-clippy/issues/13455, not much we can do about this now.

praseodym (approved):

cikzh (approved):

PR #730, opened 2024-12-10
Add API endpoint to retrieve zip with xml and pdf combined

rnijveld:
- Added a new `zip` dependency to create the zip file (doing it without a dependency, especially with compression, is hard to do)
- Added a small struct `ResultsInput` to share some code between the xml, pdf and zip endpoints. 
- I'm extracing the zip in our test, but it does use the same library as the one that created it, might not say too much about the validity of the file
- Updated urls and filenames for the results endpoints
- Remove the unnecessary content-type duplication in these endpoints for openapi
- Can be tested by going to http://localhost:3000/api/elections/4/download_zip_results
- Good test: check the hash in the PDF against the sha256sum of the XML

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -42,7 +42,7 @@ argon2 = "0.5.3"
 password-hash = { version = "0.5.0", features = ["getrandom"] }
 rand = "0.8.5"
 cookie = "0.18.1"
-zip = { version = "2.2.1", default-features = false, features = ["deflate", "time"] }
+zip = { version = "2.2.1", default-features = false, features = ["deflate", "chrono"] }
 
 [dev-dependencies]
 http-body-util = "0.1.2"
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -42,6 +42,7 @@ argon2 = "0.5.3"
 password-hash = { version = "0.5.0", features = ["getrandom"] }
 rand = "0.8.5"
 cookie = "0.18.1"
+zip = { version = "2.2.1", default-features = false, features = ["deflate", "time"] }
 
 [dev-dependencies]
 http-body-util = "0.1.2"
--- a/backend/src/election/mod.rs
+++ b/backend/src/election/mod.rs
@@ -148,7 +148,7 @@ impl ResultsInput {
     fn xml_filename(&self) -> String {
         use chrono::Datelike;
         format!(
-            "Telling_{}{}_{}.xml",
+            "Telling_{}{}_{}.eml.xml",
             self.election.category.to_eml_code(),
             self.election.election_date.year(),
             self.election.location.replace(" ", "_"),
@@ -158,7 +158,7 @@ impl ResultsInput {
     fn pdf_filename(&self) -> String {
         use chrono::Datelike;
         format!(
-            "PV_{}{}_{}.pdf",
+            "Model_Na31-2_{}{}_{}.pdf",
             self.election.category.to_eml_code(),
             self.election.election_date.year(),
             self.election.location.replace(" ", "_"),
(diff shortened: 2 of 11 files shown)


cikzh (dismissed):

backend/Cargo.toml
```diff
@@ -42,6 +42,7 @@ argon2 = "0.5.3"
 password-hash = { version = "0.5.0", features = ["getrandom"] }
 rand = "0.8.5"
 cookie = "0.18.1"
+zip = { version = "2.2.1", default-features = false, features = ["deflate", "time"] }
```
praseodym, line 45: Can we use `chrono` instead of `time`?

backend/src/election/mod.rs
```diff
@@ -103,17 +106,156 @@ pub async fn election_status(
...
+
+/// Download a zip containing a PDF for the PV and the EML with election results
+#[utoipa::path(
+    get,
+    path = "/api/elections/{election_id}/download_zip_results",
+    responses(
+        (
+            status = 200,
+            description = "ZIP",
+            content_type = "application/x-zip",
```
praseodym, line 197: The [official MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml) is `application/zip`

backend/Cargo.toml
rnijveld: I'm now using chrono, but we no longer get the modified timestamps automatically unfortunately, so I had to add some code for that.

backend/src/election/mod.rs
```diff
@@ -103,17 +106,163 @@ pub async fn election_status(
...
+            &self.results,
+            &self.summary,
+            &self.creation_date_time,
+        )
+    }
+
+    fn xml_filename(&self) -> String {
+        use chrono::Datelike;
+        format!(
+            "Telling_{}{}_{}.xml",
```
praseodym, line 151:
```suggestion
            "Telling_{}{}_{}.eml.xml",
```

backend/src/election/mod.rs
```diff
@@ -103,17 +106,163 @@ pub async fn election_status(
...
+            self.election.category.to_eml_code(),
+            self.election.election_date.year(),
+            self.election.location.replace(" ", "_"),
+        )
+    }
+
+    fn pdf_filename(&self) -> String {
+        use chrono::Datelike;
+        format!(
+            "PV_{}{}_{}.pdf",
```
praseodym, line 161:
```suggestion
            "Model_Na31-2_{}{}_{}.pdf",
```

backend/src/election/mod.rs
```diff
@@ -103,17 +106,163 @@ pub async fn election_status(
...
+    let mut cursor = std::io::Cursor::new(&mut buf);
+    let mut zip = zip::ZipWriter::new(&mut cursor);
+    let options = SimpleFileOptions::default()
+        .compression_method(zip::CompressionMethod::DEFLATE)
+        // zip file format does not support dates beyond 2107 or inserted leap (i.e. 61th) second
+        .last_modified_time(
+            chrono::Local::now()
+                .naive_local()
+                .try_into()
+                .expect("Timestamp outside of zip range"),
```
praseodym, line 242: I think the [recommended message style](https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style) is something like:
```suggestion
                .expect("Timestamp should be inside zip range"),
```

backend/src/election/mod.rs
praseodym, line 1: Maybe the PDF+EML+ZIP code is long enough to refactor into a new 'results' module?

praseodym (approved):

cikzh (approved):

PR #733, opened 2024-12-10
Update election report page to include download button for zip

rnijveld:
- This builds on top of #730
- I've updated the page to include one additional button for the combined output as a zip (and made that the primary button) while also keeping the original pdf-only download button. I've not included an xml-only download. This page is temporary anyway and will be removed as we move forward with this part of the application.
- Added an e2e test to test the download of the zip from the frontend as well

### DIFF
--- a/frontend/app/module/election/page/ElectionReportPage.tsx
+++ b/frontend/app/module/election/page/ElectionReportPage.tsx
@@ -8,6 +8,7 @@ import { Button, FormLayout, PageTitle } from "@kiesraad/ui";
 
 import cls from "./ElectionReportPage.module.css";
 
+// Prompt the user to 'download' (i.e. save) a file
 function offerDownload(blob: Blob, filename: string) {
   const file = new File([blob], filename);
   const fileUrl = window.URL.createObjectURL(file);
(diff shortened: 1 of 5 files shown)


frontend/app/module/election/page/ElectionReportPage.tsx
```diff
@@ -31,22 +50,29 @@ export function ElectionReportPage() {
...
-          }, 30000);
+          throw new Error(message);
+        }
+        filename = result.headers.get("Content-Disposition")?.split('filename="')[1]?.slice(0, -1) ?? "document";
+        return result.blob();
+      })
+      .then(
+        (blob) => {
+          offerDownload(blob, filename);
         },
```
praseodym, line 76: I think we should try to completely deduplicate the logic that is shared with the `downloadPdfResults` function.

We can make these async functions (and annotate the `onClick` function for the button as `void`) to simplify the function.

Lionqueen94 (approved):

jorisleker (approved):
downloading a zip file works! 
The actual download-page [will look different](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=1087-41062&node-type=frame&t=WX0ttTquKp7KxRRO-11), but I think this will be fixed later when we address having multiple "zittingen"

PR #749, opened 2024-12-13
Add 1st 2nd data entry progress bars for typist

Lionqueen94:
Current implementation, since the issue description would show the same value for the second and third bar:
image
@jorisleker Can you have another look at the design? Another suggestion is that the bars can be: `1e invoer bezig`, `2e invoer bezig` and `1e en 2e invoer klaar`. The downside of the `bezig` ones would be that they would go back down once they are done, which is not nice for a progress bar. Let me know what you think.


praseodym:
> Note that the `progress` field will be touched upon in #750, so it might be better to leave the backend implementation of this PR out of scope.

I don't think the progress field is used for the typist progress bars?

Lionqueen94:
> > Note that the `progress` field will be touched upon in #750, so it might be better to leave the backend implementation of this PR out of scope.
> 
> I don't think the progress field is used for the typist progress bars?

No it isn't, just the polling station data entry status is used. I do have to mention that if https://github.com/kiesraad/abacus/pull/745/ is merged first, the second progress bar needs the difference state as well.

oliver3 (approved):

frontend/app/component/election/ElectionProgress.tsx
```diff
@@ -15,10 +15,18 @@ export function ElectionProgress() {
 
   const stats: Stat[] = useMemo(() => {
     const total = statuses.length;
+    const totalFirstEntry = statuses.filter((s) =>
+      ["second_entry", "second_entry_unfinished", "second_entry_in_progress", "definitive"].includes(s.status),
+    ).length;
```
oliver3, line 20: To avoid typo's and have it a bit more explicit I would prefer something like:

```suggestion
    const firstEntryFinished: PollingStationStatus[] = [
      "second_entry",
      "second_entry_unfinished",
      "second_entry_in_progress",
      "definitive",
    ];
    const totalFirstEntry = statuses.filter((s) => firstEntryFinished.includes(s.status)).length;
```

oliver3 (approved):

praseodym (approved):

Issue #657, opened 2024-11-28
Use SQL transactions in data entry functions


praseodym:
Some data entry functions use multiple SQL queries but are not yet using SQL transactions, which could lead to race conditions. This should be fixed by introducing transactions to those functions.

To cut down on the boilerplate required for this, we should investigate if we can let Axum create the transaction so we don't have to pass it to each repository. A crate like https://github.com/digital-society-coop/axum-sqlx-tx could provide some inspiration.

We should also ensure that transaction rollbacks get handled correctly with an API error response.

praseodym:
Also, from the SIG feedback: providing a lot of parameters to the handler functions is not considered a good practice, so maybe we can implement a way to pass the repositories to the handler functions in a different way.

praseodym:
@cikzh I've moved this issue to the state machine epic too, since that will touch a lot of this logic anyway.

Issue #653, opened 2024-11-28
Create diagram with data entry states and transitions


jschuurk-kr:
Create a diagram for data entry states and transitions, similar to https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/data-entry-forms.md.

PR #750, opened 2024-12-13
Status state machine

cikzh:
Closes  https://github.com/kiesraad/abacus/issues/657, https://github.com/kiesraad/abacus/issues/759, https://github.com/kiesraad/abacus/issues/653

First implementation of the status state machine. The goal of this state machine is to move all data entry status related logic from Sqlite to Rust code. We do this so we can fully utilize the Rust type system to encode the different states and it's transitions.

We tried to keep the backend API and frontend intact as much as possible, but changes had to be made. Most notably the removal of the "unfinished" states had an impact on the frontend.

The state machine should follow [this diagram](https://github.com/kiesraad/abacus/pull/750/files#diff-9ffa9f59bdb8041845df7ee11c3d44d2055c557a24fa3ae2ee5f58086dc7848d).

As this PR is quite big, partial reviews are definitely also welcome. For example:
- Verify that the changes in the frontend test are valid 
- Feedback on the diagram https://github.com/kiesraad/abacus/issues/653
- Verify that we used database transactions where needed, so we can close [#657](https://github.com/kiesraad/abacus/issues/657)


documentatie/flowcharts/polling-station-state.md
```diff
@@ -0,0 +1,23 @@
+# Polling Station State
+
+This document describes the states a polling station can have.   
+The transition labels describe the endpoint that is used for performing the transition.
+The "save" endpoint, used to for [First/Second]EntryInProgress states is kept out, because Mermaid doesn't render transitions to itself too well.
+
+```mermaid
+stateDiagram
+NotStarted --> FirstEntryInProgress: delete
```
praseodym, line 9:
```suggestion
NotStarted --> FirstEntryInProgress: claim
```

backend/migrations/2_polling_stations.sql
```diff
@@ -11,6 +11,7 @@ CREATE TABLE polling_stations
     house_number_addition TEXT,
     postal_code           TEXT                              NOT NULL,
     locality              TEXT                              NOT NULL,
+    finished_at           INTEGER,
```
praseodym, line 14: Remove?

backend/migrations/4_polling_station_results.sql
praseodym, line 1: Undo removal

backend/migrations/2_polling_stations.sql
cikzh: Remove: 4922182c54ed183bf224ae49a50b4e0faa9eed6b

backend/migrations/4_polling_station_results.sql
cikzh: Hmmm, I think something went wrong while pushing earlier... Added back in [59a760b](https://github.com/kiesraad/abacus/pull/750/commits/59a760b2df526932ee83a037e5ceb29cbe46891a)

documentatie/flowcharts/polling-station-state.md
rnijveld: Fixed in 99290e407f1b8717bce06e7ec53d7f0495a9f3da

backend/migrations/3_polling_station_data_entries.sql
```diff
@@ -1,16 +1,8 @@
...
-    polling_station_id INTEGER NOT NULL,
-    entry_number       INTEGER NOT NULL,
-    progress           INTEGER NOT NULL,
-    data               BLOB,
-    client_state       BLOB,
-    updated_at         INTEGER NOT NULL DEFAULT (unixepoch()),
-    finalised_at       INTEGER,
+    polling_station_id INTEGER  PRIMARY KEY NOT NULL,
+    state              BLOB     NOT NULL,
+    updated_at         DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
praseodym, line 5: Why not keep the unix time? We had a discussion about this a while back.

SQLite doesn't have a `DATETIME` field, so this will be stored as a string now.

backend/src/data_entry/mod.rs
```diff
@@ -1,38 +1,95 @@
...
+    params(
+        ("polling_station_id" = u32, description = "Polling station database id"),
+        ("entry_number" = u8, description = "Data entry number (first or second data entry)"),
+    ),
+)]
+pub async fn polling_station_data_entry_get(
+    State(polling_station_data_entries): State<PollingStationDataEntries>,
+    State(polling_stations): State<PollingStations>,
+    State(elections): State<Elections>,
+    Path((id, _entry_number)): Path<(u32, EntryNumber)>, // note: we don't need the entry number here
```
praseodym, line 55: You can make the comment a `TODO` and mention #762.

backend/src/data_entry/mod.rs
```diff
@@ -236,62 +216,96 @@ pub async fn polling_station_data_entry_delete(
...
-        // get first data entry from database
-        let (_, first_data, _, _) = polling_station_data_entries
-            .get(&mut tx, id, EntryNumber::FirstEntry)
-            .await?;
-        let first_results = serde_json::from_slice::<PollingStationResults>(&first_data)?;
-
-        // only convert to result the first and second data entries are equal
-        if first_results == results {
-            polling_station_data_entries.to_result(&mut tx, id).await?
+    match entry_number {
```
praseodym, line 229: Do we want to keep the `entry_number` path parameter to this handler or let it use the entry number from the state, like in the save handler?

documentatie/flowcharts/data-entry-state.md
```diff
@@ -0,0 +1,23 @@
+# Data entry state
+
+This document describes the states a data entry can have.
+The transition labels describe the endpoint that is used for performing the transition.
+The "save" endpoint, used to for [First/Second]EntryInProgress states is kept out, because Mermaid doesn't render transitions to itself too well.
```
praseodym, line 5:
```suggestion
The "save" endpoint, used to for [First/Second]EntryInProgress states is kept out, because Mermaid doesn't render self-loops too well.
```

frontend/lib/ui/Badge/Badge.module.css
```diff
@@ -72,7 +72,7 @@
   color: var(--badge-objections);
 }
 
-.second_entry,
+.second_entry_not_started,
 .second_entry_in_progress,
 .second_entry_unfinished {
```
praseodym, line 77: This can be removed

frontend/lib/ui/Badge/Badge.module.css
```diff
@@ -60,7 +60,7 @@
   color: var(--badge-extra-entry);
 }
 
-.not_started,
+.first_entry_not_started,
 .first_entry_in_progress,
 .first_entry_unfinished {
```
praseodym, line 65: This can be removed

backend/src/data_entry/mod.rs
cikzh: Fixed in https://github.com/kiesraad/abacus/pull/750/commits/123c18020c6c3f22faf9cac76bc43e1941e7bc0a

backend/src/data_entry/mod.rs
cikzh: If I remember correctly, we implemented it like this, because we're planning to split this endpoint into separate endpoints for first and second entry, but this doesn't break frontend functionality. Splitting the endpoints depends on how we decide to refactor the frontend, though.
The save endpoint also matches on the `entry_number` path in this PR.

frontend/lib/ui/Badge/Badge.module.css
cikzh: Done in [d3c3073](https://github.com/kiesraad/abacus/pull/750/commits/d3c3073f4899e16ff8df3bf6b230b2ed19b34cef)

frontend/lib/ui/Badge/Badge.module.css
cikzh: Done in [d3c3073](https://github.com/kiesraad/abacus/pull/750/commits/d3c3073f4899e16ff8df3bf6b230b2ed19b34cef)

backend/migrations/3_polling_station_data_entries.sql
cikzh: I think we did this because it is only used in the frontend for display, and parsing/formatting the string this way with JS's `Date.toISOString()` is easier. See the line at the mention (don't know of a better way to point to a diff :) )

frontend/lib/api-mocks/RequestHandlers.ts
```diff
@@ -192,7 +196,7 @@ export const PollingStationDataEntryGetHandler = http.get<
     data: dataEntryRecord.data,
     client_state: dataEntryRecord.clientState,
     validation_results: validate(dataEntryRecord.data),
-    updated_at: Number(Date.now() / 1000),
+    updated_at: new Date().toISOString(),
```
cikzh, line 199: @praseodym see here

frontend/lib/ui/Badge/Badge.module.css
Lionqueen94: Fully fixed in fea0fec47af69848f22a0744166a1cb99d35241d

frontend/lib/ui/Badge/Badge.module.css
Lionqueen94: Fully fixed in fea0fec47af69848f22a0744166a1cb99d35241d

backend/src/data_entry/mod.rs
praseodym: I've added it to #762.

backend/migrations/3_polling_station_data_entries.sql
praseodym: I'd say `new Date(unix_epoch / 1000)` isn't that complicated either (and if we use unix time with millisecond resolution, we don't need the `/ 1000`).

Also, the `polling_station_results` table is still using unix time for the `created_at` column. I'd prefer to just stick to unix time.

rnijveld: I'd like to quote RFC3339 for this:

> Date and time formats cause a lot of confusion and interoperability
> problems on the Internet.

Just to add my reasoning for doing this:
- Unix timestamps are much harder to read as a human (horrible to debug) and aren't always unambiguous, especially when a system is misconfigured to use local time for their unix timestamps somewhere. RFC3339 timestamps could of course still be generated wrongly when the system clock is misconfigured, but at least it will be obvious much quicker when looking at the data
- RFC3339 datetimes are unambiguous (although most datetime libraries (including the chrono and time crates in Rust) don't support leap seconds, but let's not worry about that for now, unix timestamps ignore them as well)
- Our Rust code shouldn't use all those integers, but properly typed data (which is what this PR is all about)
- Sqlite doesn't do much with datatypes in table definitions anyway, an integer column can store a string just as well (try it!), this is mostly type hints for us as humans. SQLite does specify the datetime column type, it just stores it as 'whatever', but it does that for all columns anyway. The only thing you get with an integer column is that it tries some conversion shenanigans for you (and if those conversions fail it will still fall back to whichever of the five storage types of sqlite can store the data).
- The size difference between the integer and the string is pretty small anyway, a fully specified RFC3339 UTC timestamp without milliseconds is only 20 bytes of string, only 2,5 times as big as a 64 bit unix timestamp
- `polling_station_results` should most definitely also not use unix timestamps if it is up to me

We should just keep the unix timestamps to kernel interfaces and system-internal timestamps, and for the rest of us: please always use RFC3339 when sending timestamps over the network or when storing timestamps in a portable way! This was my unix timestamps are bad public service announcement, now I'm back on holiday :wave: :beach_umbrella:

praseodym: Thanks for your convincing PSA 😃 Let's switch to using RFC3339. This PR is good to go for that part and I've created #782 to switch over `polling_station_results` too.

Enjoy the rest of your holiday!

backend/src/data_entry/mod.rs
```diff
@@ -512,27 +534,29 @@ mod tests {
...
-        assert_eq!(row_count.count, 2);
-
+        // TODO: fix
+        /*
+                let row_count = query!("SELECT COUNT(*) AS count FROM polling_station_data_entries WHERE finalised_at IS NOT NULL")
+                    .fetch_one(&pool)
+                    .await
+                    .unwrap();
+                assert_eq!(row_count.count, 2);
+        */
```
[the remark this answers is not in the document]
praseodym: Fixed in fc3cd32ae5c9ae79bbebfb846b1bf12ee1fb3084

backend/tests/data_entries_integration_test.rs
```diff
@@ -251,7 +266,167 @@ async fn test_polling_station_data_entry_deletion(pool: SqlitePool) {
...
+    let status = response.status();
+    let body: ElectionStatusResponse = response.json().await.unwrap();
+
+    assert_eq!(status, StatusCode::OK);
+    assert!(!body.statuses.is_empty());
+    assert_eq!(
+        statuses[1].status,
+        DataEntryStatusName::FirstEntryInProgress
+    );
+    assert_eq!(statuses[1].first_data_entry_progress, Some(60));
```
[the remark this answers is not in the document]
praseodym: Fixed in fc3cd32ae5c9ae79bbebfb846b1bf12ee1fb3084

backend/src/data_entry/mod.rs
```diff
@@ -443,25 +464,27 @@ mod tests {
...
+        // TODO: fix
+        // let rows = query!("SELECT * FROM polling_station_data_entries")
+        //     .fetch_all(&pool)
+        //     .await
+        //     .unwrap();
+        // assert_eq!(rows.len(), 2);
+        // let first_entry = rows.iter().find(|row| row.entry_number == 1).unwrap();
+        // let second_entry = rows.iter().find(|row| row.entry_number == 2).unwrap();
+        // assert!(first_entry.finished_at.is_some());
+        // assert!(second_entry.finished_at.is_none());
```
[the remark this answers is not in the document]
praseodym: Fixed in fc3cd32ae5c9ae79bbebfb846b1bf12ee1fb3084

Lionqueen94 (approved):

Lionqueen94 (changes requested):
Screenshot from 2024-12-30 15-24-59

The timestamp format in the database when finishing a data entry is different from existing timestamps in the migrations.

praseodym:
> The timestamp format in the database when finishing a data entry is different from existing timestamps in the migrations.

Changed this in 16fb22a.

praseodym (approved):

Lionqueen94 (approved):

Issue #665, opened 2024-12-02
d2d tests: refactor fixtures, test data, helper functions


jschuurk-kr:
Based on [this comment](https://github.com/kiesraad/abacus/pull/658#discussion_r1863581839) by @jschuurk-kr form [PR 658](https://github.com/kiesraad/abacus/pull/658):

`fillDataEntryNoRecountNoDifferences()` in `data-entry.e2e.ts` needs refactoring in a new issue. These changes have given me a clearer picture of where we're outgrowing our current d2d test framework.

Some of the things I'd like to (see) tackle(d) in that refactor:

- [ ] The `fillDataEntryNoRecountNoDifferences()` function (and thus the test this code came from) does too many things. Do we need it to check focus and if the value shows up in the field? And even if so, perhaps we can split the focus etc test from the tests that cover filling first/second data entry. -> https://github.com/kiesraad/abacus/pull/897
- [ ] Moving the first data entry of the test "equal second data entry after first data entry" into a fixture. -> https://github.com/kiesraad/abacus/pull/897
- [ ] It would be useful to be able to use the same constant for both API (sending request or checking request) and UI (filling in the data entry forms). -> https://github.com/kiesraad/abacus/pull/897
- [ ] ~~`PollingStationTestData.ts` contains fixture data, a request template, a request, and an expected error response. These are different kinds of things, so should be split out.~~ -> was already done
- [ ] Clean up `frontend/e2e-tests/page-objects/data_entry/index.ts` -> https://github.com/kiesraad/abacus/pull/897
- [ ] move the `protected readonly` into the constructor of the page objects -> https://github.com/kiesraad/abacus/pull/897
- [x] Instead of hard-coding the values of `pollingStation33` in a test data file, use an API call in a fixture to retrieve the data from the backend. -> https://github.com/kiesraad/abacus/pull/754
- [x] Making the code consistent in how it checks it's on the right page (replacing `waitFor()`s with `expect`s on `toBeVisible()`. -> https://github.com/kiesraad/abacus/pull/768

PR #754, opened 2024-12-17
d2d tests: replace hard-coded pollingstation data with an api call fixture

jschuurk-kr:
partial #665

We had the data of a polling station hard-coded in `PollingStationTestData.ts`. I replaced it with a fixture that does an API call to retrieve the polling station data. This way we don't have to keep the hard-coded data in sync with the data in the backend seed data.

I had to update the eslint config, because it saw the Playwright `await use(pollingStation);` as a React hook `use()`.

I also renamed the `PollingStationTestData.ts` file to something that better matches its new contents. I will be making more changes to that file as part of the refactoring covered in #665.

Questions:
- Does the way I separated auto-fixtures and regular fixtures in `fixtures.ts` make sense? As in, is it conceptually clear? And is this a good way to do it syntax-wise?

### DIFF
--- a/frontend/.eslintrc.cjs
+++ b/frontend/.eslintrc.cjs
@@ -36,7 +36,7 @@ module.exports = {
   },
   overrides: [
     {
-      files: ["*.e2e.ts", "*PgObj.ts"],
+      files: ["*.e2e.ts", "*PgObj.ts", "e2e-tests/**/fixtures.ts"],
       extends: [
         "eslint:recommended",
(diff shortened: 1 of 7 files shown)


Lionqueen94 (approved):

praseodym (approved):

praseodym:
> * Does the way I separated auto-fixtures and regular fixtures in `fixtures.ts` make sense? As in, is it conceptually clear? And is this a good way to do it syntax-wise?

This makes sense to me!

Issue #764, opened 2024-12-18
notulen retro 16 december


jschuurk-kr:
- [x] notulen in internal repo -> https://github.com/kiesraad/abacus-internal/pull/257
- [x] aanpassingen in Abacus: documentatie/ontwikkelproces -> https://github.com/kiesraad/abacus/pull/767
- [x] aanpassingen in Abacus: epic template -> https://github.com/kiesraad/abacus/pull/766

PR #766, opened 2024-12-19
adds epic template

jschuurk-kr:
relates to #764 

This PR adds en epic template. It looks like this:
image

---

When you create an issue, you will either be presented by a modal to choose a template:
image

Or when you create an item on the board, you can open the modal by clicking the button:
image

### DIFF
--- a/.github/ISSUE_TEMPLATE/epic-template.md
+++ b/.github/ISSUE_TEMPLATE/epic-template.md
@@ -6,7 +6,6 @@ type: epic
 labels: ''
 assignees: ''
 projects: ["kiesraad/1"]
-area: epic
 
 ---
 
--- a/.github/ISSUE_TEMPLATE/epic-template.md
+++ b/.github/ISSUE_TEMPLATE/epic-template.md
@@ -1,0 +1,27 @@
+---
+name: Epic template
+about: template for epics
+title: 'Epic: '
+labels: ''
+assignees: ''
+
+---
+
+# Context
+<!--
+Describe the context of the epic. Where does it fit into the process? Which related features have been built already? Which will be built at a later time?
+-->
+
+# Architecture
+<!--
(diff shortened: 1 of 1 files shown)


Lionqueen94 (approved):

.github/ISSUE_TEMPLATE/epic-template.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Epic template
+about: template for epics
+title: 'Epic: '
+labels: ''
+assignees: ''
+
+---
```
praseodym, line 8: I think we can add `type: epic` here to automatically set the issue type too.

From the [Issue Types Public Preview  discussion](https://github.com/orgs/community/discussions/139933):
> Additionally, you can specify the type qualifier in [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) in your repositories, so that each issue created from each template automatically has the type set.

jschuurk-kr: Good find! I added three fields:
```
type: epic
projects: ["kiesraad/1"]
area: epic
```

I couldn't test the "type" in my repo outside the Kiesraad organization. The "projects" did not work in my repo. (I am sure it expect the project number.) I don't think the "area" is supported yet, so not sure if it's a good idea to add it already.

praseodym: 'Area' is a custom field in our project and I don't think it's possible to set custom fields in the issue template: https://github.com/orgs/community/discussions/63402#discussioncomment-6717854

So maybe it's better to remove that now, because the syntax might be different once it is possible to set custom fields.

Lionqueen94 (dismissed):

.github/ISSUE_TEMPLATE/epic-template.md
jschuurk-kr: I removed the 'area' field from the template.

praseodym (approved):

PR #768, opened 2024-12-19
replaces waitFor() with expect() in d2d tests

jschuurk-kr:
relates to #665

Our d2d tests used `waitFor()` to check if certain elements (esp. headings) were visible, while in other places we use an `expect().toBeVisible()`. Since our intention is to assert on visibility, `expect()` is the better way to do this. So this PR replaces all those `waitFor()` with `expect().toBeVisible()`.

### DIFF
--- a/frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
+++ b/frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
@@ -297,7 +297,7 @@ test.describe("full data entry flow", () => {
     await candidatesListPage_1.next.click();
 
     const checkAndSavePage = new CheckAndSavePage(page);
-    await checkAndSavePage.fieldset.isVisible();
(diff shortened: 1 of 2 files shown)


frontend/e2e-tests/dom-to-db-tests/data-entry.e2e.ts
```diff
@@ -297,10 +297,10 @@ test.describe("full data entry flow", () => {
     await candidatesListPage_1.next.click();
 
     const checkAndSavePage = new CheckAndSavePage(page);
-    await checkAndSavePage.fieldset.waitFor();
+    await checkAndSavePage.fieldset.isVisible();
```
Lionqueen94, line 300:
```suggestion
    await expect(checkAndSavePage.fieldset).toBeVisible();
```
shouldn't this one also have an added expect?

Lionqueen94 (approved):

praseodym (approved):

PR #780, opened 2024-12-23
Fix wrong status badge alignment on small windows

Lionqueen94:
## Issue:
Image

## Fixed:
- Fixed rendering of badge and button on data entry page on smaller screens
- Fixed h1 multi-line rendering and reduced gap around polling station name on data entry page

image

## TODO:
- Alignment of badge to the left in case of multi line h1 polling station name. @jorisleker any tips? The width of the h1 stays wide even with the content wrapped. It seems like there is no way to change this - I googeled. So either we can keep it as is, or move the badge in front of the polling station name. Those are the options I can think of.

Test locally on: `http://localhost:3000/elections/3/data-entry/5/1/recounted` with a reduced screen width.


Lionqueen94:
Alignment of badge to the left in case of multi line h1 polling station name is still not as expected. @jorisleker any tips? The width of the h1 stays wide even with the content wrapped. It seems like there is no way to change this - I googeled. So either we can keep it as is, or move the badge in front of the polling station name. Those are the options I can think of.

jorisleker:
Thanks for fixing this @Lionqueen94 

One small improvement I can think of, is to reduce the h1 line-height to 2.35rem. Bit more compact, looks better to me. 

> Alignment of badge to the left in case of multi line h1 polling station name is still not as expected. @jorisleker any tips? The width of the h1 stays wide even with the content wrapped. It seems like there is no way to change this - I googeled. So either we can keep it as is, or move the badge in front of the polling station name. Those are the options I can think of.

What happens if you put the badge _inside of the h1_ as an inline-block @Lionqueen94?

Lionqueen94:
> Thanks for fixing this @Lionqueen94
> 
> One small improvement I can think of, is to reduce the h1 line-height to 2.35rem. Bit more compact, looks better to me.
> 
> > Alignment of badge to the left in case of multi line h1 polling station name is still not as expected. @jorisleker any tips? The width of the h1 stays wide even with the content wrapped. It seems like there is no way to change this - I googeled. So either we can keep it as is, or move the badge in front of the polling station name. Those are the options I can think of.
> 
> What happens if you put the badge _inside of the h1_ as an inline-block @Lionqueen94?

@jorisleker I adjusted the line-height to 2.35rem. Can you adjust this in the design as well?
Unfortunately the behaviour is the same if I add the badge inside the h1, it just stays on the right side, no matter what I try with adjusting `display`, `justify-content` etc.

praseodym (approved):

jorisleker (approved):
@Lionqueen94 ok, let's leave it like this then, because this is an improvement over how it was 👍

PR #789, opened 2025-01-02
Added fraction module and apportionment calculation for ≥ 19 seats

Lionqueen94:


Lionqueen94:
@praseodym I added tests for all functions in fraction except `partial_cmp`, `cmp` and `eq`. Any tips on how to test those?

backend/src/apportionment/fraction.rs
```diff
@@ -0,0 +1,164 @@
...
+    use super::*;
+    use crate::apportionment::fraction::Fraction;
+
+    #[test]
+    fn test_from_count() {
+        let numerator = 5 as Count;
+        let fraction = Fraction::new(5, 1);
+        let fraction_string = format!("{}", fraction);
+        assert_eq!(fraction, Fraction::from_count(numerator));
+        assert_eq!(fraction_string, "5")
```
praseodym, line 98: The [`Display` trait](https://doc.rust-lang.org/std/fmt/trait.Display.html) automatically implement the [ToString](https://doc.rust-lang.org/std/string/trait.ToString.html) trait for the type, so this also works:

```suggestion
        assert_eq!(fraction(), "5");
```


And then you can remove the `format!` call.

Lionqueen94: Changed `format!` to `.to_string()` (which is what you mean I think). Thanks!

backend/src/apportionment/fraction.rs
```diff
@@ -0,0 +1,159 @@
...
+    use super::*;
+    use crate::apportionment::fraction::Fraction;
+
+    #[test]
+    fn test_from_count() {
+        let numerator = 5 as Count;
+        let fraction = Fraction::new(5, 1);
+        assert_eq!(fraction, Fraction::from_count(numerator));
+        assert_eq!(fraction.to_string(), "5")
+    }
```
praseodym, line 98: The parameters to `assert_eq!` in these tests is not consistent, mixing up the values passed to the left and right. I'd put the to-be-tested value on the left and the expected value on the right, e.g.:

```suggestion
    #[test]
    fn test_from_count() {
        let count = 5 as Count;
        let fraction = Fraction::from_count(count));
        assert_eq!(fraction, Fraction::new(5, 1));
        assert_eq!(fraction.to_string(), "5")
    }
```

And same for the other tests.

backend/src/apportionment/fraction.rs
```diff
@@ -0,0 +1,159 @@
...
+        Self {
+            numerator: self.numerator * other.denominator,
+            denominator: self.denominator * other.numerator,
+        }
+    }
+
+    // divide by whole number and return the whole number
+    pub fn divide_whole(&self, other: &Self) -> u64 {
+        (self.numerator * other.denominator) / (self.denominator * other.numerator)
+    }
```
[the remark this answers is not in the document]
Lionqueen94: Updated in a0fba9082ef13727cda03cedc2011ee4929f628e

praseodym (approved):

praseodym (approved):

backend/src/apportionment/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+
+mod fraction;
+
+/// Apportionment - work in progress!!
+pub fn seat_allocation(
+    seats: u64,
+    totals: &ElectionSummary,
+) -> Result<Vec<u64>, ApportionmentError> {
+    if seats < 19 {
+        return Err(ApportionmentError::SeatAllocationLessThan19SeatsNotImplemented);
```
marlonbaeten, line 13: We could use the https://doc.rust-lang.org/std/macro.unimplemented.html macro here - instead of introducing a temp error variant.

backend/src/apportionment/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+    println!("Seat allocation for 19 or more seats.");
```
marlonbaeten, line 15: Let's use the tracing create instead of `println!` https://github.com/tokio-rs/tracing
Replacing these statements with debug! allows us to be more flexible in where to direct these in the future.

marlonbaeten (approved):

backend/src/apportionment/mod.rs
Lionqueen94: Ah good to know for the future. This error will be removed in the next PR (already in progress)

backend/src/apportionment/mod.rs
Lionqueen94: I'll change the println!'s to debug! in the next PR that's in progress already

backend/src/apportionment/fraction.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn test_div_whole_number_larger_than_zero() {
+        let fraction = Fraction::new(11, 5);
+        let other_fraction = Fraction::new(1, 2);
+        let divided = fraction / other_fraction;
+        assert_eq!(divided, Fraction::new(22, 5));
+        assert_eq!(divided.to_string(), "4 2/5")
+    }
+
+    #[test]
+    fn test_div_whole_number_smaller_than_zero() {
```
jorisleker, line 150: Shouldn't this be `test_div_whole_number_smaller_than_one`?

backend/src/apportionment/fraction.rs
```diff
@@ -0,0 +1,186 @@
...
+        let fraction = Fraction::new(11, 5);
+        let other_fraction = Fraction::new(1, 2);
+        assert_eq!(
+            fraction.divide_and_return_whole_number(&other_fraction),
+            4u64
+        );
+    }
+
+    #[test]
+    fn test_div_whole_number_larger_than_zero() {
```
jorisleker, line 141: Shouldn't this be `test_div_whole_number_larger_than_one` ?

backend/src/apportionment/fraction.rs
Lionqueen94: No, it's either printed as a fraction (3/2) or as a whole number with a fraction (1 1/2). In the case of 3/2, the whole number is larger than 0 (namely 1) so we go down that route in format, and if it's 1/2 we don't have a whole number so there is just a fraction to be printed. I hope this makes sense.

backend/src/apportionment/fraction.rs
Lionqueen94: See comment above

backend/src/apportionment/fraction.rs
jorisleker: Mather of perspective. I looked at the entire fraction (1 1/2 is larger then one, 2/3 smaller then one), you just counted the wholes. With/without wholes is a nice suggestion!

PR #831, opened 2025-01-08
Simplify backend fixtures and use API calls in Playwright d2d

### LINKED ISSUE #667: simplify seeded fixtures, consider replacing test-only fixtures with API calls
>> * should the new fixture[^1] be an API call instead?

> I think API calls are a bit more maintainable because they are type checked. We can make them into reusable 'fixtures' by extracting them into separate helper functions.

>> * if it remains a fixture, should it be loaded when seeding fixtures?

> No, this way is fine to me. I think we should aim for less variety of different elections for the seeded fixtures, because as a user the current situation gets very confusing. I would prefer having only one test election (e.g. Juinen), in Dutch, to make the application with seeded fixtures easier to understand. That's something for a new issue, though.

_Originally posted by @praseodym in https://github.com/kiesraad/abacus/issues/659#issuecomment-2507575736_

[^1]: This fixture creates a data entry. It's used in the integration tests for the DELETE on polling stations.

praseodym:
- Simplify backend fixtures by using API calls for some tests
- Restructure backend fixtures to have a fixture per election instead of per database table
- Add election creation API endpoint for testing
- Use only API fixtures in Playwright d2d tests, so that we can avoid resetting the database for each test, and we can use multiple workers to execute tests concurrently. On my machine, this reduces the Playwright d2d runtime from 120s (1 worker) to 42s (6 workers).
- Update dev homepage (`/dev`) with dynamic list of elections
- Fixed routing to account setup page

Closes #667.


frontend/e2e-tests/dom-to-db-tests/fixtures.ts
```diff
@@ -2,63 +2,82 @@ import { test as base, expect } from "@playwright/test";
...
+
+    // get election details
+    const electionUrl: ELECTION_DETAILS_REQUEST_PATH = `/api/elections/${emptyElection.id}`;
+    const electionResponse = await request.get(electionUrl);
+    expect(electionResponse.ok()).toBeTruthy();
+    const election = (await electionResponse.json()) as ElectionDetailsResponse;
+
+    await use(election);
+  },
+  pollingStation1: async ({ request, election1 }, use) => {
```
oliver3, line 55: The `1` was originally the polling station with ID `1`. Maybe this can now simply become `pollingStation` and also `election` ?

backend/openapi.json
```diff
@@ -39,6 +39,55 @@
             }
           }
         }
+      },
+      "post": {
```
oliver3, line 43: Do we want to hide this POST endpoint from the OpenAPI (is that possible)? Or maybe hint that this is a test only endpoint by having a separate url with an underscore or something?

oliver3 (approved):
Nice improvements!! Just two minor suggestions.

frontend/e2e-tests/dom-to-db-tests/fixtures.ts
praseodym: Good idea, I've renamed them in cd277470a95aecdd1ad999a8db2c8649223bbdcc.

praseodym:
> If I now run `cargo run -- --reset-database --seed-data` I get only election 3. Is there any way to add a few elections for local development or is this not possible anymore with these changed fixtures? And the links on `/dev/` don't work anymore because most of them link to election 1.

You can edit `backend/src/fixtures.rs` to include more fixtures for local development. I'm curious to know what your use case is though!

I'll updated the links on `/dev`.

Lionqueen94:
> > If I now run `cargo run -- --reset-database --seed-data` I get only election 3. Is there any way to add a few elections for local development or is this not possible anymore with these changed fixtures? And the links on `/dev/` don't work anymore because most of them link to election 1.
> 
> You can edit `backend/src/fixtures.rs` to include more fixtures for local development. I'm curious to know what your use case is though!
> 
> I'll update the links on `/dev`.

I sometimes manually edit the url when testing/developing, so at least it makes more sense for me to have election 1 instead of 3, but to add some variety we might as well have the 3 elections and 6 polling stations existing in the dev environment. So updating to `const FIXTURES: &[Fixture] = load_fixtures!(["reset", "election_1", "election_2", "election_3"]);` in fixtures.rs. The link in dev for `[Stembureaus Verkiezing 3 (leeg)](http://localhost:3000/elections/3/polling-stations#coordinator)` should be removed since all 3 elections have polling stations, so we don't actually have an election without polling stations in the fixtures now.

praseodym:
I've updated the Dev Homepage in bc9f604c3e1a15e11b300ba63a46d0ab3e66a61b.

Lionqueen94 (approved):

backend/src/fixtures.rs
```diff
@@ -23,13 +23,7 @@ macro_rules! load_fixtures {
...
 /// This list should be updated manually when a new fixture is added that
 /// needs to be loaded when seeding fixtures.
-const FIXTURES: &[Fixture] = load_fixtures!([
-    "reset",
-    "elections",
-    "polling_stations",
-    "polling_station_data_entries",
-    "polling_station_results"
-]);
+const FIXTURES: &[Fixture] = load_fixtures!(["reset", "election_3"]);
```
jschuurk-kr, line 26: It would be nicer if the single election that's loaded by default has 1 instead of 3 as its id.

backend/tests/shared/mod.rs
```diff
@@ -1,13 +1,16 @@
 #![cfg(test)]
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
praseodym: This is because each integration test is compiled as its own binary, and if one of those binaries doesn't use all functions in the shared module the compiler warns about dead code.

frontend/e2e-tests/dom-to-db-tests/fixtures.ts
```diff
@@ -1,32 +1,87 @@
...
   },
+  completedElection: async ({ request, election }, use) => {
+    // finalise both data entries for all polling stations
+    for (const pollingStationId of election.polling_stations.map((ps) => ps.id)) {
+      for (const entryNumber of [1, 2]) {
+        const save_url: POLLING_STATION_DATA_ENTRY_SAVE_REQUEST_PATH = `/api/polling_stations/${pollingStationId}/data_entries/${entryNumber}`;
+        const saveResponse = await request.post(save_url, {
+          data: noRecountNoDifferencesRequest,
+        });
+        // eslint-disable-next-line playwright/no-standalone-expect
```
jschuurk-kr, line 76: Since we'll probably have more of these in the future, should we just put `/* eslint-disable playwright/no-standalone-expect */` at the top of the file?

backend/src/fixtures.rs
praseodym: Agreed, but that would require refactoring all backend tests currently using `election_1` (either to a different election id or to a different election entirely).

backend/tests/shared/mod.rs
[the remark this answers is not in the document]
jschuurk-kr: oh wow. does that warrant a code comment?

backend/src/fixtures.rs
jschuurk-kr: That shouldn't be too bad if election 1 becomes 2, 2 becomes 3, and 3 becomes 1?

I'll give it a try to see if it's really as straightforward as I think it is...

frontend/e2e-tests/dom-to-db-tests/fixtures.ts
praseodym: Done in 172e120ed087c054935398e0af80626f12852082. Weirdly enough, the linter only started complaining after I added the for loop, so I think this lint rule has some bugs.

backend/src/fixtures.rs
jschuurk-kr: I completed updating the election ids as mentioned above. Then I figured that it would also make sense to update the polling station ids, so that the election with id 1, also has the polling stations with ids 1, 2, and 3.

While I'm sure it's doable to also change the polling station ids, the test code will start to look a bit odd. Having a test with fixture `election_2` that then uses election with id 2 throughout is fine. Although even then someone might wonder why it's not using election 1.
If I also update the polling station ids, the `save` function in `backend/src/data_entry/mod.rs` then needs `Path((1, entry_number)),` changed to `Path((4, entry_number)),`. Which starts to become a bit magical number-y. Alternative would be to make the polling station id an additional parameter to the `save` function.

All of this to say:
- I'm not sure we should do this. Not because of how much effort it is right now. But because if in let's say 3 months we make another change to the fixture data that warrants moving ids around, either we have to put in the effort again, or we might as well not have done it this time. Let's not have that discussion in this PR.
- If we decide to do this, the effort and changes are too large to add them to this PR.

So I'll resolve this thread and start a Mattermost thread to collect some opinions on the merits of making the election ids and polling station ids nicely ordered.

backend/tests/shared/mod.rs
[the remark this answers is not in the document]
praseodym: I've found another workaround, namely [using `pub mod` to import the module](https://users.rust-lang.org/t/invalid-dead-code-warning-for-submodule-in-integration-test/80259). I think this is a bit more elegant than silencing the warning so I've changed to that in bacd8db86fb3811a8ccfbe54a2321c8acea81e46.

I also found that Rust has a [`reason` parameter for lint attributes](https://doc.rust-lang.org/reference/attributes/diagnostics.html#lint-reasons) that we can use for cases liket this in the future.

Lionqueen94 (approved):

jschuurk-kr (approved):

Issue #823, opened 2025-01-07
Numbers with leading zeroes are not numbers for polling station number


jschuurk-kr:
If you create or update a polling station and enter a "nummer" with leading zeroes, e.g. `001`, you get the error message "Dit is geen getal. Voer een getal in"

Firstly, it's common to have leading zeroes in a polling station number. So if we decide not to support leading zeroes, we need to inform our users clearly about this. The current error message does not do that.  
I do wonder if a better solution would not be to strip the leading zeroes instead of showing an error message.

Secondly, we should consider adding a test case to `frontend/lib/util/form.test.ts` for this behavior. The behavior is a consequence of `parseIntStrict()` which does have multiple tests to cover this. Question is if that's sufficient or if it would be helpful to add a single example test to `form.test.ts`.

Lionqueen94:
@jschuurk-kr Is there any rule in the EML regarding polling station formats or numbers and how does it work with duplication? Are number 1 and 001 considered the same number?

praseodym:
The polling station number is an integer, so `001` is saved as `1` (i.e. leading zeroes are stripped). This is also how this was designed and implemented on the data entry polling station choice page.

I agree that the behaviour on the polling station create/update page should be improved.

jorisleker:
@jschuurk-kr we've already checked how this number is stored in EML I think (at least I remember having the discussing about leading zeros). We ended up not supporting/showing leading zeros in the UI. 

My suggestion for a fix would be to just trim the leading zeros silently, and to not show an error message.

Issue #824, opened 2025-01-07
Backend fixtures and frontend disagree on value for number_of_voters of a polling station


jschuurk-kr:
The backend fixtures for polling stations (`backend/fixtures/polling_stations.sql`) have `NULL` as empty value for `number_of_voters`. If you create a polling station through the front-end and leave "Aantal kiesgerechtigden" empty, `number_of_voters` in the database contains `0`.

We should decide what the correct value is for this case and then use it consistently.

Lionqueen94:
My vote is to have the NULL value in case nothing was entered. That's the reason the database field and form field is optional.

PR #832, opened 2025-01-08
823 Polling station number should allow leading zeros

oliver3:
Closes https://github.com/kiesraad/abacus/issues/823

- `FormFieldNumber` now has a `isPollingStationNumber` boolean to use `parsePollingStationNumber()` instead of `parseIntStrict()`, which allows for leading zeros.
- Add test scenarios for `processForm()`, including commented out FIXME's that should be fixed in https://github.com/kiesraad/abacus/issues/824

### DIFF
--- a/frontend/lib/util/form.ts
+++ b/frontend/lib/util/form.ts
@@ -36,7 +36,7 @@ export type FormFieldNumber = FormFieldBase & {
   max?: number;
 };
 
-type FieldValue<F extends AnyFormField> = F extends FormFieldNumber
+export type FieldValue<F extends AnyFormField> = F extends FormFieldNumber
   ? number
   : F extends FormFieldString
     ? string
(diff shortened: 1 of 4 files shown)


Lionqueen94 (dismissed):

frontend/lib/util/form.ts
```diff
@@ -67,9 +68,13 @@ export function processForm<RequestObject>(
 
     switch (field.type) {
       case "number": {
-        const parsedValue = field.isFormatted ? deformatNumber(value) : parseIntStrict(value);
-        //parseIntStrict is used in deformatNumber as well, the result is a number or undefined.
-        if (parsedValue === undefined) {
+        const parsedValue = field.isFormatted
+          ? deformatNumber(value)
+          : field.isPollingStationNumber
+            ? parsePollingStationNumber(value)
```
lkleuver, line 74: feels weird that a "generic" util function knows what a "Polling station" is. 
could we change this either to something like parseLeadingZeroNumber
or add a new FormField.type?

lkleuver (approved):

frontend/lib/util/form.ts
oliver3: Now using `parseIntUserInput()` for all FormFieldNumber fields, also see updated PR description.

Lionqueen94 (approved):

PR #835, opened 2025-01-09
Apportionment calculation for <19 seats

### LINKED ISSUE #772: Apportionment calculation for <19 seats
Apportionment calculation for <19 seats with first tests.

Test cases:
- [ ] met restzetels, maar zonder loting

Lionqueen94:
Closes #772 

- Apportionment calculation for <19 seats has been added.
- Added tests for the above.
- Fraction has been expanded with more impls and tests.


backend/src/apportionment/mod.rs
praseodym, line 1: The `seat_allocation` function is growing rather long. I propose we split off the whole seats calculation and two remaining seats calculations into their own functions. This would also allow these functions to be tested individually if that makes things easier.

praseodym, line 1: Also, there is a lot of repetition in creating the `ElectionSummary` for each test. I think now is a good time to refactor that into a helper function, e.g. one that takes the total number of votes and the total number of votes for a variable number of political groups as parameters.

Lionqueen94: I added a helper function to create the ElectionSummary for the tests in c49e229dc801a569bb586b8fe9f17d9e6b614e85

Lionqueen94: I have extracted the whole seats calculation and surpluses calculation. Shall I extract both while loops as well? There is a slight repetition with updating the rest_seats and reducing the number of remaining seats still, but other than that the logic is already extracted into `get_pg_number_with_largest_average` and `get_pg_number_with_largest_surplus`

Lionqueen94: I also extracted the rest seat allocation to a separate function in 2f30cb965482f55450b69318c3a3fb00bf5b99c8

backend/src/apportionment/mod.rs
```diff
@@ -1,199 +1,350 @@
...
+        assert_eq!(result, Ok(vec![8, 2, 2, 1, 1, 1, 0, 0]));
+    }
+
+    #[test]
+    #[traced_test]
+    fn test_seat_allocation_less_than_19_seats_with_drawing_of_lots_error() {
+        let totals = get_election_summary(vec![500, 140, 140, 140, 140, 140]);
+        let result = seat_allocation(15, &totals);
+        assert_eq!(result, Err(ApportionmentError::DrawingOfLotsNotImplemented));
+        assert!(logs_contain(
```
cikzh, line 327: Instead checking the logs via the `traced_test` dependency, can we use something like `unimplemented!("drawing of lots")` here and asserting the panic?

cikzh (changes requested):

backend/src/apportionment/mod.rs
```diff
@@ -1,199 +1,350 @@
...
+                    "Remaining seat assigned using greatest average system to pg_number: {}",
+                    pg_number
+                );
+            }
+        }
+    }
+    Ok(rest_seats)
+}
+
 /// Apportionment - work in progress!!
```
cikzh, line 198:
```suggestion
/// Apportionment
```

I'd say the whole application is work in progress, so no need to say so explicitly :). At least add a TODO for discoverability

backend/src/apportionment/mod.rs
```diff
@@ -1,199 +1,350 @@
...
+    whole_seats: &BTreeMap<u8, u64>,
+    rest_seats: &BTreeMap<u8, u64>,
+    remaining_seats: u64,
+) -> Result<u8, ApportionmentError> {
+    let mut averages = BTreeMap::<u8, Fraction>::new();
+    for pg in pg_votes.iter() {
+        let pg_total = Fraction::from_count(pg.total);
+        let pg_seats_new = whole_seats
+            .get(&pg.number)
+            .expect("Political group should have number of whole seats")
```
cikzh, line 35: I think it's a little harsh to crash on expects here. I suggest we convert the expects in this module to `ApportionmentError`s

backend/src/apportionment/mod.rs
```diff
@@ -1,199 +1,350 @@
...
+            + 1;
+        let pg_avg_votes = pg_total / Fraction::from_u64(pg_seats_new);
+        averages.insert(pg.number, pg_avg_votes);
+    }
+    debug!("Averages: {:?}", averages);
+
+    let (&pg_number, &max) = averages
+        .iter()
+        .max_by(|(_, a), (_, b)| a.cmp(b))
+        .expect("Maximum average should be found");
```
cikzh, line 46: Same as line 35, the `None` here is caused by `averages` being empty, so the error message should reflect that

backend/src/apportionment/mod.rs
```diff
@@ -1,199 +1,350 @@
...
+    let threshold = Fraction::new(3, 4) * *quota;
+    debug!("Threshold: {}", threshold);
+    let mut surpluses = BTreeMap::<u8, Fraction>::new();
+    for pg in pg_votes.iter() {
+        let pg_total_votes = Fraction::from_count(pg.total);
+        if pg_total_votes >= threshold {
+            let pg_whole_seats = Fraction::from_u64(
+                *whole_seats
+                    .get(&pg.number)
+                    .expect("Political group should have number of whole seats"),
```
cikzh, line 81: Same as line 35

backend/src/apportionment/mod.rs
```diff
@@ -1,199 +1,350 @@
...
+fn get_pg_number_with_largest_surplus(
+    surpluses: &BTreeMap<u8, Fraction>,
+    remaining_seats: u64,
+) -> Result<u8, ApportionmentError> {
+    debug!("Surpluses: {:?}", surpluses);
+
+    let (&pg_number, &max) = surpluses
+        .iter()
+        .max_by(|(_, a), (_, b)| a.cmp(b))
+        .expect("Maximum surplus should be found");
```
cikzh, line 101: Same as line 35, the `None` here is caused by `surpluses` being empty, so the error message should reflect that

cikzh (changes requested):

backend/src/apportionment/mod.rs
praseodym: We already have an error return value for this (`ApportionmentError::DrawingOfLotsNotImplemented`) and a panic would mean that we can't assert that

backend/src/apportionment/mod.rs
praseodym: This is an invariant, a political group should always have an entry in `whole_seats` (can be 0 seats)

backend/src/apportionment/mod.rs
Lionqueen94: Surpluses cannot be empty though, this function is only called when surpluses is not empty. So maybe this code should be changed with that in mind? Or we can add an extra check for empty surpluses in this function to allow for standalone testing.

backend/src/apportionment/mod.rs
cikzh: Just discussed this with @Lionqueen94, an invariant indeed.

backend/src/apportionment/mod.rs
Lionqueen94: After a call we have decided the current implementation is fine.

backend/src/apportionment/mod.rs
```diff
@@ -1,198 +1,338 @@
...
+                let pg_number = get_pg_number_with_largest_average(
+                    pg_votes,
+                    whole_seats,
+                    &rest_seats,
+                    remaining_seats,
+                )?;
+                // TODO: The formal description states that initially rest seats can only be
+                //  assigned to unique parties, but with even more remaining seats then multiple
+                //  seats can be assigned to the same party.
+                //  Should there be any limitation built in for this theoretical case?
```
praseodym, line 186: I think we should just implement this like written in the formal description. I'd split the code in three while loops:
1. `while remaining_seats > 0 && !surpluses.is_empty()` - largest surplus iteration
2. `while remaining_seats > 0 && !unique_pgs.is_empty()` - largest average iteration with at most one seat per political group (track political groups that got one seat by removing them from a set `unique_pgs`, or something like that)
3. `while remaining_seats > 0` - largest average until done

Lionqueen94: I think I implemented it correctly now, please have a look at my solution :)

backend/src/apportionment/mod.rs
```diff
@@ -1,198 +1,349 @@
...
+            info!("======================================================");
+            debug!("Remaining seats: {}", remaining_seats);
+            if !surpluses.is_empty() {
+                // assign remaining seat to the party with the largest surplus and
+                // remove that party and surplus from the list
+                let pg_number = get_pg_number_with_largest_surplus(&surpluses, remaining_seats)?;
+                *rest_seats.entry(pg_number).or_insert(0) += 1;
+                surpluses.remove(&pg_number);
+                remaining_seats -= 1;
+                // pg_number_last_remaining_seat = Some(pg_number);
```
praseodym, line 172: remove?

backend/src/apportionment/mod.rs
```diff
@@ -1,198 +1,349 @@
...
+                    } else {
+                        None
+                    },
+                )?;
+                *rest_seats.entry(pg_number).or_insert(0) += 1;
+                remaining_seats -= 1;
+                if !unique_pgs.is_empty() {
+                    unique_pgs.retain(|&pg_num| pg_num != pg_number);
+                }
+                // pg_number_last_remaining_seat = Some(pg_number);
```
praseodym, line 200: remove?

Lionqueen94: Removed in 09e18415f56a231ca736c533fc2f824dc62d25de

backend/src/apportionment/mod.rs
```diff
@@ -1,198 +1,349 @@
...
+            let pg_number = get_pg_number_with_largest_average(
+                pg_votes,
+                whole_seats,
+                &rest_seats,
+                remaining_seats,
+                None,
+            )?;
+            *rest_seats.entry(pg_number).or_insert(0) += 1;
+            remaining_seats -= 1;
+            // pg_number_last_remaining_seat = Some(pg_number);
```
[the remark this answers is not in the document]
Lionqueen94: Removed in 09e18415f56a231ca736c533fc2f824dc62d25de

praseodym (dismissed):

praseodym (approved):

cikzh (approved):

Issue #822, opened 2025-01-07
Firefox: Scrolling up on error does not always work for polling stations create/update


jschuurk-kr:
Steps:
1. Open the page for either creating or updating a polling station.
2. Make sure the "Nummer" and "Naam" field are filled. Do __not__ click "Opslaan en toevoegen".
3. Clear the "Nummer" field.
4. Click "Opslaan en toevoegen".
5. Page scrolls up to show error.
6. Fill in "Nummer" and clear "Naam".
7. Click "Opslaan en toevoegen".
8. Page does not scroll up to show error.
9. Click "Opslaan en toevoegen".
10. Page does scroll up to show error.

At step 8 the page should scroll up to show the error.

@Lionqueen94: This only happens in Firefox and there is another scenario in which this happens only in Firefox, see below:

> Seems to be a Firefox-only issue, since I can't reproduce on Chromium.
>
> I get this sometimes, but other times Firefox does scroll up, it seems very inconsistent.
> Edit: I think I only get it after adding not a number after getting an api-error (EntryNotUnique) so:
> 1. Enter an existing polling station number and submit
> 2. Enter a letter and submit
> 3. Only then Firefox doesn't scroll, re-submit and firefox scrolls and in all other orders of filling in and submitting it also scrolls.
> 
> @jorisleker @jschuurk-kr What to do about this, it's quite an unlikely order of events and only happening on Firefox. It does not seem like it can be fixed, even scrolling twice doesn't do anything on Firefox.

_Originally posted by @Lionqueen94 in https://github.com/kiesraad/abacus/issues/839#issuecomment-2602045507_
            
> I think running into a non-unique error is a very likely reason to add a non-numeric char. I.e: I try to create polling station `12`, discover that it already exists, and then try `12a` to circumvent the error. So don't agree it is an unlikely order of events. But: the bug does not result in a blocking situation, it only occurs on Firefox, and it is unlikely that a lot of polling stations will be created in Abacus. If we don't know how to fix the bug, I would suggest leaving it as is.

_Originally posted by @jorisleker in https://github.com/kiesraad/abacus/issues/839#issuecomment-2602376246_

jschuurk-kr:
Agree that it's a minor bug in an uncommon scenario. So am in favor of leaving the behavior as-is and closing this issue. It's not worth it to invest even more time in trying to find the cause of the behavior. Also, this might get fixed auto-magically in some future release of Firefox.

praseodym:
[...]

PR #839, opened 2025-01-10
Do not show server errors when there are client-side validation errors

### LINKED ISSUE #825: Error message of duplicate polling station number shows invalid value from input field
Steps:
1. Go to /elections/1/polling-stations
2. Click "Stembureau toevoegen"
3. Fill "Nummer" with "33" (already exists) and "Name" with "some string"
4. Click "Opslaan en toevoegen"
5. Error is shown: "Er bestaat al een stembureau met nummer 33."
6. Change the "Nummer" to "asd".
7. Click "Opslaan en toevoegen"
8. Error is shown: "Er bestaat al een stembureau met nummer asd." The "Nummer" field has a red outline and the error icon is shown, but the error message about the invalid input is not shown below the input box.

If you empty the "Nummer" field, you get the error message "Er bestaat al een stembureau met nummer -1."

At step 8, the error message about the existing polling station should disappear and the error message for the input field should be shown.

lkleuver:
The polling station form didn't render errors correctly. After a server error (duplicate polling station number) the server error persists when submitting again interfering with client-side error rendering.

I've introduced a "reset" in `useCrud.ts` to have a clean flow each submit.

resolves #825


oliver3 (approved):

jschuurk-kr:
When the error is shown that there already is a polling station with that number, the "Nummer" field does not have a red border or an error icon:
image

@jorisleker Should the "Nummer" field have a red border and icon in this case?

jschuurk-kr (changes requested):
If I perform the [reproduction steps in the issue](https://github.com/kiesraad/abacus/issues/825#issue-2772570165), the correct error message is shown "Dit is geen getal. Voer een getal in", but the page does not scroll up to show that error message after clicking "Opslaan en toevoegen".

Seems to be a Firefox-only issue, since I can't reproduce on Chromium.

jorisleker:
> When the error is shown that there already is a polling station with that number, the "Nummer" field does not have a red border or an error icon:
> 
> @jorisleker Should the "Nummer" field have a red border and icon in this case?

Good catch, yes it should.

Lionqueen94:
> > When the error is shown that there already is a polling station with that number, the "Nummer" field does not have a red border or an error icon:
> > @jorisleker Should the "Nummer" field have a red border and icon in this case?
> 
> Good catch, yes it should.

This is fixed in 910591d0905812937951b35c71ff056ce0adc529

Lionqueen94:
> If I perform the [reproduction steps in the issue](https://github.com/kiesraad/abacus/issues/825#issue-2772570165), the correct error message is shown "Dit is geen getal. Voer een getal in", but the page does not scroll up to show that error message after clicking "Opslaan en toevoegen".
> 
> Seems to be a Firefox-only issue, since I can't reproduce on Chromium.

I get this sometimes, but other times Firefox does scroll up, it seems very inconsistent.

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -99,7 +100,13 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
...
                 margin="mb-md-lg"
                 defaultValue={pollingStation?.number}
-                error={validationResult.number ? t(`form.errors.${validationResult.number}`) : undefined}
+                error={
+                  validationResult.number
+                    ? t(`form.errors.${validationResult.number}`)
+                    : requestState.status === "api-error" && requestState.error.reference === "EntryNotUnique"
+                      ? " "
+                      : undefined
+                }
```
praseodym, line 109: This is not very readable now, I'd suggest moving this logic to a variable instead.

jorisleker:
> @jorisleker @jschuurk-kr What to do about this, it's quite an unlikely order of events and only happening on Firefox. It does not seem like it can be fixed, even scrolling twice doesn't do anything on Firefox.

I think running into a non-unique error is a very likely reason to add a non-numeric char. I.e: I try to create polling station `12`, discover that it already exists, and then try `12a` to circumvent the error. So don't agree it is an unlikely order of events. But: the bug does not result in a blocking situation, it only occurs on Firefox, and it is unlikely that a lot of polling stations will be created in Abacus. If we don't know how to fix the bug, I would suggest leaving it as is.

frontend/app/component/form/polling_station/PollingStationForm.tsx
Lionqueen94: Done in daf275f5314fbc79f3421523c05f1d4c6fd82154

Lionqueen94:
There is a separate issue for the Firefox scrolling bug, shall we continue the conversation there: https://github.com/kiesraad/abacus/issues/822

jorisleker:
> > > When the error is shown that there already is a polling station with that number, the "Nummer" field does not have a red border or an error icon:
> > > @jorisleker Should the "Nummer" field have a red border and icon in this case?
> > 
> > 
> > Good catch, yes it should.
> 
> This is fixed in [910591d](https://github.com/kiesraad/abacus/commit/910591d0905812937951b35c71ff056ce0adc529). Edit: I think we should show the error as described in the Alert also under the field (even though this is ugly) because otherwise `aria-errormessage` points to an element in the DOM that does not exist. This won't pass a11y checks I presume. @jorisleker what do you think? Another option is to point the `aria-errormessage` to the alert, but this will require some refactoring.

Yes agreed. Let's make sure it's not too long, as we already show the full error message at the top of the page. How about 'Voer een uniek nummer in'?

Lionqueen94:
> > > > When the error is shown that there already is a polling station with that number, the "Nummer" field does not have a red border or an error icon:
> > > > @jorisleker Should the "Nummer" field have a red border and icon in this case?
> > > 
> > > 
> > > Good catch, yes it should.
> > 
> > 
> > This is fixed in [910591d](https://github.com/kiesraad/abacus/commit/910591d0905812937951b35c71ff056ce0adc529). Edit: I think we should show the error as described in the Alert also under the field (even though this is ugly) because otherwise `aria-errormessage` points to an element in the DOM that does not exist. This won't pass a11y checks I presume. @jorisleker what do you think? Another option is to point the `aria-errormessage` to the alert, but this will require some refactoring.
> 
> Yes agreed. Let's make sure it's not too long, as we already show the full error message at the top of the page. How about 'Voer een uniek nummer in'?

Added in 68dd1f054e50507fedc66c930bef485406b2956d

Lionqueen94:
> I don't like the solution of adding `reset()` to `useCrud`, I think we can use `isValid` from `useForm` and hide server errors when `!isValid`.

Changed in 3ea5bb4c37be7c4f1df6a08bc7505773e1cd117a is this what you meant?

frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -67,9 +67,15 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
     }
   };
 
+  const numberFieldError = validationResult.number
+    ? t(`form.errors.${validationResult.number}`)
+    : isValid && requestState.status === "api-error" && requestState.error.reference === "EntryNotUnique"
+      ? t("polling_station.form.not_unique.error")
+      : undefined;
```
praseodym, line 74:
```suggestion
  let numberFieldError;
  if (validationResult.number) {
    numberFieldError = t(`form.errors.${validationResult.number}`);
  } else if (isValid && requestState.status === "api-error" && requestState.error.reference === "EntryNotUnique") {
    numberFieldError = t("polling_station.form.not_unique.error");
  }
```

praseodym (approved):

praseodym (approved):

oliver3 (approved):

jschuurk-kr (approved):

PR #852, opened 2025-01-13
Build on ubuntu-latest with musl

marlonbaeten:
Compiling for a specific libc version makes abacus less portable. The changes the target on ubuntu-latest to musl.


.github/workflows/build-lint-test.yml
```diff
@@ -19,8 +19,17 @@ jobs:
     name: Backend
```
praseodym, line 19:
```suggestion
    name: Backend (${{ matrix.os }})
```

praseodym:
Should we also pass `--target=${{ matrix.target.name }}` to other `cargo` commands so that we're not compiling and testing for one target and then recompiling to build the binary for another target?

marlonbaeten:
> Should we also pass `--target=${{ matrix.target.name }}` to other `cargo` commands so that we're not compiling and testing for one target and then recompiling to build the binary for another target?

Sure - only for `fmt` it would be weird to specify a target.

praseodym (approved):

PR #859, opened 2025-01-14
Remove sharding from Playwright d2d tests

praseodym:
Remove sharding from Playwright d2d tests so that we need fewer CI jobs.

This PR: Total duration [16m 46s](https://github.com/kiesraad/abacus/actions/runs/12766849827/usage)
`main`: Total duration [15m 1s](https://github.com/kiesraad/abacus/actions/runs/12765169639/usage)

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -136,14 +136,12 @@ jobs:
         run: npm run e2e:lib-ui
 
   playwright-d2d:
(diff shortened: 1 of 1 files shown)


jschuurk-kr:
Am in favor of removing the sharding. Comparing the durations (both of the pipeline and of the individual d2d jobs) the gains are very small. If in the future due to having more d2d tests, it becomes interesting to use sharding again, it's easy enough to enable it.

jschuurk-kr (approved):

PR #861, opened 2025-01-14
Upgrade to Node.js 22 LTS

praseodym:
Node.js 20 is out of active support (only security support), so upgrade to Node.js 22 LTS.

https://nodejs.org/en/about/previous-releases


oliver3:
We should also update the following in `package.json`

```json
"engines": {
    "node": ">=20.11.1"
},
```

jschuurk-kr (dismissed):

oliver3:
The `/docker-compose.yml` uses `image: node:20`

oliver3 (approved):

praseodym:
> We should also update the following in `package.json`
> 
> ```json
> "engines": {
>     "node": ">=20.11.1"
> },
> ```

> The `/docker-compose.yml` uses `image: node:20`

Both are done!

oliver3 (dismissed):

jschuurk-kr (dismissed):

praseodym:
Somehow the Playwright d2d tests on Windows fail now, but without a clear error message. I'll configure Playwright to pipe the stdout/stderr of the backend it starts (`webServer`) which will hopefully give some insight as to what is happening.

praseodym:
> Somehow the Playwright d2d tests on Windows fail now, but without a clear error message. I'll configure Playwright to pipe the stdout/stderr of the backend it starts (`webServer`) which will hopefully give some insight as to what is happening.

No idea what's happening, it might be easiest to reproduce this on a Windows VM and then debug it. Not high priority though.

jschuurk-kr:
> > Somehow the Playwright d2d tests on Windows fail now, but without a clear error message. I'll configure Playwright to pipe the stdout/stderr of the backend it starts (`webServer`) which will hopefully give some insight as to what is happening.
> 
> No idea what's happening, it might be easiest to reproduce this on a Windows VM and then debug it. Not high priority though.

Did a quick search on possible reasones we're getting a ECONNREFUSED on Windows. There's two, but neither seem likely:
1. The server isn't accepting connections yet. Should not be the case, because we have `port` in our Playwright config, meaning Playwright waits for that port to accept connections.
2. IPv6 vs IPv4. Node prefers IPv6 over IPv4 since NodeJS17. Should not be the case, because we're going from 20 to 22. There was a [Playwright bug](https://github.com/microsoft/playwright/issues/20784) related to this, reported by someone using NodeJS 18 on GH Actions. That bug was fixed about a year ago.

praseodym:
We can try configuring Playwright to connect to `127.0.0.1:8081` (IPv4) instead of `localhost:8081` (resolves to both IPv4 and IPv6 loopback address).

jschuurk-kr:
> We can try configuring Playwright to connect to `127.0.0.1:8081` (IPv4) instead of `localhost:8081` (resolves to both IPv4 and IPv6 loopback address).

made the change in https://github.com/kiesraad/abacus/pull/861/commits/e303017ee0e0881f94ffd7b3cf30386677c5ef7c and looks like it fixed it
image

Lionqueen94 (approved):

oliver3 (approved):

PR #877, opened 2025-01-21
Fix differences when converting json to po and back

marlonbaeten:
Covering translations from json to po and back resulted in a diff - not anymore.

Test by running:

```
cd frontend
npm run gen:po
npm run gen:translation-json
```

And check for any changes.

See issue #876


frontend/lib/i18n/locales/nl/form.json
praseodym, line 1: Looks like this file can be removed

cikzh (approved):

rnijveld (approved):

PR #881, opened 2025-01-21
No default handlers used in tests

### LINKED ISSUE #829: Vitest tests only use override request handlers and no default request handlers
Scope:
- remove `...handlers` from `setupServer` call in frontend/lib/test/server.ts
- fix any broken vitest tests
- use TypeScript OpenAPI types for all test data (const type, return type, `satisfies`)
- create folder for mock data that can be shared across all tests (vitest and playwright)
- move frontend/lib/test out of lib folder because dev dependency not app dependency

Choices:
- functions or objects for mock data

oliver3:
Resolves #829 
- Remove the default handlers from the msw server used in tests
- Replace these with one error response for all endpoints, containing the API call for easy test debugging
-  Show unhandled endpoints in test failures by showing the error response above, and a minimal ErrorBoundary
- Add specific response handlers to individual tests, keep using overrideOnce for one-time overrides for specific scenarios
- Remove pingHandler and test

All tests should still work with specified handlers

Frontend preview should work with default handlers


frontend/lib/test/server.ts
```diff
@@ -10,9 +10,13 @@
 import { delay, http, HttpResponse, JsonBodyType } from "msw";
 import { setupServer } from "msw/node";
 
-import { handlers } from "@kiesraad/api-mocks";
-
-export const server = setupServer(...handlers);
+export const server = setupServer(
+  // Tests should mock all requests explicitly, return http 500 with the url for the missing ones.
```
praseodym, line 14: Why this handler instead of the `onUnhandledRequest: "error"` that was there?

frontend/lib/api-mocks/RequestHandlers.ts
```diff
@@ -20,25 +20,6 @@ type ParamsToString<T> = {
...
-type PingParams = Record<string, never>;
-type PingRequestBody = {
-  ping: string;
-};
-type PingResponseBody = {
-  pong: string;
-};
-
-// ping handler for testing
-const pingHandler = http.post<PingParams, PingRequestBody, PingResponseBody>("/ping", async ({ request }) => {
```
praseodym, line 32: This endpoint is also used by the `MockTest` component on the `DevHomePage`. That component should also be removed if we decide that it is no longer useful.

praseodym (changes requested):

frontend/lib/test/server.ts
oliver3: I could not get it to work previously with the `onUnhandledRequest: "error"`. Now that I try again I get useful error messages, so that probably had more to do with the ErrorBoundary than this handler.

I will revert it.

frontend/lib/api-mocks/RequestHandlers.ts
oliver3: Ah missed that one. I will revert it.

praseodym: I'm ok with removing the ping endpoint as long as the `MockTest` component is also removed.

frontend/lib/test/server.ts
oliver3: Done in 943393ad3122d556087cbf28066b4d7aba33596a

frontend/lib/api-mocks/RequestHandlers.ts
oliver3: Reverted for now in 4543955a1a81e363d2a022b9bdb8f282f3c14acd

jschuurk-kr (approved):

praseodym (approved):

PR #858, opened 2025-01-14
Call login API endpoint

marlonbaeten:
Fixes #710 and #711

The login form now actually calls the login API endpoint.
Also added development endpoints to create a user or to login (without any credentials).

To Test:

```sh
# run the backend
cd backend
sqlx database reset
cargo run -- --reset-database --seed-data

# run the frontend
cd frontend
npm run dev:server
```

Or visit: https://3dd96737538f666fcb09a792a1e760435de0414b.abacus-test.nl/

Go to `/account/login` and login using "user" and "password".

PR #882, opened 2025-01-21
Add key to list items on DevHomePage

oliver3:
List elements in a map should have a key to identify them


Lionqueen94 (approved):

praseodym (approved):
@marlonbaeten also fixed this in #858, but good to have a separate PR

oliver3:
> @marlonbaeten also fixed this in #858, but good to have a separate PR

Oh did not see that.. Glad we found the same solution though :D

PR #885, opened 2025-01-22
Update to Vitest 3

oliver3:
Get all tests working again after updating to vitest 3
- [x] more strict Error comparison
- [x] mocks not being cleaned up in between tests?


praseodym:
> * mocks not being cleaned up in between tests?

In [`setup.ts`](https://github.com/kiesraad/abacus/blob/8fa32f560f442323d19ee3cc90247c7dde4ebe70/frontend/lib/test/setup.ts#L39) we currently use [`vi.clearAllMocks()`](https://vitest.dev/api/vi#vi-clearallmocks). Maybe this should instead be [`vi.resetAllMocks()`](https://vitest.dev/api/mock#mockreset) or the [`mockReset`](https://vitest.dev/config/#mockreset) setting.

praseodym (approved):

PR #896, opened 2025-01-23
adds check that polling station number is larger than 0

### LINKED ISSUE #883: Do not allow polling station numbers less than or equal to zero
Currently the "Nummer" field for polling stations accepts both 0 and negative numbers. This should not be the case.

Image

---

Can be fixed by adding `min: 1` to `number: { required: true, type: "number" },`

and changing:
```TypeScript
  let numberFieldError;
  if (validationResult.number) {
    numberFieldError = t(`form_errors.${validationResult.number}`);
  } else if (isValid && requestState.status === "api-error" && requestState.error.reference === "EntryNotUnique") {
    numberFieldError = t("polling_station.form.not_unique.error");
  }
```
based on what we decide the error message should be (see comment below).

jschuurk-kr:
It was possible to create polling stations with a "nummer" that was `0` or negative. This PR adds validation to the front-end to stop that from happening:
image

Note that I did have to override the `validationResult.number` from `form.ts` in `PollingStationForm.tsx`. Otherwise the error message would be "Waarde is te laag" instead of "Dit is geen getal. Voer een getal in".

closes #883

### DIFF
--- a/frontend/app/component/form/polling_station/PollingStationForm.tsx
+++ b/frontend/app/component/form/polling_station/PollingStationForm.tsx
@@ -3,7 +3,7 @@ import * as React from "react";
 import { isSuccess, PollingStation, PollingStationRequest, useCrud } from "@kiesraad/api";
 import { t } from "@kiesraad/i18n";
 import { Alert, Button, ChoiceList, Form, FormLayout, InputField } from "@kiesraad/ui";
-import { FormFields, useForm } from "@kiesraad/util";
+import { FormFields, useForm, ValidationError } from "@kiesraad/util";
 
 export interface PollingStationFormProps {
   electionId: number;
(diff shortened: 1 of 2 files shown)


frontend/app/component/form/polling_station/PollingStationForm.tsx
```diff
@@ -69,6 +69,10 @@ export function PollingStationForm({ electionId, pollingStation, onSaved, onCanc
 
   let numberFieldError;
   if (validationResult.number) {
+    validationResult.number =
+      validationResult.number === "FORM_VALIDATION_RESULT_MIN"
+        ? "FORM_VALIDATION_RESULT_INVALID_NUMBER"
+        : validationResult.number;
```
praseodym, line 75: I would introduce a new variable instead of changing the `validationResult.number` variable that is part of the `useForm` hook.

Lionqueen94: And then we can even use the custom message Joris mentioned before for that.

jschuurk-kr: fixed in  bd4595ef5adae80f9a8fe4872395bdbce6568ede

Lionqueen94 (approved):

praseodym (approved):

PR #904, opened 2025-01-24
Use relative paths for static assets

### LINKED ISSUE #837: Fix static assets in Ladle
Ladle currently has no access to the static assets used by the frontend via Vite.
This currently breaks font loading (and other static assets).

Ladle has its own vite.config and this needs to point to the correct "publicDir", which is /frontend/static
but the paths dont line up directly so something needs to be figured out.

praseodym:
This fixes static assets in Ladle and also works in the regular build. It also removes the unused asset `hex.svg`.

To test: the Ladle build should have the correct fonts rendered (e.g. http://localhost:61000/ladle/?story=badge--all-badges, http://ladle-static-assets.kiesraad-abacus.pages.dev/ladle/?story=badge--all-badges) and the regular build should not show any changes.

Resolves #837

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <title>Abacus</title>
-    <link rel="icon" href="/static/favicon.svg" />
+    <link rel="icon" href="./static/favicon.svg" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


jschuurk-kr:
Published ladle preview on `main`:
image

Published ladle preview of this branch:
image

jschuurk-kr (approved):
Checked some ladle and some frontend pages. The different fonts are displayed.

Lionqueen94 (approved):

PR #909, opened 2025-01-27
Rename backend to abacus

rnijveld:
- This renames the backend crate from `backend` to `abacus`, which is a better name for what our crate actually is.
- This allows filtering log messages on `RUST_LOG=abacus=trace` to only show trace messages for abacus (in contrast to `RUST_LOG=trace` which would show trace level log messages for all crates we rely on as well), instead of having to use the somewhat unintuitive `RUST_LOG=backend=trace`.
- No changes in the frontend should be needed for this change. Almost everything stays the way it was, just some updated imports.

### DIFF
--- a/backend/src/bin/abacus.rs
+++ b/backend/src/bin/abacus.rs
@@ -1,7 +1,7 @@
-use axum::serve::ListenerExt;
 #[cfg(feature = "dev-database")]
-use backend::fixtures;
-use backend::router;
+use abacus::fixtures;
+use abacus::router;
+use axum::serve::ListenerExt;
 use clap::Parser;
 use sqlx::{sqlite::SqliteConnectOptions, SqlitePool};
 use std::{
--- a/backend/src/bin/abacus.rs
+++ b/backend/src/bin/abacus.rs
@@ -1,6 +1,5 @@
 #[cfg(feature = "dev-database")]
-use abacus::fixtures;
-use abacus::router;
+use abacus::{fixtures, router};
 use axum::serve::ListenerExt;
 use clap::Parser;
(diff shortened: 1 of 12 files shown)


backend/src/bin/abacus.rs
```diff
@@ -1,7 +1,7 @@
-use axum::serve::ListenerExt;
 #[cfg(feature = "dev-database")]
-use backend::fixtures;
-use backend::router;
+use abacus::fixtures;
+use abacus::router;
```
Lionqueen94, line 3:
```suggestion
use abacus::{fixtures, router};
```

Lionqueen94 (commented):
Can we also add `RUST_LOG=abacus=trace` to `.env` or should we all add it locally?

rnijveld:
> Can we also add `RUST_LOG=abacus=trace` to `.env` or should we all add it locally?

I don't think we really need to do that, you don't need trace level logging all the time. You probably just want to add that locally whenever you're debugging something specific. The docker compose config does have logging set to debug by default though, but that logging output can be hidden a little bit easier than running abacus on a local terminal.

praseodym (approved):

backend/src/bin/abacus.rs
```diff
@@ -1,7 +1,6 @@
-use axum::serve::ListenerExt;
 #[cfg(feature = "dev-database")]
-use backend::fixtures;
-use backend::router;
+use abacus::{fixtures, router};
```
marlonbaeten, line 2: The `router` import should not be based on a feature flag.

Lionqueen94 (approved):

PR #916, opened 2025-01-27
Move breadcrumbs/links logic into `NavBar`

### LINKED ISSUE #724: Move breadcrumbs logic into NavBar component
> > Maybe we can look into creating a new component for the `<NavBar>` breadcrumb to avoid duplicating code between e.g. `PollingStationCreatePage.tsx`, `PollingStationListPage.tsx` and `PollingStationUpdatePage.tsx`?
> 
> What would help is the option to extend the current content in the NavBar, then you can add the 'base' in the layout and add an extra crumb in detail pages for example. Something to think about for a future refactor I guess.

Extending it from within child component is not trivial. But we could make the `NavBar` component create the whole breadcrumb from the current path, so that we don't have to define or extend it on every page.

_Originally posted by @praseodym in https://github.com/kiesraad/abacus/pull/701#issuecomment-2523442048_

praseodym:
Resolves #724

- Clarify that `PollingStationFormNavigation.test.tsx` is using the vanilla React Testing Library (RTL) render function and not our custom one.
- Move logic to render navigation links into NavBar, and move the NavBar rendering up to a parent component where possible. The navigation links are rendered in a new `NavBarLinks` component with several subcomponents for readability.
- Add location as an optional parameter to NavBar.
- Add test and Ladle stories for NavBar.

### DIFF
--- a/frontend/app/component/navbar/NavBarLinks.tsx
+++ b/frontend/app/component/navbar/NavBarLinks.tsx
@@ -1,0 +1,149 @@
+import { Link, useLocation } from "react-router";
+
+import { Election, useElection } from "@kiesraad/api";
+import { t } from "@kiesraad/i18n";
+import { IconChevronRight } from "@kiesraad/icon";
+
+function ElectionBreadcrumb({ election }: { election: Election }) {
+  return (
+    <>
+      <span className="bold">{election.location}</span>
+      <span>&mdash;</span>
+      <span>{election.name}</span>
+    </>
+  );
+}
+
+function DataEntryLinks() {
+  const location = useLocation();
+  const { election } = useElection();
+
+  return (
+    <>
+      <Link to={"/elections"}>{t("election.title.plural")}</Link>
+      <IconChevronRight />
+      {location.pathname.includes("/data-entry/") ? (
+        // Within the data entry, link back to the polling station choice page
+        <Link to={`/elections/${election.id}/data-entry`}>
+          <ElectionBreadcrumb election={election} />
+        </Link>
+      ) : (
+        // On the polling station choice page, display the election breadcrumb without linking
+        <span>
+          <ElectionBreadcrumb election={election} />
+        </span>
+      )}
+    </>
+  );
+}
+
+function ElectionManagementLinks() {
+  const location = useLocation();
+  const { election } = useElection();
+
+  // TODO: Add left side menu, #920
+
+  if (location.pathname.match(/^\/elections\/\d+\/?$/)) {
+    return (
+      <span>
+        <ElectionBreadcrumb election={election} />
+      </span>
+    );
+  } else {
+    return (
+      <>
+        <Link to={`/elections/${election.id}#administratorcoordinator`}>
+          <ElectionBreadcrumb election={election} />
+        </Link>
+        {location.pathname.match(/^\/elections\/\d+\/polling-stations\/\d+\//) && (
+          <>
+            <IconChevronRight />
+            <Link to={`/elections/${election.id}/polling-stations`}>{t("polling_stations")}</Link>
+          </>
+        )}
+      </>
+    );
+  }
+}
+
+function TopLevelManagementLinks() {
+  const location = useLocation();
+
+  const links = [];
+  if (location.pathname.startsWith("/elections")) {
+    links.push(
+      <span key="elections" className="active">
+        {t("election.title.plural")}
+      </span>,
+    );
+  } else {
+    links.push(
+      <Link key="elections-link" to={`/elections#administratorcoordinator`}>
+        {t("election.title.plural")}
+      </Link>,
+    );
+  }
+  if (location.pathname.startsWith("/users")) {
+    links.push(
+      <span key="users" className="active">
+        {t("users")}
+      </span>,
+    );
+  } else {
+    links.push(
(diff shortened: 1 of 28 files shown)


jschuurk-kr:
>     * How extensive do we want to test this component?

That will depend largely on if we keep the current structure with all the `if`s.

Looking at this part, does make me want to have extensive tests:
```TypeScript
if (location.pathname.startsWith("/elections/")) {
<snipped>
} else if (isAdministrator) {
    if (location.pathname.startsWith("/elections")) {
```
There's no obvious relation between the `if` and the `else if` condition. And the `if` in the `else if` is the same condition as the top `if`. That's going to be hard to maintain without tests for all the branches.

frontend/app/component/navbar/NavBarLinks.tsx
```diff
@@ -0,0 +1,145 @@
...
+        {t("logs")}
+      </span>,
+    );
+  } else {
+    links.push(
+      <Link key="logs-link" to={`/logs#administratorcoordinator`}>
+        {t("logs")}
+      </Link>,
+    );
+  }
```
Lionqueen94, line 122: Something goes wrong here that causes the `Werkplekken` and `Activiteitenlog` navlink to not be marked as active when the page is open and the navlink is still clickable.
image

praseodym: I see! I did a last minute change to `window.location` as the default location value, which causes this issue. I used `useLocation()` before, which triggers a rerender in this case, but using it as a default value was not allowed by our linting rules. 

Maybe the best fix is to remove the default value and pass `useLocation()` as a value where the `NavBar` is used.

frontend/app/component/navbar/NavBarLinks.tsx
```diff
@@ -0,0 +1,145 @@
...
+          </>
+        )}
+      </>
+    );
+  }
+}
+
+function TopLevelManagementLinks({ location }: NavBarLinksProps) {
+  const links = [];
+  if (location.pathname.startsWith("/elections")) {
```
lkleuver, line 71: Wouldn't [navlink](https://reactrouter.com/start/library/navigating#navlink) be sufficient here over if statements?

praseodym: Yes I think so, I didn't know about that. Will update, thanks!

frontend/app/component/navbar/NavBarLinks.tsx
praseodym: Fixed in 3a8770175c8a1a2bdba0560160ccf4ecbe1a08b3. Unfortunately I didn't manage to write a regression test for this because of how tricky React Router tests can be.

frontend/app/component/navbar/NavBarLinks.tsx
praseodym: Changed in 133c0a2b66943b7e38f95b871de371230287c356.

frontend/app/component/navbar/NavBarLinks.tsx
praseodym: Good catch, fixed in aa0787bf9cce0838851608068dadfb23f6ffe53d.

frontend/app/module/election/ElectionLayout.tsx
```diff
@@ -1,14 +1,18 @@
-import { Outlet } from "react-router";
+import { Outlet, useLocation } from "react-router";
 
 import { ElectionProvider, ElectionStatusProvider } from "@kiesraad/api";
 import { useNumericParam } from "@kiesraad/util";
 
+import { NavBar } from "../../component/navbar/NavBar";
```
Lionqueen94, line 6:
```suggestion
import { NavBar } from "app/component/navbar/NavBar";
```

frontend/app/module/election/page/OverviewPage.tsx
```diff
@@ -1,24 +1,25 @@
...
 
 import { ElectionStatusWithIcon } from "app/component/election/ElectionStatusWithIcon";
 import { Footer } from "app/component/footer/Footer";
-import { NavBar } from "app/component/navbar/NavBar";
 
 import { Election, useElectionList } from "@kiesraad/api";
 import { t } from "@kiesraad/i18n";
 import { Alert, PageTitle, Table, WorkStationNumber } from "@kiesraad/ui";
 
+import { NavBar } from "../../../component/navbar/NavBar";
```
Lionqueen94, line 10:
```suggestion
import { NavBar } from "app/component/navbar/NavBar";
```

Lionqueen94:
It might be good for @jorisleker to check the navbar on all the pages to confirm everything is (still) as it should be?

frontend/app/module/election/page/OverviewPage.tsx
praseodym: Changed in e6c94a4104467ee53c67de568d678f69874adabd

frontend/app/module/election/ElectionLayout.tsx
praseodym: Changed in e6c94a4104467ee53c67de568d678f69874adabd

Lionqueen94 (approved):

oliver3 (approved):

jorisleker (approved):

PR #918, opened 2025-01-27
Frontend build without dev dependencies

### LINKED ISSUE #908: Frontend build without dev dependencies
Goal: create frontend builds without npm dev dependencies (`npm ci --omit=dev`) in GitHub Actions. Reducing the number of installed dependencies to create build will reduce the attack surface during those builds.

We should create this build in a separate workflow, upload it as an Actions artifact and use it in other workflows where we need a frontend build.

In a later stage, we can consider minimizing the number of dependencies required by e.g. using [esbuild](https://esbuild.github.io) or [Rolldown](https://rolldown.rs).

praseodym:
Resolves #908

- Move all dependencies required to build frontend to npm production  dependencies.
- Change MSW import so that it is only imported when enabled.
- Change GitHub Action jobs to only install production dependencies by setting the `NODE_ENV` environment variable to `production`.

On Linux, the frontend now builds using 67 instead of 787 npm packages.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -9,11 +9,6 @@ on:
     branches:
       - main
 
-env:
-  CARGO_TERM_COLOR: always
-  ASSET_DIR: ${{ github.workspace }}/frontend/dist
-  SQLX_OFFLINE: "true"
-
 jobs:
   backend:
     name: Backend (${{ matrix.target.os }})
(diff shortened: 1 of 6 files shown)


Lionqueen94 (dismissed):

marlonbaeten (commented):
Two questions:

- Is plugin-react-swc actually used for production builds? They seem to suggest it's only for fast development rebuilds.
- Could we get rid of the cross-env dependency if we configure a default value for API_MODE?

praseodym:
> Two questions:
> 
> * Is plugin-react-swc actually used for production builds? They seem to suggest it's only for fast development rebuilds.\

I don't think it is really used for production build, but because it is imported by `vite.config.ts` it is required to run Vite.

> * Could we get rid of the cross-env dependency if we configure a default value for API_MODE?

Yes! Done in 13b8a3861d1d706ff3eeaedb1fcfefc484768851.

Lionqueen94 (approved):

oliver3 (approved):

PR #921, opened 2025-01-28
Added API endpoint for apportionment

Lionqueen94:
## Description
- Added apportionment endpoint for election returning apportionment and election summary
- Added apportionment integration tests
- Added extra party to election fixtures 2 and 3 to allow more realistic scenarios
- Added extra shared helper functions to add a `data_entry` parameter for easier testing
- Added extra `election_id` parameter to shared `create_result` function to be able to use it for different elections
- Updated Fraction to allow for correct deserialization
- Updated `response.status()` testing to be the same in all tests (asserting with equality to `StatusCode::`)


backend/src/apportionment/api.rs
```diff
@@ -0,0 +1,54 @@
...
+    State(elections_repo): State<Elections>,
+    State(polling_stations_repo): State<PollingStations>,
+    State(polling_station_results_entries_repo): State<PollingStationResultsEntries>,
+    Path(id): Path<u32>,
+) -> Result<Json<ElectionApportionmentResponse>, APIError> {
+    let election = elections_repo.get(id).await?;
+    let results = polling_station_results_entries_repo
+        .list_with_polling_stations(polling_stations_repo, election.id)
+        .await?;
+    let election_summary = ElectionSummary::from_results(&election, &results)?;
```
praseodym, line 48: We should (not in this PR) create a new function, e.g. in the summary module, to get the results for an election id. That would limit the number of imports in this file.

backend/src/apportionment/mod.rs
praseodym, line 1: Just wondering: why are all doc comments rewrapped? Did RustRover do this?

backend/tests/utils/mod.rs
```diff
@@ -1,17 +1,18 @@
...
-use sqlx::SqlitePool;
 
 pub async fn serve_api(pool: SqlitePool) -> SocketAddr {
     let app = router(pool).unwrap();
     let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
     let addr = listener.local_addr().unwrap();
-    tokio::spawn(async move {
-        axum::serve(listener, app).await.unwrap();
+    spawn(async move {
+        serve(listener, app).await.unwrap();
```
praseodym, line 15: IMO removing the `tokio::` and `axum::` prefixes makes this code less clear.

backend/tests/shared/mod.rs
```diff
@@ -88,26 +107,31 @@ pub async fn create_and_finalise_data_entry(
...
 
-pub async fn create_result(addr: &SocketAddr, polling_station_id: u32) {
-    create_and_finalise_data_entry(addr, polling_station_id, 1).await;
-    create_and_finalise_data_entry(addr, polling_station_id, 2).await;
+pub async fn create_result(
+    addr: &SocketAddr,
+    polling_station_id: u32,
+    election_id: u32,
+    data_entry: Option<DataEntry>,
+) {
```
praseodym, line 127: Adding a new function (e.g. `create_result_with_data_entry`) would have been a bit cleaner (no changes needed to tests calling this function). Same for `create_and_save_data_entry`.

praseodym (approved):
lgtm, just a few questions as comments.

backend/src/apportionment/mod.rs
Lionqueen94: No I reduced some comments from 3 to 2 lines for example, since they could still be in the max line width. This causes changes in openapi.json as well

backend/tests/shared/mod.rs
Lionqueen94: Fair point, I'll change that

backend/src/apportionment/mod.rs
praseodym: I think we're using the `rustfmt` default so the max line width would be 80.

Lionqueen94: Ah my line width is set to 100 so I used that as a guide.

backend/tests/utils/mod.rs
Lionqueen94: Re-added in 1a9227c5add7a375eaaea41c709234973d45904b

praseodym (approved):

backend/fixtures/election_2.sql
```diff
@@ -1,6 +1,6 @@
 INSERT INTO elections (id, name, location, number_of_voters, category, number_of_seats, election_date, nomination_date,
                        status, political_groups)
-VALUES (2, 'Municipal Election', 'Heemdamseburg', 100, 'Municipal', 29, '2024-11-30', '2024-11-1',
+VALUES (2, 'Municipal Election', 'Heemdamseburg', 100, 'Municipal', 15, '2024-11-30', '2024-11-1',
```
cikzh, line 3: Maybe a comment prevents a future us changing this value back to something greater than 19
```suggestion
-- number_of_seats explicitly set to a value less than 19, to be used in elections with less than 19 seats
VALUES (2, 'Municipal Election', 'Heemdamseburg', 100, 'Municipal', 15, '2024-11-30', '2024-11-1',
```

backend/src/apportionment/api.rs
```diff
@@ -0,0 +1,54 @@
+use axum::{
```
cikzh, line 1: I like this separate file for endpoints. I think we should do this for all entities (not in this PR of course :))

backend/src/election/mod.rs
```diff
@@ -1,5 +1,5 @@
 #[cfg(feature = "dev-database")]
-use axum::http::StatusCode;
+use axum::http::StatusCode; // this import needs to be separate because it's only needed for dev-database
```
cikzh, line 2: The comment is just a statement of what happens here. I'd remove it or explain _why_ it's needed for dev-database

backend/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(feature = "memory-serve")]
+use axum::http::StatusCode; // this import needs to be separate because it's only needed for memory-serve
```
cikzh, line 2: Same for this comment

backend/tests/election_integration_test.rs
```diff
@@ -1,12 +1,13 @@
 #![cfg(test)]
 
+use axum::http::StatusCode;
+use sqlx::SqlitePool;
+use test_log::test;
+
 use crate::{shared::create_result, utils::serve_api};
 #[cfg(feature = "dev-database")]
-use abacus::election::Election;
+use abacus::election::Election; // this import needs to be separate because it's only needed for dev-database
```
cikzh, line 9: This one as well

cikzh (changes requested):
Great work!

Is see that a lot of fully qualified names are moved to `use` imports, sometimes put behind a feature flag. Personally I think there is nothing wrong with using the fully qualified name in the code. Sometimes it help with ubiquitous terms like `Client`. Readability obviously shouldn't suffer though.

backend/src/apportionment/api.rs
Lionqueen94: Agreed!

backend/src/election/mod.rs
Lionqueen94: Updated in 0680eda3585e02d257a9ab3d55fe9b48abbae2f1

backend/tests/election_integration_test.rs
Lionqueen94: Changed in 0680eda3585e02d257a9ab3d55fe9b48abbae2f1

backend/src/lib.rs
cikzh: `memory-serve` is a dependency (created by @marlonbaeten btw) that serves frontend assets from memory. We don't always need the backend to do that, for example in development, when we use `vite` to serve the frontend assets.

praseodym (approved):

cikzh (approved):

PR #937, opened 2025-01-30
Updated surplus system to take political groups with surplus of 0 into account

### LINKED ISSUE #925: Update surplus system to take political groups with surplus of 0 into account
If a political group has 75% or more votes of the quota (so meets the threshold) and its surplus is zero, the group is eligible for a remaining seat.

In the current implemenation that's not the case, so the code and tests need to be updated.

Lionqueen94:
Fixed #925 

Adjusted and added tests for apportionment without remainder seats


praseodym (approved):
lgtm but I did not verify the test cases

lkleuver (approved):

PR #940, opened 2025-01-31
Added basic frontend for apportionment

### LINKED ISSUE #739: Basic frontend for apportionment
Show results and intermediate steps in frontend, see [draft design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=5452-56168&node-type=canvas))

Lionqueen94:
Closes #739 

## TODO
- [x] Add frontend tests
- [x] Update frontend based on Joris' comments and design updates
  - [x] Kunnen we voorkomen dat 'aantal zetels' bij 100% zoom over twee regels gaat lopen? ==> _should be ok now_
  - [x] Hoe ziet de kolom met de gemiddelden eruit als er 10.000 149/150 stemmen per zetel zijn? ==> _increased width so it fits_
  - [x] Wat gebeurt er met de tabel als er meer dan 1 restzetel te verdelen is? (in design wordt 'ie breder dan de container/horizontale viewport) ==> _added scroll_
  - [x] Kunnen we de top/bottom padding van de cellen zo aanpassen dat een lange lijstnaam niet leidt tot het verhogen van de rij (hebben we in inputgrid ook gedaan) ==> _done_
  - [x] List number columns should be right-aligned ==> _all right aligns correct now_
  - [x] Can we prevent the 'aantal restzetels' label from wrapping over 2 lines? ==> _should be ok now_
  - [x] DisplayFraction:
    - [x] When the numerator of a fraction is 0, we should not show the fraction because it evaluates to 0. ==> _done_
    - [x] If the average is 0, we should show the wholes as '0' I think? ==> _done_
    - [x] Fraction alignment (integer parts are all aligned to each other in the design whether they have a fraction part or not) ==> _done_
  - [x] In the design, the label 'Restzetel toegekend aan lijst' label is right-aligned. You made it left-aligned. Not sure what I like better though. ==> _all as designed now_

### Apportionment
- Added apportionment context, provider and hook
- Added DisplayFraction ui element, story and tests
- Added `elections/{election_id}/apportionment` route
- Added apportionment page with election summary and apportionment tables
- Added whole seat detail page
- Added residual seat detail page
- Add Navbar for apportionment pages (add `> Zetelverdeling` for detail pages)
- Added error when election data entry has not been finalised
- Fractions in tables are aligned as designed and horizontal scroll is added to table where needed, see:

https://github.com/user-attachments/assets/6b15289d-988c-451c-a855-83003f01b15e

### Miscellaneous
- Added css util class `font-number` setting `font-family: "GeistMono", monospace, monospace;` and `font-feature-settings: "ss09" on;`
  - Removed `font-feature-settings: "ss09" on;` from `@font-face` because this is not supported by chrome and edge

## Not in scope
- Candidate selection ("Gekozen kandidaten") including political group candidate detail pages
- Preference threshold ("Voorkeursdrempel")


backend/src/apportionment/api.rs
```diff
@@ -37,18 +40,31 @@ pub struct ElectionApportionmentResponse {
...
-        apportionment,
-        election_summary,
-    }))
+    let statuses = data_entry_repo.statuses(id).await?;
+    if statuses
+        .iter()
+        .filter(|data_entry_status| data_entry_status.status != DataEntryStatusName::Definitive)
+        .collect::<Vec<_>>()
+        .is_empty()
+    {
```
Lionqueen94, line 55: @praseodym @cikzh I came up with this if statement to check if all data entries for an election have been completed. Do you think this is a good solution? And if so, I think it's a good idea to make this into a function to be used in other endpoints as well.

praseodym: This is shorter: `if statuses.iter().all(|s| s.status == DataEntryStatusName::Definitive)`.

And good idea to create a function for this. I think that's something for a separate PR though.

jorisleker (changes requested):
## Data entry not ready
Wouldn't it be nicer to just render the zetelverdeling page with this message, without throwing an error? Technically, Abacus is not broken and does not need fixing. The data just isn't ready. Bit less alarming to the user as well. Also, a part of the message is in English instead of Dutch?
Scherm­afbeelding 2025-02-07 om 08 48 33

## Polling station choice page
Compared to `main`, the width of the polling station number column has changed. It should be narrower.
Scherm­afbeelding 2025-02-07 om 08 55 41

## Details whole seats
Scherm­afbeelding 2025-02-07 om 09 01 30

- List number column should be right-aligned 

## Details rest seats
Scherm­afbeelding 2025-02-07 om 09 03 09

- When the numerator of a fraction is 0, we should not show the fraction because it evaluates to 0. 
- If the average is 0, we should show the wholes as '0' I think?

lkleuver:
LGTM

only thing I remember from the past is that <td /> can lead to weird tables, so I would add Table.EmptyCell
but as long as it works :)

lkleuver (dismissed):

jorisleker (changes requested):
Nice work @Lionqueen94 

Just one small remark - in the _zetelverdeling_ table, I would expect the chevrons to have their own column, so the total of all seats (29 in the screenshot) aligns with the seats for each individual party. 
Scherm­afbeelding 2025-02-14 om 12 29 32

Lionqueen94:
> Nice work @Lionqueen94
> 
> Just one small remark - in the _zetelverdeling_ table, I would expect the chevrons to have their own column, so the total of all seats (29 in the screenshot) aligns with the seats for each individual party. Scherm­afbeelding 2025-02-14 om 12 29 32

I implemented this before and then undid those changes again, but I'll see if I can make an adjustment to the minimum width of that column and the alignment of the total, thanks for the review!

frontend/app/component/apportionment/Apportionment.module.css
```diff
@@ -0,0 +1,149 @@
...
+  width: 3.5rem;
+  max-width: 3.5rem;
+}
+
+.bt-1-gray {
+  border-top: 1px solid var(--bg-gray-darkest);
+}
+
+.scrollable {
+  width: 52rem;
```
lkleuver, line 22: Where does this 52rem come from? is this something unique to this page?
scrollable sounds like a global util.

lkleuver (commented):
do we want to be more strict in using CamelCase vs kebab-case?
Especially in component props this feels a bit out of sync

frontend/app/component/apportionment/Apportionment.module.css
Lionqueen94: Yes this is the set width of this page, I can split up the class into a width and scrollable

lkleuver: fine if its unique to this page, its a module right :)

Lionqueen94:
> do we want to be more strict in using CamelCase vs kebab-case? Especially in component props this feels a bit out of sync

I changed the props and variables that were snake_case to camelCase in f732f4451cbe8cb698b3d0cf0fc8f68e2931e2e5, thanks for pointing it out!

lkleuver (approved):

frontend/app/component/apportionment/WholeSeatsTable.tsx
```diff
@@ -1,23 +1,27 @@
...
 
-export function WholeSeatsTable({ finalStanding, quota }: WholeSeatsTableProps) {
+export function WholeSeatsTable({ finalStanding, politicalGroups, quota }: WholeSeatsTableProps) {
   return (
     <Table id="whole_seats_table" className={cn(cls.table, cls.wholeSeatsTable)}>
       <Table.Header>
         <Table.HeaderCell className="text-align-r">{t("list")}</Table.HeaderCell>
-        <Table.HeaderCell className="text-align-r">{t("vote_count")}</Table.HeaderCell>
+        <Table.HeaderCell>{t("list_name")}</Table.HeaderCell>
+        <Table.HeaderCell className="text-align-r w-full">{t("vote_count")}</Table.HeaderCell>
```
Lionqueen94, line 20: @oliver3 `w-full` should be on the list name column right?

frontend/lib/i18n/locales/nl/apportionment.json
```diff
@@ -0,0 +1,61 @@
...
+  "leftover_residual_seats_amount_and_information": {
+    "singular": "Hierna was er nog {num_seats} restzetel te verdelen. Deze zetel is toegewezen aan de lijst die met een zetel erbij het grootste gemiddelde aantal stemmen per zetel zouden hebben. Hierbij heeft iedere lijst niet meer dan één zetel toegewezen gekregen.",
+    "plural": "Hierna waren er nog {num_seats} restzetels te verdelen. Deze zetels zijn toegewezen aan de lijsten die met een zetel erbij het grootste gemiddelde aantal stemmen per zetel zouden hebben. Hierbij heeft iedere lijst niet meer dan één zetel toegewezen gekregen."
+  },
+  "minus": "min",
+  "not_available": "Zetelverdeling is nog niet beschikbaar",
+  "no_residual_seats_to_assign": "Er zijn geen restzetels te verdelen.",
+  "number_of_seats": "Aantal raadszetels",
+  "preference_threshold": "Voorkeursdrempel",
+  "preference_threshold_description": "25% van de kiesdeler",
```
praseodym, line 20: This is 50% for <19 seats ([Artikel P 15 Kieswet](https://wetten.overheid.nl/jci1.3:c:BWBR0004627&afdeling=II&hoofdstuk=P&paragraaf=3&artikel=P_15&z=2023-06-20&g=2023-06-20)), so we should make it a variable that is provided by the backend.

praseodym (changes requested):
Looks good, just two small things!

praseodym:
> Added css util class `font-number` setting `font-family: "GeistMono", monospace, monospace;` and `font-feature-settings: "ss09" on;`
> * Removed `font-feature-settings: "ss09" on;` from `@font-face` because this is not supported by chrome and edge

Was there some linter that complained about this, or did you observe rendering differences between browsers?

frontend/lib/ui/style/variables.css
```diff
@@ -17,7 +17,6 @@
 @font-face {
   font-family: "GeistMono";
   src: url("../../../static/font/Geist_Mono/GeistMono-Regular.otf") format("opentype");
```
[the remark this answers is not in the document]
praseodym: I've created #1052 for this because I noticed that the other fonts are also not variable.

Lionqueen94:
> > Added css util class `font-number` setting `font-family: "GeistMono", monospace, monospace;` and `font-feature-settings: "ss09" on;`
> > 
> > * Removed `font-feature-settings: "ss09" on;` from `@font-face` because this is not supported by chrome and edge
> 
> Was there some linter that complained about this, or did you observe rendering differences between browsers?

Yes I observed the differences, or more precisely, noticed that the setting in font-face wasn't working in some browsers, hence why it was by default also set for the number cells in the input grid. When we checked the font-face support it turned out to be terrible, hence the decision to make a util for it and use that instead of setting `font-feature-settings` separately everywhere we use `GeistMono`.

frontend/lib/i18n/locales/nl/apportionment.json
Lionqueen94: Changed in db9316fcdfa746e5217c42531197bea9a2c13212 (if you prefer the percentage to come from the backend, I'll arrange that in the next epic)

frontend/lib/ui/style/variables.css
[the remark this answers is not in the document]
jorisleker: > This font file doesn't have a bold style, but a bold font weight is used in apportionment. This results in ugly font rendering:
> 
> 
> 
> Screenshot 2025-02-18 at 21 47 06
> 
> 
> 
> Switching to the variable font (which is already in our repository) looks a lot better:
> 
> 
> 
> Screenshot 2025-02-18 at 21 46 55
> 
> 
> 
> ```suggestion
> 
>   src: url("../../../static/font/Geist_Mono/GeistMonoVF.ttf") format("TrueType");
> 
>   font-weight: 1 1000;
> 
> ```

If I recall correctly, the variable weight Geist Mono does not have/support the feature to turn off the dashed 0's. Can you check? Otherwise we might want to add a separate font file for the bold weights.

lkleuver (approved):

frontend/lib/i18n/locales/nl/apportionment.json
praseodym: I'd definitely prefer that because I don't want to have lt/gte 19 logic spread around the codebase.

praseodym:
> Yes I observed the differences, or more precisely, noticed that the setting in `font-face` wasn't working in some browsers, hence why it was by default also set in the css for the input cells in the input grid. When we checked the `font-face` support it turned out to be terrible, hence the decision to make a util for it and use that instead of setting `font-feature-settings` separately everywhere we use `GeistMono`. Would be awesome if we have a linter checking for this though haha.

And by 'observed the differences' you mean that the slashed zero was rendered?

frontend/lib/ui/style/util.css
praseodym, line 1: Some of these classes should be moved to `typography.css` instead.

frontend/app/module/apportionment/page/ApportionmentResidualSeatsPage.tsx
```diff
@@ -0,0 +1,128 @@
...
+  if (apportionment) {
+    const highestSurplusSteps = apportionment.steps.filter((step) => step.change.assigned_by === "HighestSurplus");
+    const highestAverageSteps = apportionment.steps.filter((step) => step.change.assigned_by === "HighestAverage");
+    return (
+      <>
+        {render_title_and_header()}
+        <main>
+          <article className={cls.article}>
+            {apportionment.residual_seats > 0 ? (
+              apportionment.seats >= 19 ? (
```
praseodym, line 75: We should let the backend return what apportionment algorithm was used instead of adding more lt/gte 19 logic in the frontend.

frontend/lib/ui/style/variables.css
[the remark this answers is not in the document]
jorisleker: Missed that '10' whilst reviewing on the slopes 🙃. Great, then let's go for the variable font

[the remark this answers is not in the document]
Lionqueen94: > Turning off the slashed zeroes (feature `ss09` on) works fine with the variable font. I've tested in Firefox, Chrome and Safari. The screenshot of the variable font above also didn't have the slashed zeroes.

Did you check this on this PR or on main? Since on the other places geist mono is used the font-feature-settings was also set in css instead of just in font-face?

Lionqueen94:
> > Yes I observed the differences, or more precisely, noticed that the setting in `font-face` wasn't working in some browsers, hence why it was by default also set in the css for the input cells in the input grid. When we checked the `font-face` support it turned out to be terrible, hence the decision to make a util for it and use that instead of setting `font-feature-settings` separately everywhere we use `GeistMono`. Would be awesome if we have a linter checking for this though haha.
> 
> And by 'observed the differences' you mean that the slashed zero was rendered?

Yes, in chrome (my main dev setup) I got the slahed zeroes, and Joris commented on it. Then looked up the font-feature-settings we had and the compatibility in font-face (because it wasn't working in my code and was in inputgrid, where it was also set in the css). So that's when I figured ok let's take it out of font-face and set it in a separate util. To prevent confusion in the future.

frontend/lib/ui/style/util.css
Lionqueen94: Done in 5e13c4d95cf27359e110b89d91d6efcd597ddb01

frontend/lib/i18n/locales/nl/apportionment.json
Lionqueen94: Removed the row for now in 5e13c4d95cf27359e110b89d91d6efcd597ddb01, will return the percentage from the backend in the next epic

frontend/app/module/apportionment/page/ApportionmentResidualSeatsPage.tsx
Lionqueen94: Good call! I'll think about a way to improve this in the next epic

praseodym (approved):
lgtm!

lkleuver (approved):

PR #945, opened 2025-02-03
Ensure that Fractions do not divide by zero

squell:
The current implementation of `Fraction` in `apportionment/fraction.rs` allows "NaN"'s to be formed, that is, a `Fraction` which has the denominator set to zero. This PR changes that by putting a type invariant in place that ensures that the denominator is always nonzero (and consequently, dividing by zero becomes a runtime `panic!`).

The current implementation of NaN's leads to surprising results; for example:

```rust
    #[test]
    fn eq_is_intransitive() {
        let one = Fraction::from_u64(1);
        let zero = Fraction::from_u64(0);
        let everything = one / zero * zero;

        // 5 is the same as 42 ?
        assert!(Fraction::from_u64(5) == everything && everything == Fraction::from_u64(42));
        
        // but of course 5 is not the same as 42
        assert!(Fraction::from_u64(5) != Fraction::from_u64(42));
    }
```
The above code violates the **Transitivity** requirement of the [`PartialEq` trait](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html) that is implemented for `Fraction`.

In practice, the current apportionment code should not be affected by this: it neither constructs nor needs `Fraction`s where the denominator is zero (there are no elections if there are 0 seats, right?). But then it is best to simply rule them out. Also the above behaviour is the exact opposite of NaN's in IEEE-754, where a NaN always compares as *unequal* to everything including itself. 

The good news is, when there is the guarantee that overflows are caught (there is `overflow-checks = true` in Cargo.toml) and that the denominator is always strictly positive, you can use a [formal correctness proof](https://github.com/squell/proofs/blob/main/Fraction/fractions.mlw) to show that the formulas given in `fraction.rs` for addition, multiplication etc. are a correct representation of the non-negative rational numbers.

##### Alternative solutions:
Instead of a type variant (which the Rust compiler doesn't check), a more explicit way to achieve the desired effect would be to use `std::num::NonZero<u64>` for the `denominator`. But that would come at the expense of readability. If you go down this route I would suggest using the `u64::checked_mul()` and `u64::checked_add` methods in every method of `Fraction` as well---for consistency and to make the arithmetical overflow checks more explicit.


praseodym:
Thanks a lot for your contribution!

> Instead of a type variant (which the Rust compiler doesn't check), a more explicit way to achieve the desired effect would be to use `std::num::NonZero<u64>` for the `denominator`. But that would come at the expense of readability. If you go down this route I would suggest using the `u64::checked_mul()` and `u64::checked_add` methods in every method of `Fraction` as well---for consistency and to make the arithmetical overflow checks more explicit.

I prefer more explicit solutions so I'm curious to see what will look like. Since the `Fraction` module is quite small I don't think it should impact readability too much. I'm not sure if that is something for this PR or for a new one, though.

praseodym (approved):

praseodym:
I agree that the more explicit solution is not very pretty and as stated doesn't have advantages over `overflow-checks = true`. However, one potential benefit is that it allows us to handle overflows more explicitly, i.e. doing something better than panicking (like `overflow-checks = true` does or what `checked_mul(..).unwrap()` would do). That's out of scope for this PR, though.

I've approved this PR, @Lionqueen94 and/or @rnijveld could you do the second review?

Lionqueen94 (approved):

Issue #953, opened 2025-02-03
Rerender in DataEntryHomePage.test.tsx starts failing when upgrading react-router to ≥7.1.4


oliver3:
See dependabot PR https://github.com/kiesraad/abacus/pull/942

PR #954, opened 2025-02-03
Keep React Router version 7.1.3

oliver3:
Update React Router to 7.1.4 or 7.1.5 causes DataEntryHomePage.test.tsx to fail. Issue https://github.com/kiesraad/abacus/issues/953 has been created to fix this.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,8 +35,9 @@ updates:
       - dependency-name: "eslint"
         update-types: ["version-update:semver-major"]
       # React Router major upgrades usually have breaking changes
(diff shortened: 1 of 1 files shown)


praseodym:
I think https://github.com/kiesraad/abacus/pull/942#issuecomment-2631215104 should fix this too.

Is there an issue in react-router for this problem?

oliver3:
> Is there an issue in react-router for this problem?

I'm not quite sure what the issue is still, but I could not find anything related in their issues.

praseodym:
https://github.com/kiesraad/abacus/pull/942#issuecomment-2631215104 resolved this, Dependabot created https://github.com/kiesraad/abacus/pull/955 without the react-router update.

PR #960, opened 2025-02-04
Extend session lifetime

marlonbaeten:
This PR adds middleware to our backend that checks whether a session is about to expire and extends the session lifetime and sends a fresh cookie to the user.

Currently the session life-time is 30 minutes and the minimum session life-time is 15 minutes. This means that every request we check whether the session is going to expire in the next 15 minutes and, if so, extend the session life-time to 30 minutes (from now) and send a new cookie to the client with a 30 minute life-time.

Note that both the cookie en the database session have life-times.

This mechanism prevents updating the database session and cookie every request. This will happen at most every 15 minutes. Also, after 30 minutes of inactivity the session will expire.

We could extend this functionality with heartbeat behavior in the client, the keep sessions alive as long as (for instance) the application has focus.

Testing this feature either involves logging in and wait 15 minutes, than do a new API request and witness a new `set-cookie` header OR look carefully at the test `test_extend_session`;


cikzh (dismissed):

backend/src/authentication/api.rs
```diff
@@ -2,11 +2,17 @@ use super::error::AuthenticationError;
 use super::session::Sessions;
 use super::user::{ListedUser, User, Users};
 use super::{SECURE_COOKIES, SESSION_COOKIE_NAME, SESSION_LIFE_TIME};
+use axum::extract::Request;
+use axum::middleware::Next;
+use axum::response::Response;
 use axum::{extract::State, response::IntoResponse, Json};
```
Lionqueen94, line 8:
```suggestion
use axum::{
  extract::{Request, State},
  middleware::Next,
  response::{IntoResponse, Json, Response}
};
```

Lionqueen94 (approved):

backend/src/authentication/api.rs
```diff
@@ -171,6 +178,34 @@ pub async fn logout(
...
+    // extend lifetime of session and set new cookie if the session is still valid and will soon be expired
+    if let Ok(Some(session)) = sessions.extend_session(session_cookie.value()).await {
+        debug!("Session extended: {:?}", session_cookie);
+
+        let mut cookie = session.get_cookie();
+        set_default_cookie_properties(&mut cookie);
+
+        debug!("Setting cookie: {:?}", cookie);
+
+        if let Ok(header_value) = cookie.encoded().to_string().parse() {
```
praseodym, line 201: `encoded()` uses the `percent-encode` feature of the `cookie`, which we should explicitly enable in our project. It currently only compiles because `axum-extra` implicitly enables the feature.

cikzh: Added in [3fd844a](https://github.com/kiesraad/abacus/pull/960/commits/3fd844a1d5f78a9a2df7902273c8bf32c3da75f4)

cikzh (dismissed):

marlonbaeten:
Thank you @cikzh <3

praseodym (dismissed):

praseodym (approved):

cikzh (approved):

PR #964, opened 2025-02-04
Simplify data entry forms

### LINKED ISSUE #850: Implement reducer solution for data entry
Implement the reducer solution that was chosen in the evaluation of the spikes (#844, #845, #851, #846).

marlonbaeten:
Fixes #850

This PR is an overhaul of the state management for the data entry forms.

It introduces a "reducer" to centralize state manipulation and simplify the data flow using events.

Apart from the central form state, the local form hooks and components are cleaned up, in particular the number of hard-to-reason-about `useEffect` and `useRef` hooks are removed.

See `frontend/app/component/form/data_entry/state` for the hooks, context and providers related to the data entry state.

Notes:

- This refactor is the MVP for less complex state management in the data entry forms, its not perfect, especially in the util functions there is still some refactoring to be done.
- This PR introduces the reducer and simplifies the central state; however, the central state is not perfectly compact as it should be.
- We focussend on rewriting the most complex parts and getting it back into `main` ASAP
- See https://react.dev/reference/react/useReducer for documentation about the `useReducer`

Other notes on React state management:

## The data flow we try to adhere to

```
┌────> State ──> View ───┐
│                        │
│                        │
└──────── Events <───────┘
```

In our case events are triggered from the UI, which are dispatched to the central reducer. The reducer updates the otherwise immutable state object. The React components render this immutable state as a UI.

## Don't overuse `useRef`

From the React documentation:

> When to Use Refs
> 
> There are a few good use cases for refs:
> 
> * Managing focus, text selection, or media playback.
> * Triggering imperative animations.
> * Integrating with third-party DOM libraries.
> 
> Avoid using refs for anything that can be done declaratively.
> 
> Don’t Overuse Refs
> 
> Your first inclination may be to use refs to “make things happen” in your app. If this is the case, take a moment and think more critically about where state should be owned in the component hierarchy. Often, it becomes clear that the proper place to “own” that state is at a higher level in the hierarchy. See the Lifting State Up guide for examples of this.

## Don't overuse `useEffect`

From the React documentation:

> You Might Not Need an Effect
> 
> Effects are an escape hatch from the React paradigm. They let you “step outside” of React and synchronize your components with some external system like a non-React widget, network, or the browser DOM. If there is no external system involved (for example, if you want to update a component’s state when some props or state change), you shouldn’t need an Effect. Removing unnecessary Effects will make your code easier to follow, faster to run, and less error-prone.

## Strive for simple components with unidirectional data flow

- Event handlers should trigger state updates
- The events objects passed from the DOM are excellent references to DOM elements and their values -> we don't need `useRef`
- Only use `useEffect` to preload API data, or trigger explicit side-effects
- Keep state minimal and calculate derived state in the render component
- The `useMemo` and `useCallback` hooks should only be used if they prevents expensive calculations - in Abacus this is almost never the case!


marlonbaeten:
After merge:

- safe form reducer
- mock data in components tests
- generalize the data entry form hooks
- de-duplicate properties in data entry forms
- write reducer tests
- cache per form section instead of gobal

Before merge:

- Fix e2e tests (Marlon)
- Enable PollingStationFormNavigation in CandidatesVotesForm (Les)
- Polling station to data entry renames (Les)
- Cleanup commented code (Les)
- Implement document.getElementById differntly in CandidatesVotesForm (MArlon)
- Rename `types.ts` and `actions.ts` etc... (Marlon)

frontend/app/component/form/data_entry/voters_and_votes/useVotersAndVotes.ts
```diff
@@ -1,60 +1,86 @@
...
+    if (!hasChanges) {
+      updateFormSection({ hasChanges: true, acceptWarnings: false, acceptWarningsError: false });
     }
+    setCurrentValues(values);
+  };
 
-    return result;
-  }, [values, setTemporaryCache, cache]);
+  const setAcceptWarnings = (acceptWarnings: boolean) => {
+    updateFormSection({ acceptWarningsError: false, acceptWarnings });
```
oliver3, line 52: In the original implementation, the error is not immediately hidden when you check the box `acceptWarnings`. Maybe we do not need `setAcceptWarnings()`.

frontend/app/component/form/data_entry/state/dataEntryUtils.test.ts
```diff
@@ -0,0 +1,338 @@
...
+    ];
+
+    expect(hasOnlyGlobalValidationResults(mixedResults)).toBe(false);
+  });
+});
+
+describe("resetFormSectionState", () => {
+  test("should reset form section state", () => {
+    const formState = {
+      ...defaultDataEntryState.formState,
```
oliver3, line 148: This is a shallow clone. Maybe it is better to make `defaultDataEntryState` (or `defaultFormState`) a function that returns a new object everytime? Then there is no need for a `structuredClone` in every test.

frontend/app/component/form/data_entry/state/reducer.test.ts
```diff
@@ -0,0 +1,166 @@
...
+import { mockElection } from "app/component/election/status/mockData";
+
+import { ApiResponseStatus } from "@kiesraad/api";
+
+import { getInitialValues } from "./dataEntryUtils";
+import dataEntryReducer, { getInitialState } from "./reducer";
+import { DataEntryAction, DataEntryState } from "./types";
+
+const initialState: DataEntryState = getInitialState(mockElection, 1, 1);
+const initialValues = getInitialValues(mockElection);
```
oliver3, line 12: Creating this as a constant reused over tests introduces the risk of changing the object in one test and having the wrong object in another. I would prefer using `getInitialValues(mockElection)` and `getInitialState(mockElection, 1, 1)` in each separate test.

praseodym (changes requested):
- [ ] When entering votes for a political group, the error message ‘Controleer het totaal van de lijst. Overleg met de coördinator als op het papier niets is ingevuld.’ appears right away instead of only after clicking ‘Volgende’. 
- [ ] Numeric formatting in form fields disappears when clicking ‘Volgende’, and doesn’t always reappear when staying on the same page due to errors or warnings.
- [ ] 'Let op: niet opgeslagen wijzigingen’ modal keeps appearing for the same page after when ‘Niet bewaren’ was chosen
  - Scenario:
    - Fill data entry up to ‘Verschillen’
    - Go back to ‘Aantal kiezers en stemmen’ using link on left side
    - Go to ‘Verschillen’ using link on left side
    - Click ‘Niet bewaren’ in 'Let op: niet opgeslagen wijzigingen’ modal
    - Go back to ‘Aantal kiezers en stemmen’ again
    - Don’t make changes and go to ‘Verschillen’ using link on left side
    - Observe that 'Let op: niet opgeslagen wijzigingen’ modal appears without making changes

frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
```diff
@@ -139,11 +57,20 @@ export function CandidatesVotesForm({ group }: CandidatesVotesFormProps) {
...
+            pg.number !== group.number ? pg : formValuesToValues(currentValues),
+          ),
+        }}
+      />
+      {formSection.isSaved && formSection.errors.length > 0 && (
+        <Feedback id="feedback-error" type="error" data={formSection.errors.map((error) => error.code)} />
       )}
-      {isSaved && hasValidationWarning && !hasValidationError && (
-        <Feedback id="feedback-warning" type="warning" data={warnings.map((warning) => warning.code)} />
+      {formSection.isSaved && formSection.warnings.length > 0 && formSection.errors.length === 0 && (
```
Lionqueen94, line 72:
```suggestion
      {formSection.isSaved && showAcceptWarnings && (
```
my suggestion would be to rename `showAcceptWarnings` to `hasWarnings`
Same suggestion for `DifferencesForm.tsx`

frontend/app/component/form/data_entry/voters_and_votes/VotersAndVotesForm.test.tsx
```diff
@@ -272,6 +317,7 @@ describe("Test VotersAndVotesForm", () => {
       expect(request_body.data).toEqual(expectedRequest.data);
     });
 
+    //TODO: fix which mock to use, confusing via ./form/testHelperFunctions and ./test.util
```
Lionqueen94, line 320: This TODO still needs to be done I think

frontend/app/component/form/data_entry/voters_and_votes/VotersAndVotesForm.test.tsx
```diff
@@ -537,7 +594,8 @@ describe("Test VotersAndVotesForm", () => {
       expect(screen.getByTestId("blank_votes_count"), "100").toHaveValue("100");
       await user.clear(screen.getByTestId("blank_votes_count"));
 
-      await waitFor(() => expect(acceptFeedbackCheckbox).not.toBeInTheDocument());
+      //TODO: check
+      //await waitFor(() => expect(acceptFeedbackCheckbox).not.toBeInTheDocument());
```
Lionqueen94, line 598: Same as above

frontend/app/component/form/data_entry/voters_and_votes/VotersAndVotesForm.test.tsx
```diff
@@ -560,15 +618,16 @@ describe("Test VotersAndVotesForm", () => {
 
       expect(feedbackWarning).toHaveTextContent(feedbackMessage);
       // All fields should be considered valid now
-      expectedValidFieldIds = expectedValidFieldIds.concat(expectedInvalidFieldIds);
+      //TODO: is this true? there is a warning in blank votes, so it should not be valid?
+      //expectedValidFieldIds = expectedValidFieldIds.concat(expectedInvalidFieldIds);
```
Lionqueen94, line 622: Once acceptWarnings is clicked, the field with a warning should become valid, so yes it's invalid before that. Right @praseodym ?

frontend/app/component/form/data_entry/state/useInitialDataEntryState.ts
```diff
@@ -0,0 +1,109 @@
...
+  dataEntry: DataEntry,
+  client: ApiClient,
+  dispatch: DataEntryDispatch,
+) {
+  if (!dataEntry.client_state) {
+    return;
+  }
+
+  // save data entry, continue is set to true to make sure polling station status changes to InProgress
+  // TODO: check whether this is always necessary
```
Lionqueen94, line 27: This TODO still needs to be solved I think

frontend/app/component/form/data_entry/state/useDataEntry.ts
```diff
@@ -0,0 +1,43 @@
...
+  useDataEntryNavigation(state, dispatch, election, pollingStationId, entryNumber);
+
+  return {
+    ...state,
+    dispatch,
+    onSubmitForm: onSubmitForm(client, requestPath, dispatch, state),
+    onDeleteDataEntry: onDeleteDataEntry(client, requestPath, dispatch),
+    onFinaliseDataEntry: onFinaliseDataEntry(client, requestPath, dispatch),
+    register: registerForm(dispatch),
+    setCache: setCache(dispatch), // TODO naar FormSection?
```
Lionqueen94, line 40: Does this TODO still have to be processed?

frontend/app/component/form/data_entry/state/reducer.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export function getInitialState(
+  election: Required<Election>,
+  pollingStationId: number,
+  entryNumber: number,
+): DataEntryState {
+  return {
+    election,
+    pollingStationId,
+    // initialData: null,
```
Lionqueen94, line 26: This comment should be removed

frontend/app/component/form/data_entry/state/reducer.ts
```diff
@@ -0,0 +1,153 @@
...
+    formState: getInitialFormState(election),
+    targetFormSectionId: INITIAL_FORM_SECTION_ID,
+    status: "idle",
+    cache: null,
+  };
+}
+
+export default function dataEntryReducer(state: DataEntryState, action: DataEntryAction): DataEntryState {
+  /// eslint-disable-next-line
+  // console.log("ACTION", action);
```
Lionqueen94, line 39: This comment should be removed

frontend/app/component/form/data_entry/state/reducer.ts
```diff
@@ -0,0 +1,153 @@
...
+        const { formState, targetFormSectionId } = buildFormState(
+          action.dataEntry.client_state as ClientState,
+          action.dataEntry.validation_results,
+          state.election,
+        );
+        return {
+          ...state,
+          formState,
+          targetFormSectionId,
+          // initialData: action.dataEntry,
```
Lionqueen94, line 53: This comment should be removed

frontend/app/component/form/data_entry/state/reducer.ts
```diff
@@ -0,0 +1,153 @@
...
+          pollingStationResults: action.dataEntry.data,
+          error: null,
+        };
+      }
+
+      return {
+        ...state,
+        formState: getInitialFormState(state.election),
+        targetFormSectionId: INITIAL_FORM_SECTION_ID,
+        // initialData: action.dataEntry,
```
Lionqueen94, line 63: This comment should be removed

frontend/app/component/form/data_entry/differences/DifferenceForm.test.tsx
```diff
@@ -9,27 +9,61 @@ import {
...
+        id: "save",
+        index: 4,
+        ...defaultFormSection,
+      },
+    },
+  },
+  targetFormSectionId: "recounted",
+  status: "idle",
+  cache: null,
+};
```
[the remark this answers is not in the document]
lkleuver: shall we discuss this at SUM? If I understand correctly @oliver3 is suggesting the opposite?

frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
lkleuver: this is a bool to indicate if the accept warnings checkbox should be shown. which is if there are warnings and no errors. 
Do agree this is getting complicated.

frontend/app/component/form/data_entry/voters_and_votes/VotersAndVotesForm.test.tsx
lkleuver: can we lift this out to a separate issue?

Lionqueen94: Yes that's what I mean sorry, check accept warnings and click volgende and then they are valid.

frontend/app/component/form/data_entry/state/reducer.ts
lkleuver: This is actually quite useful when developing, I've added a comment to the comment.
Also okay with removing it though.

frontend/app/component/form/data_entry/voters_and_votes/VotersAndVotesForm.test.tsx
lkleuver: i think we need a mock data PR/issue 😅

praseodym:
Probably a new issue (and did not check if this is also the case in `main`): when triggering an error, and clicking 'Volgende' right away without making changes, the viewport doesn't scroll up to show the error. This can be confusing because there is no visual feedback about the error on the bottom of the page.

lkleuver:
To close this PR the following codereview feedback has been moved to the following issues:
- placement and use of test data: #1112 
- pressing "volgende" twice does not scroll to top: #1113

jschuurk-kr (commented):
If I get to the Voters and Votes page, then return to the Recounted page, change the radio button, and click in the nav to go the Voters and Votes page, I don't get a pop-up to save or discard changes.

jschuurk-kr:
On the Differences page, if the error "Je kan alleen verder als je het het papieren proces-verbaal hebt gecontroleerd." is shown and the accept warnings-checkbox is ticked, changing the data in the form does not untick the accept warnings-checkbox.

lkleuver:
> If I get to the Voters and Votes page, then return to the Recounted page, change the radio button, and click in the nav to go the Voters and Votes page, I don't get a pop-up to save or discard changes.

Is this something that can be added to model based testing? changes popup inbetween sections?
It should be fixed now, recounted was missing the flag hasChanges

lkleuver:
> On the Differences page, if the error "Je kan alleen verder als je het het papieren proces-verbaal hebt gecontroleerd." is shown and the accept warnings-checkbox is ticked, changing the data in the form does not untick the accept warnings-checkbox.
> 
> The Voters and Votes pages behaves differently: when the error is shown, as soon as you tick the checkbox, the error disappears. When you then change data in the form, the text and checkbox to accept warnings disappears.

The behavior should now be the same across sections, though perhaps slightly different from main.

frontend/app/component/form/data_entry/voters_and_votes/useVotersAndVotes.ts
lkleuver: This is a behavior change thats perhaps unwanted, it's a bit finicky to fix. Do you want to change it here or a separate issue?

frontend/app/component/form/data_entry/state/dataEntryUtils.test.ts
lkleuver: I've created a separate issue for this in #1112

frontend/app/component/form/data_entry/state/dataEntryUtils.test.ts
```diff
@@ -0,0 +1,338 @@
...
+          code: "W201",
+          id: "blank_votes_count",
+        },
+      ]),
+    );
+  });
+});
+
+describe("getPollingStationSummary", () => {
+  test("getPollingStationSummary", () => {
```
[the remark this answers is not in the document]
lkleuver: this is not new now I believe, maybe part of #1112

frontend/app/component/form/data_entry/state/reducer.test.ts
lkleuver: see #1112

frontend/app/component/form/data_entry/candidates_votes/CandidatesVotesForm.tsx
lkleuver: it's in the use hook concisely now.

oliver3:
I've reached an illegal state with the steps below, which results in a backend error `Invalid data error: Data error: recounted==true but voters_recounts is None`. This does not occur on main branch.

- start data entry
- enter recounted false
- continue to voters and votes
- enter some value, do not continue
- navigate back to recounted
- enter recounted true
- navigate using the navigation list to voters and votes
- modal "unsaved changes" appears, select save
- error will be shown "Sorry, er ging iets mis, De invoer is niet geldig"
- backend throws an error "Invalid data error: Data error: recounted==true but voters_recounts is None"

oliver3 (dismissed):
Illegal state `recounted==true but voters_recounts is None` seems to be fixed!

jschuurk-kr:
> > On the Differences page, if the error "Je kan alleen verder als je het het papieren proces-verbaal hebt gecontroleerd." is shown and the accept warnings-checkbox is ticked, changing the data in the form does not untick the accept warnings-checkbox.
> > The Voters and Votes pages behaves differently: when the error is shown, as soon as you tick the checkbox, the error disappears. When you then change data in the form, the text and checkbox to accept warnings disappears.
> 
> The behavior should now be the same across sections, though perhaps slightly different from main.

Created a new issue to add RTL tests for this behavior: https://github.com/kiesraad/abacus/issues/1137

jschuurk-kr:
@lkleuver Not sure if it's in scope for this PR, but if I have an expired session, I can still navigate through the form. And as part of each new page render, I get the "De sessie is ongeldig" error popup.

praseodym:
> @lkleuver Not sure if it's in scope for this PR, but if I have an expired session, I can still navigate through the form. And as part of each new page render, I get the "De sessie is ongeldig" error popup.
> 
> I'd expect not to be able to navigate to a different page with an expired session.

Not in scope for this PR, and also not something that was changed in this PR. When navigating to other parts of the form no requests are made (only when saving, either through the 'Volgende' button or the modal) so no session validation is done.

praseodym:
It looks like the behaviour of the [X] button in the abort modal has changed compared to `main`.

Reproduce using:
1. Start data entry
2. Click one of the radio buttons on the "Is er herteld?" page
3. Click "Invoer afbreken", abort modal shows
4. Click [X] to close modal

What happens:
- Modal closes
- Form is saved and next page ("Aantal kiezers en stemmen") is shown

Expected:
- Modal closes, nothing else happens

jschuurk-kr:
> > Then I shouldn't get the "De sessie is ongeldig" error popup for each navigation action, but I do.
> 
> Hm I misread, I don't know why the "De sessie is ongeldig" error popup reappears every time. But #1138 should cover that too right? I don't think this should be a dismissible popup anyway.

Good point we should deal with both issues in #1138. Will add some words to it.

lkleuver:
> It looks like the behaviour of the [X] button in the abort modal has changed compared to `main`. (Thanks to @jschuurk-kr for finding this one)
> 
> Reproduce using:
> 
> 1. Start data entry
> 2. Click one of the radio buttons on the "Is er herteld?" page
> 3. Click "Invoer afbreken", abort modal shows
> 4. Click [X] to close modal
> 
> What happens:
> 
> * Modal closes
> * Form is saved and next page ("Aantal kiezers en stemmen") is shown
> 
> Expected:
> 
> * Modal closes, nothing else happens

In the end, it was the close button of the modal that submitted the form.
type="button" fixed it

frontend/e2e-tests/resume-data-entry.e2e.ts
```diff
@@ -49,6 +49,20 @@ test.describe("resume data entry flow", () => {
...
+    test("Closing abort modal with X only closes the modal", async ({ page, pollingStation }) => {
+      await page.goto(`/elections/${pollingStation.election_id}/data-entry/${pollingStation.id}/1/recounted`);
+      const recountedPage = new RecountedPage(page);
+      await recountedPage.no.click();
+      await recountedPage.abortInput.click();
+
+      const abortInputModal = new AbortInputModal(page);
+      await abortInputModal.close.click();
+
+      expect(recountedPage.no.isChecked()).toBeTruthy();
```
jschuurk-kr, line 62:
```suggestion
      expect(recountedPage.no).toBeChecked();
```

frontend/e2e-tests/data-entry.e2e.ts
```diff
@@ -57,7 +57,7 @@ test.describe("full data entry flow", () => {
     };
     await votersAndVotesPage.inputVotersCounts(voters);
     await votersAndVotesPage.inputVotesCounts(votes);
-    await expect(votersAndVotesPage.pollCardCount).toHaveValue(formatNumber(voters.poll_card_count));
+    await expect(votersAndVotesPage.pollCardCount).toHaveValue("1.000");
```
jschuurk-kr, line 60: What's the reason for this change?

The test is more maintainable if we use `voters.poll_card_count` instead of a hard-coded value. And `formatNumber()` works for the test using it on [line 151](https://github.com/kiesraad/abacus/blob/bd270cc7d61f365f51a3dd46645567c00eb7bb91/frontend/e2e-tests/data-entry.e2e.ts#L151C1-L151C112), so I'd expect it to work here too.

frontend/e2e-tests/resume-data-entry.e2e.ts
```diff
@@ -49,6 +49,20 @@ test.describe("resume data entry flow", () => {
     return new AbortInputModal(page);
   };
 
+  test.describe("abort data entry", () => {
+    test("Closing abort modal with X only closes the modal", async ({ page, pollingStation }) => {
```
jschuurk-kr, line 53: Am tempted to also make this test check no request was made, e.g. like this https://github.com/microsoft/playwright/issues/13284#issuecomment-1087721003

What do you think?

frontend/e2e-tests/data-entry.e2e.ts
lkleuver: was an oversight. fixed it

frontend/e2e-tests/resume-data-entry.e2e.ts
lkleuver: I've added the request check. This is actually what I mainly did to debug the issue and was quite hidden. 
I think everyone expects the "X" to do absolutely nothing.

jschuurk-kr (approved):

oliver3 (approved):

praseodym (approved):

Issue #915, opened 2025-01-27
List users page


oliver3:
New page that retrieved users from the backend and displays them in a table

Depends on #874 and #930

PR #965, opened 2025-02-05
Add user list page

oliver3:
Resolves https://github.com/kiesraad/abacus/issues/915

- Rename API endpoint `list` to `user_list`
- Add mock data and handler
- Add user list page and its test

Page can be reached from the `/dev` page.


jorisleker (dismissed):
List page looks good to me, with following review remarks:

- Have not been able to test date formatting in the 'last activity' column
- Have not been able to verify all user roles are displayed correctly (only 1 user in database)
- The 'admin' user in the database is an anonymous account, which should be impossible
- The button to add a user is lacking, but will probably be added when the 'add' feature is implemented?

oliver3:
Thanks @jorisleker ! The [frontend preview](https://b1ebf1ae.kiesraad-abacus.pages.dev/users) is also deployed and has more users.

The button will indeed be added later, and fair point about the 'admin', I'll add a fullname to make it more representable in a separate PR.

frontend/app/module/users/UserListPage.tsx
```diff
@@ -0,0 +1,54 @@
...
+          <Table id="users">
+            <Table.Header>
+              <Table.Column>{t("user.username")}</Table.Column>
+              <Table.Column>{t("role")}</Table.Column>
+              <Table.Column>{t("user.fullname")}</Table.Column>
+              <Table.Column>{t("user.last_activity")}</Table.Column>
+            </Table.Header>
+            <Table.Body className="fs-md">
+              {users.map((user) => (
+                <Table.LinkRow key={user.id} to={`${user.id}/update`}>
```
Lionqueen94, line 39: Clicking the link row gives an error, I'm assuming the update page will be implemented later?

Lionqueen94 (dismissed):

Lionqueen94 (approved):

jorisleker (approved):

jorisleker:
@oliver3 approved changed test, but now thinking: we should probably say 'yesterday 13:37' if last activity was yesterday, and not the day name. Day name for anything 2-6 days ago is fine. Before that, it should be a full date. 

Earlier assumption was that Abacus would only be used during a couple of days, but now we have more insights into the NPVV we know that even for GR2026 it will be used during at least 1 week (GSB and CSB seatings are about 1 week apart)

oliver3:
> we should probably say 'yesterday 13:37' if last activity was yesterday, and not the day name. Day name for anything 2-6 days ago is fine. Before that, it should be a full date.

Something like [this](https://github.com/kiesraad/abacus/issues/973) @jorisleker ?

PR #969, opened 2025-02-05
Fix PDF for empty polling station type

praseodym:
The Typst template did not handle `polling_station_type` being `None` properly, which is fixed by checking that the `polling_station_type` key exists.

This also contains a few minor layout changes and sets the document language to Dutch.

### DIFF
--- a/backend/templates/model-na-31-2.typ
+++ b/backend/templates/model-na-31-2.typ
@@ -1,6 +1,7 @@
 #import "common/style.typ": conf, title, mono
 #import "common/scripts.typ": *
 #let input = json("inputs/model-na-31-2.json")
+#set text(lang: "nl")
 
 #show: doc => conf(input, doc, footer: [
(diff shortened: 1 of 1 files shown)


jschuurk-kr:
@praseodym @cikzh How did you test this?

It would be useful to have a before/after pdf or before/after screenshots as part of this PR.

praseodym:
> @praseodym @cikzh How did you test this?
> 
> It would be useful to have a before/after pdf or before/after screenshots as part of this PR.

I created PDFs using `typst compile model-na-31-2.typ`:
- `main`: [model-na-31-2-main.pdf](https://github.com/user-attachments/files/18748763/model-na-31-2-main.pdf)
- This branch: [model-na-31-2-969.pdf](https://github.com/user-attachments/files/18748765/model-na-31-2-969.pdf)

lkleuver (approved):

PR #985, opened 2025-02-07
Use u32 for party & candidate numbers

michielp1807:
As mentioned in #982, currently there cannot be more than 255 parties in Abacus because a `u8` is used to store the party number. In the [OpenAPI specification](https://github.com/kiesraad/abacus/blob/2836295e0bf98da94392a3aa53e2e4fd75b33f04/backend/openapi.json#L2068) the `pg_number` is listed as `int32`, likely because OpenAPI does not have smaller integer types by default. The same thing also holds for the candidate numbers.

While it is unlikely there will be more than 255 parties in the Netherlands any time soon, using `u8`'s is in my opinion unnecessarily restrictive. The couple of bytes saved by using `u8`s will in many cases remain empty anyways due to memory alignment.


Lionqueen94 (dismissed):

praseodym (approved):
Thanks, lgtm! What do you think about adding a type alias `type PGNumber = u32;` (or similarly named)?

michielp1807:
> Thanks, lgtm! What do you think about adding a type alias `type PGNumber = u32;` (or similarly named)?

Good suggestion! I've updated the PR to use type aliases `PGNumber` and `CandidateNumber`

Lionqueen94 (commented):
Could you run `npm run gen:openapi` in the `frontend` folder to make sure the `openapi.ts` is also updated :)

michielp1807:
> Could you run `npm run gen:openapi` in the `frontend` folder to make sure the `openapi.ts` is also updated :)

Thanks for the reminder! Turns out I was missing `#[schema(value_type = u32)]` in a few places, which caused a `u32` type to be introduced in the OpenAPI specification. Now it correctly uses numbers, which means the TypeScript types can remain the same.

Lionqueen94 (approved):

praseodym (approved):

PR #990, opened 2025-02-07
Some improvements to Tables and OverviewPage

Lionqueen94:
## Description
- ElectionReportPage:
  - Added missing `Footer` to `ElectionReportPage`
- Table:
  - Made sure `Table.LinkRow` last Cell has increased padding to make sure the padding around the Chevron is as designed
  - Made sure all `Table` functions accept a `className`
  - Renamed `Table.Column` to `Table.HeaderCell` since it can also be used in a row
  - Added a `scope="row" | "col"` with a default set to `"col"` to `Table.HeaderCell` to allow row `th` usage
  - Made sure `Table.HeaderCell` can have no children, equal to the implementation of `Table.Cell`
- OverviewPage:
  - Showing a message that no elections exist on `OverviewPage` instead of throwing a harsh error, separate designs for typist and administrator are implemented
  - Added tests for `OverviewPage`
- Polling Station Choice Page:
  - Made sure there is no empty table when there are no polling stations to input on polling station choice page:
Screenshot from 2025-02-11 16-48-06
- Colors:
  - Corrected error icon color in `Alert`, `ProgressList` and `StatusList` and small alert text color
  - Set the correct gray color for the Chevron in the icon with fill because background-image color cannot be set, meaning the Chevron in the `Table.LinkRow` was black instead of gray


frontend/app/module/election/page/ElectionReportPage.tsx
```diff
@@ -82,6 +84,7 @@ export function ElectionReportPage() {
           </FormLayout.Controls>
         </article>
       </main>
+      <Footer />
```
praseodym, line 87: Maybe this could be moved up to the `ElectionLayout`, but good enough for now

frontend/lib/ui/Table/Table.tsx
praseodym, line 1: Looks like these changes are unused?

frontend/app/module/election/page/ElectionReportPage.tsx
Lionqueen94: True it can probably go there and be removed from the pages indeed

frontend/app/module/election/page/OverviewPage.tsx
```diff
@@ -49,31 +49,37 @@ export function OverviewPage() {
...
-                      userRole={isAdminOrCoordinator ? "coordinator" : "typist"}
-                    />
-                  </Table.Cell>
-                </Table.LinkRow>
-              ))}
-            </Table.Body>
-          </Table>
-        </article>
+        {!electionList.length ? (
+          <p className="mb-lg">{t("election.no_elections")}</p>
```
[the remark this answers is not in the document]
jorisleker: We actually have a [design for the empty state of the elections page](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2116-30511&t=IH9L5HGr7yf9Uk90-11) 😄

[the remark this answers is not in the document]
Lionqueen94: > We actually have a [design for the empty state of the elections page](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=2116-30511&t=IH9L5HGr7yf9Uk90-11) 😄

Is there also a design for the typist version of this page (I know, this will never happen irl)?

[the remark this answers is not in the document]
Lionqueen94: I adjusted the text to be the `<h2>` in the design. The rest of the information and the add election button will be added in https://github.com/kiesraad/abacus/issues/700 (right?)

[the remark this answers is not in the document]
jorisleker: > Is there also a design for the typist version of this page (I know, this will never happen irl)?

Good catch. I've added an [empty state for this page](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=8077-50029&t=EcSLtwBUdCIBex4b-11) as well

frontend/app/module/election/page/ElectionReportPage.tsx
Lionqueen94: Changed in 558485a170835814db0acb739f7b1c37abc419b7

frontend/app/module/election/page/OverviewPage.tsx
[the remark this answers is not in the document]
Lionqueen94: Added the page design for typist in 558485a170835814db0acb739f7b1c37abc419b7

oliver3 (dismissed):

praseodym (approved):

lkleuver (approved):

frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
```diff
@@ -77,6 +77,11 @@ export function PollingStationChoiceForm({ anotherEntry }: PollingStationChoiceF
       status: status.status,
     }));
 
+  const pollingStationsForDataEntry = pollingStations.filter((pollingStation: PollingStation) => {
+    const status = electionStatus.statuses.find((status) => status.polling_station_id === pollingStation.id)?.status;
+    return status !== "definitive" && status !== "entries_different";
+  });
+
```
praseodym, line 84: Is the list of polling stations supposed to be filtered like this? Now a typist (invoerder) won't be able to see the full list of polling stations.

Lionqueen94: This is the list of polling stations that still can be filled in, the typists are only supposed to see those. It was initially filtered out in the list component, which is uglier IMO.
The coordinator needs to solve differences so they do not need to be shown to the typists.

frontend/app/component/form/data_entry/polling_station_choice/PollingStationsList.tsx
```diff
@@ -13,17 +13,14 @@ export function PollingStationsList({ pollingStations }: PollingStationsListProp
...
+        <Table.HeaderCell>{t("polling_station.title.singular")}</Table.HeaderCell>
       </Table.Header>
       <Table.Body>
         {pollingStations.map((pollingStation: PollingStation) => {
           const status = electionStatus.statuses.find(
             (status) => status.polling_station_id === pollingStation.id,
           )?.status;
-          if (status === "definitive" || status === "entries_different") {
-            return null;
-          }
```
Lionqueen94, line 26: @praseodym See this code here regarding your comment

oliver3 (dismissed):

frontend/app/component/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
jorisleker: Yes, this is by design @praseodym.

frontend/app/module/election/OverviewLayout.tsx
```diff
@@ -1,5 +1,7 @@
 import { Outlet } from "react-router";
 
+import { Footer } from "app/component/footer/Footer";
```
jorisleker, line 3: Was there a reason we removed the footer? Can remember a discussion where @oliver3 suggested removing it to fix an issue we could not easily fix any other way. Might be mistaken. 

On preview (which is slower becasue it works with the MSW), the footer is now displayed at the top of the page while the rest of the content is loading, which leads to degraded loading experience of the page. Not sure whether this was the issue we fixed by removing the footer though.

praseodym: Yeah that's probably what why the footer was moved. If the footer is in a layout that has an outlet with a component that does more API requests, the footer will be shown while that API request is pending. 

Maybe React/React Router has a better way of blanking the page when API requests are pending?

Lionqueen94: Maybe we can put it in the `ElectionLayout` and on the `OverviewPage` instead of in the `OverviewLayout`?

praseodym: Yes that should work.

oliver3: The reason I can remember was some scrolling issue in the political groups list during data entry..

jorisleker: > The reason I can remember was some scrolling issue in the political groups list during data entry..

Aah, yes you're right. Thanks for clarifying. No objection to adding the footer to the overview pages. Would be very nice if we can solve the rendering issue

Lionqueen94: I'll move the footers back to their respective pages and add the missing footer to the ElectionReportPage

Lionqueen94: Moved the footer back in the old places and added to ElectionReportPage in a06a9ac4d70e5a68c5f8d03bc088213934ce91e8

oliver3 (approved):

jorisleker (approved):

PR #1011, opened 2025-02-10
Add user update endpoint

### LINKED ISSUE #873: Update user in backend

cikzh:
Closes #873 

Adds endpoint to update a user. Also adds the `needs_password_change` flag and sets it to true when the user is newly created, or when a temporary password is set. The handling of the flag is out of scope for this PR.

Best merged after https://github.com/kiesraad/abacus/pull/1018, so it generates the correct TS type


backend/migrations/5_users.sql
```diff
@@ -1,12 +1,13 @@
...
-    password_hash      TEXT                 NOT NULL,
-    last_activity_at   DATETIME             ,
-    updated_at         DATETIME             NOT NULL DEFAULT CURRENT_TIMESTAMP,
-    created_at         DATETIME             NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    id                     INTEGER         PRIMARY KEY AUTOINCREMENT NOT NULL,
+    username               TEXT            NOT NULL,
+    fullname               TEXT            ,
+    role                   TEXT            NOT NULL,
+    password_hash          TEXT            NOT NULL,
+    needs_password_change  BOOLEAN         NOT NULL DEFAULT TRUE,
```
lkleuver, line 8: needs_password_change could be a DATETIME?

lkleuver (approved):
soft approval

backend/migrations/5_users.sql
cikzh: What would be the reason for this?

praseodym: I think a boolean is fine.

backend/src/authentication/user.rs
```diff
@@ -185,6 +188,45 @@ impl Users {
...
 
+    /// Update a user
+    pub async fn update(
+        &self,
+        user_id: u32,
+        fullname: Option<&str>,
+        temp_password: Option<&str>,
+    ) -> Result<User, AuthenticationError> {
+        if let Some(pw) = temp_password {
+            self.set_temporary_password(user_id, pw).await?;
```
praseodym, line 199: Could you add a test that covers setting the temporary password?

frontend/lib/api/gen/openapi.ts
```diff
@@ -537,6 +544,11 @@ export interface SummaryDifferencesCounts {
   unreturned_ballots_count: SumCount;
 }
 
+export interface UpdateUserRequest {
+  fullname: string;
+  temp_password: string;
+}
```
praseodym, line 550: The OpenAPI types still need to be updated now that #1018 is merged.

praseodym (changes requested):

backend/src/authentication/user.rs
cikzh: Added in [b16517d](https://github.com/kiesraad/abacus/pull/1011/commits/b16517d637ca71bac1e367ec846169bf400ac0af)

Lionqueen94 (approved):

Issue #991, opened 2025-02-07
Flaky test: Playwright e2e authentication › login happy path


praseodym:
The 'authentication › login happy path' test (`e2e-tests\authentication.e2e.ts:10:3`) failed multiple times, but not always, in Firefox on Windows:
- https://github.com/kiesraad/abacus/actions/runs/13205374900/job/36867557887
- https://github.com/kiesraad/abacus/actions/runs/13200732829/job/36852158206
- https://github.com/kiesraad/abacus/actions/runs/13181694700/job/36794180864
- https://github.com/kiesraad/abacus/actions/runs/13210768955/job/36883785814
- https://github.com/kiesraad/abacus/actions/runs/13236735808/job/36943341433
- https://github.com/kiesraad/abacus/actions/runs/13237238693/job/36944958762

It's tricky to debug this, maybe [uploading the Playwright report as an artifact](https://playwright.dev/docs/ci-intro#setting-up-github-actions) would help here.

For the Playwright report, we need [`--reporter=html`](https://playwright.dev/docs/test-reporters#html-reporter). And we need to decide/check which folder it gets written to, so we can upload the correct folder. Output dir can be set with `PLAYWRIGHT_HTML_OUTPUT_DIR` or through the config file.

---

PR to enable Playwright reports in pipeline: https://github.com/kiesraad/abacus/pull/1010

jschuurk-kr:
Some recent runst with not all tests passing:
- failing tests on the form controller refactor: https://github.com/kiesraad/abacus/actions/runs/13283369077/job/37086848842
- flaky test: https://github.com/kiesraad/abacus/actions/runs/13267764431/job/37039845527
- 2 flaky tests: https://github.com/kiesraad/abacus/actions/runs/13266777812/job/37036437979

The flaky tests are from `resume-data-entry.e2e.ts`, so not from the login test that was fixed.

For the flaky tests the report contains the traces of the failing test, but not of the passing retry. So that matches the `trace: "retain-on-first-failure"` config.

For the failing test there is only a trace for the first failing run, not for the two subsequent retries. I think that's fine for now. It won't give us a trace if some of the runs of the same test don't fail for the same reason, but that seems very unlikely.

So with a change to the flaky test and an improvement to the Playwright pipeline reports, I think we can close this issue. Agreed, @praseodym and @marlonbaeten?

PR #1012, opened 2025-02-10
Fix race condition in e2e login test

marlonbaeten:
Possible solution for #991

The login form navigates to the setup page by default, where the success message appears. We now wait for this navigation to happen.


cikzh (approved):

lkleuver (approved):

frontend/e2e-tests/authentication.e2e.ts
```diff
@@ -31,6 +31,8 @@ test.describe("authentication", () => {
     await page.getByLabel("Wachtwoord").fill(password);
     await page.getByRole("button", { name: "Inloggen" }).click();
 
+    await page.waitForURL("/account/setup");
```
jschuurk-kr, line 34: @marlonbaeten Would have been good to add a comment here as to why we are doing a `waitForURL`.

PR #1013, opened 2025-02-10
Added update user last_activity_at

cikzh:
Updates the `last_activity_at` timestamp for a user.
The update is current called in `FromRequestParts` and `OptionalFromRequestParts` and first checks if the last activity was at least `MIN_UPDATE_LAST_ACTIVITY_AT_SECS` ago, to prevent excessive database writes. Since we already fetch the user in `FromRequestParts`, it doesn't result in extra database reads.

Currently only testing if a timestamp is set, after calling the `whoami` endpoint. After we implement authorization we could add more useful testcases.


marlonbaeten (approved):
:+1:

cikzh:
@praseodym test added in [2c24084](https://github.com/kiesraad/abacus/pull/1013/commits/2c240841fc8986ce64f44ba3cb052dcd49d2396d)

backend/tests/user_integration_test.rs
```diff
@@ -0,0 +1,84 @@
...
+    let response = reqwest::Client::new().get(&url).send().await.unwrap();
+    assert_eq!(response.status(), StatusCode::OK);
+    let body: UserListResponse = response.json().await.unwrap();
+    let user = body.users.first().unwrap();
+
+    // Panics (thus fails) when timestamp was not set
+    let first_timestamp = user.last_activity_at().unwrap();
+
+    // Test that the timestamp is not updated when we
+    // immediately call the same enpoint againg
```
Lionqueen94, line 65:
```suggestion
    // immediately call the same endpoint again
```

backend/src/authentication/user.rs
```diff
@@ -49,6 +51,29 @@ impl User {
...
+    /// Updates the `last_activity_at` field, but first checks if it has been
+    /// longer than `MIN_UPDATE_LAST_ACTIVITY_AT_SECS`, to prevent excessive
+    /// database writes.
+    pub async fn update_last_activity_at(&self, users: &Users) -> Result<(), sqlx::Error> {
+        match self.last_activity_at {
+            Some(last_activity_at)
+                if chrono::offset::Utc::now()
+                    .signed_duration_since(last_activity_at)
+                    .num_seconds()
+                    > MIN_UPDATE_LAST_ACTIVITY_AT_SECS =>
```
praseodym, line 67: The `>` sign here should be inverted (`<`). Maybe it's also cleaner to write this block as a single `if` condition instead of a `match` with two arms.

The test doesn't catch this mistake: an update was made, but the resulting `last_activity_at` timestamp (with one second resolution) is the same so the test passes. Can we somehow improve the test too?

praseodym (changes requested):

backend/src/authentication/user.rs
cikzh: Woah, thanks for the catch! I'm going to write a failing test case first, then see if I can write it up in a cleaner way.

cikzh: See: [3327dd2](https://github.com/kiesraad/abacus/pull/1013/commits/3327dd2fe8d7460a948e19ca1b8d69449953bda2)

I ended up extracting it in a function so it's unit testable. Don't be fooled by the `>` not being flipped around, the new logic expects the statement to be true now (which feels more intuitive IMO)

praseodym (approved):

Lionqueen94 (approved):

PR #1018, opened 2025-02-11
OpenAPI TS generator: no required array means all fields are optional

cikzh:
Right now the OpenAPI type generator assumes the absence of a `required` field means all fields are required, but the opposite is true; no `required` field, means all fields are optional.

### To test
1. Run the generator on the `main` branch, should result in no diff
2. Create a Rust struct with optional fields only e.g.:
```rust
struct TestStruct {
    #[schema(value_type = String, nullable = false)]
    test_field_one: Option<String>,
    #[schema(value_type = String, nullable = false)]
    test_field_two: Option<String>,
}
```
and see that the generator generates optional TS interfaces


oliver3 (approved):
image

Works!

Lionqueen94 (approved):

Issue #983, opened 2025-02-06
Save user to backend


oliver3:
Follow up of #928 saving the user to the backend

- Create user API call
- Mock data and handlers

Issue #1022, opened 2025-02-12
Return 409 Conflict for duplicate username


oliver3:
Backend should return a 409 Conflict instead of 500 when create user causes a conflict on username.

Issue #1021, opened 2025-02-12
Implement password policy


ring-ring-ring:
The current design in Figma mentions 8 letters, 2 numbers. The CISO has asked for 13 letters/numbers to stay in line with the password policy.

![Image](https://github.com/user-attachments/assets/8d64e9ab-51ab-4222-ab42-544fb94c3014)

Other validations: 
- Password should not be the same as the username
- (When changing password) Password should not be the same as the current password

Best to implement the whole validation in the backend, and use for both create user and change password

## Rules

- A minimum of 13 characters
- Characters must be in the range of `[\x20-\x7E]`
- A password cannot be the same as the username
- A password cannot be the same as the current password

oliver3:
We will accept printing ASCII characters as well as the space character, i.e. [\x20-\x7E], and will not trim the string beforehand (as we do with username and full name).

cikzh:
Updated the description with a complete list of the rules we should implement

cikzh:
@praseodym @jschuurk-kr @oliver3 @jorisleker  I guess we'll also have a minimum of 2 'regular' characters? e.g. `1234567891011` would be invalid, but `one2345678910` is valid?

praseodym:
No don't think so, I also don't remember having 2 numbers being a requirement.

cikzh:
It says so in the issue description. Can someone update the 'Rules' list so I can implement the rules that are actually needed?

jorisleker:
@ring-ring-ring Which password policy? The Kiesraad's internal password policy? Does that apply to this use case? 

Let's remember that our users have to have physical access to the data entry location (airgapped network), and only use this password a couple of times during one day. On a computer without a password manager, passkeys etc. If we make passwords requirements to strict, the logical consequence is that users will write them down. Reducing security. Is that what we want?

Is it an idea to have separate rules for the admin/coordinator roles and for typists?

praseodym:
@cikzh I think that came from the text in the Figma design, which wasn't the actual policy. I've removed it from the list.

@jorisleker The CISO has defined the minimum of 13 characters for Abacus, do you think that's too strict?

jorisleker:
Before my holiday it was 12 chars (without the minimum of 2 digits). Even 12 is long for something you will only use during this one day.

[...]

PR #1027, opened 2025-02-12
Save user to backend

oliver3:
This PR will resolve https://github.com/kiesraad/abacus/issues/983

- A new `useQueryParam` hook to more easily retrieve and clear a message sent through the url query string
- Save the user from the `UserCreateDetailsPage`, show an error if it occurs
- Show success message on the user list page
- NavBar management links for all `/users*` pages

- [x] Add `UserCreate.test.tsx` that tests the flow through the pages using the context and the api call at the end

NB
- Currently a duplicate username will lead to an error 500, that will be fixed in https://github.com/kiesraad/abacus/issues/1022


jschuurk-kr:
@oliver3 Played around for a few minutes and noticed two things:
1\. We don't seem to have any restrictions on username. If you want spaces or emoji in your username, you can! This is arguably a feature, since I don't think there are any technical reasons for such restrictions.

2\. For password length, six Abacus emoji (🧮🧮🧮🧮🧮🧮) is accepted as being 12 characters long. It's probably worth fixing to avoid the discussion "But what *is* a character?" for the minimum length requirement of passwords.

oliver3:
> @oliver3 Played around for a few minutes and noticed two things: 1. We don't seem to have any restrictions on username. If you want spaces or emoji in your username, you can! This is arguably a feature, since I don't think there are any technical reasons for such restrictions.
> 
> 2. For password length, six Abacus emoji (🧮🧮🧮🧮🧮🧮) is accepted as being 12 characters long. It's probably worth fixing to avoid the discussion "But what _is_ a character?" for the minimum length requirement of passwords.

We'll revisit the password in https://github.com/kiesraad/abacus/issues/1021#issuecomment-2656407896 and I've added a comment there that we'll go for printable ASCII and space characters.

praseodym (commented):
lgtm, two questions:
1. Do we also want a Playwright e2e test for this functionality? (cc @jschuurk-kr)
2. To be considered for a new issue in this epic: it looks like the user list page is currently (implicitly) sorted by user id, maybe sorting by role and then by username makes more sense? It looks like that is the sorting order in the Figma design.

frontend/app/module/users/create/UserCreateRolePage.tsx
praseodym, line 1: Just noticed that the font size for the error message is too big (bigger than the rest of the text and bigger than in the design) which also causes it to wrap to two lines:
Screenshot 2025-02-13 at 16 59 41
(Screenshot from Safari)

jorisleker (dismissed):
Looks good, great start!

- Agree with @praseodym on the sorting order on the overview page
- When saving/creating a user (on a slow connection), it takes quite some time for the form to be submitted. This is a bit confusing, it seems as if nothing is happening. Can we blank out the entire page or disable the submit button?

frontend/app/module/users/create/UserCreateContextProvider.tsx
```diff
@@ -1,17 +1,29 @@
...
+  const [type, setType] = useState<UserType | undefined>(undefined);
+  const [username, setUsername] = useState<string | undefined>(undefined);
 
-  function updateUser(update: Partial<User>) {
-    setUser((user) => ({ ...user, ...update }));
+  const url: USER_CREATE_REQUEST_PATH = "/api/user";
+  const userApi = useCrud<User>(url);
+
+  async function createUser(userDetails: UserDetails): Promise<ApiResult<User>> {
+    if (!role) {
```
marlonbaeten, line 17: Throwing an error in a promise will not result in the router catching this. It might be better to use return values to indicate whether this function succeeded.

frontend/app/module/users/create/UserCreateRolePage.tsx
oliver3: Good catch! Fixed it in https://github.com/kiesraad/abacus/pull/1027/commits/0481033e0c3e0c5a60bb30c9c9b807ceefa7ddc2

frontend/app/module/users/create/UserCreateContextProvider.tsx
oliver3: I've avoided this scenario now, please see https://github.com/kiesraad/abacus/pull/1027/commits/4660d437e2df6021c867437ec52972789baeb0f7

oliver3:
> @jorisleker said: "Agree with @praseodym on the sorting order on the overview page"

Sorting users has been done in https://github.com/kiesraad/abacus/pull/1027/commits/3fabae1b86e07468bccdc8c11246b9a84d33b45c

oliver3:
> @jorisleker suggested: "When saving/creating a user (on a slow connection), it takes quite some time for the form to be submitted. This is a bit confusing, it seems as if nothing is happening. Can we blank out the entire page or disable the submit button?"

I've disabled the whole form including button, same as for polling stations, in commit https://github.com/kiesraad/abacus/pull/1027/commits/5c83dc3b7823d39c963f8682ff518d76221ff115

frontend/app/module/users/UserListPage.tsx
```diff
@@ -1,12 +1,14 @@
...
-import { Button, Loader, PageTitle, Table, Toolbar } from "@kiesraad/ui";
-import { formatDateTime } from "@kiesraad/util";
+import { Alert, Button, Loader, PageTitle, Table, Toolbar } from "@kiesraad/ui";
+import { formatDateTime, useQueryParam } from "@kiesraad/util";
 
 export function UserListPage() {
   const { requestState } = useUserListRequest();
+  const [createdMessage, clearCreatedMessage] = useQueryParam("created");
 
   if (requestState.status === "loading") {
```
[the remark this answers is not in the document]
oliver3: For this page, the `useUserListRequest` is not part of some sort of context from a parent element, so the requestState.data will be empty every time the page is loaded. I don't think it is worth to store that somewhere for this one scenario, also because we'd have to add code to refresh in case it _is_ saved etc.

praseodym (approved):

PR #1033, opened 2025-02-13
Create user 409 Conflict for duplicate username

oliver3:
This is not a proper solution

### DIFF
--- a/backend/src/error.rs
+++ b/backend/src/error.rs
@@ -319,7 +319,10 @@ impl From<DataEntryTransitionError> for APIError {
 
 impl From<AuthenticationError> for APIError {
     fn from(err: AuthenticationError) -> Self {
-        APIError::Authentication(err)
+        match err {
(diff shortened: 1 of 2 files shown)


backend/src/authentication/user.rs
```diff
@@ -23,21 +23,21 @@ const MIN_UPDATE_LAST_ACTIVITY_AT_SECS: i64 = 60; // 1 minute
 /// User object, corresponds to a row in the users table
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Hash, FromRow, ToSchema)]
 pub struct User {
-    id: u32,
-    username: String,
+    pub id: u32,
```
cikzh, line 26: In this specific case I would not make them `pub`, but instead create getter methods in the `impl` block. This way we can keep them "read-only" and even but them behind `#[cfg(test)]`'s.

cikzh: See line 75 for an example for `fullname` which is read in the test like so: `user.fullname()`

Issue #997, opened 2025-02-10
Backend validation


oliver3:
Discuss and implement what the backend should validate for create and update (password length, fullname mandatory for certain roles)

cikzh:
I'll create a draft PR to do the plumbing. Let's plan a meeting asap to discuss this

PR #1036, opened 2025-02-14
Backend password validation

cikzh:
Just contains the plumbing and example validations. Actual validation implementation is to be determined and tests will be added when the validation rules are established

See #997 and #1021


backend/src/authentication/password.rs
```diff
@@ -1,25 +1,74 @@
...
+        if password.len() < MIN_PASSWORD_LEN {
+            return Err(AuthenticationError::InvalidPassword);
+        }
+
+        match old_password {
+            Some(old_pw) if verify_password(password, old_pw) => {
+                Err(AuthenticationError::InvalidPassword)
+            }
+            Some(_) | None => Ok(Self(password)),
+        }
```
praseodym: Looks like this is missing tests for the unhappy path (password too short or new password equal to old password)

cikzh: I haven't tested anything. Whoops! Putting back in draft :)

backend/src/authentication/api.rs
```diff
@@ -241,6 +241,49 @@ pub async fn development_create_user(
     Ok(StatusCode::CREATED)
 }
 
+/// Development endpoint: login as a user (unauthenticated)
```
praseodym, line 244: #1043 removed this endpoint (merged earlier today), I don't think we should be bringing it back in this PR?

backend/src/authentication/password.rs
```diff
@@ -1,25 +1,80 @@
...
 use argon2::{
     Algorithm, Argon2, Params, Version,
     password_hash::{PasswordHash, PasswordHasher, PasswordVerifier, SaltString, rand_core::OsRng},
 };
+use serde::Deserialize;
+use sqlx::Type;
 
 use super::error::AuthenticationError;
 
+pub(super) struct ValidatedPassword<'a>(&'a str);
```
praseodym, line 12: Can you add a docstring for this struct?

backend/src/authentication/password.rs
```diff
@@ -1,25 +1,80 @@
...
+        if password.len() < MIN_PASSWORD_LEN {
+            return Err(AuthenticationError::PasswordRejection);
+        }
+
+        // Password cannot be the same as the username
+        if username == password {
+            return Err(AuthenticationError::PasswordRejection);
+        }
+
+        match old_password {
```
praseodym, line 32: For consistency I'd expect a comment here

backend/src/authentication/password.rs
cikzh: [698a8b1](https://github.com/kiesraad/abacus/pull/1036/commits/698a8b19990cb27cad271b967f301214898f7f1b)

backend/src/authentication/password.rs
cikzh: Added in [698a8b1](https://github.com/kiesraad/abacus/pull/1036/commits/698a8b19990cb27cad271b967f301214898f7f1b)

cikzh:
@praseodym thanks for the feedback! I've added some comments and docstrings in other places as well

praseodym (approved):

Lionqueen94 (dismissed):

backend/src/authentication/api.rs
```diff
@@ -276,6 +276,7 @@ pub struct CreateUserRequest {
 }
 
 #[derive(Serialize, Deserialize, ToSchema)]
+#[serde(deny_unknown_fields)]
```
oliver3, line 279: Do we maybe want to configure this globally if possible?

oliver3 (dismissed):

oliver3 (approved):

praseodym (approved):

PR #1043, opened 2025-02-17
Authorization checks

### LINKED ISSUE #999: Create Rust user types for roles

### LINKED ISSUE #1000: User selection in frontend for testing/development

### LINKED ISSUE #1001: Authorization checks in backend API handlers

### LINKED ISSUE #1003: Conditional component rendering based on user role
Conditional component rendering based on user role. Mock `whoami` endpoint for Vitest where needed.

### LINKED ISSUE #1004: Show user and role in navigation bar

### LINKED ISSUE #1005: Authentication in Playwright tests
Introduce authentication with correct role in all Playwright tests.

Use storage state to store session after logging in through browser.

marlonbaeten:
- Adds authorization to backend endpoints via `FromRequestParts` of a `User`. Newtypes around `User`, e.g. Typist(pub User), provide separation of roles. 
- API requests in backend tests now provide a session cookie where needed
- Updated OpenAPI spec with 401 response where applicable
- API requests in e2e tests now provide session cookies with the correct roles
- You can login as different roles via the dev UI
- User and role are now shown in the navigation bar

To quickly login as one of the roles, use the buttons on the `/dev` page:

image

Applicable links will appear on this page once logged in.

To test:

Use the application as before, but login as one of the roles. Try to find unprotected API endpoints! There are still front-end routes that do render when not logged in as the applicable role, but their API call's will result in an error.

Note that this is a large PR, but a lot of the issues listed below are closely related or have co-dependencies.

Fixes #999
Fixes #1000
Fixes #1001
Fixes #1003
Fixes #1004
Fixes #1005


praseodym:
The links to `/users` and `/workstations` should be hidden from the navbar (links and menu) for the coordinator role.

cikzh (dismissed):

cikzh (dismissed):

frontend/app/module/election/page/ElectionHomePage.tsx
```diff
@@ -1,14 +1,19 @@
...
-import { useElection } from "@kiesraad/api";
+import { useElection, useUser } from "@kiesraad/api";
 import { t } from "@kiesraad/i18n";
 import { PageTitle } from "@kiesraad/ui";
 
 export function ElectionHomePage() {
+  const user = useUser();
   const { election } = useElection();
 
+  if (user?.role === "typist") {
```
oliver3, line 13: It seems that `useUser` is only used to check for the user role? Maybe we can repurpose it to something like 
```ts
const {isTypist, isAdministrator, isCoordinator } = useUserRole()
```

frontend/e2e-tests/state/admin.json
```diff
@@ -0,0 +1,15 @@
+{
```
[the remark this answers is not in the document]
marlonbaeten: Fixed

backend/openapi.json
```diff
@@ -2164,9 +2365,17 @@
         "type": "object",
         "required": [
           "user_id",
-          "username"
+          "fullname",
```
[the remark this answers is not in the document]
marlonbaeten: Fixed in 29d2235

frontend/app/module/election/page/ElectionHomePage.tsx
marlonbaeten: Fixed in 29d2235

backend/src/authentication/api.rs
```diff
@@ -28,14 +28,21 @@ pub struct Credentials {
...
+    pub fullname: String,
     pub username: String,
+    pub role: Role,
 }
 
 impl From<&User> for LoginResponse {
     fn from(user: &User) -> Self {
         Self {
             user_id: user.id(),
+            fullname: user
```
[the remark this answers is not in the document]
marlonbaeten: Fixed in 29d2235

backend/src/authentication/api.rs
```diff
@@ -28,7 +28,8 @@ pub struct Credentials {
 #[derive(Debug, Serialize, Deserialize, ToSchema)]
 pub struct LoginResponse {
     pub user_id: u32,
-    pub fullname: String,
+    #[schema(value_type = String, nullable = true)]
```
oliver3, line 31:
```suggestion
    #[schema(value_type = String, nullable = false)]
```

backend/src/authentication/user.rs
```diff
@@ -26,7 +26,7 @@ pub struct User {
     id: u32,
     username: String,
     #[serde(skip_serializing_if = "Option::is_none")]
-    #[schema(nullable = false)]
+    #[schema(value_type = String, nullable = true)]
```
oliver3, line 29:
```suggestion
    #[schema(value_type = String, nullable = false)]
```

backend/src/authentication/api.rs
marlonbaeten: Fixed in 293485a

cikzh (approved):

praseodym (changes requested):
The frontend doesn't handle authorization errors well. I'd expect to be redirected to the login page instead of seeing an empty page or error page. For example:
- `/elections` => error page, 401
- `/account/change-password` => empty page

Furthermore, the 'Afmelden' navbar link logs out immediately which doesn't work well on the data entry form. When I click 'Afmelden', the 'Wat wil je doen met je invoer?' modal is shown but the session is already terminated. Clicking 'Invoer bewaren' or 'Niet bewaren' in the modal no longer works at that point.

marlonbaeten:
> The frontend doesn't handle authorization errors well. I'd expect to be redirected to the login page instead of seeing an empty page or error page. For example:
> 
> * `/elections` => error page, 401
> * `/account/change-password` => empty page
> 
> Furthermore, the 'Afmelden' navbar link logs out immediately which doesn't work well on the data entry form. When I click 'Afmelden', the 'Wat wil je doen met je invoer?' modal is shown but the session is already terminated. Clicking 'Invoer bewaren' or 'Niet bewaren' in the modal no longer works at that point.

Thanks for your review!

I have added a redirect to the login page if the error boundary catches a 401 error and added explicit checks+redirects for pages where there is no API request performed. This should solve your first issue.

The second issue is solved by having an explicit logout route and component. The component will logout the user only after navigating.

See commit 7f43bb7da9ab0f5b158f82cf35f8485d62394803

praseodym (approved):
lgtm, test coverage could be a bit higher but is ok for now

PR #1060, opened 2025-02-20
Add user get endpoint and some user integration tests

oliver3:
Closes https://github.com/kiesraad/abacus/issues/1059

### DIFF
--- a/backend/tests/user_integration_test.rs
+++ b/backend/tests/user_integration_test.rs
@@ -3,7 +3,7 @@
 use abacus::authentication::UserListResponse;
 use hyper::{header::CONTENT_TYPE, StatusCode};
 use reqwest::Body;
-use serde_json::json;
+use serde_json::{json, Value};
 use sqlx::SqlitePool;
 use test_log::test;
 use utils::serve_api;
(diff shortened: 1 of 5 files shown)


backend/tests/user_integration_test.rs
```diff
@@ -59,3 +59,111 @@ async fn test_user_last_activity_at_updating(pool: SqlitePool) {
...
+        response.status(),
+        StatusCode::OK,
+        "Unexpected response status"
+    );
+    let body: UserListResponse = response.json().await.unwrap();
+    assert_eq!(body.users.len(), 1);
+    assert!(body.users.iter().any(|ps| ps.username() == "user"))
+}
+
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("users"))))]
```
cikzh, line 80: We don't need the fixture for this test

backend/tests/user_integration_test.rs
```diff
@@ -59,3 +59,111 @@ async fn test_user_last_activity_at_updating(pool: SqlitePool) {
...
+    let body: Value = response.json().await.unwrap();
+    dbg!(&body);
+
+    assert_eq!(body["id"], 1);
+    assert_eq!(body["role"], "administrator");
+    assert_eq!(body["username"], "user");
+    assert_eq!(body["fullname"], "Sanne Molenaar");
+}
+
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("users"))))]
```
cikzh, line 161: We could also leave out this fixture, any ID param should return a 404 then

backend/tests/user_integration_test.rs
```diff
@@ -59,3 +59,111 @@ async fn test_user_last_activity_at_updating(pool: SqlitePool) {
...
+    let body: Value = response.json().await.unwrap();
+    dbg!(&body);
+
+    assert_eq!(body["role"], "administrator");
+    assert_eq!(body["username"], "username");
+    assert_eq!(body["fullname"], "fullname");
+    assert!(body.get("temp_password").is_none());
+}
+
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("users"))))]
```
cikzh, line 112: We don't need this one as well

cikzh (changes requested):
LGTM, the only thing I would change is leaving out the loading of user fixtures where we don't need them

backend/tests/user_integration_test.rs
oliver3: Thanks! I removed all the fixtures as you've suggested.

Lionqueen94 (approved):

praseodym (approved):

PR #1065, opened 2025-02-21
Converted apportionment backend tests to include candidate votes

Lionqueen94:


Lionqueen94:
> This PR makes all existing tests more complicated even though the candidate votes aren't used for these scenarios. Why not change the helper function to create 50 candidates per political group, with all votes going to the first candidate? That way the tests can stay the same and list exhaustion shouldn't have any effects on these scenarios once we implement it.

Good point, I'll add the function I created later on when I implement seat assignment to candidates, for now I give every party 50 candidates with only votes for the first one as you suggested

praseodym (approved):

praseodym (dismissed):

cikzh (approved):

praseodym (approved):
lgtm, the function name `get_election_summary_with_default_50_candidates` could be a bit shorter (you can explain what a function does in a docstring, not everything has to go in the name)

cikzh (approved):

PR #1086, opened 2025-02-24
Add update user form and page

oliver3:
Resolves issue #931 

User management: update user
- fullname can be updated if it has been set before
- password can be updated after revealing input by clicking the button
- username and role cannot be updated
- Save will update user in the backend and go to user list page with an appropriate message

NB: Delete user will be done in issues #1084 and #1085


jorisleker (changes requested):
Bij niet-anonieme gebruikers staat in het design de volledige naam van de gebruiker in de h1. In de implementatie wordt altijd de username getoond.

oliver3:
> Bij niet-anonieme gebruikers staat in het design de volledige naam van de gebruiker in de h1. In de implementatie wordt altijd de username getoond.

Die had ik gemist thanks, opgelost in https://github.com/kiesraad/abacus/pull/1086/commits/42d188acc4fdbd942c385bf4ebffe7d44ad7b365

frontend/app/module/users/update/UserUpdateForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    setEditPassword(true);
+                  }}
+                >
+                  <IconPencil />
+                  {t("users.change_password")}
+                </Button>
+              </div>
+            )}
+
+            <div className="mb-lg">
```
lkleuver, line 104: Maybe this can be something in FormLayout?

lkleuver (commented):
lgtm once test fix is merged.

lkleuver (approved):

jorisleker (changes requested):
Previously requested change regarding the h1 looks good now!

Found something else though: I can create two users with the following usernames:
- Anoniemus1
- anoniemus1

This will **not** result in a 'username not unique error'. I think the username should be case insenstive, so these names should be considered the same.

If I create users with exactly the same name, I get a 500 server error, instead of a 'this username already exists' error.

oliver3:
> If I create users with exactly the same name, I get a 500 server error, instead of a 'this username already exists' error.
> 
> edit: strictly speaking, this is out of scope for your issue because the bug is already in the codebase, so feel free to ignore this review and fix this in another pull request.

I cannot reproduce the Error 500..

I've added a new issue to discuss and implement the case (in)sensitivity of the username: https://github.com/kiesraad/abacus/issues/1091

jorisleker (approved):

PR #1097, opened 2025-02-27
Rename user.json to account.json

oliver3:
Rename and sort the `user.json` i18n file.


Lionqueen94 (approved):

praseodym (approved):
Maybe the component location (`component/form/user/..`) should also be changed?

oliver3:
> Maybe the component location (`component/form/user/...`) should also be changed?

Yes good idea, there will be further changes so I will keep it in mind

Issue #1091, opened 2025-02-26
Case (in)sensitive username


oliver3:
Found something else though: I can create two users with the following usernames:
- Anoniemus1
- anoniemus1

This will **not** result in a 'username not unique error'. I think the username should be case insenstive, so these names should be considered the same.

If I create users with exactly the same name, I get a 500 server error, instead of a 'this username already exists' error. 

_Originally posted by @jorisleker in https://github.com/kiesraad/abacus/pull/1086#pullrequestreview-2644791023_

oliver3:
@jorisleker Would the following work?
- Store username case sensitive, show in lists, header etc with the correct casing
- Compare username case insensitive for: create user (duplicate username), login

jorisleker:
> [@jorisleker](https://github.com/jorisleker) Would the following work?
> 
> * Store username case sensitive, show in lists, header etc with the correct casing
> * Compare username case insensitive for: create user (duplicate username), login

Yes, that would work!

PR #1125, opened 2025-03-04
Return 409 Conflict for duplicate username

### LINKED ISSUE #1022: Return 409 Conflict for duplicate username

marlonbaeten:
Fixes #1022

This makes sure there are no duplicate usernames (case-insensitive) and that trying to create a user with a duplicate username results in a 409 server response.

Test by create two users with the same username.


praseodym:
Logging in is still case sensitive, I think that will be resolved by #1091?

praseodym (approved):

rnijveld (dismissed):

praseodym (changes requested):
Please also add the 409 to the OpenAPI spec (thanks @oliver3)

oliver3 (approved):

PR #1144, opened 2025-03-07
Add user id to the first and second data entry

### LINKED ISSUE #1128: Expand data entry state with user information
- Add user id to data entry state enum structs (similar to progress percentage)
- Track which user id is doing/has done first data entry and second data entry

praseodym:
- Move data entry API handlers to a separate module
- Add user id to the first and second data entry
  And return an error when updating a data entry as a different user.

Resolves #1128


backend/src/data_entry/api.rs
```diff
@@ -0,0 +1,662 @@
+use crate::data_entry::status::CurrentDataEntry;
+use crate::{
+    APIError,
+    authentication::{Typist, User},
+    data_entry::{
+        PollingStationResults, ValidationResults,
+        entry_number::EntryNumber,
+        repository::PollingStationDataEntries,
+        status::{ClientState, DataEntryStatus, DataEntryStatusName},
```
Lionqueen94, line 9:
```suggestion
        status::{ClientState, CurrentDataEntry, DataEntryStatus, DataEntryStatusName},
```

backend/src/data_entry/api.rs
```diff
@@ -0,0 +1,662 @@
+use crate::data_entry::status::CurrentDataEntry;
```
Lionqueen94, line 1:
```suggestion
```

backend/src/data_entry/api.rs
praseodym: Changed in 851f7d1af407f09060766bd3002c4e1a710ef432

backend/src/data_entry/api.rs
praseodym: Changed in 851f7d1af407f09060766bd3002c4e1a710ef432

cikzh (approved):

rnijveld (approved):

PR #1154, opened 2025-03-10
Audit log UI

### LINKED ISSUE #1078: Audit log frontend page (excl. filters)

marlonbaeten:
Tweaks the audit log functionality and adds an UI (Audit log event table).

Fixes #1078

Not that the filter button is part of issue #1079

To test:

Login as administrator or coordinator and view the `/logs` page.
Currently only audit log events are added when logging-in or -out, so you could do this a couple of times to get a longer list. Events are paginated (with 10 per page). The details button also shows the metadata and IP address of events.


jorisleker:
@marlonbaeten lekker bezig.

Small remark: can you replace the 'details' button with a chevron, just like we do in other tables?

jorisleker:
Some questions:

**Table**
- Why was the column 'Gebeurtenis' added? Seems to me that the content used there should be in the column 'Toelichting'. Saves some space / an empty column.

**Modal**
- Does the modal need a primary call to action? Don't think so; the 'X' might be enough. Saves some space
- In the details modal there is space to display the full date/time: 20 maart 2025 12:23:01 (full monthname, year and time including seconds).
- What does 'Aantal ingelogde gebruikers' mean? (now displays '0' all the time). Is the number of sessions the user has? If so, would change the copy to 'Aantal ingelogde sessies'
- 'Sessieduur' is '0', What will we display here? In what unit? seconds, minutes, hours?

marlonbaeten:
> Some questions / suggestions:
> 
> **Table**
> 
> * Why was the column 'Gebeurtenis' added? Seems to me that the content used there should be in the column 'Toelichting'. Saves some space / an empty column.

'Gebeurtenis' is the audit event type, which is always present, toelichting is an optional message which is used for some audit events. I could move this to the details section however.

> **Modal**
> 
> * Does the modal need a primary call to action? Don't think so; the 'X' might be enough. Saves some space

Changed in 494944d 👍🏻 

> * In the details modal there is space to display the full date/time: 20 maart 2025 12:23:01 (full monthname, year and time including seconds).

Check - added a new date time format method in 494944d 👍🏻 

> * What does 'Aantal ingelogde gebruikers' mean? (now displays '0' all the time). Is the number of sessions the user has? If so, would change the copy to 'Aantal ingelogde sessies'

The number of sessions active at that time (excluding the new session). We will use this for statistics in the future.

> * 'Sessieduur' is '0', What will we display here? In what unit? seconds, minutes, hours?

Also used for statistics in the future, the unit is seconds. We show these fields in de audit log since any additional recorded data might be relevant dor diagnostics. There was a bug in this statistic that is fixed in 494944d, also added the unit to the label.

frontend/lib/util/format.ts
```diff
@@ -66,25 +66,43 @@ function isYesterday(date: Date): boolean {
   );
 }
 
-export function formatDateTime(date: Date) {
-  const today = new Date();
+export function formatDateTimeFull(date: Date) {
```
praseodym, line 69: Can you add a simple unit test for this function?

jorisleker:
> The number of sessions active at that time (excluding the new session). We will use this for statistics in the future.

Active sessions for the current user, active sessions for all users (including other sessions for this user), or something else? The label isn't very clear in this situation

marlonbaeten:
> > The number of sessions active at that time (excluding the new session). We will use this for statistics in the future.
> 
> Active sessions for the current user, active sessions for all users (including other sessions for this user), or something else? The label isn't very clear in this situation

"Aantal ingelogde gebruikers"  = total number of logged in users / active sessions at that time. Any suggestions for a more clear label?

jorisleker:
> 'Gebeurtenis' is the audit event type, which is always present, toelichting is an optional message which is used for some audit events. I could move this to the details section however.

Yeah would move the toelichting to the details section.

Noticed that we start paginating when there are more then 10 events now. Would recommend showing at least 200 or more items on one page (as much as we can without slowing down the page to much). That way, filtering + control F will have some effect when looking for specific events.

frontend/lib/util/format.ts
marlonbaeten: Added some unit tests in [b0e0d9f](https://github.com/kiesraad/abacus/pull/1154/commits/b0e0d9fa7ad4b5ac44992f14699b0d5298c00d0e)

marlonbaeten:
@jorisleker thanks for the feedback!

- I've changed the metric to include the newly created session. That would make more sense. We need the metric on the login event since we want to keep track of the maximum number of concurrent sessions.
- The "toelichting" field is removed from the table in b0e0d9f
- The default items per page is changed to 200, although I could make this user configurable in the front-end when implementing the filter functionality (item per page can be passed the the api endpoint).

jorisleker (approved):
👍

Issue #1019, opened 2025-02-11
Direct user to change password screen


cikzh:
When the `needs_password_change` flag is set, the user should be directed to the change password screen.

Depends on #873

Issue #792, opened 2025-01-05
Epic: user management


praseodym:
# Context

[Design in Figma](https://www.figma.com/design/xHDfsv69Nhmk3IrWC0303B/Public---Kiesraad---Abacus-Software-voor-verkiezingsuitslagen-en-zetelverdeling?node-id=280-21844&t=hyq09FyDbA4Ebug8-4)

# Architecture

# Scope

Related to #673, #676 

In scope:
- List of users
  - username, role, full name, last activity
- Create single user - wizard, see Figma
  - Roles: typist (invoerder), coordinator, administrator
  - Anonymous account: full name field not shown, user must add full name on first login
  - Username
  - Full name has to filled out for admin & coordinator accounts
  - Input field for temporary password (user must change password on first login), with minimum length
- Update single user (admin):
  - One "form" screen 
  - read-only username
  - change full name
  - change role
  - change password button that shows temp password field in-place 
  - Delete user, with confirmation modal

Out of scope: 
- Login instructions button
- Create multiple users
- Import users
- Generating temporary password
- ~~Login (#673)~~
- Profile pages for logged in user
- New adjustment to password policy (kiesraad/abacus-internal#283)

jschuurk-kr:
@oliver3 @cikzh @jorisleker For the user page, will there be a search and/or filter option?

With the large number of users in a municipality and the separation of CSB and GSB coordinators and typists, I do think we need something to make it easier to navigate the list.

praseodym:
Currently there is no search and/or filter included in the design. It could be a nice addition, but IMO not a must have that we should implement now. @Lionqueen94 also asked about search/filtering on the polling stations page, see [comment in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=5442-50666#994218964).

jorisleker:
Search/filter is not foreseen here, and not worth the effort IMO. Number of users will probably never exceed ca 60-70, and in most cases not more then 5-10. As long as we keep it on one page, scrolling (or using control+F) will be much easier and faster than finding and discovering how to use a custom search functionality.

jorisleker:
Use case on the polling station page could be stronger, but even there, not sure our user benefit would outweigh our dev effort.

PR #1156, opened 2025-03-10
Rejecting users that need to change their password

### LINKED ISSUE #1020: User should not be able to do anything when the `needs_password_change` flag is set

cikzh:
Fixes #1020

- Adds a check in `(Optional)FromRequestParts` to check if the user needs to change their password
- Updates user fixtures to explicitly set `needs_password_change` to false
- Adds integration test

I'm not sure about the implementation of the exception for the /user/account path in `FromRequestParts`, let me know if there is a better way


jschuurk-kr:
@cikzh How does the user experience this? Do they get redirected to a page that explains they need to change their password? Or is that part of the user management epic #792?

rnijveld:
Why are we not doing this just at login? Once someone is logged in I wouldn't want their authentication credentials to matter anymore? And if we do I would recommend we create a 'remove sessions for user' functionality instead that is called whenever a password reset is required?

cikzh:
Decided that we'll implement it this way. Frontend implementation in this issue: #1158

oliver3:
Frontend implementation has been done in  #1019

frontend/e2e-tests/authentication.e2e.ts
jschuurk-kr, line 1: We need an e2e tests for first login where you have to set a new password.

frontend/e2e-tests/authentication.e2e.ts
```diff
@@ -8,10 +8,10 @@ test.describe("authentication", () => {
     await page.getByLabel("Wachtwoord").fill("AdminPassword01");
     await page.getByRole("button", { name: "Inloggen" }).click();
 
-    await page.waitForURL("/account/setup");
+    await page.waitForURL("/elections");
 
     // TODO: use new page object when we know which page to render
-    await expect(page.getByRole("alert")).toContainText("Inloggen gelukt");
+    await expect(page.getByText("Sanne Molenaar(Beheerder)")).toBeVisible();
```
jschuurk-kr, line 14: Probably better to do this in a separate issue/PR, but it looks like it's time to introduce a page object for the top nav bar, with the other page objects inheriting from it.

`DataEntryBasePgObj.ts` has a `clickElectionInNavBar`, which then should be moved to that new page object.

cikzh: I've looked at it briefly, but I agree that it's probably best to do it in a separate PR :). I've created https://github.com/kiesraad/abacus/issues/1178 for this

frontend/e2e-tests/authentication.e2e.ts
cikzh: Added a test in https://github.com/kiesraad/abacus/issues/1178

frontend/e2e-tests/page-objects/NavBarPgObj.ts
jschuurk-kr, line 1: Good start, what still needs to be done:
- make page objects inherit from this one, so we don't use this page object directly in tests
- create a better locator for the user name and role, which will need a change to the page

Up to you if you want to do that in this PR, or in issue https://github.com/kiesraad/abacus/issues/1178.

cikzh: I'm going to do it in a separate PR, removed the issue from the Fixes in the description so it won't be closed

cikzh: Also updated #1178's description with this feedback

Lionqueen94 (approved):

frontend/e2e-tests/authentication.e2e.ts
```diff
@@ -23,4 +30,57 @@ test.describe("authentication", () => {
...
+    // Login as admin and create a user
+    await page.goto("/account/login");
+    await page.getByLabel("Gebruikersnaam").fill("admin");
+    await page.getByLabel("Wachtwoord").fill("AdminPassword01");
+    await page.getByRole("button", { name: "Inloggen" }).click();
+
+    const navBarPgObj = new NavBarPgObj(page);
+    await expect(navBarPgObj.logout).toBeVisible();
+
+    const username = createRandomUsername("Invoerder");
```
oliver3, line 44: You can use the `user` fixture instead of creating a new user through the UI, like for instance [here](https://github.com/kiesraad/abacus/blob/main/frontend/e2e-tests/users.e2e.ts#L92)

oliver3 (changes requested):

frontend/e2e-tests/authentication.e2e.ts
cikzh: I explicitly opted out of using the user fixtures, as I want to be absolutely sure I'm using a freshly created user. If we make changes in the user creation flow in the future, I want this test to fail.

backend/src/authentication/user.rs
```diff
@@ -155,11 +155,12 @@ where
         };
 
         match users.get_by_session_key(session_cookie.value()).await {
-            Ok(user) => {
+            Ok(user) if !user.needs_password_change => {
```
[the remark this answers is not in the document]
cikzh: This check if left after changing the implementation and needs to be removed.

frontend/e2e-tests/authentication.e2e.ts
```diff
@@ -23,4 +30,57 @@ test.describe("authentication", () => {
...
+    // Logout
+    await navBarPgObj.logout.click();
+
+    // Login as the newly created user
+    await page.goto("/account/login");
+    await page.getByLabel("Gebruikersnaam").fill(username);
+    await page.getByLabel("Wachtwoord").fill(tempPassword);
+    await page.getByRole("button", { name: "Inloggen" }).click();
+
+    // Fill out the account setup page
```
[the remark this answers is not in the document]
cikzh: I already have this (and other candidates) fixed in a different branch. See this discussion: https://github.com/kiesraad/abacus/pull/1156#discussion_r1998466909

frontend/e2e-tests/page-objects/NavBarPgObj.ts
```diff
@@ -0,0 +1,16 @@
...
+export class NavBarPgObj {
+  readonly logout: Locator;
+  readonly navigation: Locator;
+
+  constructor(protected readonly page: Page) {
+    this.navigation = page.getByRole("navigation", { name: "primary-navigation" });
+    this.logout = this.navigation.getByRole("link", { name: "Afmelden" });
+  }
+
+  async clickElectionInNavBar(electionLocation: string, electionName: string) {
```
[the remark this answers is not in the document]
cikzh: Good idea, I'll do the fix in #1178. I'll add a ToDo in that issue

backend/src/authentication/user.rs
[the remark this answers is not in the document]
cikzh: Removed in [b7ab061](https://github.com/kiesraad/abacus/commit/b7ab061331fa13994e325fac1363b00db28057d4)

marlonbaeten (dismissed):

marlonbaeten (approved):

oliver3 (approved):

Lionqueen94 (approved):

PR #1159, opened 2025-03-11
Remove client side password validation from user management pages

### LINKED ISSUE #1140: Use backend password validation only, for create and update user
Remove `validatePassword.ts` and use backend validation only

oliver3:
Resolves #1140 

For both create and update user:
- Moved api call logic from page to form to be able to handle `PasswordRejection` api error
- Remove password length validation code
- Update tests accordingly

Can be tested by creating and updating a user in user management. The current password length requirement in the backend is 13 chars.


frontend/app/module/users/create/UserCreateDetailsForm.tsx
```diff
@@ -0,0 +1,145 @@
...
+type ValidationErrors = Partial<CreateUserRequest>;
+
+export function UserCreateDetailsForm({ role, showFullname, onSubmitted }: UserCreateDetailsFormProps) {
+  const [validationErrors, setValidationErrors] = useState<ValidationErrors | null>(null);
+
+  const { create, requestState } = useCrud<User>("/api/user" satisfies USER_CREATE_REQUEST_PATH);
+  const [usernameUniqueError, setUsernameUniqueError] = useState<string | undefined>(undefined);
+  const [error, setError] = useState<AnyApiError>();
+
+  if (error && !(error instanceof ApiError)) {
```
lkleuver, line 31: doesn't useCrud handle this with a param?

lkleuver (approved):
LGTM with small comment

frontend/app/module/users/create/UserCreateDetailsForm.tsx
oliver3: I think useCrud() should already throw indeed, but TypeScript still thinks that `ApiResult` can be `AnyApiError`. I will try and fix this in a separate issue.

PR #1187, opened 2025-03-17
Rename and restructure api hooks

oliver3:
- Rename `useApi` to `useApiClient`
- Rename `useApiRequest` to `useInitialApiGet`
- Move and refactor api types


frontend/lib/api/useInitialApiGet.ts
```diff
@@ -0,0 +1,91 @@
...
+      controller.abort(DEFAULT_CANCEL_REASON);
+    };
+  }, [client, path, throwFatalErrors]);
+
+  return {
+    requestState,
+    refetch,
+  };
+}
+
```
Lionqueen94, line 83: it would be nice to add a comment that fatal errors are returned by this function

oliver3: I've tried to give a better description for the two flavours in https://github.com/kiesraad/abacus/pull/1187/commits/7537c0949f942b57ffbc28745776d5298c5bbab5

Lionqueen94 (approved):

PR #1200, opened 2025-03-18
scroll to top if errors

### LINKED ISSUE #1113: Data-entry: pressing "volgende" twice does not scroll to top
when triggering an error, and clicking 'Volgende' right away without making changes, the viewport doesn't scroll up to show the error. This can be confusing because there is no visual feedback about the error on the bottom of the page.

lkleuver:
This change scrolls to top when there are errors when you press the "volgende" button a second time without changing something.

It's a minor tweak discovered in the refactoring of data entry.
Note that it should not scroll on warnings due to the "accept warnings" checkbox being at the bottom of the screen.

How to test:
go to voters and votes and add numbers that don't add up, press "volgende", scroll down and press it again.

closes #1113

### DIFF
--- a/frontend/app/component/form/data_entry/state/useDataEntryFormSection.ts
+++ b/frontend/app/component/form/data_entry/state/useDataEntryFormSection.ts
@@ -58,7 +58,12 @@ export function useDataEntryFormSection<FORM_VALUES>({
     data: Partial<PollingStationResults>,
     options?: SubmitCurrentFormOptions,
   ): Promise<boolean> => {
-    return await onSubmitForm(data, { ...options, showAcceptWarnings });
+    const result = await onSubmitForm(data, { ...options, showAcceptWarnings });
+    if (formSection.errors.length) {
+      // scroll to top when there are errors, this is mainly necesarry when users click "volgende" a second time without chaning anything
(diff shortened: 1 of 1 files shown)


jorisleker (approved):
Tested on voters and votes and on a candidate list. Works as expected on both

frontend/app/component/form/data_entry/state/useDataEntryFormSection.ts
```diff
@@ -58,7 +58,12 @@ export function useDataEntryFormSection<FORM_VALUES>({
     data: Partial<PollingStationResults>,
     options?: SubmitCurrentFormOptions,
   ): Promise<boolean> => {
-    return await onSubmitForm(data, { ...options, showAcceptWarnings });
+    const result = await onSubmitForm(data, { ...options, showAcceptWarnings });
+    if (formSection.errors.length) {
+      // scroll to top when there are errors, this is mainly necesarry when users click "volgende" a second time without chaning anything
+      window.scrollTo(0, 0);
```
oliver3, line 64: Is the `window.scrollTo(0, 0);`  in the effect below now redundant?

oliver3 (approved):

PR #1203, opened 2025-03-18
Add user audit events to authentication module

marlonbaeten:
WIP


backend/src/lib.rs
```diff
@@ -125,8 +127,8 @@ pub fn router(pool: SqlitePool) -> Result<Router, Box<dyn Error>> {
 
     let app = app
         .layer(TraceLayer::new_for_http())
-        .layer(middleware::from_fn_with_state(
-            pool.clone(),
+        .layer(middleware::map_response_with_state(
```
marlonbaeten, line 130: Note that our extend-session middleware now is  a `map_response_with_state` which is better suited than the `middleware::from_fn_with_state`, since we can inject FromRequest types in `map_response_with_state`.

praseodym:
Is there a way to reduce the number of parameters passed to the API route handler functions? Abacus already scores below average on the 'Unit Interfacing' metric in Sigrid, and adding the `AuditService` parameter to many more handlers makes it score even worse.

cikzh (dismissed):

praseodym:
This PR has merge conflicts (from merging #1216).

praseodym:
> > Is there a way to reduce the number of parameters passed to the API route handler functions? Abacus already scores below average on the 'Unit Interfacing' metric in Sigrid, and adding the `AuditService` parameter to many more handlers makes it score even worse.
> 
> The problem is that the `AuditService` is constructed from the current request, since it has to keep track of, for instance, the current IP address. So we can't make it a part of the state.
> 
> I do not think that dependency injection is, per definition, a bad thing.

It makes sense and agreed that dependency injection is not a bad thing, but it would be better if we could do it without adding more and more parameters to each API handler. For now this is fine, though.

cikzh (approved):

backend/src/authentication/api.rs
```diff
@@ -243,28 +258,40 @@ async fn logout(
...
-    let mut res = next.run(req).await;
+pub async fn extend_session(
+    State(users): State<Users>,
+    State(sessions): State<Sessions>,
+    jar: CookieJar,
+    audit_service: AuditService,
+    mut response: Response,
+) -> Response {
+    // let jar = CookieJar::from_headers(req.headers());
+    // let mut res = next.run(req).await;
```
praseodym, line 269: remove this?

praseodym (approved):

cikzh (approved):

Issue #1137, opened 2025-03-06
Add RTL tests for hiding the option to accept warnings after making a change in data entry


jschuurk-kr:
When one of the data entry pages has a warning, we show a checkbox with the text "Ik heb de aantallen gecontroleerd met het papier en correct overgenomen." If you resubmit, an additional error is shown "Je kan alleen verder als je het papieren proces-verbaal hebt gecontroleerd.". These should disappear, as soon as you make a change to any data entry field.

While this behavior is implemented correctly, we currently don't have tests for this. So for each of the pages that can display warnings (Recounted, VotersAndVotes and Differences) we should implement two RTL tests:
- the checkbox with text disappears upon change
- the checkbox with text and the error message disappear upon change 

Assumption is that whether the checkbox is checked or not, has no effect on the behavior.

jschuurk-kr:
In the model-based e2e tests for warning we have the state "voterVotesPageWarningCorrected" which is reached by submitting data that results in a warning, then changing the data to be valid. That state checks if:
- the warning message is visible
- the accept warnings checkbox and text are hidden

So what's not covered is (1) the error message about accepting warnings; (2) the behavior of any other page than the Voters and Votes page.

PR #1205, opened 2025-03-18
Add tests to check that the accept warnings checkbox disappears

oliver3:
Resolves https://github.com/kiesraad/abacus/issues/1137
- Add `aria-describedby` to Alert that points to alert children
- Fix text: "het ~~het~~ papieren proces-verbaal"
- Add tests to check that the accept warnings checkbox disappears if some data will be entered on the page.


frontend/lib/ui/Alert/Alert.tsx
```diff
@@ -16,13 +16,15 @@ export interface AlertProps {
 }
 
 export function Alert({ type, onClose, children, margin, variant = "default" }: AlertProps) {
+  const id = React.useId();
+
   return (
-    <div className={`${cn(cls.alert, cls[type], variant)} ${margin}`} role="alert">
+    <div className={cn(cls.alert, cls[type], margin, variant)} role="alert" aria-describedby={id}>
```
Lionqueen94, line 22: I don't think margin can be in cn() since it's a css class described in `util.css`

oliver3: I've checked and the value of `margin` can be one of the strings `mb-md`, `mb-md-lg` and `mb-lg`, which are all classes in the `util.css`

Lionqueen94: Ah I thought it won't work to use classes from util.css in `cn()`, but if that does work then perfect

frontend/lib/i18n/locales/nl/differences.json
```diff
@@ -8,6 +8,6 @@
   "too_many_ballots_handed_out_count": "Te veel uitgereikte stembiljetten",
   "other_explanation_count": "Andere verklaring voor het verschil",
   "no_explanation_count": "Geen verklaring voor het verschil",
-  "continue_after_check": "Je kan alleen verder als je het het papieren proces-verbaal hebt gecontroleerd.",
+  "continue_after_check": "Je kan alleen verder als je het papieren proces-verbaal hebt gecontroleerd.",
```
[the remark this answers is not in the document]
oliver3: Good idea thanks, moved the text to `data_entry.json` in https://github.com/kiesraad/abacus/pull/1205/commits/1bf9c0e642cfb256699d3c1919bcbddd0f65e052

cikzh (approved):

Lionqueen94 (approved):

PR #1211, opened 2025-03-19
Explicitly exit with statuscode, use fetch

oliver3:
- Fix incidental hanging of script by using `exit()`
- Use statuscode to stop the npm run script if the backend is not running
- Use fetch instead of http module

### DIFF
--- a/frontend/scripts/check_server.js
+++ b/frontend/scripts/check_server.js
@@ -1,19 +1,11 @@
-const http = require("http");
-
 const PORT = 8080;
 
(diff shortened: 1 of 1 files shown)


praseodym (dismissed):
lgtm!

I think the old version also continued when the backend was not running, it just logged that it was not running. I don't mind this change but wanted to point it out.

praseodym (approved):

cikzh (approved):

Issue #1149, opened 2025-03-10
Filter data entry choice list based on first data entry user


praseodym:
Filter polling stations choice list so that first data entry user cannot do second data entry.

jschuurk-kr:
@jorisleker After a user has done the first entry for a polling station, should that polling station be filtered out from the polling station list for that user? Or should it be shown, but with a specific label?

Currently the page says "Bekijk de lijst met alle stembureaus", so filtering out some polling stations would contradict that text.

jorisleker:
> [@jorisleker](https://github.com/jorisleker) After a user has done the first entry for a polling station, should that polling station be filtered out from the polling station list for that user? Or should it be shown, but with a specific label?

The list should only show polling stations that the current user could fill in. If they have done the first entry for a polling station, then that polling station isn't a valid option for them anymore. So it should be left out. 

Agree that the label isn't perfect, we could change it to 'Toon de lijst met nog in te vullen stembureau's', but on the other hand: I don't think the current label will lead to problems.

Issue #1130, opened 2025-03-05
Error handling when already claimed


praseodym:
Error handling when chosen polling station is already claimed but user still started data entry, e.g. because the polling station choice page was loaded before another user claimed that data entry.

There should be another request to check the latest status of the polling station data entry when the user is typing the polling station numbers (see comments below).

[Design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=4983-24212&t=ZPSTdEWO8BIVqwrj-11)

Note: design should only be used when redirecting back to the polling station choice page. When the user enters the polling station number in the input field, use [this design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=137-5073&t=ZPSTdEWO8BIVqwrj-11) instead

jorisleker:
@praseodym is my conclusion that the lookup on the polling station choice page is done from 'cache' / the list as it was the moment the polling station choice page was loaded, and not 'live' (no api request when the user types a number?) 

If so: I think we might want to reconsider that decision. There might be a significant time interval between the user finishing a data entry, and getting a new _proces verbaal_. So the chance the list that's available to the frontend is out of sync is quite big.

praseodym:
Yes that is the case, there are no further requests once the page is loaded. We could do a request when the user starts typing, but what about the list of polling stations ("Bekijk de lijst met alle stembureaus") at the bottom of the page?

jorisleker:
Think doing that request would solve most issues. 
The list of all polling stations is hidden by default. We could do the same request when the user clicks the link?

jorisleker:
Decided in standup
- when opening a polling station that was already claimed by someone else
- redriect. to polling station choice page
- show alert on polling station choice page

@jorisleker will add design for alert

jorisleker:
@lkleuver design added, see issue description

PR #1231, opened 2025-03-24
Feedback when polling station in progress

### LINKED ISSUE #704: Feedback when polling station in progress
Original issue #388

When searching for a polling station, you can run into a couple of scenario's that we don't have proper feedback for yet.

1. Polling station is found, but assigned to another user (regardless whether it's first or second data entry):
image

2. Polling station is found, has entry data and is assigned to current user (status batch for first and second entry has pencil icon): 
image

The design of the 'search by polling station number' component can be found [on the components page in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=137-5073&t=fwYzqCKFITZhwsl2-4)

Make sure scenario 2 only shows if the current user is assigned to the polling station.

Also in scope: filter "Je hebt nog een openstaande invoer" for only the current user.

lkleuver:
This PR integrates the logged in user with the PollingStationChoice form, showing related tags/warnings for in progress and checks if a different user is already working on data entry.

There are two states introduced:
- In progress with different user 
- In progress by current user

# Requires extra attention / feedback:

The PollingStationChoicePage.tsx currently returns "unauthenticated" when there is no user.
Is this something that will be handled up the chain in future releases?

The tests mock useUser currently, it seems preferable to use a test util for this to mock a logged in user, TestUserProvider currently conflicts with the way the tests are setup already including an ApiProvider.

closes #704


oliver3:
If you manage to start data entry that is already claimed by somebody else, you get no error message, just a blank screen on the "recounted" page

frontend/src/components/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
```diff
@@ -61,6 +61,17 @@ export function PollingStationChoiceForm({ anotherEntry }: PollingStationChoiceF
       return;
     }
 
+    if (pollingStationStatus) {
+      if (
+        (pollingStationStatus === "first_entry_in_progress" && statusEntry.first_entry_user_id !== user?.user_id) ||
+        (pollingStationStatus === "second_entry_in_progress" && statusEntry.second_entry_user_id !== user?.user_id)
+      ) {
+        setAlert(DEFINITIVE_POLLING_STATION_ALERT);
```
oliver3, line 69: This message is "Het stembureau dat je geselecteerd hebt kan niet meer ingevoerd worden", maybe we can make it more generic? In any case, the `DEFINITIVE_POLLING_STATION_ALERT` name is not accurate.

frontend/src/components/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
```diff
@@ -61,6 +61,17 @@ export function PollingStationChoiceForm({ anotherEntry }: PollingStationChoiceF
       return;
     }
 
+    if (pollingStationStatus) {
+      if (
+        (pollingStationStatus === "first_entry_in_progress" && statusEntry.first_entry_user_id !== user?.user_id) ||
+        (pollingStationStatus === "second_entry_in_progress" && statusEntry.second_entry_user_id !== user?.user_id)
```
oliver3, line 67: This logic is now done in several locations, and will probably have to be done for the `PollingStationsList` again. Now that the status logic has become more complex, can we extract that to one location, either a function or one datastructure that can be used for the different places:
- unfinished message
- polling station selector message
- error message
- list of selectable polling stations

oliver3 (changes requested):

frontend/src/components/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
oliver3: image

praseodym:
> What should be shown for the "Kies het stembureau" list if another typist is doing data entry? Should if also be filtered out? @jorisleker image

It should be filtered out, see https://github.com/kiesraad/abacus/issues/1149#issuecomment-2729134110. We don't have an issue in this epic for this (yet?).

frontend/src/components/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
lkleuver: I've added a separate alert for in progress different user

lkleuver:
> > What should be shown for the "Kies het stembureau" list if another typist is doing data entry? Should if also be filtered out? @jorisleker
> 
> It should be filtered out, see [#1149 (comment)](https://github.com/kiesraad/abacus/issues/1149#issuecomment-2729134110). We don't have an issue in this epic for this (yet?).

I've filtered out in progress by different user

oliver3 (dismissed):
Looks good! Logic is more centralised now, and the list is filtered.

praseodym (dismissed):

frontend/src/components/form/data_entry/polling_station_choice/util.ts
```diff
@@ -0,0 +1,56 @@
+import { DataEntryStatusName, ElectionStatusResponseEntry, LoginResponse, PollingStation } from "@/api";
+
+export enum PollingStationUserStatus {
+  MISSING = "MISSING",
+  IN_PROGRESS_CURRENT_USER = "IN_PROGRESS_CURRENT_USER",
+  IN_PROGRESS_OTHER_USER = "IN_PROGRESS_OTHER_USER",
+  FINISHED = "FINISHED",
+  UNKNOWN = "UNKNOWN",
+}
```
praseodym, line 9:
```suggestion
export enum PollingStationUserStatus {
  MISSING,
  IN_PROGRESS_CURRENT_USER,
  IN_PROGRESS_OTHER_USER,
  FINISHED,
  UNKNOWN,
}
```
There is also no reason for the enum members to be upper case.

praseodym:
> Both issues can be prevented by doing an api-request to fetch a fresh list of polling stations. The UI (spinner) of the polling station choice page was designed with this lookup in mind.

This was already discussed and should be part of #1130, another issue in this epic.

frontend/src/components/form/data_entry/polling_station_choice/util.ts
lkleuver: when you dont set a value the debug messages are a bit less easy to read (ints) so thats why I added it.
as for upper case. it's a convention I'm used to, but i'll make it conform the rest of the app.

lkleuver: renamed

oliver3 (approved):

frontend/src/components/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx
```diff
@@ -49,38 +64,49 @@ export function PollingStationChoiceForm({ anotherEntry }: PollingStationChoiceF
...
+    (pollingStation) => pollingStation.userStatus === PollingStationUserStatus.InProgressCurrentUser,
+  );
 
-  const pollingStationsForDataEntry = pollingStations.filter((pollingStation: PollingStation) => {
-    const status = electionStatus.statuses.find((status) => status.polling_station_id === pollingStation.id)?.status;
-    return status !== "definitive" && status !== "entries_different";
-  });
+  if (!user) {
+    return <div>Unauthenticated</div>;
+  }
```
praseodym, line 109: I only just noticed this logic. The correct way to handle this would be to redirect to the login page, but because this page is also doing an authenticated API request I think this might already be handled by the `ErrorBoundary`?

https://github.com/kiesraad/abacus/blob/93c6c781fffa286d5fcf74509607e1396bd77e11/frontend/src/app/ErrorBoundary.tsx#L11-L15

Lionqueen94: Fixed in 2ee4b8f79edb112c5ad0ad799670d5950850c7a7
Now a navigate to the login page as in `AccountSetupPage.tsx` (if it should be removed there as well I can do that)

PR #1240, opened 2025-03-25
Add audit event logging to all state changing endpoints

### LINKED ISSUE #1077: Add audit events to all other handlers

marlonbaeten:
WIP

Fixes #1077


praseodym (approved):

frontend/src/lib/i18n/locales/nl/log.json
```diff
@@ -30,8 +30,7 @@
   "event": {
```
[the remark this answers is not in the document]
marlonbaeten: Fixed in 7aed7b14a2e6e2435e1fc237b376a61e55b22370

backend/src/audit_log/audit_event.rs
```diff
@@ -0,0 +1,207 @@
...
+use utoipa::ToSchema;
+
+use crate::{
+    ErrorResponse,
+    authentication::LoginResponse,
+    data_entry::PollingStationDataEntry,
+    election::{Election, ElectionCategory, ElectionStatus},
+    error::ErrorReference,
+    polling_station::{PollingStation, PollingStationType},
+};
```
praseodym, line 14: Can we find a way to create these details structs without a reference to each module? Each of these module already has a reference to `audit_log`, so we now have cyclic dependencies.

marlonbaeten: Yeah that is not great, but we either way need the types, like `ElectionStatus` or `PollingStationType`. Alternatively I could not use these types but store them as a string. Than I could move al the From<T> stuff to an Into<T>, which makes the cyclic dependencies go away. What do you prefer @praseodym?

marlonbaeten: Hmm, we would also get a lot of clippy warnings, since `From` is preferred over `Into`

praseodym: What about making all types in the `*Details` structs a type from `std` (e.g. make `election_category` and `election_status` in `ElectionDetails` a `String`) and then moving the `From` trait implementation to the module (e.g. `impl From<Election> for ElectionDetails` to the `election` module)? That should remove the references from the `audit_log` module to every other module.

marlonbaeten: So exactly what I did in https://github.com/kiesraad/abacus/commit/ebb5a162d13ad77b31463e090566e344e15b2c37? Then I will push that commit to this branch.

praseodym: Yes that looks good!

backend/src/data_entry/api.rs
```diff
@@ -101,6 +103,12 @@ async fn polling_station_data_entry_claim(
     // Save the new data entry state
     polling_station_data_entries.upsert(id, &new_state).await?;
 
+    let data_entry = polling_station_data_entries.get_row(id).await?;
+
+    audit_service
+        .log(&AuditEvent::DataEntryClaimed(data_entry.into()), None)
+        .await?;
```
praseodym, line 110: I think it'd be better to create the audit event from `DataEntryStatus`, and also log the current status.

marlonbaeten: We keep track of "progress" which could translate to a status. I could also keep a separate field with the status string.

We either way need to keep track of the polling station this relates to, that's why I used "PollingStationDataEntry".

marlonbaeten: Added status label in 453bf5549efe69e129387b2fa21182f87b3ed719

praseodym (dismissed):

backend/src/audit_log/service.rs
```diff
@@ -38,11 +39,11 @@ where
 
 impl AuditService {
     #[cfg(test)]
-    pub fn new(log: AuditLog, ip: IpAddr, user: User) -> Self {
+    pub fn new(log: AuditLog, user: User, ip: Option<IpAddr>) -> Self {
```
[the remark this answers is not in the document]
marlonbaeten: This methods is only for tests. An audit event needs a user ([or it will throw an error](https://github.com/kiesraad/abacus/pull/1240/files#diff-83844f37e5ec681fb66524aa162f270913aaff89fe3c2950375eabac16568fecL94)). The IP address is always optional.

praseodym (approved):

Lionqueen94 (approved):

PR #1252, opened 2025-03-25
Added data entry difference state transitions

### LINKED ISSUE #653: Create diagram with data entry states and transitions

### LINKED ISSUE #1243: Update state machine diagram
https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/data-entry-state.md

cikzh:
Fixes #1243  

The diagram looks _much_ better, even with the looping `save` transitions, using:
```
---
config:
  layout: elk
---
```
Unfortunately the `elk` layout is not available on GitHub. If anyone knows how to change the layout algorithm, let me know

### DIFF
--- a/documentatie/flowcharts/data-entry-state.md
+++ b/documentatie/flowcharts/data-entry-state.md
@@ -5,19 +5,22 @@ The transition labels describe the endpoint that is used for performing the tran
 The "save" endpoint, used to for [First/Second]EntryInProgress states is kept out, because Mermaid doesn't render self-loops too well.
 
 ```mermaid
-stateDiagram
-  direction LR
+stateDiagram-v2
+  [*] --> FirstEntryNotStarted
+  FirstEntryNotStarted --> FirstEntryInProgress: claim
+  #FirstEntryInProgress --> FirstEntryInProgress: save
+  FirstEntryInProgress --> SecondEntryNotStarted: finalise
+  FirstEntryInProgress --> FirstEntryNotStarted: delete
+  SecondEntryNotStarted --> SecondEntryInProgress: claim
+  #SecondEntryInProgress --> SecondEntryInProgress: save
   state is_equal <<choice>>
-  FirstEntryNotStarted --> FirstEntryInProgress:claim
-  #FirstEntryInProgress --> FirstEntryInProgress:save
-  FirstEntryInProgress --> SecondEntryNotStarted:finalise
-  FirstEntryInProgress --> FirstEntryNotStarted:delete
-  SecondEntryNotStarted --> SecondEntryInProgress:claim
-  is_equal --> Definitive:equal? yes
-  is_equal --> EntriesDifferent:equal? no
-  #SecondEntryInProgress --> SecondEntryInProgress:save
-  SecondEntryInProgress --> is_equal:finalise
-  SecondEntryInProgress --> SecondEntryNotStarted:delete
-  EntriesDifferent --> FirstEntryInProgress: discard both entries
-  EntriesDifferent --> SecondEntryInProgress:keep one entry
+  SecondEntryInProgress --> is_equal: finalise
+  SecondEntryInProgress --> SecondEntryNotStarted: delete
+  state resolve <<choice>>
+  EntriesDifferent --> resolve: resolve
+  resolve --> SecondEntryNotStarted: keep one entry
+  resolve --> FirstEntryNotStarted: discard both entries
+  is_equal --> Definitive: equal? yes
+  is_equal --> EntriesDifferent: equal? no
+  Definitive --> [*]
 ```
--- a/documentatie/flowcharts/data-entry-state.md
+++ b/documentatie/flowcharts/data-entry-state.md
@@ -24,3 +24,7 @@ stateDiagram-v2
   is_equal --> EntriesDifferent: equal? no
   Definitive --> [*]
 ```
+
+When resolving differences between the first and second entry (`EntriesDifferent` state), the coordinator can choose to
+keep one of the entries or discard both. If one of the entries is kept, the other entry is deleted. The remaining entry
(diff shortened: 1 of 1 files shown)


documentatie/flowcharts/data-entry-state.md
jschuurk-kr, line 1: GitHub now seems to render the file in a way that makes it unclear what the start and end states are:
image

documentatie/flowcharts/data-entry-state.md
```diff
@@ -6,20 +6,18 @@ The "save" endpoint, used to for [First/Second]EntryInProgress states is kept ou
...
+  FirstEntryInProgress --> SecondEntryNotStarted:finalise
+  FirstEntryInProgress --> FirstEntryNotStarted:delete
+  SecondEntryNotStarted --> SecondEntryInProgress:claim
+  is_equal --> Definitive:equal? yes
+  is_equal --> EntriesDifferent:equal? no
+  #SecondEntryInProgress --> SecondEntryInProgress:save
+  SecondEntryInProgress --> is_equal:finalise
+  SecondEntryInProgress --> SecondEntryNotStarted:delete
+  EntriesDifferent --> FirstEntryInProgress: discard both entries
+  EntriesDifferent --> SecondEntryInProgress:keep one entry
```
jschuurk-kr, line 22: I think it would be useful to add some text below the diagram about the difference between keeping the first or the second entry. It probably should not be included in the diagram itself, because they results in the same state change.

documentatie/flowcharts/data-entry-state.md
praseodym: /Moving `state is_equal <<choice>>` to the top of the Mermaid definition causes that state to be rendered at the left of the diagram, which is not what we want.

Also, we can add explicit start and end states using `[*]`. See my other comment.

documentatie/flowcharts/data-entry-state.md
praseodym, line 22: After deleting an entry/both entries, the state machine should be in the 'not started' state because there is no-one working on that entry:

```suggestion
  EntriesDifferent --> FirstEntryNotStarted: discard both entries
  EntriesDifferent --> SecondEntryNotStarted:keep one entry
```

documentatie/flowcharts/data-entry-state.md
praseodym: With `config: elk` on https://mermaid.live it looks like this, which I don't think is a lot better:
State diagram with elk renderer

praseodym: Moving `state is_equal <<choice>>` to the top of the Mermaid definition causes that state to be rendered at the left of the diagram, which is not what we want.

Also, we can add explicit start and end states using `[*]`. See my other comment.

praseodym: Changed in 5951ffda22a2b6277f35ab23fa30fe5f24ed2c58

documentatie/flowcharts/data-entry-state.md
praseodym: Fixed in 5951ffda22a2b6277f35ab23fa30fe5f24ed2c58

praseodym: Added a text in c1abb9766b735a8585674559e18785926f60da16, please review.

praseodym:
I've taken the liberty to update the diagram (5951ffda22a2b6277f35ab23fa30fe5f24ed2c58 and c1abb9766b735a8585674559e18785926f60da16).

documentatie/flowcharts/data-entry-state.md
```diff
@@ -5,21 +5,26 @@ The transition labels describe the endpoint that is used for performing the tran
...
-SecondEntryInProgress --> SecondEntryNotStarted: delete
-#EntriesDifferent --> EntriesDifferent: save
-# Will be Implemented in #130: EntriesDifferent --> NotStarted: delete
-EntriesDifferent --> Definitive: resolve
+stateDiagram-v2
+  [*] --> FirstEntryNotStarted
+  FirstEntryNotStarted --> FirstEntryInProgress: claim
+  #FirstEntryInProgress --> FirstEntryInProgress: save
+  FirstEntryInProgress --> SecondEntryNotStarted: finalise
+  FirstEntryInProgress --> FirstEntryNotStarted: delete
```
jorisleker, line 13: 'discard' or 'abort' would be a name that aligns a bit better with the language used in the user interface (applies to line 18 as well).

jorisleker (approved):
Maybe rename 'delete', but can live with this as well :-)

documentatie/flowcharts/data-entry-state.md
praseodym: Good suggestion, changed in 8d7a376ac46a13d6fabc78dbdd48f00a2481c19d

marjoleintamis (commented):
1 taalkundige suggestie gedaan, verder prima :)

documentatie/flowcharts/data-entry-state.md
```diff
@@ -5,21 +5,26 @@ The transition labels describe the endpoint that is used for performing the tran
 The "save" endpoint, used to for [First/Second]EntryInProgress states is kept out, because Mermaid doesn't render self-loops too well.
```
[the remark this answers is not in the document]
praseodym: Dankje, verwerkt in 46a3196c48de24c555a734323c4a7c67a4ed30f8!

marjoleintamis (approved):

jorisleker (approved):

oliver3 (approved):

praseodym (approved):

PR #1257, opened 2025-03-26
Add typist name to election status page

oliver3:
- Determine which typist to show in the election status page for "in_progress" and "first_entry_finished"
- Retrieve users and show these extra columns with proper widths
- Remove vertical padding from table cells to avoid rows becoming too high when text wrap is needed
- Add data to status mocks

### DIFF
--- a/frontend/src/components/election/status/ElectionStatus.tsx
+++ b/frontend/src/components/election/status/ElectionStatus.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 
-import { DataEntryStatusName, Election, ElectionStatusResponseEntry, PollingStation } from "@kiesraad/api";
+import { DataEntryStatusName, Election, ElectionStatusResponseEntry, PollingStation, User } from "@kiesraad/api";
 import { t } from "@kiesraad/i18n";
 import { IconPlus } from "@kiesraad/icon";
 import {
@@ -31,9 +31,14 @@ export interface ElectionStatusProps {
   election: Required<Election>;
   pollingStations: PollingStation[];
   navigate: (path: string) => void;
+  users: User[];
 }
 
-export function ElectionStatus({ statuses, election, pollingStations, navigate }: ElectionStatusProps) {
+interface PollingStationWithStatusAndTypist extends PollingStation, Partial<ElectionStatusResponseEntry> {
+  typist?: string;
+}
+
+export function ElectionStatus({ statuses, election, pollingStations, navigate, users }: ElectionStatusProps) {
   const categoryCounts: Record<StatusCategory, number> = React.useMemo(
     () =>
       Object.fromEntries(statusCategories.map((cat) => [cat, statusCount(statuses, cat)])) as Record<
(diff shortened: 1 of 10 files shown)


frontend/src/components/election/status/ElectionStatus.tsx
```diff
@@ -58,9 +63,40 @@ export function ElectionStatus({ statuses, election, pollingStations, navigate }
...
+    switch (status.status) {
+      case "first_entry_in_progress":
+        typistId = status.first_entry_user_id;
+        break;
+      case "second_entry_in_progress":
+        typistId = status.second_entry_user_id;
+        break;
+      case "second_entry_not_started":
+        typistId = status.first_entry_user_id;
+        break;
```
praseodym, line 81: What about:
```suggestion
      case "first_entry_in_progress":
      case "second_entry_not_started":
        typistId = status.first_entry_user_id;
        break;
      case "second_entry_in_progress":
        typistId = status.second_entry_user_id;
        break;
```

praseodym, line 81: What about:
```suggestion
      case "first_entry_in_progress":
      case "second_entry_not_started":
        typistId = status.first_entry_user_id;
        break;
      case "second_entry_in_progress":
        typistId = status.second_entry_user_id;
        break;
```

frontend/src/components/election/status/ElectionStatus.tsx
```diff
@@ -58,9 +63,40 @@ export function ElectionStatus({ statuses, election, pollingStations, navigate }
     return data;
   }, [statuses, categoryCounts]);
 
-  const pollingStationsWithStatuses = pollingStations.map((ps) => {
+  function getTypistName(status: ElectionStatusResponseEntry | undefined) {
```
praseodym, line 66: We could move this function out of the `ElectionStatus` component by passing `users` as a function argument. This would reduce the amount of code within the component.

praseodym (dismissed):
lgtm, only two suggested improvements

frontend/src/components/election/status/ElectionStatus.tsx
oliver3: It is pretty specific for this `ElectionStatus` though, showing e.g. the first entry user when the status is `second_entry_not_started`.

frontend/src/components/election/status/ElectionStatus.tsx
oliver3: Thanks, done in https://github.com/kiesraad/abacus/pull/1257/commits/0536722ec59ab2294b5da978f2d5e5ef403d7350

frontend/src/components/election/status/ElectionStatus.tsx
praseodym: It can still be in this file, but outside of the component. Similar to the `statusCount` function.

oliver3: Ah those are two different things of course, I misunderstood. Good idea and done in https://github.com/kiesraad/abacus/pull/1257/commits/dbdc3093e98463c83bf4500a7cb4575ec61e3a50

praseodym (approved):

jorisleker (approved):

PR #1259, opened 2025-03-26
Fix flaky e2e tests

oliver3:
In recent PR's I see a lot of flaky tests for the model based tests on MacOS. It seems that filling in a polling station number is not working and the test cannot continue to data entry. This potential fix uses `pressSequentially` to enter the polling station number.

image

Additional changes:
- add an `expect()` after the `pressSequentially` for the name of the polling station being shown
- increase local e2e test timeout to 20 seconds, so you can run them all for a single browser engine
- add an `expect()` to the left-nav-panel icon test (test was flaky at that point when running locally)

### DIFF
--- a/frontend/e2e-tests/data-entry.e2e.ts
+++ b/frontend/e2e-tests/data-entry.e2e.ts
@@ -117,7 +117,7 @@ test.describe("full data entry flow", () => {
 
     const pollingStationChoicePage = new PollingStationChoicePage(page);
     await expect(pollingStationChoicePage.fieldset).toBeVisible();
-    await pollingStationChoicePage.selectPollingStationAndClickStart(pollingStation.number);
+    await pollingStationChoicePage.selectPollingStationAndClickStart(pollingStation);
 
     const recountedPage = new RecountedPage(page);
     await expect(recountedPage.fieldset).toBeVisible();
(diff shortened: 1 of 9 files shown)


jschuurk-kr (dismissed):

praseodym (commented):
I guess this works, but maybe waiting for the correct feedback text to appear is a better approach? We're already using this in another e2e test:
https://github.com/kiesraad/abacus/blob/7d88c32fcfa47210c30da7fec3a2ea89d86aa3e8/frontend/e2e-tests/data-entry.e2e.ts#L32-L36

oliver3:
There’s something funny on that page, it now fails with

```
Warning: Can't perform a React state update on a component that hasn't mounted yet. This indicates that you have a side-effect in your render function that asynchronously later calls tries to update the component. Move this work to useEffect instead.
    at PollingStationChoiceForm (/home/runner/work/abacus/abacus/frontend/src/components/form/data_entry/polling_station_choice/PollingStationChoiceForm.tsx:16:37)
```

jschuurk-kr:
> I guess this works, but maybe waiting for the correct feedback text to appear is a better approach? We're already using this in another e2e test:
> 
> https://github.com/kiesraad/abacus/blob/7d88c32fcfa47210c30da7fec3a2ea89d86aa3e8/frontend/e2e-tests/data-entry.e2e.ts#L32-L36

Are we ok with having an `expect()` in a page object? Alternative would be to use a `waitFor()`:
```TypeScript
// waitFor()
await this.pollingStationFeedback.filter( { hasText: pollingStation.name}).waitFor( {timeout: 1000} );

// expect()
await expect(this.pollingStationFeedback).toContainText(pollingStation.name);
```
In general I'm not a fan of assertions in page objects, but this is a helper method and the `expect()` does give clearer feedback if the element is present, but contains a different text.

praseodym:
> In general I'm not a fan of assertions in page objects, but this is a helper method

That's why I think it's fine to do that.

jschuurk-kr:
> > In general I'm not a fan of assertions in page objects, but this is a helper method
> 
> That's why I think it's fine to do that.

done in d40a21d363da4235bd01f4fb836ee19133f92c03

jschuurk-kr (dismissed):

PR #1270, opened 2025-03-27
Run PDF gen using typst binary instead of embedded typst by default

rnijveld:
Typst is an enormous dependency that requires around 200 dependencies and causes my rust-analyzer instance to be killed all the time by the Linux OOM (out of memory) killer. While I would say that this is technically more of a rust-analyzer and rust compiler problem in that it can't handle the size of our project, it is still something I run into on an almost daily basis.

Meanwhile we don't do that much PDF generation during regular development. And if you are actually developing PDFs you don't really need to touch the typst code at all. This made me look at how easy it was to actually split of typst during regular development and just call the typst CLI binary directly. It turns out: not that hard at all, as long as we don't mind the binary invocation to be a little heavy handed: I just copy over the entire templates directory to a temporary location and replace the input with the actual input in that temporary location.

Meanwhile, CI still produces the binaries with typst support build-in, meaning that there should be no changes at all for our release binaries with this. But it would be good to at least test that development still works as expected.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -81,7 +81,7 @@ jobs:
       - name: Check Clippy without default features
         run: cargo --verbose --locked clippy --target=${{ matrix.target.name }} --all-targets --no-default-features -- -D warnings
       - name: Run tests
-        run: cargo --verbose --locked test --target=${{ matrix.target.name }} -- --nocapture
+        run: cargo --verbose --locked test --target=${{ matrix.target.name }} --features embed-typst -- --nocapture
       - name: Build
         run: cargo --verbose --locked build --features memory-serve,embed-typst --target=${{ matrix.target.name }}
       - uses: actions/upload-artifact@v4
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -17,7 +17,7 @@ jobs:
     name: Backend (${{ matrix.target.os }})
     strategy:
       matrix:
-        target: 
+        target:
           - os: ubuntu-latest
             name: x86_64-unknown-linux-musl
             binary: backend/target/x86_64-unknown-linux-musl/debug/abacus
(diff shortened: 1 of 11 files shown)


cikzh (approved):

praseodym:
> Typst is an enormous dependency that requires around 200 dependencies and causes my rust-analyzer instance to be killed all the time by the Linux OOM (out of memory) killer. While I would say that this is technically more of a rust-analyzer and rust compiler problem in that it can't handle the size of our project, it is still something I run into on an almost daily basis.

I've never had these issues. Do you know how much RAM rust-analyzer uses on your machine before it gets OOM killed? And do other team members have similar problems?

cikzh:
I have this problem as well. As a workaround, a colleague installed a user-space OOM killer for me, so it terminates rust-analyzer before other applications. This helps in that my editor doesn't hang anymore, but it still results in a loop where rust-analyzer keeps rebooting and shutting down when OOM.
While debugging, we watched rust-analyzer's memory usage in htop and in just keeps on growing until OOM when we do actions like Find References

rnijveld:
> > Typst is an enormous dependency that requires around 200 dependencies and causes my rust-analyzer instance to be killed all the time by the Linux OOM (out of memory) killer. While I would say that this is technically more of a rust-analyzer and rust compiler problem in that it can't handle the size of our project, it is still something I run into on an almost daily basis.
> 
> I've never had these issues. Do you know how much RAM rust-analyzer uses on your machine before it gets OOM killed? And do other team members have similar problems?

Everyone at tweede golf has experienced this problem. None of us know exactly how much memory was used because our systems become unresponsive while the OOM killer isn't triggered yet. But at least my machine has 32GB available and is using around 10-15GB normally. I personally specifically experience issues when trying to run cargo test (with more recent abacus checkouts I have to actually exit rust-analyzer before I run cargo test, otherwise I know my system freezes, but it also occurs at other random times). At that time several additional rust compilers try and run in parallel, and there might be some kind of interaction with rust-analyzer at that time that causes this behavior.

Aside from this (not so great) behavior, this also really helps on my machine with compile speed, specifically when linking it saves seconds and that really helps with the feedback-loop while developing.

With these patches I don't experience any problems at all fortunately. If these changes are too much for now I can also just submit a couple of small changes that at least allow a very simple local patch-set for me instead.

jschuurk-kr:
> > Typst is an enormous dependency that requires around 200 dependencies and causes my rust-analyzer instance to be killed all the time by the Linux OOM (out of memory) killer. While I would say that this is technically more of a rust-analyzer and rust compiler problem in that it can't handle the size of our project, it is still something I run into on an almost daily basis.
> 
> I've never had these issues. Do you know how much RAM rust-analyzer uses on your machine before it gets OOM killed? And do other team members have similar problems?

Minor data point: when running `cargo test` on my machine, total memory usage peaked at 21GB the first time and at 25GB the second time. So I think the reason I've not been running into this problem is that I 'm lucky to have a machine with 64GB of memory.

praseodym (approved):
I never noticed rust-analyzer using that much memory on my machine (it's usually using around 1-2 GB) and never experienced crashes/OOMs, but if you are having problems and this PR helps then I'm totally fine with merging it.

PR #1206, opened 2025-03-18
Audit log filter component

marlonbaeten:
WIP

PR #1282, opened 2025-03-31
Fix dialog modal

marlonbaeten:
Currently the Modal components closes itself using the dialog API: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

At the same time we keep state in react to either hide or show the dialog - this can cause issues when quickly reopening modals.

This PR removed the code that closes the dialog with the JavaScript API; which is not necessary for our use-cases.

To test: open a modal, like when leaving a data entry without saving or in de audit log.


jschuurk-kr (changes requested):
If I close the modal by hitting the Escape-key, I can't open a new modal.

marlonbaeten:
> If I close the modal by hitting the Escape-key, I can't open a new modal.
> 
> Same behavior exists without this fix in the branch of #1206

Interesting: I didn't know this feature of dialog. I've added a handler for the cancel event to call the correct `onClose` callback. This prevents the dialog from being hidden.

lkleuver (approved):

frontend/src/components/ui/Modal/Modal.tsx
```diff
@@ -13,29 +13,57 @@ export interface ModalProps {
   children?: ReactNode;
 }
 
+/**
+ * Modal component
+ *
+ * @param {string} title - The title of the modal.
+ * @param {boolean} noFlex - If true, the modal will not use flexbox for is contents layout.
+ * @param {function} onClose - Callback function to be called when the modal is closed.
```
oliver3, line 21:
```suggestion
 * @param {function} onClose - Callback function to be called when the modal should be closed.
```

Lionqueen94 (approved):
Looks and works well!

jschuurk-kr (approved):

PR #1291, opened 2025-04-01
Added ZeroVotesCast error

### LINKED ISSUE #1214: Add error when all political groups got 0 votes

Lionqueen94:
Closes #1214


cikzh (approved):
Only one nitpick: I'm not sure if `UNPROCESSABLE_ENTITY` is the right status code to return, but I don't know what status code would be better... I also think it's to unimportant to keep the PR hanging.

LGTM overall!

praseodym (approved):

PR #1293, opened 2025-04-01
Add error reference for 'already claimed'

praseodym:
Add error reference for 'already claimed' so that the frontend can detect this specific error.

Relates to #1130

### DIFF
--- a/frontend/src/api/gen/openapi.ts
+++ b/frontend/src/api/gen/openapi.ts
@@ -410,6 +410,7 @@ export type ErrorReference =
   | "AllListsExhausted"
   | "ApportionmentNotAvailableUntilDataEntryFinalised"
   | "DatabaseError"
+  | "DataEntryAlreadyClaimed"
   | "DrawingOfLotsRequired"
   | "EntryNotFound"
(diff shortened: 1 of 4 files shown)


frontend/src/api/gen/openapi.ts
```diff
@@ -410,6 +410,7 @@ export type ErrorReference =
   | "AllListsExhausted"
   | "ApportionmentNotAvailableUntilDataEntryFinalised"
   | "DatabaseError"
+  | "DataEntryAlreadyClaimed"
```
oliver3, line 413: An appropriate message has to be added to `frontend/src/lib/i18n/locales/nl/error.json`

oliver3 (changes requested):

frontend/src/api/gen/openapi.ts
oliver3: Fixed in https://github.com/kiesraad/abacus/pull/1293/commits/3f2226158076105fe295f052c2da8d4fb6d1a3a6

lkleuver (approved):

PR #1295, opened 2025-04-01
Remove `FormSectionReference` from data entry

praseodym:
The `type` in `FormSectionReference` was not used so all usages could be replaced with `FormSectionId`.


oliver3 (approved):
Another improvement! Thanks

PR #1297, opened 2025-04-02
Run PR+merge queue workflow on Linux only

praseodym:
The macOS and Windows workflow runs didn't find any bugs so far, and especially the Windows runner can be slower than the other runners. It's wasteful to run tests on all three platforms without providing any real benefit, so this PR changes the 'Build, lint & test' PR and merge queue workflow to only run on Linux. The weekly build will continue to build and test for Linux, macOS and Windows.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -14,20 +14,8 @@ env:
 
 jobs:
   backend:
-    name: Backend (${{ matrix.target.os }})
-    strategy:
-      matrix:
-        target:
-          - os: ubuntu-latest
-            name: x86_64-unknown-linux-musl
-            binary: backend/target/x86_64-unknown-linux-musl/debug/abacus
-          - os: macos-latest
-            name: aarch64-apple-darwin
-            binary: backend/target/aarch64-apple-darwin/debug/abacus
-          - os: windows-latest
-            name: x86_64-pc-windows-msvc
-            binary: backend/target/x86_64-pc-windows-msvc/debug/abacus.exe
-    runs-on: ${{ matrix.target.os }}
+    name: Backend
+    runs-on: ubuntu-latest
     defaults:
       run:
         working-directory: ./backend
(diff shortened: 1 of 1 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -116,11 +103,8 @@ jobs:
           path: frontend/dist
 
   playwright-ladle:
-    name: Playwright ladle tests
-    strategy:
-      matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-    runs-on: ${{ matrix.os }}
+    name: Playwright Ladle tests
+    runs-on: ubuntu-latest
```
jschuurk-kr, line 107: We don't run these in the Release workflow. So let's keep them here?

praseodym: So far we haven't seen any differences between platforms, do we expect to find any with the Ladle tests that would make it a good idea to keep testing this on macOS and Windows?

jschuurk-kr: Fair point. We do run against three browsers on Linux and the OS should really not matter for Ladle tests.

.github/workflows/build-lint-test.yml
```diff
@@ -141,20 +125,16 @@ jobs:
...
           files: frontend/playwright.ladle.junit.xml
           disable_search: true
           fail_ci_if_error: true
 
   playwright-e2e:
-    name: Playwright e2e tests (${{ matrix.os }})
+    name: Playwright e2e tests
     needs:
       - backend
-      - frontend # for the tests, frontend build is included in backend job
```
jschuurk-kr, line 153: Is there a point in running the e2e tests if some of the frontend tests fail? I guess that raises the question if we see `need` as only a technical dependency (can it run?) or also as an information relation (should it run?).

This dependency on the front-end job was also removed from the deploy job. There I see how it could be useful to have the deployment for investigating failures in the pipeline.

praseodym: I thought it could be useful to run the Playwright e2e tests too get as much information on what failed, but we can put this `needs` back too.

jschuurk-kr: Sometimes useful, sometimes distracting? Let's merge the PR as it is and see what happens.

jschuurk-kr (approved):

oliver3 (approved):

praseodym:
I've just changed the required status checks for the repository, so the Linux-only tests are now the required ones:
Status checks that are required

PR #1299, opened 2025-04-03
Types and type predicate functions for specific SeatChangeStep types

oliver3:
- Add types for specific versions of the SeatChangeStep, depending on the kind of `change` inside
- Add predicate functions to be able to filter an array resulting in the correct type
- Change Table props to use specific types
- Update tests and testdata

Questions: is this an improvement? Are the names okay?

### DIFF
--- a/frontend/src/features/apportionment/components/Apportionment.stories.tsx
+++ b/frontend/src/features/apportionment/components/Apportionment.stories.tsx
@@ -1,10 +1,10 @@
 import { Story } from "@ladle/react";
 
 import { PoliticalGroup } from "@/api";
-import { HighestAverageStep } from "@/features/apportionment/utils/seat-change";
 
 import * as gte19Seats from "../testing/19-or-more-seats";
 import * as lt19Seats from "../testing/less-than-19-seats";
+import { HighestAverageStep } from "../utils/seat-change";
 import { ApportionmentTable } from "./ApportionmentTable";
 import { ElectionSummaryTable } from "./ElectionSummaryTable";
 import { FullSeatsTable } from "./full_seats/FullSeatsTable";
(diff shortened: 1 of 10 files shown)


frontend/src/features/apportionment/components/Apportionment.stories.tsx
```diff
@@ -1,6 +1,7 @@
 import { Story } from "@ladle/react";
 
 import { PoliticalGroup } from "@/api";
+import { HighestAverageStep } from "@/features/apportionment/utils/seat-change";
```
oliver3, line 4: missed one..

frontend/src/features/apportionment/utils/seat-change.ts
```diff
@@ -0,0 +1,25 @@
...
+export function isHighestAverageStep(step: SeatChangeStep): step is HighestAverageStep {
+  return step.change.changed_by === "HighestAverageAssignment";
+}
+export function isLargestRemainderStep(step: SeatChangeStep): step is LargestRemainderStep {
+  return step.change.changed_by === "LargestRemainderAssignment";
+}
+export function isAbsoluteMajorityStep(step: SeatChangeStep): step is AbsoluteMajorityStep {
+  return step.change.changed_by === "AbsoluteMajorityReassignment";
+}
+export function isListExhaustionRemovalStep(step: SeatChangeStep): step is ListExhaustionRemovalStep {
```
Lionqueen94, line 23: I suggest naming the above functions and types with Assignment/Reassignment, so it's more consistent with Removal in the naming of this function and step. Otherwise great improvement I think!

Lionqueen94 (dismissed):

frontend/src/features/apportionment/utils/seat-change.ts
oliver3: Thanks! Updated the type and function names. Now doubting if the table props should be changed accordingly :)

Lionqueen94 (approved):

frontend/src/features/apportionment/utils/seat-change.ts
Lionqueen94: Yeah I was just thinking the same haha

Lionqueen94 (approved):

praseodym (approved):

Issue #1247, opened 2025-03-25
Highlight consequences of users choice


cikzh:
When the user selects the radio button to either keep one of the entries, or discard them both

PR #1304, opened 2025-04-03
DifferencesTable and ResolveDifferencesTables components

oliver3:
Implements https://github.com/kiesraad/abacus/issues/1246

Not implemented yet : styling to indicate keeping first or second data entry, this will be done in https://github.com/kiesraad/abacus/issues/1247

---

New component `DifferencesTable`:
- Labelled by heading
- Styling for the data entry differences page
- Show only rows with differences between first and second
- Show "gap row" for one or more hidden rows in between rows that do have differences
- Renders nothing if there are no differences at all

New component `ResolveDifferencesTables`:
- Show `DifferencesTable` for all sections
- Show `DifferencesTable` for all political groups
- [Ladle story](https://dcf21c28.kiesraad-abacus.pages.dev/ladle/?story=app--resolve-differences--resolve-differences-tables)

Translation between `PollingStationResults` structure and data entry sections and fields, to be able to iterate over section and fields in `ResolveDifferencesTables`.
- Object to "configure" data entry fields with code and corresponding paths within `PollingStationResults` 
- Function to get the field value according to its path
- Translations corresponding to paths in `PollingStationResults`

### DIFF
--- a/frontend/src/features/resolve_differences/utils/dataEntry.ts
+++ b/frontend/src/features/resolve_differences/utils/dataEntry.ts
@@ -1,0 +1,67 @@
+import { PollingStationResults } from "@/api";
+import { FormSectionId } from "@/components/form/data_entry/state/types";
+
+type ObjectPath<T> = {
+  // For each entry, if it is an object (including optional)
+  [K in keyof T]: NonNullable<T[K]> extends object
+    ? T[K] extends unknown[]
+      ? never // Skip arrays, which are also objects
+      : `${K & string}.${Extract<keyof NonNullable<T[K]>, string>}` // "parent.child" template literal for string keys
+    : K & string; // else, the key itself when it is a string
+}[keyof T]; // Get all values together
+
+type SectionId = Extract<FormSectionId, "recounted" | "voters_votes_counts" | "differences_counts">;
+
+export type PollingStationResultsPath = NonNullable<ObjectPath<PollingStationResults>>;
+
+export interface DataEntrySection {
+  id: SectionId;
+  fields: { code?: string; path: PollingStationResultsPath }[];
+}
+
+export const sections: DataEntrySection[] = [
+  {
+    id: "recounted",
+    fields: [{ path: "recounted" }],
+  },
+  {
+    id: "voters_votes_counts",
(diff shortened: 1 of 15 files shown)


frontend/src/features/resolve_differences/utils/dataEntry.ts
```diff
@@ -0,0 +1,67 @@
...
+      { code: "E", path: "votes_counts.votes_candidates_count" },
+      { code: "F", path: "votes_counts.blank_votes_count" },
+      { code: "G", path: "votes_counts.invalid_votes_count" },
+      { code: "H", path: "votes_counts.total_votes_cast_count" },
+      { code: "A.2", path: "voters_recounts.poll_card_count" },
+      { code: "B.2", path: "voters_recounts.proxy_certificate_count" },
+      { code: "C.2", path: "voters_recounts.voter_card_count" },
+      { code: "D.2", path: "voters_recounts.total_admitted_voters_count" },
+    ],
+  },
```
praseodym, line 43: In the data entry form these fields are on the same page but with a header ("Toegelaten kiezers na hertelling door gemeentelijk stembureau") before A.2 to D.2:
Screenshot 2025-04-03 at 17 06 29

This is more clear than the current implementation of the differences table where it is not clear that fields A.2 to D.2 are recounted values:
Screenshot 2025-04-03 at 17 06 21

However, in the new models the fields A.2 to D.2 are removed. This means there won't have a need for that header and I recommend we don't implement it either.

praseodym (approved):
Looks good!

frontend/src/features/resolve_differences/utils/dataEntry.ts
praseodym: I thought we could try to summarize those questions to a single label that would fit in the 'Omschrijving' column. But agreed, we can design that later.

jorisleker: yeah I think that would work. Especially if we add the number of the question (i.e 2.2 or 2.3) in the 'code' column
Scherm­afbeelding 2025-04-03 om 21 17 04

PR #1306, opened 2025-04-04
Restructure feature data_entry

### LINKED ISSUE #1192: Restructure feature data_entry
For this feature:
- [x] Remove `index.ts` files
- [x] Move all components (including Pages) to `components`
- [x] Move hooks and contexts to `hooks`
- [x] Move test util files and data to `testing`
- [x] Check if there are files in `src/api` and `src/lib/` that should move to this feature
- [x] Check if imports can be shortened to using `@/` (check for `@kiesraad/`, `../../../../..`)
- [x] Merge module css files
- [x] Update sigrid.yaml

Lionqueen94:
Closes #1192

Also extracted polling station choice to a separate feature and renamed to data entry choice


praseodym:
Can you check if we can easily move `polling_station_choice` into its own feature?

praseodym:
Also `usePollingStationStatus` and `useDebouncedCallback` can be moved from `lib/util/hook` to the data entry and data entry choice features respectively.

frontend/src/features/data_entry/components/utils.ts
praseodym, line 1: This file can be moved to `src/features/data_entry/utils`

praseodym (commented):
In general: some files in `frontend/src/features/data_entry/stores/` should be moved to `../utils`, some others (`use*`, contexts) to `../hooks` and others (providers) to `../components`. We haven't used `stores` anywhere else and I don't think we have a good reason to use it only for the data entry feature.

praseodym (dismissed):

praseodym (approved):

PR #1313, opened 2025-04-07
Use `Swatinem/rust-cache` action for CI

praseodym:
Use the [`Swatinem/rust-cache`](https://github.com/Swatinem/rust-cache) action in our CI workflows. This action manages the Cargo caches a lot better than our current approach, greatly reducing our cache size which increases performance (downloading and extracting the cache takes time) and avoids caches from being purged (because GitHub limits the total cache size for a repository to 10GB).

In my tests the savings is as follows:
- Build, lint & test: 2.4-4.4GB -> 880MB, 1min10s -> 11s to download
- Codecov: 3GB -> 640MB, 50s -> 8s to download

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -30,36 +30,25 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 22
-      - name: Install dependencies
+      - name: Install npm dependencies
         run: npm ci
         working-directory: ./frontend
-      - name: Build
+      - name: Build frontend
         run: npm run build
         working-directory: ./frontend
       # Compile and test backend
+      - name: Install musl
+        run: sudo apt-get update && sudo apt-get install -y musl-tools
       - name: Setup Rust
         shell: bash
         run: >
           rustup update stable &&
           rustup default stable &&
-          echo RUST_VERSION=$(rustc --version | cut -d' ' -f2) >> $GITHUB_ENV
+          rustup target add x86_64-unknown-linux-musl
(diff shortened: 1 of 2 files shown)


Lionqueen94 (approved):

.github/workflows/build-lint-test.yml
```diff
@@ -30,36 +30,25 @@ jobs:
...
-            ~/.cargo/.crates.toml
-            ~/.cargo/.crates2.json
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            ./backend/target/
-          key: ${{ runner.os }}-cargo-${{ env.RUST_VERSION }}-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: ${{ runner.os }}-cargo-${{ env.RUST_VERSION }}-
-      - name: Install musl
```
marlonbaeten, line 59: We have to keep `sudo apt-get update && sudo apt-get install -y musl-tools` in order to compile for musl.

marlonbaeten: It might not the installed on the runner: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

marlonbaeten: See also https://github.com/actions/runner-images/issues/5770#issuecomment-1160218740

praseodym: It's still there, just moved up by a few lines

PR #1315, opened 2025-04-07
Only fetch a user once per request

marlonbaeten:
We inject the authenticated user using the `FromRequestParts` trait on User. Since there are now multiple middlewares and extractors the trait implementation will be called multiple times, resulting in the same code / queries being executed multiple times for a single request. In some cases the session and user are queried 3 times for a single request.

This PR adds an `inject_user` middleware that adds the session and user as an extension to the request. The `FromRequestParts` and `OptionalFromRequestParts` implementations use this injected user directly and do not have to execute database queries every time.


cikzh (approved):
Did some changes during pair programming

cikzh (approved):

praseodym (approved):

PR #1323, opened 2025-04-08
Show abort dialog on check and save form

### LINKED ISSUE #1316: Bug: geen modal bij 'Invoer afbreken' op pagina 'Controleren en opslaan'
Reproduceren:

- Voer een pv in, klik nog niet op `Opslaan`
- Klik op de pagina **Controleren en opslaan** op  `Invoer afbreken`
- Er wordt niet gevraagd of je de gegevens wilt bewaren, maar je gaat direct naar de pagina met stembureaus. De invoer wordt wel opgeslagen, je kunt dan ook weer op het stembureau klikken om door te gaan met invoeren.

marlonbaeten:
Fixes #1316 

To test: try to abort on the check and save section of the form (Controleren en opslaan). An abort dialog should appear.

### DIFF
--- a/frontend/src/features/data_entry/components/DataEntryNavigation.tsx
+++ b/frontend/src/features/data_entry/components/DataEntryNavigation.tsx
@@ -30,6 +30,7 @@ export function DataEntryNavigation({ onSubmit, currentValues = {} }: DataEntryN
 
     if (
       status === "deleted" ||
+      status === "finalising" ||
       status === "finalised" ||
       status === "aborted" ||
       currentLocation.pathname === nextLocation.pathname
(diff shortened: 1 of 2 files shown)


praseodym:
"Invoer bewaren" on the check and save page now finalises the data entry, which I don't think is the expect behaviour. I'd expect it to be essentially a no-op that redirects to the data entry choice page (because the data entry has already been saved).

marlonbaeten:
> "Invoer bewaren" on the check and save page now finalises the data entry, which I don't think is the expect behaviour. I'd expect it to be essentially a no-op that redirects to the data entry choice page (because the data entry has already been saved).

So show a different dialog? Maybe you could clarify the expected behavior in the issue #1316 Or should we ask @jorisleker ?

jorisleker:
> I'd say we show same dialog, but the "Invoer bewaren" button shouldn't finalise the data entry but just save the current state. But I'm curious to know what @jorisleker thinks.

I am with @praseodym here: Same dialog, "invoer bewaren" should not finalize the data entry.

praseodym (changes requested):
See comments

marlonbaeten:
> I'd say we show same dialog, but the "Invoer bewaren" button shouldn't finalise the data entry but just save the current state. But I'm curious to know what @jorisleker thinks.

Fixed in [ec169ec](https://github.com/kiesraad/abacus/pull/1323/commits/ec169ecae8328e11fe5f2bf515097aad4efc6002)

PR #1329, opened 2025-04-08
Filter data entry choice for second entry 

### LINKED ISSUE #1149: Filter data entry choice list based on first data entry user

### LINKED ISSUE #1150: Feedback when current user cannot do second data entry
Feedback when current user cannot do second data entry because they already did the first data entry, but tries to do so anyway (enters polling station number on the polling station choice page).

Image

[See Figma for design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=8705-49192&t=2xszvMoEmkadjrtF-4)

praseodym:
Second data entry user should be a different user than the one that did the first data entry.

Resolves #1149
Resolves #1150


frontend/src/features/data_entry_choice/components/DataEntryChoiceForm.tsx
```diff
@@ -166,6 +167,7 @@ export function DataEntryChoiceForm({ anotherEntry }: DataEntryChoiceFormProps)
             >
               {t("start")}
             </Button>
+            ff
```
Lionqueen94, line 170:
```suggestion
```

praseodym: Oops 😅

jorisleker (approved):
LGTM

frontend/src/features/data_entry_choice/utils/util.ts
```diff
@@ -49,8 +50,19 @@ export function getPollingStationWithStatusList({
...
     return result;
   });
 }
+
+export function getUrlForDataEntry(
+  electionId: number,
+  pollingStationId: number,
+  pollingStationStatus?: DataEntryStatusName,
+): string {
+  const entryNumber = pollingStationStatus?.startsWith("second_entry") ? 2 : 1;
```
lkleuver, line 66: feels a bit iffy to do string compares.

lkleuver (approved):
LGTM

small comment but not sure what would be a better approach

PR #1330, opened 2025-04-08
Move `useFormKeyboardNavigation` to data entry feature

praseodym:
`useFormKeyboardNavigation` is only used for data entry, so move it to the data entry feature.


Lionqueen94 (approved):

lkleuver (approved):
lgtm

PR #1350, opened 2025-04-10
Added inline Alert component

cikzh:
Added attribute to Alert component to render an inline variant of it.
Opted for an attribute instead of a variant, so it we get inline versions of all variants "for free".

PR is best reviewed by viewing the component in Ladle


praseodym:
https://inline-alert.kiesraad-abacus.pages.dev/ladle/?story=alert--inline-alert

praseodym:
Compared to the design in Figma I think there should be a bit more space between the header and the description text.

PR #1352, opened 2025-04-11
Show warning when data entry is already claimed

### LINKED ISSUE #1130: Error handling when already claimed

lkleuver:
These changes introduce an alert warning when a user tries to start data entry on polling station that is already claimed.

tests:
- added rtl test for rendering the alert
- added e2e test for checking that the redirect works with the correct path information

closes #1130


frontend/src/features/data_entry_choice/components/DataEntryChoicePage.tsx
```diff
@@ -30,12 +30,27 @@ export function DataEntryChoicePage() {
...
+  }
+
+  function getPollingStationNumber(pollingStationId: string | number | null) {
+    if (!pollingStationId) {
+      return 0;
+    }
+    pollingStationId = typeof pollingStationId === "string" ? parseInt(pollingStationId) : pollingStationId;
+    const pollingStation = pollingStations.find((ps) => ps.id === pollingStationId);
+    return pollingStation ? pollingStation.number : 0;
+  }
```
praseodym, line 52: It would be cleaner to put the polling station number in a variable directly instead of using a function that is called from the TSX

praseodym (commented):
#1130 also mentions "There should be another request to check the latest status of the polling station data entry when the user is typing the polling station numbers (see comments below).", do you want to add that to this PR or pick that up in a new issue?

frontend/src/features/data_entry_choice/components/DataEntryChoicePage.tsx
lkleuver: fixed in [240ca64](https://github.com/kiesraad/abacus/pull/1352/commits/240ca64e810eeb0ebc35d84b0f358dc7120c5b08)

praseodym (approved):

Lionqueen94 (approved):

praseodym:
> #1130 also mentions "There should be another request to check the latest status of the polling station data entry when the user is typing the polling station numbers (see comments below).", do you want to add that to this PR or pick that up in a new issue?

Created #1363 for this

PR #1374, opened 2025-04-15
Cleanup election status and link to differences page

marlonbaeten:
Scope:
- Cleanup of `ElectionStatus` component
- Added Placeholder page for "Resolve Differences" page
- Added link to "Resolve Differences" page

Fixes #1245

To test:

- Load [election_1_data_entry_differences.sql](https://github.com/kiesraad/abacus/blob/4aef12562aa65c38e5273398cbd650514ae525b6/backend/fixtures/election_1_data_entry_differences.sql)
- Login as coordinator
- Go to the election status page
- Click the first row "Verschil invoer 1 en 2"
- See a placeholder table appear

PR #1382, opened 2025-04-15
Endpoint for polling station data entry status and hook to call this endpoint

### LINKED ISSUE #1244: Endpoint for differences checker

marlonbaeten:
Fixes #1244

Scope:
- Implements a new endpoint: `GET /api/polling_stations/{polling_station_id}/data_entries`
- Added hook to `ResolveDifferencesPage` called `usePollingStationDataEntryDifferences` that handles the logic of fetching the newly created endpoint and handles any error scenario's.

 To test:

- Load [election_1_data_entry_differences.sql](https://github.com/kiesraad/abacus/blob/4aef12562aa65c38e5273398cbd650514ae525b6/backend/fixtures/election_1_data_entry_differences.sql)
- Login as coordinator
- Go to the election status page
- Click the first row "Verschil invoer 1 en 2"
- See the differences table appear with backend data (in contrast with #1374 in which the table contains mock data)


praseodym (commented):
Note that this PR has a merge conflict that needs to be resolved.

cikzh (approved):

backend/src/data_entry/api.rs
```diff
@@ -45,6 +46,25 @@ pub fn router() -> OpenApiRouter<AppState> {
...
+    get,
+    path = "/api/polling_stations/{polling_station_id}/data_entries",
+    responses(
+        (status = 200, description = "Get data entries for polling station", body = DataEntryStatus),
+        (status = 401, description = "Unauthorized", body = ErrorResponse),
+        (status = 500, description = "Internal server error", body = ErrorResponse),
+    ),
+)]
+async fn polling_station_data_entry_status(
+    _user: User,
```
praseodym, line 59: This endpoint should only be accessible to the coordinator.

backend/src/data_entry/api.rs
```diff
@@ -45,6 +46,25 @@ pub fn router() -> OpenApiRouter<AppState> {
...
+    State(data_entry_repo): State<PollingStationDataEntries>,
+    Path(id): Path<u32>,
+) -> Result<Json<DataEntryStatus>, APIError> {
```
praseodym, line 62: This adds all data entry status enum variants to the OpenAPI spec, even though we only need the first and second data entries in case of differences. Would it be better to create a simple struct with the first and second data entry, and return that instead?

oliver3:
Does this endpoint return the data entry status or the data entries ? I see both names used in this PR and I think we should pick one.

backend/src/data_entry/api.rs
cikzh: Fixed in [933f803](https://github.com/kiesraad/abacus/commit/933f803e31b784595fc4170eaa7a4615203b5bca)

cikzh:
> Does this endpoint return the data entry status or the data entries ? I see both names used in this PR and I think we should pick one, I would say status.

The endpoint is about the "status of data entry of a certain polling station", so I think status is better as well. I'll change it

backend/src/data_entry/api.rs
cikzh: The endpoint doesn't necessarily assume it's being used for differences right now. We could of course do that and add a check in the backend to make sure the status is actually in the `EntriesDifferent` state. So I think the question is whether we think we want a generic endpoint for getting the data entry state, or a specific endpoint for difference resolving.

praseodym: I don't like the `DataEntryStatus` enum leaking into the frontend, but having a super-specific endpoint like 'get the two data entries in case they are different' isn't too appealing either. So let's keep the endpoint generic.

Lionqueen94 (approved):

praseodym (approved):

PR #1387, opened 2025-04-16
Add e2e test for second data entry user being different

### LINKED ISSUE #1151: Add e2e test for second data entry user being different
Add e2e test that second data entry user has to be different from the user doing the first data entry.

lkleuver:
Add e2e test for data entries requiring first and second data entry to have different users.

Note:
Has slightly empty test for starting data entry when this is the case.

closes #1151

### DIFF
--- a/frontend/e2e-tests/tests/data-entry.e2e.ts
+++ b/frontend/e2e-tests/tests/data-entry.e2e.ts
@@ -946,19 +946,13 @@ test.describe("api error responses", () => {
     await expect(dataEntryHomePage.fieldset).toBeVisible();
   });
 
-  test("UI Error: Second data entry user must be different from first entry", async ({ page, pollingStation }) => {
-    await page.route(`*/**/api/polling_stations/${pollingStation.id}/data_entries/1/claim`, async (route) => {
-      await route.fulfill({
-        status: 409,
-        json: {
-          error: "Conflict",
-          fatal: false,
-          reference: "SecondEntryNeedsDifferentUser",
-        },
-      });
-    });
-
-    await page.goto(`/elections/${pollingStation.election_id}/data-entry/${pollingStation.id}/1/recounted`);
+  test("UI Error: Second data entry user must be different from first entry", async ({
+    page,
+    pollingStationFirstEntryDone,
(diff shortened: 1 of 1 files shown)


frontend/e2e-tests/tests/data-entry.e2e.ts
```diff
@@ -945,4 +945,37 @@ test.describe("api error responses", () => {
...
+    await page.route(`*/**/api/polling_stations/${pollingStation.id}/data_entries/1/claim`, async (route) => {
+      await route.fulfill({
+        status: 409,
+        json: {
+          error: "Conflict",
+          fatal: false,
+          reference: "SecondEntryNeedsDifferentUser",
+        },
+      });
+    });
```
jschuurk-kr, line 959: For an e2e test let's create the situation in which the application returns this response, instead of mocking it.

lkleuver: yeah I realized later that it kind of defeats the purpose. 
I've changed it.

frontend/e2e-tests/tests/data-entry.e2e.ts
```diff
@@ -945,4 +945,31 @@ test.describe("api error responses", () => {
...
+    page,
+    pollingStationFirstEntryDone,
+  }) => {
+    await page.goto(
+      `/elections/${pollingStationFirstEntryDone.election_id}/data-entry/${pollingStationFirstEntryDone.id}/1/recounted`,
+    );
+    const recountedPage = new RecountedPage(page);
+
+    // Data entry currently returns "null" for all responses without results
+    await expect(recountedPage.next).toBeHidden();
```
jschuurk-kr, line 959: Created an issue for this bug where the user is not shown a proper error message: https://github.com/kiesraad/abacus/issues/1389.

jschuurk-kr, line 959:
```suggestion
    await expect(recountedPage.fieldset).toBeHidden();
```

jschuurk-kr (approved):

Lionqueen94 (approved):

PR #1392, opened 2025-04-17
Handle wrong data entry URL errors (reload finalised data entry, invalid entry number)

### LINKED ISSUE #1389: Error message should be shown if user revisits a url for a data entry they did previously
If a user did the first data entry and then revisits the url of one of the forms, e.g. `/elections/1/data-entry/4/1/recounted`, the `POST` to the claim endpoint returns a `409`:
```json
{"error":"First entry already finalised","fatal":false,"reference":"InvalidStateTransition"}
```
and the user is shown a blank page except for the top nav bar.

We need to show a proper error message.

Lionqueen94:
Closes #1389 

## Description
- Removed unused ErrorReferences
- Added `DataEntryAlreadyFinalised` ErrorReference
- Added handling of `DataEntryAlreadyFinalised` and `InvalidStateTransition` errors
- Added `--features embed-typst` to cargo build for local e2e tests


jschuurk-kr:
Tried out the scenario described in #1389. Result of returning to a data entry url you have completed is:

image

Behavior is the same if you go to the same url with a different user.

I think showing a warning is fine. We do this for "Een andere invoerder is bezig met dit stembureau." as well.

If I then complete second data entry and go the url of the first or second data entry, I get "Abacus is stuk" "error.polling_station_already_definitive". So would be nice to also implement a warning for this scenario.

If I change the data entry to an invalid number, I get a 400 "Invalid URL: invalid entry number \`3`" response and am shown the blank page.

Since it does take some effort from the user to trigger this behavior, I wouldn't want to invest too much time in showing different warnings for different errors. So my proposal would be to add a warning for `error.polling_station_already_definitive` and something generic for all the rest. The "Abacus is stuk"-page is actually not that bad for the generic purpose, since it has a back-button on the page that will get you right back to where you where. The copy is not ideal for this kind of scenario, though.
@praseodym What do you think?

frontend/e2e-tests/page-objects/data_entry/DataEntryHomePgObj.ts
```diff
@@ -34,11 +37,15 @@ export class DataEntryHomePage {
       name: "Je invoer is opgeslagen",
     });
     this.resumeDataEntry = page.getByRole("heading", { level: 2, name: "Je hebt nog een openstaande invoer" });
+    this.dataEntryWarningAlertTitle = page.getByTestId("dataEntryWarningAlertTitle");
```
jschuurk-kr, line 40: `name` can also take Regex, so we can do:
```suggestion
    this.dataEntryWarningAlertTitle = page.getByRole("heading", {level: 2, name: /^Je kan stembureau \d+ niet invoeren.$/})
```

frontend/e2e-tests/page-objects/data_entry/DataEntryHomePgObj.ts
```diff
@@ -34,11 +37,15 @@ export class DataEntryHomePage {
...
+    this.dataEntryWarningAlertDescription = page.getByTestId("dataEntryWarningAlertDescription");
```
jschuurk-kr, line 41: I have a slight preference to check if `this.alertDataEntryWarning` contains the description text over introducing a locator based on an `id`. For a user looking at the page it's more important that the alert with the right title contains the description than that an element with a specific id shows the description.

praseodym:
> If I then complete second data entry and go the url of the first or second data entry, I get "Abacus is stuk" "error.polling_station_already_definitive". So would be nice to also implement a warning for this scenario.

Agreed that we should add a proper warning for this. I think the current error comes from this code:
https://github.com/kiesraad/abacus/blob/f38bf25b405e87dbbd552f3776bb6eb903f3c2be/frontend/src/features/data_entry/components/DataEntryLayout.tsx#L27-L29
We should remove this code and handle this case in the `DataEntryProvider` instead. (I don't like the `DataEntryLayout` using the `pollingStationStatus`, I feel like it could get out of sync with the data entry itself.)

> If I change the data entry to an invalid number, I get a 400 "Invalid URL: invalid entry number `3`" response and am shown the blank page. [..] The "Abacus is stuk"-page is actually not that bad for the generic purpose, since it has a back-button on the page that will get you right back to where you where. 

We shouldn't show a blank page, throwing an error to show the generic "Abacus is stuk" error is fine.

frontend/e2e-tests/page-objects/data_entry/DataEntryHomePgObj.ts
Lionqueen94: Changed in 1112b3ea1e859892cb9bd5dd0b722701b3258288

frontend/e2e-tests/page-objects/data_entry/DataEntryHomePgObj.ts
Lionqueen94: Changed in 1112b3ea1e859892cb9bd5dd0b722701b3258288

Lionqueen94:
> > If I then complete second data entry and go the url of the first or second data entry, I get "Abacus is stuk" "error.polling_station_already_definitive". So would be nice to also implement a warning for this scenario.
> 
> Agreed that we should add a proper warning for this. I think the current error comes from this code:
> 
> https://github.com/kiesraad/abacus/blob/f38bf25b405e87dbbd552f3776bb6eb903f3c2be/frontend/src/features/data_entry/components/DataEntryLayout.tsx#L27-L29
> 
> We should remove this code and handle this case in the `DataEntryProvider` instead. (I don't like the `DataEntryLayout` using the `pollingStationStatus`, I feel like it could get out of sync with the data entry itself.)
> 
> > If I change the data entry to an invalid number, I get a 400 "Invalid URL: invalid entry number `3`" response and am shown the blank page. [..] The "Abacus is stuk"-page is actually not that bad for the generic purpose, since it has a back-button on the page that will get you right back to where you where.
> 
> We shouldn't show a blank page, throwing an error to show the generic "Abacus is stuk" page is fine.

I now check for the entry number where the original check for definitive was located and show the Not Found page with "Ongeldige invoer voor dit stembureau". I can change this to the "Abacus is stuk" page if this is better suited. What do you think?
The `entry_number.rs` doesn't return `EntryNumberNotSupported` unfortunately, so it's a bit less nice to capture in the DataEntryProvider useEffect

praseodym:
> I now check for the entry number where the original check for definitive was located and show the Not Found page with "Ongeldige invoer voor dit stembureau". I can change this to the "Abacus is stuk" page if this is better suited. What do you think? @jschuurk-kr @praseodym

That's fine!

I still see a blank page when I edit the URL to a second data entry while working on the first data entry (`InvalidStateTransition` gets returned from the claim API), can we add better error handling in that code path?

Lionqueen94:
> > I now check for the entry number where the original check for definitive was located and show the Not Found page with "Ongeldige invoer voor dit stembureau". I can change this to the "Abacus is stuk" page if this is better suited. What do you think? @jschuurk-kr @praseodym
> 
> That's fine!
> 
> I still see a blank page when I edit the URL to a second data entry while working on the first data entry (`InvalidStateTransition` gets returned from the claim API), can we add better error handling in that code path?

Handled in 7301612c85a6daa26aa85d1dd2ad6927e41176d7 also, and did some code cleanup as well

jorisleker (approved):
LGTM
- when I open the recounted page for a polling station I've already completed, I am redirect to the polling station choice page, with an appropriate alert
- when i use an invalid entry number or invalid polling station choice id, I get a 'abacus is stuk' page, with appropriate error message

oliver3:
As discussed, fatal errors during data entry are not handled at all at the moment, but we decided to create a separate issue for that https://github.com/kiesraad/abacus/issues/1405
Depending on the solution for that issue, we could maybe remove the error handling specific for an illegal data entry number.

oliver3 (approved):

Issue #1333, opened 2025-04-09
Restore front-end test coverage after replacing RTL integration test with e2e test


jschuurk-kr:
Follow-up of issue https://github.com/kiesraad/abacus/issues/898 and PR https://github.com/kiesraad/abacus/pull/1314

### Scope
Test code coverage dropped significantly for some files after replacing `DataEntry.integration.test.tsx` with e2e tests. Goal of this issue is to restore the code coverage to its previous level as much as possible/sensible.

See https://app.codecov.io/gh/kiesraad/abacus/commit/35ae00707e2236fd14cf7461d6de0e02507337a9/indirect-changes

Test code coverage comparison of branch `898-change-dataentryintegrationtesttsx-to-playwright-tests` with `main`:
- [x] code coverage src/api
  - [x] ApiClient.ts: 2 lines down -> https://github.com/kiesraad/abacus/pull/1345
- [x] code coverage src/app
- [x] code coverage src/components: 112 lines down
  - [x] ErrorModal.tsx: 11 lines down -> https://github.com/kiesraad/abacus/pull/1344
  - [x] StickyNav.tsx: 37 lines down -> ❌ 1 line left after merging other PRs, [line sets `maxHeight`](https://app.codecov.io/gh/kiesraad/abacus/blob/main/frontend%2Fsrc%2Fcomponents%2Fui%2FAppLayout%2FStickyNav.tsx)
  - [x] PollingStationNumber.tsx: 2 lines down -> ❌ not worth it, very simple ui component
  - [x] ProgressList.tsx: 28 lines down -> https://github.com/kiesraad/abacus/pull/1411
  - [x] ProgressListScroll.tsx: 34 lines down -> https://github.com/kiesraad/abacus/pull/1411 ([4 lines left to cover](https://app.codecov.io/gh/kiesraad/abacus/blob/main/frontend%2Fsrc%2Fcomponents%2Fui%2FProgressList%2FProgressListScroll.tsx) related to height and scroll)
- [x] code coverage src/features: 237 lines down
  - [x] data_entry/components: 207 lines down
    - [x] CandidatesVotesPage.tsx: 8 lines down -> https://github.com/kiesraad/abacus/pull/1365
    - [x] useVotersAndVotes: 1 line down -> https://github.com/kiesraad/abacus/pull/1427
    - [x] AbortDataEntryControl.tsx: 9 lines down -> https://github.com/kiesraad/abacus/pull/1437
[...]

PR #1411, opened 2025-04-23
restore code coverage of DataEntryProgress and ProgressList

jschuurk-kr:
partial #1333 

### Scope
- restore test coverage of DataEntryProgress: was 9 of 106 lines, is now 106 of 106 lines.
- restore test coverage for ProgressList: was 13 of 41 lines, is now 41 of 41 lines.
- small refactorings of DataEntryProgress.tsx
- simplify test util `overrideServerClaimDataEntryResponse()` by removing a parameter

### Questions
-  Are the refactorings in DataEntryProgress.tsx worth the risk?

### Comment
The `ProgressList` can set an "idle" class on an element twice. Once if it's not the active item and a second time if `menuStatusForFormSection()` in DataEntryProgress sets the `status` to "idle".  So you can get `idle idle disabled`, where the first "idle" means not active and the second "idle" means not error/warning/accept/...
~Let's use two different classes for this when we make the forms more generic.~ resolved in 47c2718c


Lionqueen94:
Can we just add or remove "active" if an element becomes active or idle. I agree that a double "idle" is quite bad

frontend/src/features/data_entry/components/DataEntryProgress.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+    const checkAndSaveLink = within(checkAndSave).getByRole("link", { name: "Controleren en opslaan" });
+    expect(checkAndSaveLink).toBeVisible();
+    expect(checkAndSaveLink).toHaveAttribute(
+      "href",
+      `/elections/${electionId}/data-entry/${pollingStationId}/${entryNumber}/save`,
+    );
+  });
+
+  test("Mismatch between election data and formState", async () => {
```
oliver3, line 235: Not sure what risk this test covers, as the formState should be initially created from the election data.

oliver3 (approved):
Looks good! Two remarks that are not important enough to prevent merging

frontend/src/features/data_entry/components/DataEntryProgress.test.tsx
jschuurk-kr: Not sure either.

It does gets us coverage for the line `if (!formSection) return null;`. And it demonstrates the behavior in case of an unlikely mistmatch, for whatever that's worth.

frontend/src/features/data_entry/testing/mock-data.ts
```diff
@@ -50,7 +50,7 @@ export function getDefaultFormSection(id: FormSectionId, index: number): FormSec
   return {
     id,
     index,
-    title: "Toegelaten kiezers en uitgebrachte stemmen",
+    title: "Toegelaten kiezers en uitgebrachte stemmen", // TODO: shouldn't this be dynamic?
```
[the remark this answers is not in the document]
jschuurk-kr: changed to "title of the form section" in 5030d172

jschuurk-kr:
> Can we just add or remove "active" if an element becomes active or idle. I agree that a double "idle" is quite bad

done in 47c2718c with help from Mark

Lionqueen94 (approved):

oliver3 (approved):

PR #1427, opened 2025-04-28
restore code coverage: useVotersAndVotes (cache)

lkleuver:
Adds a test for the cache flow in useVotersAndVotes which was not yet covered.

I've added a separate cache.test.tsx file since the other data entry sections also miss this coverage. 
This will go away once we make it more generic.


Lionqueen94 (approved):

frontend/src/features/data_entry/hooks/useDataEntry.ts
```diff
@@ -24,9 +24,13 @@ export default function useDataEntry(
...
   pollingStationId: number,
   entryNumber: number,
+  initialDataEntryState?: DataEntryState,
 ): DataEntryStateAndActions {
   const client = useApiClient();
-  const [state, dispatch] = useReducer(dataEntryReducer, getInitialState(election, pollingStationId, entryNumber));
+  const [state, dispatch] = useReducer(
+    dataEntryReducer,
+    initialDataEntryState || getInitialState(election, pollingStationId, entryNumber),
+  );
```
jschuurk-kr, line 33: I'm not sure what to think of this change, since we're adding behavior to production code for testing purposes, while it's not obvious that it's for testing purposes-only.

I'm not sure what an alternative solution would be, though. Maybe mocking `getInitialState()` in the reducer?

@praseodym @oliver3 Do you have any ideas?

lkleuver: this is how it was implemented before. 
That does not mean it's correct or doesn't need to change of course

oliver3: Together with @lkleuver we came to the conclusion that mocking the data entry reducer through `DataEntryContext` would be better indeed. @lkleuver is working on another PR where he is setting that up and can be a blue print for this PR as well.

Lionqueen94 (commented):
I think this PR needs a rebase because commits from main have been added

frontend/src/features/data_entry/hooks/useDataEntry.ts
lkleuver: I've changed the test, since mocking DataEntryContext would require to rewrite the entire test suite, I've moved this test to cache.test.tsx since other use..hooks also miss the cache check.

However since the plan is to make these more generic i've skipped these for now (it was never covered outside voters and votes so it's technically not part of the "epic".

oliver3 (approved):

Lionqueen94 (approved):

PR #1429, opened 2025-04-28
Moved impls for ApiError to respective modules

Lionqueen94:
- Moved impls for ApiError to respective modules
- Reordered imports in all Rust files where not consistent into 3 groups:
1.    `std`, `core` and `alloc`,
2.    external crates,
3.    `self`, `super` and `crate` imports.
- Moved imports from crate to super where possible

### DIFF
--- a/backend/src/error.rs
+++ b/backend/src/error.rs
@@ -1,3 +1,5 @@
+use std::error::Error;
+
 use axum::{
     Json,
     extract::rejection::JsonRejection,
@@ -8,7 +10,6 @@ use hyper::header::InvalidHeaderValue;
 use quick_xml::SeError;
 use serde::{Deserialize, Serialize};
 use sqlx::Error::RowNotFound;
-use std::error::Error;
 use tracing::error;
 use utoipa::ToSchema;
 use zip::result::ZipError;
--- a/backend/src/error.rs
+++ b/backend/src/error.rs
@@ -350,12 +350,6 @@ impl From<SeError> for APIError {
     }
 }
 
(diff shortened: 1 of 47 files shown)


backend/src/error.rs
```diff
@@ -357,40 +349,12 @@ impl From<SeError> for APIError {
...
-    fn from(err: AuthenticationError) -> Self {
-        APIError::Authentication(err)
-    }
-}
-
 impl From<ZipError> for APIError {
     fn from(err: ZipError) -> Self {
         APIError::ZipError(err)
     }
 }
```
praseodym, line 356: `ZipError` is only used in the `report` module so it can be moved there

praseodym (changes requested):
I noticed you sorted and regrouped imports, but applied your own grouping. A common practice is to use three groups:

1.    `std`, `core` and `alloc`,
2.    external crates,
3.    `self`, `super` and `crate` imports.

`rustfmt` can fix this for us by using the [`group_imports = StdExternalCrate`](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#StdExternalCrate%5C%3A) option, but unfortunately that feature is not yet stable. However, you can run it once using `cargo +nightly fmt -- --config group_imports=StdExternalCrate`.

backend/src/error.rs
Lionqueen94: Changed in 8c69acb345cadced8b2ce344c820300b6a83f1d7
The import stays because it is handled in the into_response

cikzh (approved):

PR #1430, opened 2025-04-28
Renamed data_entry_choice feature to data_entry_home

### LINKED ISSUE #1336: Rename `data_entry_choice` feature to `data_entry_home`
> @oliver3 and I had a 'long' discussion about the name, which resulted in me adding the question to the PR.
> 
> My thoughts on the name:
> - `PollingStationChoicePage` works fine as a name for the page object, but it's too different from `data_entry_choice` in the frontend. And I agree with the rename in the frontend away from "polling station".
> - `DataEntryChoicepage` I find confusing in the e2e tests. The concept of polling station is a lot more prominent on that page than the concept of data entry. So this feels to me like a code concept leaking into a non-code context. (edit: You don't pick a data entry. You pick a polling station and the application decides which data entry it is.)
> - `DataEntryHomePage` seemed like a good middle ground between the two other options.

_Originally posted by @jschuurk-kr in https://github.com/kiesraad/abacus/pull/1314#discussion_r2034710233_

Depends on #1329 (to avoid merge conflicts)

Lionqueen94:
Closes #1336


oliver3 (approved):
Looks good! Also agree with your decisions regarding DataEntryHome vs. PollingStationChoice.

praseodym (approved):

PR #1431, opened 2025-04-28
Add ProgressList.OverviewItem

oliver3:
For the overview on the data entry resolve differences page as shown below. Ladle can be seen at https://0bca3129.kiesraad-abacus.pages.dev/ladle/?story=progress-list--overview-progress-list.

image

### DIFF
--- a/frontend/src/components/ui/ProgressList/ProgressList.module.css
+++ b/frontend/src/components/ui/ProgressList/ProgressList.module.css
@@ -29,12 +29,20 @@
     scroll-margin-top: 0.8rem;
     scroll-margin-bottom: 0.8rem;
 
-    &:hover {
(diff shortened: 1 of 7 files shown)


Lionqueen94 (commented):
Looks good, but doesn't the OverviewItem also need a hover? @jorisleker

praseodym:
Does this component have the same scroll behaviour as the normal ProgressList component (see its Ladle story)?

oliver3:
Thanks @Lionqueen94 and @praseodym, both would be improvements over the list without any behaviour in this implementation, and after discussing with @jorisleker it is okay not to have those right now.

Lionqueen94 (approved):

jorisleker (approved):

PR #1453, opened 2025-04-30
restore code coverage: DataEntryNavigation

lkleuver:
### Scope

restore test coverage of DataEntryNavigation as part of issue #1333


Lionqueen94 (commented):
Nice addition!

There is still one state not covered according to codecov, the `blocker.proceed()` when `state === "aborted"`, see:
image

lkleuver:
> Nice addition!
> 
> There is still one state not covered according to codecov, the `blocker.proceed()` when `state === "aborted"`, see: image
> 
> Edit: Or maybe this state is actually never able to be triggered?

It only triggers when the reducer/action side is faulty

Lionqueen94:
> > Nice addition!
> > There is still one state not covered according to codecov, the `blocker.proceed()` when `state === "aborted"`, see: image
> > Edit: Or maybe this state is actually never able to be triggered?
> 
> It only triggers when the reducer/action side is faulty.
> 
> edit: I've added a small test for one scenario, but status === "aborted" can't really be the case I think, since the blocker wont block on that status and thus the modal wont be rendered to trigger the function.
> 
> I wonder if we could remove the if statement in the component

Yeah it makes sense that "aborted" will not be the status if the onSubmit "aborting" fails, so this case seems impossible to me too

praseodym (approved):

PR #1462, opened 2025-05-02
Election Create routing

cikzh:
Added actual navigation for the election create pages and put it in its own feature.


praseodym (changes requested):
I see this on `/elections/create/check-definition` after uploading an election definition EML file:
Screenshot 2025-05-02 at 17 46 39

praseodym (changes requested):
Navigation works fine now!

The `NavBar` still needs some attention:
- on `/elections/create` the admin links (Verkiezingen/Gebruikers/Werkplekken/Logs) are shown
- on `/elections/create/check-definition` (the next page), no links are shown

Looking at the Figma design, the admin links need to be shown for all steps too.

jorisleker (changes requested):
Sorry @cikzh, the design was updated whilst you were building it:
- one step was added to this flow ('Aantal kiesgerechtigden').
- items after the current step are disabled (just like in data entry)
- items before the current step have the check icon 

Some other things that I noticed:
- Items in the menu that are not clickable should not have the hover effect

cikzh:
I'll put the PR in draft until I've implemented it with the new design

cikzh:
@praseodym Fixed links and also added a check for the user's role.

@jorisleker Added menu item and added hardcoded `disabled` attributes. The logic for displaying the menu items (including the checkmarks) will be implemented in another PR once we more pages than just the first.

jorisleker (dismissed):

jorisleker (approved):

praseodym (approved):

Issue #1331, opened 2025-04-09
Refactor e2e tests: helper functions, file size


jschuurk-kr:
Follow-up of issue https://github.com/kiesraad/abacus/issues/898 and PR https://github.com/kiesraad/abacus/pull/1314

### Scope
- check and improve usage of frontend/e2e-tests/helpers-utils/e2e-test-helpers.ts
- split up large e2e test files: resume-data-entry.e2e.ts and data-entry.e2e.ts
- [x] replace some of the API respons mocks in data-entry.api-errors.e2e.ts with actions -> https://github.com/kiesraad/abacus/pull/1465
- [x] importing from `@kiesraad`, relates to https://github.com/kiesraad/abacus/issues/1226
  - `import { PollingStation, PollingStationResults } from "@kiesraad/api";` -> `... from "@/api";
-  [x] small refactorings mentioned in https://github.com/kiesraad/abacus/pull/1465
- remove tests that are also covered by the model-based tests 
- [x] move helper functions that use the API to their own file -> https://github.com/kiesraad/abacus/pull/1532
- anything else?

PR #1465, opened 2025-05-08
e2e tests: replace two mocks with fixtures and some small refactorings

jschuurk-kr:
partial #1331 (There's more to do, but this PR is large enough as it is.)

### Scope
- move data entry e2e tests to separate folder
- move data entry api error response e2e tests to separate file
- replace two mocked responses in data entry api error response e2e tests with fixtures
- move the two data entry api error response e2e tests that do need mocked responses to bottom of the file
- rename `user` fixture to `newTypist`

It's probably easier to review per commit than the whole PR at once.

### Question
I introduced two new fixtures: `typistOne` and `typistTwo`. They log in using the `storageState` and return a `page` (to do browser things) and a `request` (to do API things). This was needed for the test where one user claims the data entry and then a second user tries to also claim the data entry.

Question is if we want to stop using `test.use({ storageState: "e2e-tests/state/typist.json", });` in favor of these fixtures. Except for cases where you need separate users in a single test, the only advantage would be readability. Instead of having the user account at the top of the file, it would be in the fixtures (i.e. arguments) for each test.


Lionqueen94 (approved):

praseodym (approved):
> Question is if we want to stop using test.use({ storageState: "e2e-tests/state/typist.json", }); in favor of these fixtures. Except for cases where you need separate users in a single test, the only advantage would be readability. Instead of having the user account at the top of the file, it would be in the fixtures (i.e. arguments) for each test.

I prefer the fixtures because I think they are more readable, but I also don't think this needs to be refactored right now. We can start using these fixtures when writing new tests.

PR #1466, opened 2025-05-08
Add headers for security hardening

### LINKED ISSUE #508: Add CSP headers to served html / SVG documents
Seeing the requested functionality:
  [S] De frontend is bereikbaar vanuit de backend: de applicatie kan als één geheel gedraaid worden.

It would be wise to add CSP headers limiting the scope of where 'active content' is allowed to be loaded from. I expect the more capable system-admin to put a proxy on front of the container. But I wouldn't count on all admin's to do so, especially if the docker container seems to 'just work' (tm)

### LINKED ISSUE #1425: Add HTTP headers for security hardening
E.g. https://owasp.org/www-project-secure-headers/

praseodym:
This PR will add relevant headers from https://owasp.org/www-project-secure-headers/ to all responses from Abacus.

Also add a [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control) header to prevent caching of API responses. [`memory-serve` already sets](https://docs.rs/memory-serve/latest/memory_serve/#configuration-options) Cache-Control header for static files, so no change is needed there.

Closes #1425
Closes #508


Lionqueen94 (dismissed):

jschuurk-kr:
I ran the [oshp-validator](https://github.com/oshp/oshp-validator) and not all tests passed. Tests are defined in [this file](https://github.com/oshp/oshp-validator/blob/main/tests_suite.yml). See attached file for the test output (had to get creative with the format, because I can't upload the html report): [20250508-output-oshp-validator.md](https://github.com/user-attachments/files/20102521/20250508-output-oshp-validator.md).

Will check the failing tests against the OWASP recommendations and the implementation.

praseodym:
I've added the remaining headers. The `oshp-validator` doesn't report anything unexpected now:

```
> ./venom run --var="target_site=http://127.0.0.1:8080" --var="logout_url=api/user/logout" tests_suite.yml
	  [trac] writing venom.0.log
 • HTTP security response headers test suites (tests_suite.yml)
 	• Strict-Transport-Security FAIL
 		• http
 		  Testcase "Strict-Transport-Security", step #0-0: Assertion "result.headers.Strict-Transport-Security ShouldNotBeNil" failed. expected: Not Nil but is was (tests_suite.yml:16)
 		  Testcase "Strict-Transport-Security", step #0-0: Assertion "result.headers.Strict-Transport-Security ShouldEqual \"max-age=31536000; includeSubDomains\"" failed. expected: max-age=31536000; includeSubDomains  got: <nil> (tests_suite.yml:17)
 	• X-Frame-Options PASS
 	• X-Content-Type-Options PASS
 	• Content-Security-Policy PASS
 	• X-Permitted-Cross-Domain-Policies PASS
 	• Referrer-Policy PASS
 	• Clear-Site-Data PASS
 	• Cross-Origin-Embedder-Policy PASS
 	• Cross-Origin-Opener-Policy PASS
 	• Cross-Origin-Resource-Policy PASS
 	• Permissions-Policy PASS
 	• Cache-Control FAIL
 		• http
 		  Testcase "Cache-Control", step #0-0: Assertion "result.headers.Cache-Control ShouldEqual \"no-store, max-age=0\"" failed. expected: no-store, max-age=0  got: max-age:300, private (tests_suite.yml:138)
 	• Feature-Policy-should-not-exist PASS
 	• Public-Key-Pins-should-not-exist PASS
 	• Expect-CT-should-not-exist PASS
 	• X-XSS-Protection-should-not-exist PASS
final status: FAIL
```

I had to change `method: GET` to `method: POST` for the `Clear-Site-Data` check on the logout endpoint, because use HTTP POST for the logout API request.

backend/src/authentication/api.rs
```diff
@@ -1,5 +1,6 @@
 use axum::{
     extract::{Path, State},
+    response::AppendHeaders,
     response::{IntoResponse, Json},
```
Lionqueen94, line 4:
```suggestion
    response::{AppendHeaders, IntoResponse, Json},
```

praseodym: Fixed in 7fbadaa16b887fb41d5ec1fdb4b3b89bb34b8ef6

Lionqueen94 (approved):

marlonbaeten:
Note that abacus-test.nl will also add security headers, not to be confused with the headers set by the backend.

marlonbaeten (approved):
Nice work!

Lionqueen94 (approved):

marlonbaeten:
I have found an inline image - which is denied by this policy. I propose to change the inline image to an icon file, instead of changing the policy. I will make a PR for this.

praseodym:
> I have found an inline image - which is denied by this policy. I propose to change the inline image to an icon file, instead of changing the policy. I will make a PR for this.

Which one?

marlonbaeten:
> > I have found an inline image - which is denied by this policy. I propose to change the inline image to an icon file, instead of changing the policy. I will make a PR for this.
> 
> Which one?

https://github.com/kiesraad/abacus/pull/1481

PR #1470, opened 2025-05-09
Errors for hash validation

cikzh:
Implements part of #1457 

Checks if the length of the input is exactly 4 characters, shows an error otherwise.
Actually checking the input is out of scope and will implemented after #889


praseodym (approved):
@cikzh I fixed the `no-unsafe-type-assertion` lint error in 94bb68f38c280f5d96bf5f5271e13836e0a6f06b

jorisleker (approved):
Tested and invalid length of input message LGTM
Assume that error messages where the input is 4 chars but does not match the hash will be added after the backend api returns a repons.

PR #1471, opened 2025-05-09
Update validation logic to be able to check for differences with the first data entry and add warning W.001

### LINKED ISSUE #1439: Backend: Update validation logic to be able to check for differences with the first data entry
- Keep the validation logic in `validation.rs` as is
- Implement validate trait on `DataEntryStatus`:
  - Call validation logic on `PollingStationResults`
  - If 2nd data entry, call compare function for 2nd data entry differences
    - Return warning with field names with differences in case of differences with first data entry
- Except in summary module, always call validate on `DataEntryStatus` instead of on `PollingStationResults`

### LINKED ISSUE #1441: Frontend: Add warning and update acceptWarnings checkbox text

Lionqueen94:
Closes #1439 
Closes #1441 

Screenshot recounted page:
image

Screenshot voters and votes page:
image


jorisleker (approved):
LGTM, 
- ✅ warning is shown only when there are no validation errors
- ✅ warning is shown only when there are differences with first entry
- ✅ no warning is shown when I enter a 0 in second data entry and left field empty in second data entry, so check is on interpreted value, not literal
- ✅ am able to accept warnings and continue
- ✅ am able to save input with differences warnings

Issue #1477, opened 2025-05-12
Implement side menu routing logic for election creation


cikzh:
- Keep track of election creation state
- Checks and arrows

PR #1484, opened 2025-05-12
Implemented progress list state and added polling station role stub page

cikzh:
Closes https://github.com/kiesraad/abacus/issues/1477

Adds state for the progress list menu of the Election Create pages.

Note that the "Volgende" button in the CheckElectionDefinition component currently is mocked. Actual functionality to the button will be added after #889.
 
Also note that currently the state is kept when using the browser's navigation buttons, but lost when explicitly refreshing the page.


jorisleker (commented):
Navigation state looks good!
One thing I noticed (and not sure if this was supposed to be implemented, or is something that will be added later): 

- If i use 'volgende' to go to the polling station role page, ...
- and then click 'verkiezinginsdefinitie' in the navigation, ...
- I go back to that page (expected)
- but the form is empty (not expected)
- and the 'polling station role' page is disabled in the navigation (not expected)

(if not implemented yet in this PR, I'll approve. If this was supposed to work as expected, then please fix :-))

cikzh:
@jorisleker This is not implemented yet. We don't actually check (and save) the hash with the backend yet. This will be implemented after #889

jorisleker (approved):

praseodym (approved):

jorisleker (approved):

PR #1485, opened 2025-05-12
Added e2e test for second data entry that is different from first data entry

Lionqueen94:
## Questions
1. ~@jschuurk-kr Is this the desired location for this test, or do you prefer it in one of the other e2e test files?~
2. ~@jschuurk-kr Should there be any check on if the data entries are now marked as different for the coordinator? If so, can I change roles in a test?~

### DIFF
--- a/frontend/e2e-tests/helpers-utils/e2e-test-helpers.ts
+++ b/frontend/e2e-tests/helpers-utils/e2e-test-helpers.ts
@@ -1,9 +1,9 @@
 import { expect, Page } from "@playwright/test";
 import { CandidatesListPage } from "e2e-tests/page-objects/data_entry/CandidatesListPgObj";
 import { CheckAndSavePage } from "e2e-tests/page-objects/data_entry/CheckAndSavePgObj";
-import { DataEntryBasePage } from "e2e-tests/page-objects/data_entry/DataEntryBasePgObj";
 import { DataEntryHomePage } from "e2e-tests/page-objects/data_entry/DataEntryHomePgObj";
 import { DifferencesPage } from "e2e-tests/page-objects/data_entry/DifferencesPgObj";
+import { NavigationPanel } from "e2e-tests/page-objects/data_entry/NavigationPanelPgObj";
 import { RecountedPage } from "e2e-tests/page-objects/data_entry/RecountedPgObj";
 import { VotersAndVotesPage } from "e2e-tests/page-objects/data_entry/VotersAndVotesPgObj";
 
@@ -44,7 +44,8 @@ export async function fillDataEntryPages(page: Page, results: PollingStationResu
 }
 
 export async function fillCandidatesListPages(page: Page, results: PollingStationResults) {
-  const candidateListNames: string[] = await new DataEntryBasePage(page).navPanel.allListNames();
+  const candidateListNames: string[] = await new NavigationPanel(page).allListNames();
+
   // make sure the form has the same number of political groups as the input data
   expect(candidateListNames.length).toBe(results.political_group_votes.length);
 
--- a/frontend/e2e-tests/helpers-utils/e2e-test-helpers.ts
+++ b/frontend/e2e-tests/helpers-utils/e2e-test-helpers.ts
@@ -1,6 +1,7 @@
 import { expect, Page } from "@playwright/test";
 import { CandidatesListPage } from "e2e-tests/page-objects/data_entry/CandidatesListPgObj";
 import { CheckAndSavePage } from "e2e-tests/page-objects/data_entry/CheckAndSavePgObj";
+import { DataEntryBasePage } from "e2e-tests/page-objects/data_entry/DataEntryBasePgObj";
 import { DataEntryHomePage } from "e2e-tests/page-objects/data_entry/DataEntryHomePgObj";
 import { DifferencesPage } from "e2e-tests/page-objects/data_entry/DifferencesPgObj";
 import { RecountedPage } from "e2e-tests/page-objects/data_entry/RecountedPgObj";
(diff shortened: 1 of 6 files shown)


jschuurk-kr:
@Lionqueen94 To use different roles in an e2e test you can use the [`typistTwo` fixture](https://github.com/kiesraad/abacus/blob/main/frontend/e2e-tests/fixtures.ts#L93) and create a similar one for the coordinator.

However, I'm not sure we should do all of this in a single test. Will need to think about how I'd want to split it up. Or feel free to suggest something!

jschuurk-kr:
@Lionqueen94 As to your questions:

> Is this the desired location for this test, or do you prefer it in one of the other e2e test files?

Location is fine as-is. The data-entry.e2e.tst file is too long. And the arguably the test for triggering W.001 should be in the "errors and warnings" describe. But thinking about how this could be improved takes more time than I have right now.

> Should there be any check on if the data entries are now marked as different for the coordinator? If so, can I change roles in a test?

There should be a check, since `resolve-differences.e2e.ts` sets up the different data entries through the API. To use different roles in an e2e test you can use the [`typistTwo` fixture](https://github.com/kiesraad/abacus/blob/main/frontend/e2e-tests/fixtures.ts#L93) and create a similar one for the coordinator.

Finally, for splitting up the test, I'm thinking it should be two tests:
1. second entry with differences and accepted warnings, so it shows up for the coordinator
2. second entry with differences, but warnings are corrected instead of accepted

frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
```diff
@@ -424,6 +426,157 @@ test.describe("second data entry", () => {
...
+      ["Je invoer is opgeslagen", "Geef het papieren proces-verbaal terug aan de coördinator."].join(""),
+    );
+
+    // check if data entries are marked as definitive on coordinator status page
+    const coordinatorPage = coordinator.page;
+    await coordinatorPage.goto(`/elections/${pollingStation.election_id}/status`);
+
+    const electionStatusPage = new ElectionStatus(coordinatorPage);
+    await expect(electionStatusPage.definitive).toBeVisible();
+    await electionStatusPage.definitive.getByRole("row", { name: pollingStation.name }).click();
```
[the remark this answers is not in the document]
Lionqueen94: Nvm fixed it... somehow the `{ page }` does not work, I needed to use `role.page` for both

Lionqueen94:
> @Lionqueen94 As to your questions:
> 
> > Is this the desired location for this test, or do you prefer it in one of the other e2e test files?
> 
> Location is fine as-is. The data-entry.e2e.tst file is too long. And the arguably the test for triggering W.001 should be in the "errors and warnings" describe. But thinking about how this could be improved takes more time than I have right now.
> 
> > Should there be any check on if the data entries are now marked as different for the coordinator? If so, can I change roles in a test?
> 
> There should be a check, since `resolve-differences.e2e.ts` sets up the different data entries through the API. To use different roles in an e2e test you can use the [`typistTwo` fixture](https://github.com/kiesraad/abacus/blob/main/frontend/e2e-tests/fixtures.ts#L93) and create a similar one for the coordinator.
> 
> Finally, for splitting up the test, I'm thinking it should be two tests:
> 
>     1. second entry with differences and accepted warnings, so it shows up for the coordinator
> 
>     2. second entry with differences, but warnings are corrected instead of accepted

I changed it to 2 separate tests and check the status page for both :) Ready for review now

frontend/e2e-tests/helpers-utils/e2e-test-helpers.ts
```diff
@@ -35,12 +36,20 @@ export async function fillDataEntryPages(page: Page, results: PollingStationResu
...
-  const candidateListNames: string[] = await differencesPage.navPanel.allListNames();
+  await fillCandidatesListPages(page, results);
+
+  const checkAndSavePage = new CheckAndSavePage(page);
+  await expect(checkAndSavePage.fieldset).toBeVisible();
+  return checkAndSavePage;
+}
+
+export async function fillCandidatesListPages(page: Page, results: PollingStationResults) {
+  const candidateListNames: string[] = await new DataEntryBasePage(page).navPanel.allListNames();
```
jschuurk-kr, line 47: I prefer to never instantiate a BasePage page object, because they're intended to be used through inheritance. There doesn't seem to be a way to enforce this, though.

Whether that's a good rule or not, since we only need an instance of the `NavigationPanel` here, we should instantiate that one directly.

frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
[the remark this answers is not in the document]
jschuurk-kr: I think what happens is that the `page` is logged in as the typist, because of the `test.use()` at the top of the file and the blank page is shown briefly as the typist goes to the election status page right before they are redirected to the login page.

frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
```diff
@@ -424,6 +426,157 @@ test.describe("second data entry", () => {
...
```
jschuurk-kr, line 511: We need to do something smart in the page object for locating a specific row in a specific table. Can take a look at it later today.

frontend/e2e-tests/helpers-utils/e2e-test-helpers.ts
oliver3: Ah good one! Changed into NavigationPanel in https://github.com/kiesraad/abacus/pull/1485/commits/f35a6ec50b493df18097e132f8d49bcafffcc9b5

frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
oliver3: I've tried some options in another PR and all extra methods, helper functions and custom matchers felt a bit overkill or would prevent using it simply as a locator.
I changed it now to check if the table contains the name of the polling station.

jschuurk-kr (approved):

lkleuver (approved):

Issue #1224, opened 2025-03-24
Restructure i18n


oliver3:
Our `src/lib/i18n` folder contains both the tooling and the translation json files for the whole application.

In the https://github.com/kiesraad/abacus/tree/bulletproof-react-structure branch this has been split up into `src/utils/i18n`, generic json files in `src/assets/i18n/nl/*` and json files per feature, e.g. `src/features/polling_stations/assets/i18n/nl.json`

- [x] decide upon structure
- [ ] move files to their new location
- [ ] make the scripts work
- [ ] update all documentation where `lib/i18n` is present

Current proposal:
- dedicated `i18n` folder within each feature, containing a json per locale, e.g. `features/polling_stations/i18n/nl.json`
- top level i18n folder containing nl.ts and generic bundles. To be decided: only one file or allow multiple files

lkleuver:
Has this been discussed yet?
I personally worry that the proposal will make it hard to have all files translated since it would be quite hidden where all translations files are.

Also, there needs to be a barrel file somewhere I believe, for typescript validation of translation keys.

Lionqueen94:
Another good thing to note is that currently the `generic.json` file cannot contain key-dictionary pairs, only key-value pairs. Otherwise running `npm run gen:translation-json` does not result in the correct files (the key of the dictionary value becomes a new file).

PR #1486, opened 2025-05-13
Restructure i18n

oliver3:
Resolves https://github.com/kiesraad/abacus/issues/1224
- Move `src/lib/i18n` to `src/i18n`
- Rename `i18n.ts` to `translate.ts` and `index.ts` to `i18n.ts`
- Remove re-exporting and fix circular import
- Update all imports
- Update scripts, sigrid config & documentation accordingly

Question: is this new location okay? It is not a folder that the Bulletproof React describes but I feel it is an important enough part of our application to justify introducing the `i18n/` folder.


Lionqueen94 (approved):
I think this location is fine, nice work

PR #1490, opened 2025-05-14
Allow inlined images in CSP

praseodym:
Vite automatically inlines images to `data:` URLs, which was broken by the strict CSP in #1466. This PR fixes this by explicitly allowing inlined images from `data:` URLs. The (insufficient) fix in #1481 is undone.


oliver3:
Tested locally with `./pull-and-run` and everything is working now!

oliver3 (approved):

cikzh (approved):

PR #1503, opened 2025-05-16
switch to eslint flat config file

### LINKED ISSUE #580: Migrate to ESLint flat config
(Split off from #213)

ESLint is moving towards a new flat configuration format. We're using the older eslintrc configuration format, so we need to migrate. There is a [migration guide](https://eslint.org/docs/latest/use/configure/migration-guide), but due to the number of ESLint plugins we use this turns out to be a bit tricky. The configuration migrator (`@eslint/migrate-config`) doesn't result in a working config.

Until this is done, running `npm run lint` reports the following warning: `
(node:17894) ESLintRCWarning: You are using an eslintrc configuration file, which is deprecated and support will be removed in v10.0.0. Please migrate to an eslint.config.js file. See https://eslint.org/docs/latest/use/configure/migration-guide for details. An eslintrc configuration file is used because you have the ESLINT_USE_FLAT_CONFIG environment variable set to false. If you want to use an eslint.config.js file, remove the environment variable. If you want to find the location of the eslintrc configuration file, use the --debug flag.
`

jschuurk-kr:
closes #580 

### Scope

Switch to the flat config file that's the default since eslint 9. The trick seems to be to use `tseslint` from `typescript-eslint` in the config file.

Apart from creating the new config file based on the old one, I had to make the following changes:
- add an `override` for the `globals` package, since there's a dumb typo in the version we were using, see https://github.com/sindresorhus/globals/issues/268#issuecomment-2437327512
- ignore an import linting rule from in `frontend/scripts/gen_icons.js` and `frontend/scripts/gen_po.js`

### Testing
Please test this thoroughly by checking if the different kinds of files (as defined through the `files` and `ignores`in each section of the config) have the correct linting rules applied.

You can do this by either adding code that violates a rule, or by adding an unneeded `// eslint-disable-next-line <some rule>` to a file. In both cases eslint should show an error or warning.

### Questions
- Is there a better solution for the new ignores in `frontend/scripts/gen_icons.js` and `frontend/scripts/gen_po.js`?


marlonbaeten (dismissed):
Tested by running:
```shell
cd frontend
npm install
npm lint
```

And repeating the last command with various formatting errors. Works :+1:

It does also catch some indentation errors that the previous configuration seems to allow. Although there is a performance regression, on my machine `npm run lint` takes about 20% longer.

praseodym (dismissed):

praseodym:
> * Is there a better solution for the new ignores in `frontend/scripts/gen_icons.js` and `frontend/scripts/gen_po.js`?

I've fixed the ignore in `gen_icons.js` and added a comment in `gen_po.js` (that one is a bit tricker to fix, might not be worth it right now).

> Although there is a performance regression, on my machine `npm run lint` takes about 20% longer.

I also noticed that ESLint is even slower now :(

praseodym (approved):

praseodym:
Fixed a merge conflict for `package-lock.json` in 7d7af94.

frontend/eslint.config.js
```diff
@@ -0,0 +1,140 @@
...
+      tseslint.configs.strictTypeChecked,
+      importPlugin.flatConfigs.recommended,
+      importPlugin.flatConfigs.typescript,
+      reactRefresh.configs.recommended,
+      jsxA11y.flatConfigs.recommended,
+      eslintPluginPrettierRecommended,
+      reactHooks.configs["recommended-latest"],
+    ],
+    rules: {
+      "no-undef": "off",
```
oliver3, line 44: I'd prefer to keep it on and I see no errors if I do so

frontend/eslint.config.js
```diff
@@ -0,0 +1,140 @@
...
+      "@typescript-eslint/restrict-template-expressions": "off",
+    },
+  },
+  {
+    files: ["src/components/ui/**/*.e2e.ts"],
+    extends: [playwright.configs["flat/recommended"]],
+    rules: {
+      // Needed for Ladle, page.waitForSelector("[data-storyloaded]")
+      "playwright/no-wait-for-selector": "off",
+      "@typescript-eslint/no-floating-promises": "error",
```
oliver3, line 133: Annoying that these e2e rules have to be repeated again. We might be able to extract some common rules but not sure if it is worth it.

frontend/eslint.config.js
jschuurk-kr: typescipt-eslint recommends doing it this way: https://typescript-eslint.io/troubleshooting/faqs/eslint/#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors

But as you say, enabling it does not result in any issues, so am fine with keeping it on.

frontend/eslint.config.js
```diff
@@ -0,0 +1,140 @@
+// @ts-check
+import eslint from "@eslint/js";
+import tsParser from "@typescript-eslint/parser";
+// @ts-ignore
```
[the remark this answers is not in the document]
jschuurk-kr: Agree for eslint, but not for the IDE. (I think?)

frontend/eslint.config.js
jschuurk-kr: I suspect the `"@typescript-eslint/no-floating-promises": "error"` can be removed, since it's these files also maths the section with `files: ["*.e2e.ts", "e2e-tests/**/*"]`.

marlonbaeten (dismissed):

frontend/eslint.config.js
jschuurk-kr: fixed in dd2bd33bea82e46da16509c209ff56f4fcfc0298

frontend/eslint.config.js
jschuurk-kr: line that disables "no-undef" removed in 5a8b78687675e2641becb847d60730ed30fa0898

frontend/eslint.config.js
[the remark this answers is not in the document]
jschuurk-kr: removed the `@ts-check` and `@ts-ignore` in 6b5bc1ea8c03f526ea8aa2e08a524916bf8370de

typescript-eslint has it in [their example config](https://typescript-eslint.io/getting-started/#step-2-configuration), but am not sure what value it adds. We can always re-introduce it later, if we need it.

frontend/eslint.config.js
oliver3: Oh interesting, so the idea is that the `tsc` should catch them and has better judgement. In any case, let's avoid globals altogether indeed.

oliver3 (approved):

frontend/eslint.config.js
jschuurk-kr: @oliver3 Wait, so I should restore the `"no-undef": "off",`?

And we'll still have the `globals` for `files: ["**/*.js"],` regardless.

praseodym (approved):

frontend/eslint.config.js
jschuurk-kr: keeping "no-undef": "off" out of the file after discussion with Oliver

PR #1504, opened 2025-05-16
Fixed mutation testing issues in authentication backend module

Lionqueen94:

### DIFF
--- a/backend/src/authentication/mod.rs
+++ b/backend/src/authentication/mod.rs
@@ -31,7 +31,6 @@ pub const SECURE_COOKIES: bool = false;
 
 #[cfg(test)]
 mod tests {
-    use api::{AccountUpdateRequest, Credentials, UserListResponse};
     use axum::{
         Router,
         body::Body,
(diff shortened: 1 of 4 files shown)


praseodym:
@Lionqueen94 this branch has a merge conflict

backend/src/authentication/mod.rs
```diff
@@ -87,6 +91,16 @@ mod tests {
...
 
+    #[test]
+    fn test_session_life_time() {
+        assert_eq!(SESSION_LIFE_TIME, TimeDelta::seconds(60 * 30));
+    }
+
+    #[test]
+    fn test_session_min_life_time() {
+        assert_eq!(SESSION_MIN_LIFE_TIME, TimeDelta::seconds(60 * 15));
+    }
```
praseodym, line 102: I think we should be testing whether these constants are used correctly, i.e. session life time is initially set correctly and it is renewed if there is less than the minimum left. Just checking whether constants are set is not useful. @jschuurk-kr what do you think?

jschuurk-kr: Agree that testing the value of the constant is not valuable. The mutation testing tool reporting the line with constant, feels like an issue with the tool to me. I suspect it wouldn't have reported it, if we set the content (less readably) as a single value instead of as a multiplication.

Lionqueen94: Yeah it's because of the multiplication that it's a mutant because the specific value was nog checked. I can see if I can check the usage of the constant

Lionqueen94: There is already an assert on the use of SESSION_LIFE_TIME in the cookie, so it's odd that it gave a mutant on that constant:
`assert!(response_cookie.contains(&format!("Max-Age={}", SESSION_LIFE_TIME.num_seconds())));`
The middleware is tested well so I don't see a need to add more tests related to these constants.

@praseodym @jschuurk-kr Shall I remove the tests for the specific constants or leave them in the tests?

jschuurk-kr: @Lionqueen94 I think it's best to remove them

praseodym: Removed in e6be792f

praseodym (approved):

PR #1507, opened 2025-05-19
Add second admin and coordinator to backend fixtures

jschuurk-kr:
Motivation for this change is me wondering what happens if two coordinators try to resolve differences between data entries 1 and 2 at the same time. So I want to have two coordinators and two admins in our backend fixtures.

### Scope
- add second admin to backend fixtures and dev page
- add second coordinator to backend fixtures and dev page
- update `admin`, `coordinator` and `typist` fixtures to have `1` in their name and password
- fix tests: backend tests and e2e tests
- more `let cookie = login_as_admin(app.clone()).await;` in `backend/src/authentication/mod.rs`

### DIFF
--- a/backend/fixtures/users.sql
+++ b/backend/fixtures/users.sql
@@ -1,10 +1,14 @@
 INSERT INTO users (id, username, fullname, role, password_hash, needs_password_change)
 -- Passwords:
 -- admin: 'AdminPassword01'
+-- admin2: 'Admin2Password01'
 -- coordinator: 'CoordinatorPassword01'
+-- coordinator: 'Coordinator2Password01'
 -- typist: 'TypistPassword01'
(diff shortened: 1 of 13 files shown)


backend/fixtures/users.sql
```diff
@@ -1,10 +1,14 @@
 INSERT INTO users (id, username, fullname, role, password_hash, needs_password_change)
 -- Passwords:
 -- admin: 'AdminPassword01'
+-- admin2: 'Admin2Password01'
 -- coordinator: 'CoordinatorPassword01'
+-- coordinator: 'Coordinator2Password01'
```
Lionqueen94, line 6:
```suggestion
-- coordinator2: 'Coordinator2Password01'
```

praseodym (approved):

PR #1519, opened 2025-05-20
Increase request size limit and improve error message

### LINKED ISSUE #1469: Show error when user tries to upload a file that is too big
Right now when a user tries to upload a file that is too big, it gets a generic "Abacus is stuk" error page.

For this issue:

- [x] Find out what the file size limit is

> Axum has a default limit of 2MB, since we are not running any reverse proxy this is the only limit. We post XML files as strings encoded in JSON, there are no noteworthy limits when handling this kind of data.

- [x] Set the file size limit to something that is appropriate for our use cases

> The largest EML file we could find is `Kandidatenlijsten_TK2017_Arnhem.eml.xml` which is around 400 kB. We make an overestimation of any future EML file by assuming this list could grow to 2.5 times the size and that any future added metadata could inflate the file size further by of a factor 12, which is the smallest [abundant number](https://en.wikipedia.org/wiki/Abundant_number). This means we should support file uploads of 12 MB max.

> PR #1519 increases the size limit to 12 MB

- [x] Show an error message with useful instructions to the user

> PR #1519 adds a nicely formatted error message for HTTP 413 (Payload Too Large) errors.

Example trigger for this error:
- Go to `/elections/create` as an administrator
- Provide a large file to the Election Definition upload field

marlonbaeten:
Fixes #1469 

File uploads were limited to 2 MB, this PR increases this size to 12 MB. We think 12 MB covers any reasonably use case by a margin.

This also adds a nice error message for HTTP 413 Payload Too Large responses, by decorating the error response and rendering a error message in the upload form.

To test:
- login as admin
- go to `/elections/create`
- upload any test file larger than 12 MB
- notice the error message

### DIFF
--- a/backend/tests/election_integration_test.rs
+++ b/backend/tests/election_integration_test.rs
@@ -112,13 +112,13 @@ async fn test_election_create_works(pool: SqlitePool) {
 
 #[test(sqlx::test(fixtures(path = "../fixtures", scripts("users"))))]
 #[cfg(feature = "dev-database")]
-async fn test_election_create_payload_limit(pool: SqlitePool) {
+async fn test_election_create_payload_too_large(pool: SqlitePool) {
     use abacus::MAX_BODY_SIZE_MB;
     use reqwest::Body;
 
     let addr = serve_api(pool).await;
 
-    // Create a > MAX_BODY_SIZE_MB payload
+    // Create a payload that is larger than MAX_BODY_SIZE_MB
     let body = Vec::from_iter((0..MAX_BODY_SIZE_MB * 1024 * 1024 + 1).map(|_| b'a'));
 
     let url = format!("http://{addr}/api/elections");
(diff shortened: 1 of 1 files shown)


Lionqueen94 (commented):
I think it would be good to add a test for the 413 error message in the frontend

marlonbaeten:
> I think it would be good to add a test for the 413 error message in the frontend and backend

That's fair 🫡 

Added in [3e5eeb6](https://github.com/kiesraad/abacus/pull/1519/commits/3e5eeb63352721ed320abc13e983a37cac023638)

backend/tests/election_integration_test.rs
```diff
@@ -110,6 +110,58 @@ async fn test_election_create_works(pool: SqlitePool) {
...
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("users"))))]
+#[cfg(feature = "dev-database")]
+async fn test_election_create_payload_too_large(pool: SqlitePool) {
+    use abacus::MAX_BODY_SIZE_MB;
+    use reqwest::Body;
+
+    let addr = serve_api(pool).await;
+
+    // Create a payload that is larger than MAX_BODY_SIZE_MB
+    let body = Vec::from_iter((0..MAX_BODY_SIZE_MB * 1024 * 1024 + 1).map(|_| b'a'));
```
cikzh, line 122: I love that we test this by literally screaming to the backend until it can't take it anymore. For full effect we could use a capital 'A' :)

cikzh (approved):

Lionqueen94 (approved):

PR #1520, opened 2025-05-20
Add FirstEntryHasErrors state to state diagram

oliver3:
Part of https://github.com/kiesraad/abacus/issues/1493 

Adding `FirstEntryHasErrors` state and associated transitions and choices, to match flow https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/afhandeling-fouten-verschillen-coordinator.md

Q:
Are `discard first entry` and `resume first entry` needed as distinct transition names or can they be `discard` and `claim` ?


praseodym (approved):
> Q: Are `discard first entry` and `resume first entry` needed as distinct transition names or can they be `discard` and `claim` ?

It could be `discard` and `resume` but definitely not `claim`, because the coordinator that is resolving the error state is not claiming the data entry (it stays assigned to the typist that was working on it).

jorisleker (approved):

Issue #1421, opened 2025-04-28
Epic: continue data entry on errors


praseodym:
Epic kick-off: Wednesday 14 May 2025 13:30 @oliver3 @lkleuver @jschuurk-kr @jorisleker @marjoleintamis 

# Context

'Foutopsporingsmodus'

# Architecture

# Scope

[Design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=7850-19268&m=dev) (Rollen -> Invoerder -> Concept - foutopsporingsmodus)

- flow chart is already available: https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/afhandeling-fouten-verschillen-coordinator.md
- new state that blocks second entry, shown in 'errors and warnings' election status
- coordinator needs to be able to check entry: split out to #1546

praseodym:
I've moved all issues related to the coordinator functionality to #1546, so that the scope of both epics is better defined.

PR #1529, opened 2025-05-22
Add status FirstEntryHasErrors

oliver3:
- Add new data entry state`FirstEntryHasErrors`
- Incoming state transitions as documented in our [state diagram](https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/data-entry-state.md)
  - When we finalise a `FirstEntryInProgress` and it still contains errors, go to `FirstEntryHasErrors` instead of `SecondEntryNotStarted`, 
  - When keep_second_entry is called with a second data entry that has errors, go to `FirstEntryHasErrors`
- Update tests accordingly
- Update Badge component that handles all possible states

This PR is not testable through the UI currently but is an integral part of [the epic "foutopsporing"](https://github.com/kiesraad/abacus/issues/1421) and will be tested through the other issues
The Badge can be seen in Ladle

### DIFF
--- a/backend/src/data_entry/status.rs
+++ b/backend/src/data_entry/status.rs
@@ -40,6 +40,7 @@ impl From<ValidationResults> for DataEntryTransitionError {
 pub enum DataEntryStatus {
     FirstEntryNotStarted, // First entry has not started yet
     FirstEntryInProgress(FirstEntryInProgress),
+    FirstEntryHasErrors(FirstEntryHasErrors),
     SecondEntryNotStarted(SecondEntryNotStarted),
     SecondEntryInProgress(SecondEntryInProgress),
     EntriesDifferent(EntriesDifferent),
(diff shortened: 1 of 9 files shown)


backend/src/data_entry/status.rs
praseodym, line 1: You'll also need to add the transition to `FirstEntryHasErrors` in the `keep_second_entry` function

praseodym (changes requested):

backend/src/data_entry/status.rs
oliver3: Implemented in https://github.com/kiesraad/abacus/pull/1529/commits/e7d87ed224adbfe81f0cb58f4cf0614b4b6702e8 thanks!

praseodym (approved):

Lionqueen94 (approved):

Issue #890, opened 2025-01-22
Save election definition from frontend


praseodym:
Expand frontend page to call API endpoint to save the previewed election definition.

PR #1541, opened 2025-05-26
Validate and save endpoint for EML import

cikzh:
Fixes #889 
Fixes #1480

- Verify endpoint now also verifies user provided hash
- Added endpoint that saves election
- CheckElectionDefinition now verifies the user provided hash by doing the backend call
- Tests

To test, you can test by uploading `src/eml/tests/eml110a_test.eml.xml`. The hash for that file is:
```
84c9 caba ff33 6c42 9825 b20c 2ba9 1ceb 3c61 9b99 8af1 a57e cf00 8930 9bce 0c33
```
This first stub should be `9825` and the second `8af1`.

You can also test with a different (valid) EML file and discovering the hash by executing `sha256sum path/to/file` in the command line.

PR #1549, opened 2025-05-26
Refactor election creation state management and add save call

cikzh:
Implements #890 
- Refactored state management. We're now keeping the state in the Provider via reducers. Does not yet implement the save call to the backend.
- Adds save call to backend to save election. Navigates to election listing after.

### DIFF
--- a/frontend/src/features/election_create/components/CheckHash.tsx
+++ b/frontend/src/features/election_create/components/CheckHash.tsx
@@ -10,7 +10,7 @@ import { formatDateFull } from "@/utils/format";
 
 import { RedactedHash, Stub } from "./RedactedHash";
 
-interface CheckElectionDefinitionProps {
+interface CheckHashProps {
   date: string;
   title: string;
   fileName: string;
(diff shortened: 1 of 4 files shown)


praseodym:
Also in https://github.com/kiesraad/abacus/pull/1541#discussion_r2108536128 you mentioned that you would update the docstring in this PR, but I don't see that change here.

frontend/src/features/election_create/components/CheckHash.tsx
```diff
@@ -1,24 +1,67 @@
...
+        if (typeof value !== "string" || value.length !== 4) {
+          newStub.error = t("election.check_eml.check_hash.hint");
+        } else {
+          completeHash[stub.index] = value;
+        }
+        setStubs(newStubs);
+      }
+    });
+
+    // Only submit when there a no errors
```
bramtweedegolf, line 60: typo: when there _are_ not errors

cikzh:
> lgtm, the only thing I noticed is that the elections overview page has to be explicitly refreshed before it shows the newly created election.

I've created https://github.com/kiesraad/abacus/issues/1573 to fix this. I'll update the docstring and the comment typo there.

PR #1561, opened 2025-05-27
Air Gap detection backend functionality (DNS lookup only)

### LINKED ISSUE #1538: Create airgap task
- Mechanism to run a airgap detection task every X seconds
- Implement a simple DNS check using `dns_lookup`

### LINKED ISSUE #1539: Create axum middleware to block request when airgap violation is detected

marlonbaeten:
Fixes #1538 and fixes #1539 

Implements an air gap detection mechanism that, when enabled by the feature flag or CLI flag, periodically checks whether the device connected to the internet.

Currently only a DNS lookup is implemented. Other checks will be added in #1559

You can test this by running with the "airgap-detection" or calling the CLI with the `--airgap-detection` flag:

```shell
cargo run -- --airgap-detection
```

When you are connected to the internet this should give a 503 error on any API call. Note that handling this error in the frontend will be implemented in #1540

The middleware will also block any requests in the unlikely scenario that the airgap detection thread has crashed.


bramtweedegolf (approved):
We could add more checks in the future; but this seems to me to be a simple and effective solution. Nice!

PR #1576, opened 2025-06-02
Added alerts to election status page to show after resolving differences

Lionqueen94:
## Screenshots
image

image

### DIFF
--- a/frontend/src/features/election_status/components/ElectionStatusPage.tsx
+++ b/frontend/src/features/election_status/components/ElectionStatusPage.tsx
@@ -58,11 +58,11 @@ export function ElectionStatusPage() {
       </header>
       {differencesResolvedAlert && (
         <Alert type="success" onClose={closeSuccessAlert}>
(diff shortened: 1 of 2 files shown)


jorisleker:
@Lionqueen94 I've added designs for these cases [here](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=141-7305&t=h5Yy7k7bYWPpV1LZ-11).

To avoid confusion and having to explain things like 'second entry now is first entry', I suggest not using 'first' and 'second data entry' after the differences have been resolved. See Figma for details.

jorisleker:
@Lionqueen94 Updated the copy of the header in [the alert on this page](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=10401-68536&t=oYQGxKuYsAqEYQSV-4) as well (same change, don't want to use 'eerste' or 'tweede')

jorisleker (approved):

PR #1577, opened 2025-06-02
Add e2e test for election creation

### LINKED ISSUE #1415: Election definition tests
At least an e2e test that'll walk through the entire flow. Also cover error cases.

cikzh:
Fixes #1415

Adds e2e tests for election definition uploading, hash checking, and saving a new election


marlonbaeten (approved):
"valid, but incorrect file" is a bit of a confusing description - but I get it :+1:

PR #1578, opened 2025-06-03
Additional air-gap checks

marlonbaeten:
In OSV2020 airgaps are detected using PING and by performing HTTP requests.

Performing a ping is not implemented in the Rust standard library. It also would involve root permissions if performed on a raw socket in a Linux system. Since including a external dependency for this is too much of a burden for this feature, we do not ping external systems.

Instead 3 types of simple checks are performed:

- A DNS resolve using the default resolver
- Establishing a TCP connection over IPv4
- Establishing a TCP connection over IPv6

Since the DNS query (usually) is performed over UDP these methods cover the two fundamental internet protocols.

Note that performing a HTTP request (like in OSV2020) has more points of failure than just establishing a TCP connection, while a HTTP request (apart form HTTP3) always starts of with a TCP connection. So only checking the final result of a HTTP request has a higher chance of false negative in a air-gap detection context.

To test:

Run abacus with the air-gap detection:

```shell
RUST_LOG=abacus=trace cargo run -- --airgap-detection
```

This should output the following:
```shell
ERROR abacus::airgap::detect: Airgap violation detected, abacus is connected to the internet!
TRACE abacus::airgap::detect: TCP IPv4 connections: 2
```

Bonus points for blocking TCP traffic in your firewall and seeing the  following output:
```shell
ERROR abacus::airgap::detect: Airgap violation detected, abacus is connected to the internet!
TRACE abacus::airgap::detect: DNS lookup success: 3
```

Code coverage is not perfect since, in a test , the detection will fail early.

### DIFF
--- a/backend/src/airgap/detect.rs
+++ b/backend/src/airgap/detect.rs
@@ -1,9 +1,9 @@
 use std::{
-    net::ToSocketAddrs,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, TcpStream, ToSocketAddrs},
     sync::{Arc, RwLock, atomic::AtomicBool},
     time::{Duration, Instant},
 };
-use tracing::error;
+use tracing::{error, trace};
 
 #[derive(Debug, Clone)]
 pub struct AirgapDetection {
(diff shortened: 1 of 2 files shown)


backend/src/airgap/detect.rs
```diff
@@ -52,20 +70,55 @@ impl AirgapDetection {
...
+            self.set_last_check();
+            return;
+        }
 
-                self.set_airgap_violation_detected(true);
-            }
-            Err(_) => {
-                self.set_airgap_violation_detected(false);
-            }
+        // persorm a DNS lookup using the default resolver
```
bramtweedegolf: Small typo: perform

bramtweedegolf (approved):

backend/src/airgap/detect.rs
marlonbaeten: Fixed in [26bbd5d](https://github.com/kiesraad/abacus/pull/1578/commits/26bbd5d6a1ba02d7d8e6970b62999ae14bc3f3e4)

bramtweedegolf (approved):

bramtweedegolf (approved):

backend/src/airgap/detect.rs
```diff
@@ -50,23 +68,62 @@ impl AirgapDetection {
...
+        }
+
+        self.set_last_check();
+    }
 
-        match dns_lookup_result {
-            Ok(_) => {
-                // if we get a result from the DNS lookup, the system is not in an airgap
-                error!("Airgap violation detected, abacus is connected to the internet!");
+    /// Detects if the system is in an airgap by attempting to connect to IPv3, IPv6 addresses and performing DNS lookups
```
praseodym, line 83:
```suggestion
    /// Detects if the system is in an airgap by attempting to connect to IPv4, IPv6 addresses and performing DNS lookups
```

PR #1583, opened 2025-06-05
Adjusted errors and warnings to have a version for typist and one for coordinator

Lionqueen94:
Different copy for coordinator will be added in another issue


praseodym (dismissed):
lgtm. Note that the data entry components are only used by the typist, so having a user role check is not really needed.

Lionqueen94:
> lgtm. Note that the data entry components are only used by the typist, so having a user role check is not really needed.

True, but the coordinator will be able to view the data entry when this epic is finished, and we still have to decide how to deal with this (new components or a variant). It can always be changed when we work on this.

frontend/src/features/data_entry/components/differences/DifferencesForm.tsx
```diff
@@ -46,10 +48,20 @@ export function DifferencesForm() {
       />
       {error instanceof ApiError && <ErrorModal error={error} />}
       {formSection.isSaved && !formSection.errors.isEmpty() && (
-        <Feedback id="feedback-error" type="error" data={formSection.errors.getCodes()} />
+        <Feedback
+          id="feedback-error"
+          type="error"
+          data={formSection.errors.getCodes()}
+          userRole={userRole.isCoordinator ? "coordinator" : "typist"}
```
cikzh: You could pass in the user's role here directly if you retrieve it via `useUser` instead of `useUserRole`, or we could add the user's role in the `return` in the useUserRole hook.
It would save use a lot of extra checks and we wouldn't have to make the assumption that "not-a-coordinator" is always a typist.

cikzh (commented):
Looks good overall, but we could save ourselves some unnecessary checks if we want

frontend/src/features/data_entry/components/differences/DifferencesForm.tsx
Lionqueen94: True, but the issue is that administrator will never have access to this, but it will then have to be a role option in the Feedback component. How should I deal with this? Just show the same information to both "coordinator" and "administrator"? Or an empty component for "administrator"?

cikzh: > Just show the same information to both "coordinator" and "administrator"?

I would say that is the way to go, as that is what would happen now anyway. You could then also use `userRole.isTypist()` instead of `userRole === "typist"` in the Feedback component.

frontend/src/components/ui/Feedback/Feedback.tsx
```diff
@@ -13,19 +13,20 @@ interface FeedbackProps {
...
   data?: ClientValidationResultCode[];
+  isTypist: boolean;
 }
 
-export function Feedback({ id, type, data }: FeedbackProps) {
+export function Feedback({ id, type, data, isTypist }: FeedbackProps) {
   const feedbackHeader = useRef<HTMLHeadingElement | null>(null);
   const link = (children: ReactElement) => <Link to={`../voters-and-votes`}>{children}</Link>;
-
+  const role = isTypist ? "typist" : "coordinator";
```
cikzh:
```suggestion
  // NOTE: non-typist roles are always mapped to a coordinator here
  const role = isTypist ? "typist" : "coordinator";
```

Lionqueen94: I merged main and changed the code and added the note

praseodym (approved):

Issue #1514, opened 2025-05-20
Coordinator: errors and warnings detail pages


oliver3:
- "Read only" view of data entry
- Coordinator texts
- Navigation icons are different from data entry

NB: `Badges B1-3.1 en 3.2` etc are for new models and not for this epic

PR #1588, opened 2025-06-06
Make data entry structure declarative

### LINKED ISSUE #1547: Proof of concept: declarative data entry form
Proof of concept for declarative configuration of the data entry form, where the form fields are defined by a TypeScript type instead of hardcoded pages in TSX. This should make it easier to implement differences between CSO and DSO data entry and make it easier to define variants such as the read only view (#1514) and differences table view.

praseodym:
This PR refactors the data entry feature to define the form structure in a declarative way. The data entry structure is defined in `structure.ts` and rendered using `DataEntrySection.tsx` and `DataEntrySubsections.tsx`. This makes it easier to change the structure in the future (e.g. new models), makes it easier to introduce variants of the form (e.g. variant for DSO) and allows us to make a read-only variant (#1514).

The UI and overall user experience should be unchanged, this is something to pay attention to when testing.

Some components have not yet been refactored to use the new declarative setup, such as the `DataEntryProgress` and the `resolve_differences` feature. These will still have to be changed in subsequent PRs.

Closes #1547


frontend/src/features/data_entry/components/candidates_votes/CandidatesVotesPage.test.tsx
```diff
@@ -30,16 +29,22 @@ describe("Test CandidatesVotesPage", () => {
...
+    vi.spyOn(console, "error").mockImplementation(() => {});
     renderPage();
 
     await waitFor(() => {
-      expect(screen.getByText("Geen lijst gevonden voor 123")).toBeVisible();
+      expect(
+        screen.getByText(
+          "Error thrown during render: Form section political_group_votes_123 not found in data entry structure",
+        ),
+      ).toBeVisible();
```
Lionqueen94, line 42: Do we want to make this more user friendly and in Dutch or will this never realistically be shown anyway?

Lionqueen94 (commented):
Code looks good, haven't tested functionally yet, awesome improvement!

cikzh (changes requested):
Encountered a bug when doing a manual data entry side-by-side with a `main` build via abacus-test.nl and a build from this branch locally to check for visual differences. 
When trying to submit "Lijst 2", the backend returns an error:

```
 Failed to deserialize the JSON body into the target type: data.political_group_votes[0].total: invalid type: string "", expected u32 at line 1 column 533
```
`main` submits the list just fine, while entering the same data.
I didn't dive into the bug yet to check if it's related to this PR though.

praseodym:
> Encountered a bug when doing a manual data entry side-by-side with a `main` build via abacus-test.nl and a build from this branch locally to check for visual differences. When trying to submit "Lijst 2", the backend returns an error:
> 
> ```
>  Failed to deserialize the JSON body into the target type: data.political_group_votes[0].total: invalid type: string "", expected u32 at line 1 column 533
> ```
> 
> `main` submits the list just fine, while entering the same data. I didn't dive into the bug yet to check if it's related to this PR though.

That's indeed a bug in this branch, I think because I cleaned up the mapping code this morning and didn't manually test everything (and the automated tests didn't change this). I'll investigate and fix.

frontend/src/features/data_entry/components/candidates_votes/CandidatesVotesPage.test.tsx
praseodym: It should only be shown when manually messing around with the URL, so I didn't think it was worth translating.

praseodym:
> > Encountered a bug when doing a manual data entry side-by-side with a `main` build via abacus-test.nl and a build from this branch locally to check for visual differences. When trying to submit "Lijst 2", the backend returns an error:
> > ```
> >  Failed to deserialize the JSON body into the target type: data.political_group_votes[0].total: invalid type: string "", expected u32 at line 1 column 533
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `main` submits the list just fine, while entering the same data. I didn't dive into the bug yet to check if it's related to this PR though.
> 
> That's indeed a bug in this branch, I think because I cleaned up the mapping code this morning and didn't manually test everything (and the automated tests didn't change this). I'll investigate and fix.

Should be fixed with bccae296fa0a2a19cd7831cc1a80949dd28fabc0. The issue was that sometimes values from other form sections persisted in `currentValues`, which messed up the mapped result object.

Lionqueen94 (approved):
I did some functional testing and didn't encounter any issues

cikzh (approved):
Me neither, bug is gone and visually it looks identical to main as far as I could see

Issue #1377, opened 2025-04-15
Frontend page to upload candidate list


rnijveld:
Frontend page to upload candidate list file (and previously selected from previous step election definition), upload to backend to parse its content and show that to user (preview of the file that was uploaded), and enter hash. Do not save yet.

Issue #1376, opened 2025-04-15
API endpoint to parse candidate list for import


rnijveld:
API endpoint that takes all the input from the election definition, and additionally gets the input from the candidate list, validates that the election definition and candidate list are matching, and reports back the contents and hash for the candidate list. Does not save to the database yet.

PR #1596, opened 2025-06-10
1377 frontend page to upload candidate list

bramtweedegolf:
Implements [#1376](https://github.com/kiesraad/abacus/issues/1376) and [#1377](https://github.com/kiesraad/abacus/issues/1377)

After uploading an election definition, the user can upload a candidate list, must then enter a valid hash and if successful is sent to the check and save page where the election is created.

Note: Validation for the candidate list in the backend will be added in  [#1589](url)

### DIFF
--- a/backend/src/election/api.rs
+++ b/backend/src/election/api.rs
@@ -172,7 +172,6 @@ pub async fn election_import_validate(
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
-#[serde(rename_all = "camelCase")]
 pub struct ElectionDefinitionImportRequest {
     election_hash: [String; crate::eml::hash::CHUNK_COUNT],
     election_data: String,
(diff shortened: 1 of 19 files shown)


backend/src/election/api.rs
```diff
@@ -129,8 +128,13 @@ pub async fn election_create(
 pub struct ElectionDefinitionValidateRequest {
```
praseodym: This struct should be named differently to reflect that it also takes a candidate list, not just the election definition EML.

Same for the comment on L142-143 which only mentions the election definition.

praseodym (approved):

Lionqueen94 (commented):
Nice work! I think it would be good to increase the code coverage a bit more with some extra tests.

backend/src/election/api.rs
bramtweedegolf: Done!

bramtweedegolf:
Good point! Added more frontend tests; more backend tests will be added when we create more XML files in [this issue](https://github.com/kiesraad/abacus/issues/1589).

praseodym (approved):

Lionqueen94 (approved):

PR #1609, opened 2025-06-12
`DataEntryProgress` with data entry structure

praseodym:
Refactor the `DataEntryProgress` component to use the declarative data entry structure. UI/UX should be the same compared to `main`.


bramtweedegolf (approved):

praseodym:
Test failed due to #1623

PR #1610, opened 2025-06-12
Data entry section routing using section IDs

praseodym:
Refactor data entry section routes to use form section IDs.


frontend/src/features/data_entry/components/DataEntryProgress.test.tsx
```diff
@@ -203,21 +203,21 @@ describe("Test DataEntryProgress", () => {
     expect(votersAndVotesLink).toBeVisible();
     expect(votersAndVotesLink).toHaveAttribute(
       "href",
-      `/elections/${electionId}/data-entry/${pollingStationId}/${entryNumber}/voters-and-votes`,
+      `/elections/${electionId}/data-entry/${pollingStationId}/${entryNumber}/voters_votes_counts`,
```
Lionqueen94, line 206: We should only use dashes in user facing URLs right?

Lionqueen94 (approved):

frontend/src/features/data_entry/components/DataEntryProgress.test.tsx
praseodym: Preferably, but that would mean changing the form section ids everywhere. Maybe we can do that when we implement the new models.

frontend/src/features/data_entry/components/DataEntrySection.tsx
```diff
@@ -89,10 +93,16 @@ export function DataEntrySection({ sectionId }: { sectionId: FormSectionId }) {
       <DataEntryNavigation onSubmit={onSubmit} currentValues={currentValues} />
       {error instanceof ApiError && <ErrorModal error={error} />}
       {formSection.isSaved && !formSection.errors.isEmpty() && (
-        <Feedback id="feedback-error" type="error" data={formSection.errors.getCodes()} userRole={user.role} />
+        <Feedback id="feedback-error" type="error" data={memoizedErrorCodes} userRole={user.role} shouldFocus={true} />
```
Lionqueen94, line 96: The default of shouldFocus is true so it's not needed to add it here (but for clarity it's fine to leave it)

praseodym: I prefer it this way because it's more explicit

Lionqueen94 (approved):

bramtweedegolf (approved):

PR #1628, opened 2025-06-17
Generate test election helper tool

rnijveld:
A helper binary (run with `cargo run --bin gen-test-election`) that generates an election, the political groups and candidates in those groups, and a set of polling stations for the election. The size of the election can be specified with a couple of arguments to the binary. Intended for load testing and for generating testing data.

I've attempted to create some weird cases in the test data as well, including polling stations with quotes in their name, first and last names with special characters, including and excluding last name prefixes, sometimes omitting the first name entirely for some political groups. The same applies to names of localities etc. 

Names are all made up. Political groups and polling stations have names somewhat related to the election process. Locality names should all not exist. Street names are based on the most popular street names in the Netherlands. First and last names are based on some hand-picked entries from several Dutch name generators. Most of the rest is just random numbers.


praseodym (approved):

bramtweedegolf (approved):
Very nice!

PR #1634, opened 2025-06-17
Removed differences badge

Lionqueen94:
As discussed with @jorisleker: the differences badge has been removed


backend/.sqlx/query-4660bfcdf0bdee50df05c66848f10b73d40829367369694cc0a563f476ec3a38.json
praseodym, line 1: I think this change was not intended?

praseodym: Hm #1628 has the same change, never mind I guess

Lionqueen94: This is still from the sql query fix I think. It shows up on any branch when `cargo sqlx prepare -- --all-targets --all-features` is run

jorisleker (changes requested):
Status overview looks good now,

On the resolve differences page, we should not show a badge in the header (we are not looking at the first or second data entry, but at both of them)

Scherm­afbeelding 2025-06-17 om 13 48 49

Lionqueen94:
> Status overview looks good now,
> 
> On the resolve differences page, we should not show a badge in the header (we are not looking at the first or second data entry, but at both of them)
> 
> Scherm­afbeelding 2025-06-17 om 13 48 49

@praseodym would this be smarter to edit in your refactoring branch to prevent potential conflicts? Or I'll just wait for your PR to be merged first, that's probably best

praseodym:
I think it shouldn't conflict

PR #1639, opened 2025-06-17
Add zebra striping to political group data entry

praseodym:
Add back zebra striping to the political group data entry forms, which was accidentally removed in #1547.


praseodym:
Rebased and ready for review now that #1613 is merged.

oliver3 (approved):

PR #1651, opened 2025-06-20
Small documentation correction: missing a seat in test text of full seats for this test case

Hilbrand:


ring-ring-ring:
Goed gespot, dank je wel!!

Issue #1654, opened 2025-06-20
Coordinator errors and warnings overview


praseodym:
"Alle fouten en waarschuwingen" like controleren en opslaan (typist), updated via #1496 , with different coordinator texts

https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=10369-31482&m=dev

Issue #1544, opened 2025-05-26
Add copy for all errors and warnings for coordinator to `validatieregels-plausibiliteitschecks-tellingen.md`


Lionqueen94:
Change the copy in `feedback.json` as well

praseodym:
Because we are updating the coordinator copy at the same time as changing the validation rules for the new models (#1611), this will be implement in epics #1545 and #798.

Issue #1512, opened 2025-05-20
Coordinator page navigation menu


oliver3:
Navigation menu
- Clickable "nav" links to errors and warnings detail / section pages
- Document when to show what icon in the nav in documentation:
  - switch(section):
    - section is current page --> current icon, highlight, no link, break
    - section has errors --> error icon, break
    - section has warnings --> warning icon, break
    - default: no icon

[As designed in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=10369-31460&t=n8uBMUduu8x5IdNI-11)

PR #1656, opened 2025-06-20
Add read-only data entry view to resolve errors page

### LINKED ISSUE #1514: Coordinator: errors and warnings detail pages

praseodym:
Add read-only data entry view to resolve errors page: closes #1514. For now all data entry sections are shown on the main page, navigation will be added later (see below).

There are Ladle stories for the [read-only input grid](https://data-entry-read-only.kiesraad-abacus.pages.dev/ladle/?arg-readOnly=true&story=input-grid--default-grid) and the [read-only data entry view](https://data-entry-read-only.kiesraad-abacus.pages.dev/ladle/?arg-sectionId=voters_votes_counts&story=features--resolve-errors--read-only-section).

21aa7c11ad0a3ad928ce1ef33da85ccf5c1cc019 also fixes an issue for the regular data entry, where warning icons are not shown for any fields when an error is present on that page (this used to be the desired behaviour before #1421).

Note that:
- the errors and warnings overview is to be done in #1654
- the left-side navigation menu to individual data entry sections is to be done in #1512
- the coordinator validation copy is to be done in #1544

Please review per commit, because the first two commits contain refactoring to move files from the data entry feature to the app-wide utils and types directories.


oliver3 (dismissed):
Code looks good to me, tested it and the errors and warnings seem to be all there.

Only wondering if the design of the "recounted" section is good enough, what do you think @jorisleker ?

image

praseodym:
> Code looks good to me, tested it and the errors and warnings seem to be all there.
> 
> Only wondering if the design of the "recounted" section is good enough, what do you think @jorisleker ?

Do you mean the "Controleer of je antwoord gelijk is aan het papieren proces-verbaal" or something else?

Also note that these sections will be change when we implement the new models, so if it is minor textual changes I don't think it's worth fixing at this point.

oliver3:
> Do you mean [...]

Mainly the "Controleer of je antwoord gelijk is aan het papieren proces-verbaal" indeed, but also more general, as there is no visual design for a radio button that should appear read-only. 

We can also revisit this once the final coordinator texts and the new models are there indeed.

praseodym:
> Mainly the "Controleer of je antwoord gelijk is aan het papieren proces-verbaal" indeed, but also more general, as there is no visual design for a radio button that should appear read-only.

The styling is actually different, you can see that in the ['Default radio' Ladle story](https://data-entry-read-only.kiesraad-abacus.pages.dev/ladle/?story=checkbox-and-radio--default-radio). 

By the way, I also tried adding a variant with the [`readonly` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly) but then the radio button can receive focus, which I found a bit out of place when the input grid rows render a simple `<span>` that cannot receive focus.

jorisleker:
> Only wondering if the design of the "recounted" section is good enough, what do you think @jorisleker ?
LGTM! We might revisit the disabled radio button styling later, but for now, this looks good enough for me. 

Regarding using `disabled` vs `readonly` on the radios @praseodym: 
For accessiblity reasons, I think `readonly` is [the better choice](https://stackoverflow.com/questions/77132461/when-i-should-use-readonly-instead-of-disabled) here: Using `disabled` will hide the field and it's contents completely for screenreaders, which is not what we want; you should be able to evaluate how the form was filled in. The contents of other inputs are visible to screenreaders, because we are treating those as regular content (by not using an input with disabled state, but a span).

praseodym:
Apparently the radio input [do not support `readonly` at all](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly). And from the [MDN `:read-only` CSS pseudo-class docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:read-only):
> Elements on which the HTML attribute [readonly](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly) doesn't have an effect (such as [`<input type="radio">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/radio), [`<input type="checkbox">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/checkbox), and all other non-form elements) are also selected by the :read-only pseudo-class.

So I think we have to stick with this.

oliver3 (approved):

PR #1662, opened 2025-06-23
Allow exporting of a generated test election to EML files

### LINKED ISSUE #1661: Export of generated data to EML from the gen-test-election tool

rnijveld:
Fixes #1661 

- Added some logic that allows converting an abacus election back to the EML files that it consists of
- Added some CLI parameters to configure the behavior of the export (only if a path is specified an export should be made)
- Added a way to configure the number of seats in the election
- Updated some defaults to generate a little smaller elections by default
- Can be tested by running `cargo run --bin gen-test-election -- --export-definition path/to/somewhere` (also check the other options by running `cargo run --bin gen-test-election -- --help`)


marlonbaeten (approved):
Works!

bramtweedegolf (approved):

PR #1666, opened 2025-06-23
Moved existing e2e test into full data entry flow test section and added check for alert after submit

Lionqueen94:
Moved existing e2e test into full data entry flow test section and added check for alert after submit

Also fixed which header is shown after submitting a data entry with errors or differences

### DIFF
--- a/frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
+++ b/frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
@@ -389,6 +389,78 @@ test.describe("full data entry flow", () => {
     await expect(dataEntryHomePage.fieldsetNextPollingStation).toBeVisible();
     await expect(dataEntryHomePage.dataEntrySaved).toBeVisible();
   });
+
+  test("submit with accepted errors on voters and votes, differences and candidate votes pages", async ({
+    page,
+    pollingStation,
+  }) => {
+    await page.goto(`/elections/${pollingStation.election_id}/data-entry/${pollingStation.id}/1/recounted`);
+
+    const recountedPage = new RecountedPage(page);
+    await recountedPage.checkNoAndClickNext();
+
+    const votersAndVotesPage = new VotersAndVotesPage(page);
+    const voters = {
+      poll_card_count: 100,
+      proxy_certificate_count: 0,
+      voter_card_count: 0,
+      total_admitted_voters_count: 100,
+    };
+    const votes = {
+      votes_candidates_count: 100,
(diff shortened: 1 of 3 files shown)


frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
```diff
@@ -389,6 +389,78 @@ test.describe("full data entry flow", () => {
...
+    await differencesPage.next.click();
+
+    const candidatesListPage_1 = new CandidatesListPage(page, 1, "Lijst 1 - Political Group A");
+
+    await candidatesListPage_1.fillCandidatesAndTotal([737, 153], 890);
+    await candidatesListPage_1.next.click();
+    await candidatesListPage_1.checkAcceptErrorsAndWarnings();
+    await candidatesListPage_1.next.click();
+
+    await expect(candidatesListPage_1.error).toBeVisible();
```
jschuurk-kr: I'm confused how `candidatesListPage_1.error` and `checkAndSavePage.fieldset` should be visible without there being an action in between. Should the check on `candidatesListPage_1.error` happen right before `candidatesListPage_1.checkAcceptErrorsAndWarnings()`?

frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
```diff
@@ -389,6 +389,78 @@ test.describe("full data entry flow", () => {
...
+    const checkAndSavePage = new CheckAndSavePage(page);
+    await expect(checkAndSavePage.fieldset).toBeVisible();
+
+    const listItemsVotersAndVotes = page
+      .getByTestId("save-form-summary-list-voters_votes_counts")
+      .getByRole("listitem");
```
jschuurk-kr: Locators should be in the page object, not in the test. Same applies for line 447.

frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
Lionqueen94: Very correct, changed in e2cbc2127b57ca9c04076c9d2f70abeed7ad0d57

frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
Lionqueen94: I added the locators to the page object and added a check for the last list item as well in e2cbc2127b57ca9c04076c9d2f70abeed7ad0d57

PR #1669, opened 2025-06-24
Add generation of election results for test elections

rnijveld:
- This add some additional parameters to the gen-test-election binary that allow creating some data entry values (and corresponding values in the results table) for a test election while it is being generated.
- The data is generated according to a power law distribution (i.e. first one gets the most and having a long tail of lower and lower amounts), the slope of the distribution can be configured, but not the order of the groups. For candidates the first candidate always gets the highest amount.
- When writing files to disk, a json file for model na-31-2 is also generated to allow easy testing
- Best way to test is to generate an election with `cargo run --bin gen-test-election -- --with-data-entry` and to log in as coordinator to view the apportionment pages.
- Added a couple of names that use unicode characters that do not fit in the first byte
- You can specify the percentage of polling stations that should have completed their first and second data entry, see the `--help` for details


marlonbaeten (approved):
Would be cool to have some differences and "volmachten" etc.

PR #1671, opened 2025-06-24
1660 refactor election create e2e test

jschuurk-kr:
### In scope
- put all strings related to an eml file in a single re-usable object and put those in a file in the test-data folder
- refactor `OverviewPgObj.ts`: `this.importedElections = page.getByText("Gemeenteraad Amsterdam 2022");`
- refactor all locators in the test itself
- remove two single-line methods and unused `hash` locators

### Not in scope
**check on the overview page if the specific election has been added, taking into account parallel test execution**
Reason: only solution I can think of is copying the election definition eml and changing the election name to something unique for each test. This would also require recalculating the hash. That seems more trouble than it's worth.

A different and lesser option is to check if the final POST of election creation returns a 201. I don't think we need to do this. I'm fine with relying on the check that we return to the overview page and rely on our error handling in the frontend that returing to the overview page implies the 201 response.

### Comments
I have mixed feelings about changing `await expect(page.getByText(eml110a.filename)).toBeVisible();`
to `await expect(overviewPage.main).toContainText(eml110a.filename);`. I'm in favor of the change because it refactors the locator out of the test and as a result I could make the `expect` express the intention of the test a tiny bit better. But I wish I had been able to find a better locator than `overviewPage.main`.

### DIFF
--- a/frontend/e2e-tests/page-objects/election/OverviewPgObj.ts
+++ b/frontend/e2e-tests/page-objects/election/OverviewPgObj.ts
@@ -7,23 +7,15 @@ export class OverviewPgObj {
   readonly header: Locator;
   readonly alert: Locator;
   readonly create: Locator;
-  readonly importedElections: Locator;
-  readonly readyStates: Locator;
+  readonly elections: Locator;
+  readonly electionsInReadyState: Locator;
 
   constructor(protected readonly page: Page) {
     this.navBar = new NavBar(page);
     this.header = page.getByRole("heading", { name: "Beheer verkiezingen" });
     this.alert = page.getByRole("heading", { name: "Je account is ingesteld" });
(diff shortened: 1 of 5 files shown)


oliver3 (approved):
Good improvements! 

About your comment, we could add an `id` on the parts that we want to check in the `CheckHash` component and use those. Or maybe `aria-label` attributes that explain what the two lines of the "card" are. Or even change the design to make it explicit what these lines are and using that (similar to the label "Digitale vingerafdruk (hashcode):")

image

bramtweedegolf (approved):
Excellent! I'll check if any of my related PRs can benefit from similar improvements. (After we merge this PR.)

I agree that "expect(overviewPage.main).toContainText(eml110a.filename)" is a little better / clearer.

bramtweedegolf (approved):

Lionqueen94 (approved):

PR #1676, opened 2025-06-24
PDF template for model Na 31-2

### LINKED ISSUE #1672: Implement new model template for Na 31-2

marlonbaeten:
New Typst template for model Na 31-2

Fixes #1672

Model content:

[Na 31-2 incl. Wpv voorstelPv gsb-cso.pdf](https://github.com/user-attachments/files/20884353/Na.31-2.incl.Wpv.voorstelPv.gsb-cso.pdf)

Model design:

[cso-gsb-pv.pdf](https://github.com/user-attachments/files/20884356/cso-gsb-pv.pdf)

New output:

[model-na-31-2.pdf](https://github.com/user-attachments/files/20884342/model-na-31-2.pdf)

Implementation notes:
- The page "Controles in opdracht van het centraal stembureau" moved to a separate template
- I tried keeping the model itself as simple as possible, styling and scripting is moved to the "common" folder
- The `votes_table` has some complexity related to rows per table/column and balancing columns

You can try out the template in the Typst web environment: https://typst.app/project/pf0ekPUcoIdxIaMpKR83po (@praseodym can give you the correct permissions)


jorisleker:
@marlonbaeten I've reviewed the output PDF. Made some notes in the attached pdf

Main issues I found:
- max-width for paragraph texts is missing
- date formatting on cover page should be 'woensdag 26 maart 2025' instead of '26-03-2025'
- numbers should be formatted with thousands seperator
- 'De lijst gaat door op de volgende pagina -->' missing for lists with >50 candidates
- How candidates are divided over columns; align with Abacus (zebra striping + spacer rows)

[reviewed - model-na-31-2.pdf](https://github.com/user-attachments/files/20904071/reviewed.-.model-na-31-2.pdf)

jorisleker:
And for the sake of readabilty, was wondering wether we should use a `-` instead of a 0 for candidates with zero votes. Think this will mainly help the second data entry in HSB/CSB cases, where you use a printed document as your input.

jorisleker (changes requested):
ssee previous comment

marlonbaeten:
@jorisleker 

I have implemented your comments. Some notes:

- Max-width of 75% is applied to all paragraphs and headers of level 3
- Numbers have a thousand separator and a 0 is rendered as -
- For the votes table I have implemented a algorithm to balance the columns and limit them to max 25 items. I have fixed an of-by-one error: if the number of remaining rows is even each column will have the same number of rows, if it is odd the first column get one row more.
- Note that Typst could also break the columns based on the remaining space / page height, but that would cause "unbalanced" columns.

See [na-31-2.final_version.really-final.last-version.pdf](https://github.com/user-attachments/files/20976856/na-31-2.final_version.really-final.last-version.pdf)

backend/templates/common/scripts.typ
```diff
@@ -242,14 +270,14 @@
...
-              str(c.votes)
+              fmt-number(c.votes)
             })),
           )
 
           rc += 1
 
           if togo > 0 {
-            if calc.rem(togo, half_remainder) == 0 {
+            if calc.rem(togo  +1, half_remainder) == 0 {
```
Lionqueen94:
```suggestion
            if calc.rem(togo + 1, half_remainder) == 0 {
```

Lionqueen94:
We changed the number rendering in Abacus to render each number with the same width, do we want to / are we able to do this in Typst as well? This would make the polling station number list look better I think. @jorisleker

marlonbaeten:
> We changed the number rendering in Abacus to render each number with the same width, do we want to / are we able to do this in Typst as well? This would make the polling station number list look better I think. @jorisleker

Yes I'm aware and I already added this. See:

https://github.com/kiesraad/abacus/blob/models-june-2025-typst/backend/templates/common/scripts.typ#L80
https://github.com/kiesraad/abacus/blob/models-june-2025-typst/backend/templates/common/scripts.typ#L272

I will also add this for the polling station numbers :+1:

praseodym:
Looking at this version I think it'd be more consistent to always break the candidate votes columns at 25/25/15/15 instead of letting Typst balance the columns. Because the layout for every political group will be the same, I think this will make the model easier to use when filling in by hand (in case of hand counting) and easier to use when doing data entry at the CSB level.

jorisleker:
> Looking at this version I think it'd be more consistent to always break the candidate votes columns at 25/25/15/15 instead of letting Typst balance the columns. Because the layout for every political group will be the same, I think this will make the model easier to use when filling in by hand (in case of hand counting) and easier to use when doing data entry at the CSB level.

Agree with this. This will lead to fewer columns and fewer instances where we need subtotals fields. We should not show the subtotals field for <=25 candidates, because then there is only one column, and total == subtotal. 

Note: we should update the rendering of lists in Abacus to use the same algorithm for rendering the spacer row between candidates to do that after the 25th, 50th and 65th candidate.

marlonbaeten:
@jorisleker @praseodym I've implemented the 25/25/15/15  columns as discussed. The rows are now always 23pt high, which works for a names spanning two lines. A name spanning 3 lines will break the layout.

praseodym (dismissed):

marlonbaeten:
@praseodym @jorisleker I have changes `-` to `0` - can you review again?

praseodym:
Why was it changed back, I thought rendering `-` was intended (https://github.com/kiesraad/abacus/pull/1676#issuecomment-3004743760)?

marlonbaeten:
> Why was it changed back, I thought rendering `-` was intended ([#1676 (comment)](https://github.com/kiesraad/abacus/pull/1676#issuecomment-3004743760))?

@jorisleker asked me on mattermost to change it back to `0`
Edit: aha, but only for the subtotals - I will change that back again :-)

jorisleker (approved):

praseodym (approved):

PR #1678, opened 2025-06-25
Retrieve user names on coordinator pages using getName from new useUsers hook

oliver3:
- Add a new Provider and Layout to retrieve users once
- Expose users through a context, together with a `getName()` function to more easily get the user's name to display
- Hook to get the `getName()` function
- Use all of this on the coordinator status and resolve pages

To test:
Different pages (status page, resolve errors, resolve differences) should still show the user's name and the code is more clean, readable and therefore maintainable


praseodym (approved):

bramtweedegolf (approved):
Look good!

PR #1679, opened 2025-06-25
Dedicated get resolve differences endpoint and only return status name

oliver3:
First commit:
- Add dedicated endpoint to get the differences for resolve differences
- Use endpoint for resolve differences
- Remove now unused get data entries / data entry status endpoint

Second commit:
- Change response for data entry finalise and resolve differences & errors endpoint to only return status name, so that the data entry status object is not exposed in the API anymore and has to be mocked etc

### DIFF
--- a/backend/.sqlx/query-4660bfcdf0bdee50df05c66848f10b73d40829367369694cc0a563f476ec3a38.json
+++ b/backend/.sqlx/query-4660bfcdf0bdee50df05c66848f10b73d40829367369694cc0a563f476ec3a38.json
@@ -11,7 +11,7 @@
       {
         "name": "event: serde_json::Value",
         "ordinal": 1,
-        "type_info": "Null"
+        "type_info": "Text"
       }
     ],
     "parameters": {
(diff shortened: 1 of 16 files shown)


bramtweedegolf (approved):
Looks good!

backend/.sqlx/query-4660bfcdf0bdee50df05c66848f10b73d40829367369694cc0a563f476ec3a38.json
praseodym, line 1: It's annoying that this specific sqlx cache file gets changed back and forth

praseodym: Created #1683

oliver3: Oh I assumed it was every time a different sqlx that only got changed forth, never checked my assumption.. Thanks!

PR #1667, opened 2025-06-23
Add create election abort button and confirmation modal on navigation

bramtweedegolf:
Added an "Afbreken" button for the election creation flow, which shows a confirmation modal.
The same modal is used when a user navigates away from the election creation screens.

Figma [link](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=1455-13075&m=dev)

Closes #1640

Note: issue #1645 will add some testing that will improve the coverage of this PR

PR #1680, opened 2025-06-25
Election create: make sure navigation and browser back and forward work as intended

bramtweedegolf:
Make sure navigating between the election creation routes works as intended.
Note: if a user moves backwards in the wizard, he/she is forced to restart from that point. 

- When a user uploads an election and a candidate list, then goes back to election; do not show the hash page but show a new election upload.
- When a user uploads an election and a candidate list, then goes back to election; clear the state and restart a new upload. 
- When the user successfully saves an election, clear the state. Using the browser back button should send the user to the upload new election form.

Fixes: #1645


frontend/e2e-tests/tests/election-create.e2e.ts
```diff
@@ -152,4 +152,199 @@ test.describe("Election creation", () => {
...
+    await checkDefinitionPage.hashInput1.fill(eml110a.hashInput1);
+    await checkDefinitionPage.hashInput2.fill(eml110a.hashInput2);
+    await checkDefinitionPage.next.click();
+
+    // Candidate page
+    const uploadCandidateDefinitionPage = new UploadCandidateDefinitionPgObj(page);
+    await expect(uploadCandidateDefinitionPage.header).toBeVisible();
+    await uploadCandidateDefinitionPage.uploadFile(page, eml230b.path);
+    const checkCandidateDefinitionPage = new CheckCandidateDefinitionPgObj(page);
+    await expect(checkCandidateDefinitionPage.header).toBeVisible();
```
jschuurk-kr: Since we perform these steps in quite a few steps, I think it would be good to add some page object methods and/or helper functions for this.

The cleanest way to do this is to add a method to the page object that performs all relevation actions on the page (there are some examples in data entry) and one or (if useful) several helper functions in `helpers-utils/e2e-test-browser-helpers.ts` to combine steps across pages.
That might be overengineering it a bit, though. So up to you what to do exactly, as long as the tests become shorter.

bramtweedegolf: Thanks! I was already considering refactoring the tests. I don't think it's overengineering, since the tests will only be getting longer as we implement more of the election upload. :)

How do your feel about this refactor? -> [4205855](https://github.com/kiesraad/abacus/pull/1680/commits/42058557eefe55626fc2290322281b8079d27b06)

Also; I previously created some helpers for this test:  frontend/src/features/election_create/components/ElectionCreate.test.tsx
How does that look to you?

frontend/e2e-tests/helpers-utils/e2e-test-browser-helpers.ts
```diff
@@ -73,3 +80,35 @@ export async function fillDataEntryPagesAndSave(page: Page, results: PollingStat
...
+
+export async function uploadCandidatesAndInputHash(page: Page) {
+  const overviewPage = new OverviewPgObj(page);
+
+  // Candidate page
+  const uploadCandidateDefinitionPage = new UploadCandidateDefinitionPgObj(page);
+  await expect(uploadCandidateDefinitionPage.header).toBeVisible();
+  await uploadCandidateDefinitionPage.uploadFile(page, eml230b.path);
+  await expect(overviewPage.main).toContainText(eml230b.filename);
+  await expect(overviewPage.main).toContainText(eml230b.electionDate);
```
jschuurk-kr: This is not the `overviewPage`, which shows the list of created elections, but the "Verkiezing toevoegen" page that contains the wizard for adding elections. (It doesn't help that the heading of the page changes after importing the election definition.)

jschuurk-kr:
If I upload an election definition and then go back to "Verkiezingsdefinitie", I am required to upload a new election definition, but the heading still mentions the old election: "Toevoegen: Gemeenteraad Amsterdam 2022"

jschuurk-kr (changes requested):
> When a user uploads an election and a candidate list, then goes back to election; do not show the hash page but show a new election upload.
> When a user uploads an election and a candidate list, then goes back to election; clear the state and restart a new upload.

This is a destructive action for which we don't present a confirmation dialog. Also, all similar-looking other navigation panels in the application do not discard data. We either need to add a confirmation dialog or allow navigation without discarding data.

Or am I missing an interaction between this PR and https://github.com/kiesraad/abacus/pull/1667?

jschuurk-kr:
Not related to the scope of this PR: The "Bestand kiezen"-button does not have a light blue highlight when I put focus on it with the Tab-key.

jschuurk-kr:
If I upload the election definition and the candidate list and then click "Type stemopneming", the url changes to http://localhost:3000/elections/create/counting-method-type and I get a page that says "Pagina niet gevonden. De pagina /dev is niet gevonden."

Not sure if it's worth the effort to change this, but I would have prefered building this epic without assuming that future epics are already in place.

frontend/e2e-tests/tests/election-create.e2e.ts
jschuurk-kr: Refactor of the e2e tests looks good! I do think comments like "// upload election and check hash" can be removed when the helper is called.

jschuurk-kr: Helpers in ElectionCreate.test.tsx look good too!

It did take me a while to realize that the comment "We leave them in here so we can test their absence"  refered to their absence in the rendered component.

frontend/e2e-tests/helpers-utils/e2e-test-browser-helpers.ts
bramtweedegolf: Fixed in [1d3f672](https://github.com/kiesraad/abacus/pull/1680/commits/1d3f67257164123fe4a4d69d157aed2855891a94)

bramtweedegolf:
> > When a user uploads an election and a candidate list, then goes back to election; do not show the hash page but show a new election upload.
> > When a user uploads an election and a candidate list, then goes back to election; clear the state and restart a new upload.
> 
> This is a destructive action for which we don't present a confirmation dialog. Also, all similar-looking other navigation panels in the application do not discard data. We either need to add a confirmation dialog or allow navigation without discarding data.
> 
> Or am I missing an interaction between this PR and #1667?
> 
> (edit) Using the browser's back and forward buttons does let me navigate back and then forward again while retaining my data.

During the epic kickoff we initially did not think it would be worth the effort to implement something like this. But it would be relatively easy to implement after we merge this PR and #1667. So I created a new issue, to discuss whether we want to add this: #1732

bramtweedegolf:
> Not related to the scope of this PR: The "Bestand kiezen"-button does not have a light blue highlight when I put focus on it with the Tab-key.

Well spotted, I created a new issue: #1733

bramtweedegolf:
> If I upload the election definition and the candidate list and then click "Type stemopneming", the url changes to http://localhost:3000/elections/create/counting-method-type and I get a page that says "Pagina niet gevonden. De pagina /dev is niet gevonden."
> 
> Not sure if it's worth the effort to change this, but I would have prefered building this epic without assuming that future epics are already in place.

Agreed :) Won't fix in this epic, but something to pay attention to in the future.

praseodym (approved):
lgtm taking into account the follow-up issues

praseodym:
@bramtweedegolf this PR has a merge conflict after merging #1667

rnijveld (approved):

Issue #1683, opened 2025-06-26
SQLx cache file gets changed repeatedly


praseodym:
The `query-4660bfcdf0bdee50df05c66848f10b73d40829367369694cc0a563f476ec3a38.json` SQLx cache file gets changed back and forth

_Originally posted by @praseodym in https://github.com/kiesraad/abacus/pull/1679#discussion_r2168797764_

PR #1684, opened 2025-06-27
Fix Clippy uninlined_format_args warnings

praseodym:
Fix uninlined_format_args lints as reported by Clippy since Rust 1.88.

### DIFF
--- a/backend/.sqlx/query-4660bfcdf0bdee50df05c66848f10b73d40829367369694cc0a563f476ec3a38.json
+++ b/backend/.sqlx/query-4660bfcdf0bdee50df05c66848f10b73d40829367369694cc0a563f476ec3a38.json
@@ -11,7 +11,7 @@
       {
         "name": "event: serde_json::Value",
         "ordinal": 1,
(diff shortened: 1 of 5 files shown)


backend/.sqlx/query-4660bfcdf0bdee50df05c66848f10b73d40829367369694cc0a563f476ec3a38.json
```diff
@@ -11,7 +11,7 @@
       {
         "name": "event: serde_json::Value",
         "ordinal": 1,
-        "type_info": "Text"
+        "type_info": "Null"
```
jschuurk-kr, line 14: How is this change related to fixing the uninlined_format_args warnings?

praseodym: It is not, this was caused by Lefthook updating the SQLx cache on pre-commit. That this file changes often is a bug: #1683.

PR #1686, opened 2025-06-27
increase number of political groups in e2e test data set

### LINKED ISSUE #1681: e2e test data: increase number of political groups
Currently our e2e tests use the election defined in `electionRequest` in `frontend/e2e-tests/test-data/request-response-templates.ts`. This election contains only one political group. That's not realistic enough.

So the test data needs to be expanded to 2 or 3 political groups, at least one of which is a blank list.

TODO:
- extend test data and fix any tests that break -> https://github.com/kiesraad/abacus/pull/1686
- confirm translation of "blanco lijst" and add it to woordenlijst-EN-NL.md -> https://github.com/kiesraad/abacus/pull/1682

jschuurk-kr:
closes #1681 

### Scope
- add second, unnamed list to test data set for e2e tests
- fix all e2e tests that broke because of the additional list
- rename `emptyRequest` to `requestWithError` because that's what it is and improve it
- rename `NavigationPanel` page object to `ProgressList` so it matches the component name and so it's no longer easy to confuse it with the `NavBar`
- add "bevat een waarschuwing" to the test "check icons for accept, active, empty, error, warning, unsaved statuses"

Changes are done through several commits. First I added the unnamed list, but always with 0 votes, and got all the tests to pass again. Then I did all the other things, including having votes for the unnamed list.

When we make the changes for the new models and errors/warnings, we should consider adding a third list to the test data set.

### DIFF
--- a/frontend/e2e-tests/test-data/request-response-templates.ts
+++ b/frontend/e2e-tests/test-data/request-response-templates.ts
@@ -195,7 +195,6 @@ export const noRecountNoDifferencesDataEntry: PollingStationResults = {
   ],
 };
 
-// TODO: check contents based on usage
 export const noRecountNoDifferencesRequest: DataEntry = {
   progress: 83,
   data: noRecountNoDifferencesDataEntry,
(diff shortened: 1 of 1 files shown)


praseodym (approved):
lgtm!

Lionqueen94 (approved):

PR #1694, opened 2025-06-30
Write to audit log when air gap violation is detected

### LINKED ISSUE #1636: airgap detection should write to auditlog
Airgap detection currently does not write to the log as it closes the API. It would be nice if it could write to the log (both the audit log and the application log), at least when it triggers and when it stops.

marlonbaeten:
Fixes #1636

- Write to audit log when air gap violation is detected
- Write to audit log when air gap violation is resolved
- Write `error` and `info` level messages to the console if the status changes (instead of `trace` and `warn` level)

Implementation notes:

- A user is now optional for the audit log, most changes are related to this
- A MPSC channel is introduced to send status updates from the sync context / thread where the checks happen to the async / tokio runtime where we write to the audit log table

To test:

- Disable your internet connection
- Start the application with the "--airgap-detection" flag
- Enable your internet connection
- Wait for abacus to detect this
- Disable you connection again
- Wait for abacus to detect this
- Check the audit log, it should show the following:

image

Test coverage is not perfect since it is hard to air-gap Github action runners.


praseodym:
@marlonbaeten this PR has failing tests

backend/src/airgap/detect.rs
```diff
@@ -212,10 +278,33 @@ mod tests {
         assert_eq!(res.status(), StatusCode::SERVICE_UNAVAILABLE);
     }
 
+    #[test(sqlx::test)]
+    async fn test_log_status_changes_to_audit_log(pool: SqlitePool) {
```
Lionqueen94, line 282: Can you expand this test to check for the logging of AirGapViolationResolved as well?

Lionqueen94 (commented):
Works well and the log looks good, just have one comment

backend/src/airgap/detect.rs
marlonbaeten: So the only way of testing this at this moment is disabling the internet connection. We do not have this option in a github action runner....

Lionqueen94: Ah true :/

Lionqueen94 (dismissed):

Lionqueen94 (approved):

PR #1703, opened 2025-07-02
Fix repeated SQLx cache changes and use SQLite STRICT tables

### LINKED ISSUE #1683: SQLx cache file gets changed repeatedly

praseodym:
- **Update event column type from JSONB to TEXT in audit_log**
  The JSONB column type confused SQLx when the table is empty.

- **Update SQLx cache prepare pre-commit command**
  - Only use default features to avoid a memory-serve build on commit.
  - Use a temporary database to ensure all migrations have been applied and
  prevent any influence from existing data in the database.

- **Use STRICT SQLite tables**

  See https://www.sqlite.org/stricttables.html for more info.

  - Use `TEXT` instead of `DATETIME` and use SQLx `NaiveDate` type
  - Use `TEXT` instead of `BLOB` for JSON and use SQLx `Json` type
  
Resolves #1683


praseodym:
Pro tip: don't test Lefthook with `--all-files` because it will run commands in the wrong directory. Test using `lefthook run pre-commit --force` instead, but note Prettier won't work because it doesn't get any changed files as arguments.

oliver3 (approved):
Strict table and generating sqlx caches from a separate db seem like good improvements!

praseodym:
> * **Update event column type from JSONB to TEXT in audit_log**
>       The JSONB column type confused SQLx when the table is empty.

When this PR is merged, I will rebase #1706 to fix the audit log to use actual JSONB instead of a 'JSONB' typed column storing regular text.

Issue #1515, opened 2025-05-20
Coordinator: navigate from resolve differences to errors and warnings


oliver3:
If applicable

PR #1704, opened 2025-07-02
Go to resolve errors after resolve differences when keeping second data entry with errors

oliver3:
- Closes https://github.com/kiesraad/abacus/issues/1515
- Add e2e scenario, fixture and do some refactoring
- Use new state returned by the backend to decide where to go after resolve differences

Question: Is this refactor better? are these data entry response templates names better?

To test: 
- do a first data entry and then a second data entry that has an error and therefore is different from the first
- resolve the differences as a coordinator by keeping the second data entry
- see that the resolve errors page is now shown instead of the coordinator status page

### DIFF
--- a/frontend/e2e-tests/fixtures.ts
+++ b/frontend/e2e-tests/fixtures.ts
@@ -14,17 +14,14 @@ import {
 } from "@/types/generated/openapi";
 
 import { DataEntryApiClient } from "./helpers-utils/api-clients";
-import {
-  completePollingStationDataEntries,
-  completePollingStationDataEntriesWithDifferences,
-  loginAs,
-} from "./helpers-utils/e2e-test-api-helpers";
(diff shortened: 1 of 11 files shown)


praseodym (changes requested):
Tested, redirect works! But the [alert from the design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=10401-68917&t=Wd5zd4LOHuBDnyem-4) is still missing on the resolve errors page.

oliver3:
> Tested, redirect works! But the alert from the design is still missing on the resolve errors page.

Decided to create a separate issue #1705 for this part of the requirements that I completely missed

praseodym (approved):

Lionqueen94 (approved):

PR #1706, opened 2025-07-02
Update audit log to store events as SQLite JSONB 

praseodym:
Using JSONB is more efficient when querying fields inside the JSON event, such as done in the filtering functions.

For more information about SQLite JSONB, see: https://sqlite.org/json1.html#jsonb


Lionqueen94 (approved):

praseodym:
> Surprised me that JSONB was allowed for STRICT but ok!

What do you mean?

PR #1710, opened 2025-07-03
Added flowchart for committee session state

Lionqueen94:


jorisleker (changes requested):
Would change two things:

- 'delete all polling stations' --> 'delete last polling station' (as suggested by @praseodym), since we do not have a batch delete feature and it's really about not having any left
- In the footnote: "currently it's only possible to delete polling stations that do not have a data entry in progress". Would replacd this '... that do nat have data entry', because 'in progress' suggest that you can delete it when data entry is finalized, which isn't the case.

marjoleintamis (dismissed):

jorisleker (dismissed):

jorisleker (approved):

praseodym (approved):

jorisleker (approved):

Issue #1705, opened 2025-07-02
Show notification on resolve errors page when coming from resolve differences


oliver3:
Separated from #1515, see comment https://github.com/kiesraad/abacus/pull/1704#pullrequestreview-2979320278

Including an investigation for an alternative way of dealing with cross-page messages

We should look into fixing https://github.com/kiesraad/abacus/issues/1731 in this issue/PR as well

PR #1711, opened 2025-07-03
Messages provider, hook and component to show a message to the user on the next page

oliver3:
In preparation of #1705 (Show notification on resolve errors page): new way of providing a message to the user on the next page, e.g. for notifying that the polling station was successfully created

- Context and provider that keep an array of messages
- Hook with method to push a message for the next page, to be used on the "from" page
- Component that retrieves and clears all messages from the context and shows them, to be used on the "to" page
- Use for the polling station pages
  - Change PUT polling station endpoint to return the updated polling station

NB: titles in the alert for polling stations are now changed to `1.25rem` which is actually according to design

#### To test
Create, update or delete a polling station and notice that you receive the correct message on the polling station list page. If you navigate back and forth to that page (e.g. by editing  a polling station and cancelling), the message is not shown anymore.


Lionqueen94 (approved):
Works great!

PR #1655, opened 2025-06-20
Replaced missing total error with new F.402 validation error

Lionqueen94:

PR #1713, opened 2025-07-04
Fix data entry progress list resizing and scrolling

### LINKED ISSUE #1712: Data entry progress list always shows scroll bar with many lists in Firefox
Scrollbar is shown even with no scrollable content:
Image

No scrollbar is shown in Google Chrome.

The `StickyNav` component uses JavaScript to resize the progress list when scrolling down, to compensate for the disappearing header. This JavaScript logic is what causes the scrollbar to appear on Chrome.

praseodym:
- Simplify sticky nav max height handling, and use CSS variable for top offset. This fixes #1712. I couldn't find a solution without JavaScript.
- Simplify progress list item scroll behaviour
- Remove domtoren util
- Restore the progress list scroll gradient that was broken (probably due to the LightningCSS migration), see screenshot below

Can be tested by creating an election that has more political groups than fit in the height of your browser window (e.g. `cargo run --bin gen-test-election -- --political-groups 50`) and then using the data entry.

Progress list scroll gradient:
Screenshot 2025-07-04 at 14 47 05

### DIFF
--- a/frontend/src/components/ui/ProgressList/ProgressList.module.css
+++ b/frontend/src/components/ui/ProgressList/ProgressList.module.css
@@ -206,18 +206,18 @@ section.scroll {
     background: linear-gradient(to top, var(--bg-gray) 0%, rgba(249, 250, 251, 0) 100%);
   }
 
-  &:global(.showTopGradient) {
+  &.showTopGradient {
     &::before {
       opacity: 1;
     }
   }
-  &:global(.showBottomGradient) {
+  &.showBottomGradient {
     &::after {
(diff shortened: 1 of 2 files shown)


praseodym:
> I'm not sure if this broke now (probably didn't) but when I get the F.402 error the error is not scrolled into view. Same for the error when you have not checked the accept errors/warnings checkbox.

Looks like this is broken in `main` and didn't work in #1655 either. I've created #1723 for that.

Lionqueen94 (approved):

PR #1729, opened 2025-07-07
Small improvement in frontend openapi generator

bramtweedegolf:
Fixed an issue in the openapi generator, where if the backend returns an Option<Vec<...>>, the typescript type becomes Unknown[].

In the openapi XML, the Option<Vec<...>> becomes type: ["array", "null"]:
openapi1

The rest of the generated types remain unaffected:

openapi3

### DIFF
--- a/frontend/scripts/openapi/generator.ts
+++ b/frontend/scripts/openapi/generator.ts
@@ -176,12 +176,20 @@ function tsType(s: ReferenceObject | SchemaObject | undefined): string {
 
   if (Array.isArray(s.type)) {
     type = s.type
-      .map((t: NonArraySchemaObjectType | "null") => {
+      .map((t: NonArraySchemaObjectType | "null" | "array") => {
         if (t === "null") {
           return "null";
         }
 
-        return tsType({ type: t });
+        if (t === "array") {
+          if ("items" in s) {
+            return tsType({ type: t, items: s.items });
+          } else {
+            // Cannot determine type; type remains "unknown"
+          }
+        } else {
+          return tsType({ type: t });
+        }
       })
       .join(" | ");
   }


frontend/scripts/openapi/generator.ts
```diff
@@ -176,12 +176,20 @@ function tsType(s: ReferenceObject | SchemaObject | undefined): string {
...
-        return tsType({ type: t });
+        if (t === "array") {
+          if ("items" in s) {
+            return tsType({ type: t, items: s.items });
+          } else {
+            // Cannot determine type; type remains "unknown"
+          }
+        } else {
+          return tsType({ type: t });
+        }
```
praseodym, line 192: Maybe this can be simplified by calling `tsType({ type: s });` instead of `tsType({ type: t });`, because `tsType` already has logic to handle an array type. Did you try that?

bramtweedegolf: Good idea! I just tried; unfortunately I don't think we can. When the function tsType is called we have these params:

```
{
  s: {
    type: [ 'array', 'null' ],
    items: { '$ref': '#/components/schemas/PollingStationRequest' }
  },
  t: 'array'
}
```

So we are already running tsType({ type: s }), which is looping over the types in [ 'array', 'null' ].
We need to look at the $ref in items; the only case that does that is this one:

```
switch (s.type) {
  ...
  case "array":
      type = `${tsType(s.items)}[]`;
      break;
  ...
}
```

which does another recursive call that triggers this case:

```
if ("$ref" in s) {
  return s.$ref.substring(s.$ref.lastIndexOf("/") + 1);
}
```

Which finally yields a type after 2 recursive subcalls.

We could maybe change line 186.

`return tsType({ type: t, items: s.items });`

to
`
return '${tsType( s.items )}[]';`

To remove one of the recursive subcalls. Unfortunately that doesn't really make the code any simpler / more readable.

PR #1753, opened 2025-07-08
Switch to Storybook

praseodym:
Remove Ladle and add Storybook, resolves #1725.

- Storybook can be started using `npm run storybook`.
- See [How to write stories](https://storybook.js.org/docs/writing-stories). Storybook can automatically create controls for most component props, which reduces the need to explicitly create stories for variants of a component. The current stories have been mostly migrated from Ladle, so they are not always the best starting point when writing a new story.
- Playwright Ladle tests have been rewritten using [Storybook interaction tests](https://storybook.js.org/docs/writing-tests/interaction-testing) (Vitest in browser mode).
- A static preview deployment is available at `/storybook` on the preview URL, e.g. https://storybook.kiesraad-abacus.pages.dev/storybook/
- Storybook reduces the number of npm dependencies:
  - Ladle (`main`): 827 packages
  - Ladle removed: 575 packages
  - Storybook added: 638 packages

Issue #1750, opened 2025-07-08
Add committee session details page


Lionqueen94:
[Figma design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=9860-56278&m=dev)

Issue #1604, opened 2025-06-10
Pause/resume data entry


praseodym:
Ability to pause and resume data entry for the current committee session.

jschuurk-kr:
Findings so far:

1. The text for the coordinator "Het is nu niet mogelijk om nieuwe stembureaus in te voeren." is confusing. Typists can no longer do data entry, but the coordinator is still able to add polling stations. (cc @jorisleker)

2. If the coordinator clicks "Stembureaus" on the status page, adds a new polling station and then returns to the status page, the newly added polling station is not present on the status page.
 
3. If the coordinator clicks "Stembureaus" on the status page, then adds a new polling station and wants to get back to the status page, they need to first use the NavBar to go the main election page and then click there to get to the status page. It would be nice to be able to return directly. (cc @jorisleker)

4. If data entry is paused, typists do not get a 409 for the /elections/1/data-entry page. Yet they can't navigate to that page from the page that lists all the elections. This is inconsistent.

Lionqueen94:
> Findings so far:
> 
>     1. The text for the coordinator "Het is nu niet mogelijk om nieuwe stembureaus in te voeren." is confusing. Typists can no longer do data entry, but the coordinator is still able to add polling stations. (cc [@jorisleker](https://github.com/jorisleker))
> 
>     2. If the coordinator clicks "Stembureaus" on the status page, adds a new polling station and then returns to the status page, the newly added polling station is not present on the status page.
> 
>     3. If the coordinator clicks "Stembureaus" on the status page, then adds a new polling station and wants to get back to the status page, they need to first use the NavBar to go the main election page and then click there to get to the status page. It would be nice to be able to return directly. (cc [@jorisleker](https://github.com/jorisleker))
> 
>     4. If data entry is paused, typists do not get a 409 for the /elections/1/data-entry page. Yet they can't navigate to that page from the page that lists all the elections. This is inconsistent.

4. This will be changed in #1852, a modal will be implemented for this specific case.

jorisleker:
> 1. The text for the coordinator "Het is nu niet mogelijk om nieuwe stembureaus in te voeren." is confusing. Typists can no longer do data entry, but the coordinator is still able to add polling stations. (cc [@jorisleker](https://github.com/jorisleker))

We only show the 'het is niet mogelijk om nieuwe stembureaus in te voeren' message to the typist, right? For them, this message is correct. Don't this will confuse coordinators.  
  
[...]

PR #1754, opened 2025-07-08
Added committee session frontend

### LINKED ISSUE #1602: Committee session frontend

### LINKED ISSUE #1699: Change status overview breadcrumb
Status overview subpages should have name of the committee session (e.g. "Eerste zitting").

See [Figma design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=10401-64583&t=yc8tTFQkVYC0gWfc-4), specifically [this comment thread](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=10401-64583#1273253537).

Lionqueen94:
Closes #1699 
Closes #1602


frontend/src/components/layout/ElectionLayout.tsx
```diff
@@ -12,8 +13,10 @@ export function ElectionLayout() {
   return (
     <ElectionProvider electionId={electionId}>
       <ElectionStatusProvider electionId={electionId}>
-        <NavBar location={location} />
-        <Outlet />
+        <CommitteeSessionListProvider electionId={electionId}>
```
Lionqueen94: @praseodym I only need this provider for the ElectionHomePage, is there any way to not add this provider to this layout so it's needed for all child pages as well, making the endpoint need User instead of Coordinator permission. Or do you have another suggestion how to improve this?

praseodym: What about adding the provider to the `ElectionHomePage` instead of to the layout?

Lionqueen94: Refactored as discussed, thanks!

backend/src/committee_session/api.rs
```diff
@@ -59,9 +59,10 @@ pub async fn election_committee_session_list(
     Path(election_id): Path<u32>,
 ) -> Result<Json<CommitteeSessionListResponse>, APIError> {
     elections_repo.get(election_id).await?;
-    let committee_sessions = committee_sessions_repo
+    let mut committee_sessions = committee_sessions_repo
         .get_election_committee_session_list(election_id)
         .await?;
+    committee_sessions.reverse();
```
praseodym: Can't we sort in the database using SQL `ORDER BY`?

frontend/src/components/committee_session/CommitteeSessionStatus.tsx
```diff
@@ -0,0 +1,43 @@
...
+    case "data_entry_not_started":
+      return t(`committee_session_status.${role}.data_entry_not_started`);
+    case "data_entry_in_progress":
+      return t(`committee_session_status.${role}.data_entry_in_progress`);
+    case "data_entry_paused":
+      return t(`committee_session_status.${role}.data_entry_paused`);
+    case "data_entry_finished":
+      return t(`committee_session_status.${role}.data_entry_finished`);
+  }
+}
```
praseodym: Doesn't ``t(`committee_session_status.${role}.${status}`)`` work?

frontend/src/features/election_management/components/CommitteeSessionCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+
+export interface ButtonLink {
+  label: string;
+  to: To;
+}
+
+function ButtonLinkList(buttonLinks: ButtonLink[], firstRowBold: boolean) {
+  return (
+    <Table className="buttonLinks">
```
praseodym: We're only using the `buttonLinks` class in this `CommitteeSessionCard` component, so we should define the class in `CommitteeSessionCard.module.css` instead of in `Table.module.css`.

We can then also use the `cls.buttonLinks` pattern to import the class so that it doesn't need to be `:global`.

praseodym (commented):
I noticed that `/elections/1` no longer works as an administrator due to 401 from `/api/elections/1/committee_sessions`, which is problematic because the administrator needs to be able to manage polling stations until data entry is started.

frontend/src/features/election_management/components/CommitteeSessionCard.tsx
Lionqueen94: I tried this, but I had some issues with `tr.rowLink` in the styling, but might just be able to use `tr` instead. Will try!

Lionqueen94: Fixed in 1c481e9fcb5c70d0c2eba21cce246514955f9038

frontend/src/components/committee_session/CommitteeSessionStatus.tsx
Lionqueen94: Fixed in 1c481e9fcb5c70d0c2eba21cce246514955f9038

backend/src/committee_session/api.rs
Lionqueen94: Of course this makes more sense haha, fixed in 1c481e9fcb5c70d0c2eba21cce246514955f9038

praseodym (dismissed):

praseodym (approved):

praseodym:
@Lionqueen94 #1753 has been merged, can you port the Ladle story to Storybook?

jorisleker:
> * Question for @jorisleker Should the "Details van de zitting" link always be visible, even when the details have been filled in already? Relevant for issue [Add committee session details page #1750](https://github.com/kiesraad/abacus/issues/1750)

Yes it's okay to always have this link visible, I think we should decide on the committee session detail page whether the user is allowed to edit, or should only be able to view the details

jorisleker (changes requested):
Reviewed on ladle, some things I noticed:

- startdate/time are always visible now. they should not be visible if they have not been entered (likely in 'created' and 'not_started' states)
- "paused" does not have a link to resume ([see how it should be](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=9485-34701&t=p7b2XThXRYjDp4G3-11))
- Can't see the effect of the 'currentSession' boolean?

frontend/.storybook/preview.tsx
```diff
@@ -20,6 +20,10 @@ const preview: Preview = {
       },
     },
   },
+  initialGlobals: {
+    // 👇 Set the initial background color
+    backgrounds: { value: "light" },
+  },
```
praseodym: This changes the default background colour for all stories, is that intended?

Lionqueen94:
> Reviewed on ladle, some things I noticed:
> 
>     * startdate/time are always visible now. they should not be visible if they have not been entered (likely in 'created' and 'not_started' states)
> 
>     * "paused" does not have a link to resume ([see how it should be](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=9485-34701&t=p7b2XThXRYjDp4G3-11))
> 
>     * Can't see the effect of the 'currentSession' boolean?

- when you remove the date and time in the controls they remove from the card.
- pausing and resuming will be implemented in issue #1604 
- `currentSession=true` makes the top link bold on the card, this is more visible when you have 2 sessions that both have links

Also, what do you think of the hover state in the screenshot in the description of the PR? @jorisleker

frontend/.storybook/preview.tsx
Lionqueen94: I thought it just sets the default background to light theme. I need it to be able to override the background color to grey to see my white card. Do you have a better way for this?

praseodym: See https://storybook.js.org/docs/essentials/backgrounds#defining-the-background-for-a-story, so use `globals` in your story and don't modify `preview.tsx`

jorisleker:
Re: hover state. Forgot that in the design, thanks for spotting. Can you increase padding so the background fill extends all the way to the separator (top and bottom equal padding)?

Lionqueen94:
> Re: hover state. Forgot that in the design, thanks for spotting. Can you increase padding so the background fill extends all the way to the separator (top and bottom equal padding)?

Padding is currently a `gap: 0.5rem`, otherwise it was going to be tricky when to add the padding or when not (first and last element). But I'll see if I can change anything

frontend/.storybook/preview.tsx
Lionqueen94: Without the `initialGlobals` in `preview.tsx` the `globals` setting in a story does not seem to work :/

praseodym: Fixed it in 180cfa9abe77fc533e384e92e9149491e45f3023

praseodym (approved):

jorisleker (approved):

praseodym (approved):

Issue #1632, opened 2025-06-17
Frontend page to import polling stations for coordinator


bramtweedegolf:
Depends on https://github.com/kiesraad/abacus/pull/1949, https://github.com/kiesraad/abacus/pull/1954, https://github.com/kiesraad/abacus/pull/1955 and https://github.com/kiesraad/abacus/pull/1955

PR #1758, opened 2025-07-09
1632 frontend page to import polling stations for coordinator

bramtweedegolf:
After the election and candidate list upload, a list of polling stations can be uploaded.
The list of polling stations does not have to have the same election id as the previously uploaded files.
Uploading the polling stations can be skipped. 

TODO:
- Styling of several buttons
- Add last frontend e2e tests

Fixes https://github.com/kiesraad/abacus/issues/1630 and https://github.com/kiesraad/abacus/issues/1632

Depends on https://github.com/kiesraad/abacus/pull/1680 and ttps://github.com/kiesraad/abacus/pull/1667


backend/src/election/api.rs
```diff
@@ -215,7 +222,20 @@ pub async fn election_import_validate(
         election = EML230::from_str(&data)?.add_candidate_lists(election)?;
     }
 
-    Ok(Json(ElectionDefinitionValidateResponse { hash, election }))
+    // parse and validate polling stations, and update number of voters
+    let polling_stations;
+    if let Some(data) = edu.polling_station_data.clone() {
```
rnijveld: I don't think we need to clone here, because we're just using references anyway?

backend/src/election/api.rs
```diff
@@ -255,7 +277,24 @@ pub async fn election_import(
     let mut new_election = EML110::from_str(&edu.election_data)?.as_abacus_election()?;
     new_election = EML230::from_str(&edu.candidate_data)?.add_candidate_lists(new_election)?;
 
-    let election = elections_repo.create(new_election).await?;
+    // Process polling stations
+    let mut polling_places = None;
+    if let Some(polling_station_data) = edu.polling_station_data.clone() {
```
rnijveld: Same as previous comment

backend/src/eml/eml_110.rs
```diff
@@ -180,6 +186,55 @@ impl EML110 {
...
+        &self,
+        mut election: crate::election::NewElection,
+    ) -> std::result::Result<crate::election::NewElection, EMLImportError> {
+        // we need to be importing from a 110b file
+        if self.base.id != "110b" {
+            return Err(EMLImportError::Needs110b);
+        }
+
+        // Fill in number of voters
+        if let Some(voters) = &self.contest().max_votes {
```
rnijveld, line 227: I think we want to change `max_votes` in the `Contest` struct to parse as `u64` or something like that already, that saves some checking here.

backend/src/eml/eml_110.rs
```diff
@@ -397,6 +452,36 @@ pub struct PollingStation {
...
+            number: self
+                .physical_location
+                .polling_station
+                .id
+                .parse()
+                .or(Err(EMLImportError::InvalidPollingStation))?,
+            number_of_voters: None,
+            polling_station_type: None,
+            address: "".to_string(),
+            postal_code: match self.physical_location.address.locality.postal_code.clone() {
```
rnijveld, line 477: I think postal code can just remain empty, I don't think we should fail importing if a postal code is missing. cc @praseodym

backend/src/polling_station/repository.rs
```diff
@@ -131,6 +131,42 @@ impl PollingStations {
...
 
+    /// Create many polling stations for an election
+    pub async fn create_many(
+        &self,
+        election_id: u32,
+        new_polling_stations: Vec<PollingStationRequest>,
+    ) -> Result<Vec<PollingStation>, sqlx::Error> {
+        // Max number of parameters for a statement
+        // See: https://www.sqlite.org/limits.html
+        const BIND_LIMIT: usize = 999;
```
rnijveld: As an alternative to this loop, we could also start a transaction here and just execute a single query for each polling station, that should perform almost as fast as this solution, while not having the somewhat fragile bind limit calculation.

If we're not doing that, I would at least add several warning comments to make sure to update the bind limit calculation when adding or removing fields.

backend/src/eml/eml_110.rs
praseodym: Agreed

backend/src/polling_station/structs.rs
```diff
@@ -23,7 +23,7 @@ pub struct PollingStation {
     #[schema(nullable = false)]
     pub polling_station_type: Option<PollingStationType>,
     pub address: String,
-    pub postal_code: String,
+    pub postal_code: Option<String>,
```
praseodym: I don't think we need to make this an `Option`, because strings can be left empty and we currently already use the postal code field that way in the frontend.

praseodym:
Can you remove commits that were merged in #1667 from this PR?

frontend/e2e-tests/page-objects/election/create/CheckPollingStationDefinitionPgObj.ts
```diff
@@ -0,0 +1,17 @@
...
+export class CheckPollingStationDefinitionPgObj {
+  readonly header: Locator;
+  readonly showMore: Locator;
+  readonly next: Locator;
+  readonly table: Locator;
+  readonly stations: Locator;
+
+  constructor(protected readonly page: Page) {
+    this.header = page.getByRole("heading", { level: 2, name: "Controleer stembureaus" });
+    this.showMore = page.getByTestId("show-more");
```
jschuurk-kr:
```suggestion
    this.showMore = page.getByRole("button", { name: /Toon alle \d+ stembureaus/})
```

frontend/e2e-tests/tests/election-create.e2e.ts
```diff
@@ -30,9 +33,42 @@ test.describe("Election creation", () => {
...
+    // upload election and check hash
+    await uploadElectionAndInputHash(page);
+
+    // upload candidates list and check
+    await uploadCandidatesAndInputHash(page);
+
+    // skip polling stations
+    const uploadDefinitionPage = new UploadPollingStationDefinitionPgObj(page);
+    await expect(uploadDefinitionPage.header).toBeVisible();
+    await uploadDefinitionPage.skipButton.click();
```
jschuurk-kr:
```suggestion
    const uploadPollingStationsPage = new UploadPollingStationDefinitionPgObj(page);
    await expect(uploadPollingStationsPage.header).toBeVisible();
    await uploadPollingStationsPage.skipButton.click();
```

frontend/e2e-tests/tests/election-create.e2e.ts
```diff
@@ -213,19 +250,16 @@ test.describe("Election creation", () => {
...
+    await expect(uploadCandidateDefinitionPage.header).toBeVisible();
 
     // Back button
     await page.goBack();
 
     // We should be back at the candidate page
-    const uploadCandidateDefinitionPage = new UploadCandidateDefinitionPgObj(page);
-    await expect(uploadCandidateDefinitionPage.header).toBeVisible();
+    const uploadDefinitionPage = new UploadDefinitionPgObj(page);
+    await expect(uploadDefinitionPage.header).toBeVisible();
```
jschuurk-kr: The comment is now incorrect. I think the best solution is to rename `UploadDefinitionPgObj` to `UploadElectionDefinitionPgObj` and remove the comment.

(I think in general as a team we have the habbit to overcomment tests slightly.)

frontend/e2e-tests/tests/election-create.e2e.ts
```diff
@@ -323,4 +360,80 @@ test.describe("Election creation", () => {
...
+
+    // upload wrong file
+    const uploadDefinitionPage = new UploadPollingStationDefinitionPgObj(page);
+    await expect(uploadDefinitionPage.header).toBeVisible();
+    await uploadDefinitionPage.uploadFile(page, eml110a.path);
+    await expect(uploadDefinitionPage.main).toContainText(eml110a.filename);
+    await expect(uploadDefinitionPage.error).toBeVisible();
+  });
+
+  test("show more button should show full list of polling stations", async ({ page }) => {
```
jschuurk-kr, line 382: For this test and the next one ("no show more button should be visible is <10 polling stations"), can't we cover that in an RTL test?

jschuurk-kr: @bramtweedegolf What was the resolution here?

frontend/src/features/election_create/components/UploadPollingStationDefinition.module.css
```diff
@@ -0,0 +1,11 @@
+button {
```
oliver3: A bit surprising to me that this does not fail, but this should be the class `.button` (with a dot) if I understand correctly. Now all buttons in the application are styled like this.

backend/src/election/api.rs
```diff
@@ -170,12 +172,17 @@ pub struct ElectionAndCandidateDefinitionValidateRequest {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = Option<Vec<String>>, nullable = false)]
+    polling_station_data: Option<String>,
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
 pub struct ElectionDefinitionValidateResponse {
     hash: RedactedEmlHash,
     election: NewElection,
+    polling_stations: Option<Vec<PollingStationRequest>>,
```
oliver3, line 188:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    #[schema(nullable = false)]
    polling_stations: Option<Vec<PollingStationRequest>>,
```

backend/src/election/api.rs
```diff
@@ -226,6 +246,7 @@ pub struct ElectionAndCandidatesDefinitionImportRequest {
     election_data: String,
     candidate_hash: [String; crate::eml::hash::CHUNK_COUNT],
     candidate_data: String,
+    polling_station_data: Option<String>,
```
oliver3, line 257:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    #[schema(nullable = false)]
    polling_station_data: Option<String>,
```

backend/src/election/api.rs
```diff
@@ -170,12 +172,17 @@ pub struct ElectionAndCandidateDefinitionValidateRequest {
     #[schema(value_type = Option<Vec<String>>, nullable = false)]
     candidate_hash: Option<[String; crate::eml::hash::CHUNK_COUNT]>,
     candidate_data: Option<String>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = Option<Vec<String>>, nullable = false)]
```
oliver3: Why `Vec`?

frontend/src/features/election_create/components/ElectionCreateContextProvider.tsx
```diff
@@ -47,6 +61,7 @@ function reducer(state: ElectionCreateState, action: ElectionCreateAction): Elec
       return {
         ...state,
         election: action.response.election,
+        pollingStations: null,
```
oliver3: So in this case pollingStations is set to null, but e.g. for `SELECT_CANDIDATES_DEFINITION` this does not happen. Is this intentional?

frontend/src/features/election_create/components/UploadPollingStationDefinition.tsx
```diff
@@ -0,0 +1,170 @@
...
+  async function onFileChange(e: React.ChangeEvent<HTMLInputElement>) {
+    const currentFile = e.target.files ? e.target.files[0] : undefined;
+    if (currentFile !== undefined) {
+      const data = await currentFile.text();
+      const response = await create({
+        election_hash: state.electionDefinitionHash,
+        election_data: state.electionDefinitionData,
+        candidate_hash: state.candidateDefinitionHash,
+        candidate_data: state.candidateDefinitionData,
+        polling_station_data: data,
```
oliver3, line 48: Apparently Serde is pretty forgiving, because if you would `satisfies ElectionAndCandidateDefinitionValidateRequest` this, it would complain that `polling_station_data` has to be an array and not a string.... (see also remark about the `value_type = Option<Vec<String>>`)

backend/src/polling_station/repository.rs
oliver3: It seems that this value can be set a lot higher, if I understand correctly:

`SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0` (from https://www.sqlite.org/limits.html)

backend/src/polling_station/repository.rs
```diff
@@ -131,6 +131,42 @@ impl PollingStations {
...
+
+        let mut stations: Vec<PollingStation> = Vec::new();
+
+        let batches = new_polling_stations.chunks(BIND_LIMIT / 8);
+        for batch in batches {
+            let mut inserted: Vec<PollingStation> = QueryBuilder::new(
+                "INSERT INTO polling_stations (election_id, name, number, number_of_voters, polling_station_type, address, postal_code, locality) "
+              ).push_values(batch.iter(), |mut b, station| {
+                b.push_bind(election_id)
+                 .push_bind(station.name.clone())
```
oliver3: Not sure if it matters (was mostly curious myself) but this can be `&station.name` instead of clone, i.e. the compiler and tests do not complain :)

backend/src/election/api.rs
```diff
@@ -258,11 +280,28 @@ pub async fn election_import(
     let mut new_election = EML110::from_str(&edu.election_data)?.as_abacus_election()?;
     new_election = EML230::from_str(&edu.candidate_data)?.add_candidate_lists(new_election)?;
 
+    // Process polling stations
+    let mut polling_places = None;
+    if let Some(polling_station_data) = edu.polling_station_data {
+        new_election =
+            EML110::from_str(&polling_station_data)?.update_number_of_voters(new_election)?;
```
oliver3: A bit confusing that it both mutates the `new_election` and also returns it. Not sure if this can be improved

frontend/src/features/election_create/components/ElectionCreate.test.tsx
```diff
@@ -336,36 +389,338 @@ describe("Election create pages", () => {
...
 
-    // Expect stub to be highlighted
-    expect(screen.getByText("1")).toHaveRole("mark");
-    expect(screen.getByText("2")).not.toHaveRole("mark");
+    // The modal should have triggered
+    expect(await screen.findByRole("heading", { level: 2, name: "Niet opgeslagen wijzigingen" })).toBeVisible();
+  });
 
-    // Override again
+  test("It does not show the confirmation modal when attempting to navigate away if nothing was done", async () => {
```
oliver3: I think there was a merge error, tests are duplicated in this file.

frontend/src/features/election_create/components/ElectionCreate.test.tsx
```diff
@@ -336,36 +389,338 @@ describe("Election create pages", () => {
...
+
+    // Close button should keep us at the import page
+    const closeButton = screen.getByText("Annuleren");
+    expect(closeButton).toBeVisible();
+    await user.click(closeButton);
+    expect(await screen.findByRole("heading", { level: 2, name: "Controleer kandidatenlijst" })).toBeVisible();
+  });
+
+  test("That the confirmation modal delete button closes the modal", async () => {
+    vi.spyOn(console, "warn").mockImplementation(() => {});
```
oliver3: It seems there are no console.warn or console.error calls in the code.

Please remove all spyOn console and check the one console.warning that seems actually valid:

```
Matched leaf route at location "/elections" does not have an element or Component. This means it will render an <Outlet /> with a null value by default resulting in an "empty" page.

```

oliver3 (changes requested):

frontend/src/features/election_create/components/UploadPollingStationDefinition.module.css
praseodym: A class is nicer, but this might work because of CSS modules

oliver3: It does not unfortunately, buttons throughout the application now have hover underlines.

backend/src/polling_station/repository.rs
praseodym: You mentioned this in the PR description:
> Saving 400+ polling stations in separate queries took about 8s, so I added a 'create_many' in the repo

That's 20ms per polling station, I'm really surprised SQLite would be that slow. Do you have any idea why why it would be that slow?

Maybe it's worth trying again now that we turned on WAL mode (#1831).

rnijveld (approved):

frontend/src/features/election_create/components/ElectionCreate.test.tsx
oliver3: Warning does not affect the test I guess..

frontend/e2e-tests/tests/election-create.e2e.ts
bramtweedegolf: @jschuurk-kr Sorry, apparently my responses were stuck in 'pending' and were never sent. 

We could cover this in a RTL test; but the RTL tests use mock data, while the e2e tests use the EML files. We already have some EML files with >10 polling stations, while creating mock data for a RTL test is more bothersome. Do we want to do that? (If so, I will create a separate issue.)

jschuurk-kr: I don't think it being more bothersome is a very strong argument.

However, because of a number of arguments I don't feel like enumerating here, I'm fine with leaving these tests as they are now.

PR #1764, opened 2025-07-11
Add change committee session status endpoint

Lionqueen94:


jschuurk-kr:
> I added tests for the committee session status being incorrect for all frontend pages, but somehow codecov still says the throw new Error(...) is never triggered, do you have any idea what's wrong @jschuurk-kr?

Am not sure. Maybe it's not registering for a similar reason as what's mentioned in this explanation about vitest's [`toThrowError()`](https://vitest.dev/api/expect.html#tothrowerror): "You must wrap the code in a function, otherwise the error will not be caught, and test will fail. This does not apply for async calls as rejects correctly unwraps the promise."

The test does fail if you if you add an `expect` to `expectErrorPage()` that should resolve to `false`?

Lionqueen94:
> > I added tests for the committee session status being incorrect for all frontend pages, but somehow codecov still says the throw new Error(...) is never triggered, do you have any idea what's wrong @jschuurk-kr?
> 
> Am not sure. Maybe it's not registering for a similar reason as what's mentioned in this explanation about vitest's [`toThrowError()`](https://vitest.dev/api/expect.html#tothrowerror): "You must wrap the code in a function, otherwise the error will not be caught, and test will fail. This does not apply for async calls as rejects correctly unwraps the promise."
> 
> The test does fail if you if you add an `expect` to `expectErrorPage()` that should resolve to `false`?

Turns out the tests work fine, but codecov does not think the line is covered. Created a separate issue #1813 to change the errors to 403 errors and show an error modal on the Data Entry pages

marjoleintamis (dismissed):
Flowchart gecheckt ivm documentatie, ziet er goed uit!

backend/fixtures/election_6.sql
```diff
@@ -0,0 +1,50 @@
...
+                "last_name": "Doe",
+                "locality": "Rotterdam",
+                "gender": null
+              }
+            ]
+          }
+        ]');
+
+INSERT INTO committee_sessions (id, number, election_id, location, start_date, start_time)
+VALUES (7, 1, 6,'', '', '');
```
praseodym, line 50: Doesn't this also need a committee session status?

backend/src/polling_station/repository.rs
```diff
@@ -4,6 +4,7 @@ use sqlx::{SqlitePool, query, query_as};
 use super::structs::{PollingStation, PollingStationRequest};
 use crate::AppState;
 
+#[derive(Clone)]
```
praseodym: Looks like we don't need this

backend/fixtures/election_6.sql
Lionqueen94: No the default status is 'created' and I need one fixture to have this status for a committee session

backend/src/polling_station/repository.rs
Lionqueen94: Correct, it was needed in an earlier iteration but not anymore now, removed in 594e780bce904ece1035230779ce1614213dc483

jschuurk-kr:
Probably not in scope of this PR, but vertical alignment is off in two places in the election overview page
image

jschuurk-kr:
Admins and coordinators can reach /elections/2/status before "Start steminvoer" has been clicked. The status page then has "Klaar voor invoer" in the top-right corner, which is ~~incorrect~~ correct, but confusing to me. Because "invoer" can not start until "Start steminvoer" has been clicked.

jschuurk-kr:
If I delete all polling stations for an election, the status of the session does return to "Voorbereiden", but the number of polling stations on the election overview page (e.g. /elections/2) still shows the old number of polling stations.

jschuurk-kr:
As admin I have a "Steminvoer starten" button, but when I click it, I get a 401 Unauthorized response and am redirected to the login page.

Additionally, my session still persists. So even though I can't return to the election overview page by using the browser's back-button, I can return to it by entering the right url in the browser's address bar.

Lionqueen94:
> Probably not in scope of this PR, but vertical alignment is off in two places in the election overview page image

Nope it's not, this is by design :sweat_smile: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=9780-54761&m=dev

jschuurk-kr:
> > Probably not in scope of this PR, but vertical alignment is off in two places in the election overview page image
> 
> Nope it's not, this is by design 😅 https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=9780-54761&m=dev

And turns out I'm not the first one to ask. :-D So let's have the conversation in Figma and not in this PR.

Lionqueen94:
> Admins and coordinators can reach /elections/2/status before "Start steminvoer" has been clicked. The status page then has "Klaar voor invoer" in the top-right corner, which is ~incorrect~ correct, but confusing to me. Because "invoer" can not start until "Start steminvoer" has been clicked.

@jorisleker shall I also add a "Start steminvoer" button to the `ElectionStatusPage` next to the "Klaar voor invoer" status?

Lionqueen94:
> As admin I have a "Steminvoer starten" button, but when I click it, I get a 401 Unauthorized response and am redirected to the login page.
> 
> Additionally, my session still persists. So even though I can't return to the election overview page by using the browser's back-button, I can return to it by entering the right url in the browser's address bar.

This is fixed in 5559f327b188e406e3e9d88efc3c952b2580aad7. The buttons the admin should not have access to are now not shown to the admin.

Lionqueen94:
> If I delete all polling stations for an election, the status of the session does return to "Voorbereiden", but the number of polling stations on the election overview page (e.g. /elections/2) still shows the old number of polling stations.

This ba6319e6cf49b9028b9d9c2f0d59428e162447ca seems to fix it. I couldn't reproduce it with deleting polling stations but could reproduce it with adding polling stations. The number seems to always be correct now, please retest :pray:

jschuurk-kr:
If I create an eletion without any polling stations, the election status overview (/elections/2) does not clearly tell me what my next step should be:
image

@jorisleker Should there be a button to add polling stations in the card of the session? Or should the "Stembureaus" row in  "Over deze verkiezing" be highlighted?

jschuurk-kr:
Probably not important enough to fix:
1. Create an election with a single polling station.
2. As coordinator go to the election status overview. It will have the button "Start steminvoer".
3. As admin remove the polling station.
4. As coordinator click "Start steminvoer".

Result: the coordinator ends up on the polling station status overview with the election in status "Voorbereiden". The page shows the text "Er zijn nog geen stembureaus toegevoegd voor deze verkiezing."

Arguably the coordinator should get an error message when they click "Start steminvoer", because the election does not progress to the next phase.

jschuurk-kr:
> > If I delete all polling stations for an election, the status of the session does return to "Voorbereiden", but the number of polling stations on the election overview page (e.g. /elections/2) still shows the old number of polling stations.
> 
> This [ba6319e](https://github.com/kiesraad/abacus/commit/ba6319e6cf49b9028b9d9c2f0d59428e162447ca) seems to fix it. I couldn't reproduce it with deleting polling stations but could reproduce it with adding polling stations. The number seems to always be correct now, please retest 🙏

Can't reproduce it anymore. :confetti_ball:

About the deleting: difference might be that I was testing with multiple accounts?

jorisleker:
> If I create an eletion without any polling stations, the election status overview (/elections/2) does not clearly tell me what my next step should be: image
> 
> @jorisleker Should there be a button to add polling stations in the card of the session? Or should the "Stembureaus" row in "Over deze verkiezing" be highlighted?

@Lionqueen94 I've added an alert for this case (first session, no polling stations) with a link to manage manage polling stations. See [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12389-83905&t=Bf8FAvSSKAe1GhAI-11)

backend/tests/shared/mod.rs
```diff
@@ -216,6 +219,17 @@ pub async fn create_result_with_non_example_data_entry(
...
 
+pub async fn change_status_committee_session(
+    pool: SqlitePool,
+    committee_session_id: u32,
+    status: CommitteeSessionStatus,
+) -> CommitteeSession {
+    CommitteeSessions::new(pool.clone())
+        .change_status(committee_session_id, status)
+        .await
+        .unwrap()
```
praseodym: In integration tests we should be using API endpoints, not calling the repository directly

backend/tests/election_integration_test.rs
```diff
@@ -194,11 +194,18 @@ async fn test_election_details_not_found(pool: SqlitePool) {
 }
 
 #[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_2", "users"))))]
-async fn test_election_pdf_download(pool: SqlitePool) {
-    let addr = serve_api(pool).await;
+async fn test_election_pdf_download_works(pool: SqlitePool) {
+    let committee_sessions_repo = CommitteeSessions::new(pool.clone());
```
praseodym: Same here: we shouldn't be using the repository in integration tests

backend/tests/polling_station_integration_test.rs
```diff
@@ -44,13 +45,20 @@ async fn test_polling_station_listing(pool: SqlitePool) {
     )
 }
 
-#[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_2", "users"))))]
-async fn test_polling_station_creation(pool: SqlitePool) {
-    let addr = serve_api(pool).await;
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_6", "users"))))]
+async fn test_polling_station_creation_for_committee_session_with_created_status(pool: SqlitePool) {
+    let committee_sessions_repo = CommitteeSessions::new(pool.clone());
```
praseodym: And same here (and other instances in this file): we shouldn't be using the repository in integration tests

Lionqueen94:
> > Admins and coordinators can reach /elections/2/status before "Start steminvoer" has been clicked. The status page then has "Klaar voor invoer" in the top-right corner, which is ~incorrect~ correct, but confusing to me. Because "invoer" can not start until "Start steminvoer" has been clicked.
> 
> @jorisleker shall I also add a "Start steminvoer" button to the `ElectionStatusPage` next to the "Klaar voor invoer" status?

Will add this in issue #1604

Lionqueen94:
> > If I create an eletion without any polling stations, the election status overview (/elections/2) does not clearly tell me what my next step should be: image
> > @jorisleker Should there be a button to add polling stations in the card of the session? Or should the "Stembureaus" row in "Over deze verkiezing" be highlighted?
> 
> @Lionqueen94 I've added an alert for this case (first session, no polling stations) with a link to manage manage polling stations. See [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12389-83905&t=Bf8FAvSSKAe1GhAI-11)

Added in 05bed368bfd91a88bc4284268bbfa705422b2460

backend/tests/shared/mod.rs
Lionqueen94: Oeh yes that's a big mistake of mine, corrected everywhere in 0c4f06e510794dc7ca3b0c299880da55a24a81f1
Thanks!

backend/tests/election_integration_test.rs
Lionqueen94: Oeh yes that's a big mistake of mine, corrected everywhere in 0c4f06e510794dc7ca3b0c299880da55a24a81f1
Thanks!

backend/tests/polling_station_integration_test.rs
Lionqueen94: Oeh yes that's a big mistake of mine, corrected everywhere in 0c4f06e510794dc7ca3b0c299880da55a24a81f1
Thanks!

praseodym (approved):
lgtm!

PR #1778, opened 2025-07-14
Add counting method to election structs

### LINKED ISSUE #1772: Add DSO/CSO to data model for elections

rnijveld:
Fixes #1772 

- Backend structs updated with CSO/DSO counting methods. 
- Should not result in any visible change in the frontend yet
- All elections are currently made with CSO counting method


praseodym:
@rnijveld linting fails, I think you need to add the counting method to some test/mock data

PR #1794, opened 2025-07-17
move button to create polling station to the left

jschuurk-kr:
### Scope
This makes the position of the button consistent with the elections and user management pages.

This was [discussed with Joris](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=1121-9960#1325271076), who has updated the design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=5442-50666&t=anpcKMQSFRViMZH4-0

### Question
The PollingStationListPage used `Toolbar.Section`. The LogsHomePage used `ToolbarSection`. They are the same thing because Toolbar does `Toolbar.Section = ToolbarSection;`. Which of these is the prefered way?

Because either `Toolbar.Section = ToolbarSection;` needs to be removed, or the LogsHomePage needs to use `Toolbar.Section`.

### Pictures
image

image

image

### DIFF
--- a/frontend/src/components/ui/Toolbar/Toolbar.stories.tsx
+++ b/frontend/src/components/ui/Toolbar/Toolbar.stories.tsx
@@ -3,7 +3,7 @@ import type { Meta, StoryFn, StoryObj } from "@storybook/react-vite";
 import { IconPlus } from "@/components/generated/icons";
 
 import { Button } from "../Button/Button";
-import { Toolbar, ToolbarSection } from "./Toolbar";
+import { Toolbar } from "./Toolbar";
 
 type Props = {
   pos: "start" | "center" | "end";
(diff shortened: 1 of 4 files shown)


praseodym:
> The PollingStationListPage used `Toolbar.Section`. The LogsHomePage used `ToolbarSection`. They are the same thing because Toolbar does `Toolbar.Section = ToolbarSection;`. Which of these is the prefered way?
> 
> Because either `Toolbar.Section = ToolbarSection;` needs to be removed, or the LogsHomePage needs to use `Toolbar.Section`.

We should be using `Toolbar.Section` everywhere, that's also how we do this for e.g. `Table`. I've fixed it in 557274098c5a8897bd2f41f8ba89e814ed47a890.

Lionqueen94 (approved):

PR #1803, opened 2025-07-19
handle all non available states in PollingStationChoiceForm

### LINKED ISSUE #1707: PollingStationChoiceForm's handleSubmit() does not account for all polling station non-available states
On the PollingStationChoiceForm a user can select a polling station by entering the polling station number and clicking "Beginnen". However, a polling station can be in a state where the user is not allowed to start data entry. (These polling stations are also filtered out from the list of polling stations lower on the page.)

For some of those states, when the user clicks "Beginnen" we show an alert with a state-specific text:

Image

However, for two of these states we don't: `SecondEntryNotAllowed` and `HasErrors`. For those states we show a generic warning banner at the top of the page, saying "Je kan stembureau 1 niet invoeren. Er is een ongeldige actie uitgevoerd." These states should also result in a specific error message after the user clicks "Beginnen".

jschuurk-kr:
closes #1707

### Scope

- Make PollingStationChoiceForm handle all error scenarios when a user clicks "Beginnen" after inputting the number of a polling station they are not allowed to do data entry for.
- Parametrize the tests of these error scenarios
- Get rid of `extendedPollingStationMockData` (PollingStationMockData.ts), because only one polling station left in it

### Question
There are some inconsistencies in the messages we show to the user, see `frontend/src/i18n/locales/nl/polling_station_choice.json`:
1. After submitting a polling station number (so clicking "Beginnen"), we sometimes put the polling station name in the alert and sometimes we don't.
2. In the PollingStationChoice Selector the polling station name is in a `<strong>`, in the PollingStationChoiceForm it is not. (Requires a bigger change than just adding the `<strong>`.)

Should we address this this week in a separate PR, or is this good enough as-is? cc @jorisleker

### DIFF
--- a/frontend/src/features/data_entry_home/components/PollingStationChoiceForm.test.tsx
+++ b/frontend/src/features/data_entry_home/components/PollingStationChoiceForm.test.tsx
@@ -104,8 +104,6 @@ describe("Test PollingStationChoiceForm", () => {
 
     test.each([
       // TODO add tests for:
-      // "first_entry_in_progress" by other user => invalid, PollingStationUserStatus.InProgressOtherUser
-      // "first_entry_has_errors" => not handled on submit, PollingStationUserStatus.HasErrors
       // "second_entry_not_started" 1st entry by same user => not handled on submit, PollingStationUserStatus.SecondEntryNotAllowed
       {
         testDescription: "invalid",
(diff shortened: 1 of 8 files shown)


Lionqueen94 (approved):

oliver3 (approved):

jorisleker:
@jschuurk-kr could you add screenshots of the messages you are referring to? Having a bit of a hard time finding out what exactly your comment is about 🙏

jschuurk-kr:
> @jschuurk-kr could you add screenshots of the messages you are referring to? Having a bit of a hard time finding out what exactly your comment is about 🙏

@jorisleker That's a lot of work. :sweat_smile:

As a compromise, here's the parametrized test where you can see all the messages: https://github.com/kiesraad/abacus/blob/d47b500a75f74461e7d5c4c418946664e4b763aa/frontend/src/features/data_entry_home/components/PollingStationChoiceForm.test.tsx#L105
`selectorFeedback` is what you see after inputting a polling station number. `submitFeedback` is what you see after clicking "Beginnen".

Here's a screenshot for the one instance where we use `<strong>`:
image

jschuurk-kr:
Merging this PR because the [open discussion point](https://github.com/kiesraad/abacus/pull/1803#issuecomment-3096697653) will not affect this PR. (If changes are needed, those will be done in a new PR.)

PR #1810, opened 2025-07-21
Throw and handle forbidden errors

### LINKED ISSUE #1798: Use 403 Forbidden instead of 401 Unauthorized when credentials are valid
The backend currently always returns 401 Unauthorized when visiting a page that the user has no permission for (e.g. election management when logged in as a typist). Instead, we should return 403 Forbidden, so that a better error message can be shown.

From https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/401:

> A 401 Unauthorized is similar to the [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/403) response, except that a 403 is returned when a request contains valid credentials, but the client does not have permissions to perform a certain action.

We should also handle 403 Forbidden in the frontend with an error screen instead of redirecting to the login page.

marlonbaeten:
Fixes #1798

The backend will now return a 403 Forbidden error if the user is logged in but does not have the correct role.
The frontend will show the error instead of redirecting the user to the login page.

Error page:

image

### DIFF
--- a/frontend/src/components/error/FatalErrorPage.tsx
+++ b/frontend/src/components/error/FatalErrorPage.tsx
@@ -16,12 +16,12 @@ interface FatalErrorPageProps {
   error?: Error;
 }
 
-export function FatalErrorPage({ title, message, code, reference, error }: FatalErrorPageProps) {
+export function FatalErrorPage({ title = "error.title", message, code, reference, error }: FatalErrorPageProps) {
   return (
     <AppLayout>
       {/* Show NavBar for / to avoid call to useElection outside ElectionProvider */}
       <NavBar location={{ pathname: "/" }} />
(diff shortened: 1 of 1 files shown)


praseodym:
I think the 403 code should also be added to the `responses` in the `utoipa::path` attributes.

backend/src/authentication/api.rs
```diff
@@ -467,7 +473,7 @@ mod tests {
 
     use crate::authentication::{SECURE_COOKIES, api::set_default_cookie_properties};
 
-    #[test(sqlx::test)]
```
marlonbaeten, line 470: Since the database is not used for this test, this can be a tokio test (normal async test) instead of an sqlx test.

cikzh (approved):

PR #1822, opened 2025-07-22
Use watch channel instead of a mpsc channel for airgap detection updates

marlonbaeten:
Possible fix for a flaky test: https://github.com/kiesraad/abacus/actions/runs/16420154869/job/46396218293?pr=1799

Replace the mpsc channel for a watch channel, which should be more reliable when communicating from an sync to an async context.

### DIFF
--- a/backend/src/airgap/detect.rs
+++ b/backend/src/airgap/detect.rs
@@ -4,7 +4,7 @@ use std::{
     sync::{Arc, RwLock, atomic::AtomicBool},
     time::{Duration, Instant},
 };
-use tokio::sync::mpsc::{self, Sender};
+use tokio::sync::watch::{self, Sender};
 use tracing::{error, info, trace, warn};
 
 use crate::audit_log::{AuditEvent, AuditLog};
@@ -59,7 +59,7 @@ impl AirgapDetection {
     /// Starts the airgap detection in a background task.
     /// It will periodically check for airgap violations by attempting to connect to a known server.
     pub async fn start(pool: SqlitePool) -> AirgapDetection {
-        let (tx, mut rx) = mpsc::channel::<AirGapStatusChange>(1);
+        let (tx, mut rx) = watch::channel(AirGapStatusChange::AirGapViolationResolved);
 
         let airgap_detection = AirgapDetection {
             enabled: true,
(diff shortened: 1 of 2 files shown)


praseodym:
> Replace the mpsc channel for a watch channel, which should be more reliable when communicating from an sync to an async context.

Can you give some pointers to documentation that describes it's more reliable or more suitable for this use case?

marlonbaeten:
> Can you give some pointers to documentation that describes it's more reliable or more suitable for this use case?

In general, this channel is less complex, since it does not have to deal with capacity or back-pressure etc.

The documentation for the watch channel explicitly states that the sender and receiver are thread safe: https://docs.rs/tokio/latest/tokio/sync/watch/index.html#thread-safety 

The documentation for the mpsc channel implies that in a non-tokio runtime (like our detection thread) there might be a risk of starvation: https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html#multiple-runtimes

marlonbaeten:
> The only cause that I can think of is that the runtime is shut down, but the os thread is still running briefly. So if the perform_detection is called at exactly that moment, we get the panic because the channel is closed and the test fails.

Note we never panic when the channel is closed - we only log an error.

The reason for a separate thread for air-gap detection is that we use standard library sync DNS resolve method. Doing this in a tokio context would require a DNS library dependency.  Also, this thread should exactly run for the lifetime of the application, so creating any shutdown mechanisms are not necessary.

marlonbaeten:
I think I have found the reason for the flaky test: it's a race condition in the test itself. The airgap detection only registers the "last check time" after sending the update over the channel. The test immediately checks the "last check time" after an airgap violation occurred.

I'll add a fix for that race condition in this PR.

marlonbaeten:
The change to the watch channel is still an improvement over the mpsc channel.

marlonbaeten:
I also increased the timeout to 10 seconds for this test, since a bad network connection could cause a timeout.

praseodym:
> The reason for a separate thread for air-gap detection is that we use standard library sync DNS resolve method. Doing this in a tokio context would require a DNS library dependency.

What about https://docs.rs/tokio/latest/tokio/net/fn.lookup_host.html?

marlonbaeten:
> > The reason for a separate thread for air-gap detection is that we use standard library sync DNS resolve method. Doing this in a tokio context would require a DNS library dependency.
> 
> What about https://docs.rs/tokio/latest/tokio/net/fn.lookup_host.html?

I was not aware that tokio included a wrapper. Thanks!

That could work - but we would also have to add additional logic for TCP timeouts.

I don't really see the advantage of refactoring the thread to a tokio task at this point, but maybe we can discuss this during a meeting.

bramtweedegolf (approved):
I think the change `assert!(violation_detected, "Airgap detection did not run");` is a great improvement for the test.

Using the watch instead of mpsc channel may not be related to the problem, but is probably more appropriate here, so I am fine with merging that change too.

I would be fine with merging this fix without the refactor, and maybe discuss the refactor at some point in the future as a nice-to-have.

praseodym:
> I would be fine with merging this fix without the refactor, and maybe discuss the refactor at some point in the future as a nice-to-have.

Agreed!

praseodym (approved):

Issue #1731, opened 2025-07-07
Username missing in alert


jorisleker:
### Tested on:
- latest release build ([38fe4595](https://github.com/kiesraad/abacus/commit/38fe45954e5bfb8e291a8a84424c6b9b235283df))

### Reproduction steps
- Create a data entry, make no mistakes (use invoerder1)
- Create a second data entry for the same polling station, make sure there is an error (use invoerder 2)
- Log in as coordinator
- Resolve the difference by picking the second data entry
- Observed: the alert shows 'Gebruiker' 
- Expected: the alert shows the name of invoerder 2

<img width="1467" alt="Image" src="https://github.com/user-attachments/assets/212f1c53-7c03-4ae8-98da-5aff132e16df" />

Note; in the design, the coordinator is immediately redirected to the 'foutopsporingsmodus' if the remaining data entry contains an error. The current implementation redirects to the statusoverview page (but I think fixing this is part of #1705)

PR #1827, opened 2025-07-22
Resolve errors and resolve differences messages using useMessages

oliver3:
- Implements https://github.com/kiesraad/abacus/issues/1705
- Fixes https://github.com/kiesraad/abacus/issues/1731

Changes:
- Show message (alert) coming from resolve differences, on resolve errors page
- Show messages from resolve differences and resolve errors on election status page using `pushMessage()` and `<Messages>`
  - Fix showing correct typist name corresponding to data entry that was selected in resolve differences
- Update vitest tests
- Add new message to Playwright test, rest is still working!

To test:
- Create two data entries with differences, see as a coordinator that after resolving differences, the message on the status page corresponds to the choice made
- Create two data entries with differences and an error in the second data entry. Choose the second data entry as coordinator on the resolve differences and see that the message is shown on the resolve errors page that comes afterwards.
- Try to reproduce https://github.com/kiesraad/abacus/issues/1731

### DIFF
--- a/frontend/e2e-tests/page-objects/election/ResolveErrorsPgObj.ts
+++ b/frontend/e2e-tests/page-objects/election/ResolveErrorsPgObj.ts
@@ -1,6 +1,7 @@
 import { type Locator, type Page } from "@playwright/test";
 
 export class ResolveErrorsPgObj {
+  readonly alertDifferencesResolved: Locator;
   readonly title: Locator;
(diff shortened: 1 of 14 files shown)


Lionqueen94 (approved):

praseodym (approved):
lgtm, couldn't reproduce the missing username in the alert

PR #1830, opened 2025-07-22
Download empty report Na 31-2 Bijlage 1

### LINKED ISSUE #1825: Knowledge sharing CSO report templates during first committee session

marlonbaeten:
Fixes #1825

## Changes

- Adds a section to the election detail page "Lege modellen voor deze verkiezing"
- Adds a download button for "Na 31-2 Bijlage 1" (one PDF per polling station in a ZIP file)
- Refactor of the `pdf_gen` module to better handle multiple models and better async support
- Added a `zip` module to reuse the boilerplate code for creating a ZIP file response
- Created an data input struct for "Na 31-2 Bijlage 1"

## Performance

- During development downloading a ZIP with 100 PDF files takes around 4000 ms (using external typst)
- In a development build with `embed-typst` the same ZIP takes about 1700 ms te generate
- In a production build with `embed-typst` it takes about 1600 ms
- Downloading a ZIP with 3 PDF's takes about 60 ms

## Testing

- Run the application
- Login as a coordinator
- Navigate to `/elections/1` 
- Click the link at the bottom of the page to download a ZIP
- Validate that the zip contains valid PDF files with the correct contents for that polling station

## Screenshot

image


backend/test_na_31_2_bijlage1.zip
praseodym, line 1: I don't think we need to commit this file, and the test should probably write it to a temp dir (or keep it in memory)

frontend/e2e-tests/page-objects/election/ElectionHomePgObj.ts
```diff
@@ -0,0 +1,9 @@
+import { Locator, Page } from "@playwright/test";
+
+export class ElectionHome {
+  readonly downloadBijlage1: Locator;
+
+  constructor(protected readonly page: Page) {
+    this.downloadBijlage1 = page.getByRole("cell", { name: "Na 31-2 Bijlage" });
```
praseodym:
```suggestion
    this.downloadBijlage1 = page.getByRole("cell", { name: "Na 31-2 Bijlage 1" });
```

praseodym:
I think we should be able to [stream the ZIP file](https://docs.rs/zip/latest/zip/write/struct.ZipWriter.html#method.new_stream) so that we don't have to keep all PDF files in memory, but that would require structuring the code differently so I'm not sure if it's worth it.

marlonbaeten:
> I think we should be able to [stream the ZIP file](https://docs.rs/zip/latest/zip/write/struct.ZipWriter.html#method.new_stream) so that we don't have to keep all PDF files in memory, but that would require structuring the code differently so I'm not sure if it's worth it.

I'd be happy to implement this in a new PR 👍🏻

praseodym:
@marlonbaeten this branch has conflicts

PR #1831, opened 2025-07-22
Always use SQLite in WAL mode

praseodym:
Always use SQLite in [WAL mode](https://sqlite.org/wal.html): "WAL is significantly faster in most scenarios."

SQLx doesn't use WAL mode by default, but the [`journal_mode` function](https://docs.rs/sqlx/0.8.6/sqlx/sqlite/struct.SqliteConnectOptions.html#method.journal_mode) can be used to enable it. Interestingly, the SQLx documentation mentions that "any commands in `sqlx-cli` which create a SQLite database will create it in WAL mode.", so only if the database was created using the SQLx CLI, Abacus would already use WAL mode.

A k6 load test with 1000 parallel users[^1] shows a huge performance boost when using WAL mode: p99 http request duration decreases from 1s to 5ms. This is also clearly noticeable when browsing the election status page during the load test.

Without WAL:
no-wal

With WAL:
wal

[^1]: on my laptop on battery 🔋

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -10,7 +10,10 @@ use axum::{
     serve::ListenerExt,
 };
 use hyper::http::{HeaderName, HeaderValue, header};
-use sqlx::{SqlitePool, sqlite::SqliteConnectOptions};
+use sqlx::{
+    SqlitePool,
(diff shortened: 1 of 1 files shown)


oliver3 (approved):
🚀

PR #1840, opened 2025-07-24
Start, pause, resume and finish data entry

Lionqueen94:
…ttee session

### DIFF
--- a/frontend/src/components/ui/AppLayout/AppLayout.module.css
+++ b/frontend/src/components/ui/AppLayout/AppLayout.module.css
@@ -21,6 +21,9 @@
         gap: 0.5rem;
         line-height: 1.75rem;
         align-items: center;
+        a:visited {
+          color: var(--link-default);
+        }
       }
     }
     section:global(.smaller-gap) {
--- a/frontend/src/features/election_management/components/CommitteeSessionCard.module.css
+++ b/frontend/src/features/election_management/components/CommitteeSessionCard.module.css
@@ -1,5 +1,5 @@
 .card {
-  width: 25.33rem;
+  width: 26rem;
   background-color: var(--base-white);
   padding: 1.75rem 2rem;
   display: flex;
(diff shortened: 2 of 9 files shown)


jorisleker (dismissed):

frontend/src/features/election_status/components/ElectionStatusPage.test.tsx
```diff
@@ -52,25 +63,180 @@ async function renderPage() {
...
+      CommitteeSessionStatusChangeRequestHandler,
       ElectionListRequestHandler,
       ElectionRequestHandler,
       ElectionStatusRequestHandler,
       UserListRequestHandler,
     );
   });
 
-  test("Finish input not visible when data entry is in progress", async () => {
+  test("Page render when committee session status is created for coordinator", async () => {
```
oliver3: It looks like this is completely the same for coordinator and admin, maybe we can use `test.each` to test both roles? Maybe this also helps for the other tests in this file.

frontend/src/features/election_status/components/ElectionStatusPage.test.tsx
```diff
@@ -126,4 +437,42 @@ describe("ElectionStatusPage", () => {
 
     await expectForbiddenErrorPage();
   });
+
+  test("Shows error page when election status change call returns an error", async () => {
```
oliver3, line 409: I think that some tests can be more focused, e.g. this test is to check if errors are handled correctly and does not have to check the rendering of alerts and buttons etc

frontend/src/features/election_status/components/ElectionStatusPage.test.tsx
Lionqueen94: I changed 2 tests for both users into test.each in 53071edf8cee7794887b830e10c8eab41e2ae260

frontend/src/features/election_status/components/ElectionStatusPage.test.tsx
Lionqueen94: In 53071edf8cee7794887b830e10c8eab41e2ae260 I removed a few checks from this test, except one, because the page still needs to await the render to make sure the button can be found

jschuurk-kr:
oops! [commented on the issue](https://github.com/kiesraad/abacus/issues/1604#issuecomment-3149957460) instead of the PR...

marlonbaeten (approved):
👍🏻

Issue #1815, opened 2025-07-21
Frontend: Change number of eligible voters


Lionqueen94:
- Add frontend page and link it --> Nearly the same as #1633, 
  - make a reusable component with a configurable button label, field description and endpoint

https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=9529-61657&m=dev

Issue #1633, opened 2025-06-17
Frontend page to set number of voters


bramtweedegolf:
Depends on #1714 
 

- Number of voters screen ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=10100-124594&m=dev)) --> Nearly the same as #1714, make a reusable component with a configurable button label and field description

PR #1841, opened 2025-07-25
Add form and page to edit number of voters for a committee session

oliver3:
- Implements https://github.com/kiesraad/abacus/issues/1815
- Move all routes for election management to `features/election_management/routes.tsx`
- Add form to edit number of voters, reusable for https://github.com/kiesraad/abacus/issues/1633
- Add page that uses the form and submits to backend and navigate back
- Update `<InputField>` with a `fieldWidth` that does not wrap the label for a "narrow" input


frontend/src/features/election_management/components/NumberOfVotersForm.tsx
```diff
@@ -0,0 +1,53 @@
...
+              label={t("number_of_voters")}
+              hint={hint}
+              fieldWidth="full-field-with-narrow-input"
+              numberInput
+              defaultValue={defaultValue}
+            />
+
+            <FormLayout.Controls>
+              <Button size="lg" type="submit">
+                {t("save")}
```
Lionqueen94: The button text needs to be customizable too, since it's "next" for the import flow, see https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=10100-124743&m=dev

frontend/src/features/election_management/components/NumberOfVotersForm.tsx
```diff
@@ -0,0 +1,53 @@
+import { FormEvent } from "react";
```
Lionqueen94: This form needs to be moved to the components folder to make is accessible for the election import flow page as well. Maybe it can be added to `/src/components/committee_session`?

frontend/src/features/election_management/components/NumberOfVotersForm.tsx
oliver3: Thanks, added it to the props.

frontend/src/features/election_management/components/NumberOfVotersForm.tsx
oliver3: Moved it to `components/election` as it will be used by the features `election_management` and `election_create`. I've also added a story with test for the form itself.

marlonbaeten (approved):
Works!

Issue #1487, opened 2025-05-13
Bug: Formatted NumberInput does not clear in playwright tests


Lionqueen94:
Fill call should already clear the field, but it doesn't, a specific clear call doesn't either. Happens for number >= 1000 (formatted with the `.`)

![Image](https://github.com/user-attachments/assets/afad5a24-df77-49b7-b436-1525b301fe69)

oliver3:
This is a Playwright on chromium only bug, and really fixing would mean changing how the NumberInput works. A workaround is mentioned in the NumberInput e2e test itself, closing this for now.

Issue #360, opened 2024-09-23
Experiment: replace '.' with thin space in numberinput


jorisleker:
We currently use `.`'s as thousands separators in numerical inputs. Mainly to improve legibility. But it also results in some counter intuitive behavior when editing values in those inputs because we ignore the `.` when the user is typing (jumping over it while using backspace and delete, but not when using the arrow keys. 

We might make it bit more inituitive whilst still achieving our legibility goal by replacing the `.` with a unicode `three-per-em-space` character (U+2004) ` `.  (note: this char is rendered the same with as a regular space in GitHub's monospaced font, but Space Grotesk renders it correctly.

Comparison:
`12.345`
`12 345`

Screenshots:
![Scherm­afbeelding 2024-09-23 om 09 16 02](https://github.com/user-attachments/assets/7506552e-4ad6-4dde-914b-6953ee6e4fc5)

![Scherm­afbeelding 2024-09-23 om 09 16 46](https://github.com/user-attachments/assets/e06b7db8-b49f-493b-8731-630ddc7dd0b8)

PR #1842, opened 2025-07-25
NumberInput formatting using overlay

oliver3:
### Context
At the moment, the NumberInput shows a formatted number when the input has no focus by changing the value to the formatted number, and changing it back to unformatted when receiving focus. 

This causes extra complexity for things like keeping the selection and using the value of this NumberInput (it has to be deformatted for each usage). It also causes timing issues in the tests (see e.g. [Bug: Formatted NumberInput does not clear in playwright tests](https://github.com/kiesraad/abacus/issues/1487)).

Also, I strongly feel that how the number is shown to the user is an implementation detail and that the rest of the code should be unaware of this formatting. This will also make changing it e.g. for [Experiment: replace '.' with thin space in numberinput](https://github.com/kiesraad/abacus/issues/360) much easier.

### Change
In this PR, the formatted number is shown with an overlay over the input when it does not have focus, and removes that overlay when it receives focus. With this way of showing the number, the value of the input does not have to be changed. Automated tests that do want to check if the formatting is shown, have to assert the overlay's text instead of the input value, and all application code should now be unaware of formatting and deformatting numbers.

### Testing
This NumberInput is part of the data entry and should be very thoroughly tested, on different browsers and with sorts of user interaction (keyboard, mouse, copy paste, ...?)

The NumberInput is also used in the polling station form, and will be used for number of voters when https://github.com/kiesraad/abacus/pull/1841 will be merged.

### DIFF
--- a/frontend/src/components/ui/InputGrid/InputGrid.module.css
+++ b/frontend/src/components/ui/InputGrid/InputGrid.module.css
@@ -42,19 +42,24 @@
         border-right: 1px solid var(--bg-gray-darkest);
         padding: 0;
         input {
-          text-align: right;
           width: 100%;
           height: 4.5rem;
           border: 2px solid transparent;
-          padding: 1rem 1.5rem;
-          font-size: var(--font-size-body);
-          font-style: normal;
           &:focus,
           &:focus-visible {
             outline: none;
             border-color: var(--border-color-focus);
           }
         }
+
+        input,
+        :global(.formatted-overlay) {
+          text-align: right;
+          padding: 1rem 1.5rem;
+          font-size: var(--font-size-body);
+          font-style: normal;
+        }
+
         &.readOnly span {
           display: block;
           height: 4.5rem;
--- a/frontend/src/components/ui/NumberInput/NumberInput.module.css
+++ b/frontend/src/components/ui/NumberInput/NumberInput.module.css
@@ -1,0 +1,23 @@
+.container {
+  position: relative;
+
+  /* Global to be able to give custom styling */
+  :global(.formatted-overlay) {
+    position: absolute;
+    background: white;
+    pointer-events: none;
+
+    /* Same as input border */
+    inset: 2px;
+
+    /* Same as reset.css input */
+    padding: 1px 2px;
+
+    display: flex;
+    > span {
+      /* Vertical center, horizontal grow to be able to align right */
+      align-self: center;
+      flex-grow: 1;
+    }
+  }
+}
--- a/frontend/src/components/ui/NumberInput/NumberInput.stories.tsx
+++ b/frontend/src/components/ui/NumberInput/NumberInput.stories.tsx
@@ -9,23 +9,26 @@ export const DefaultNumberInput: StoryObj = {
   },
   play: async ({ canvas, userEvent, step }) => {
     const input = canvas.getByRole("textbox");
+    const FORMATTED_OVERLAY = "test-formatted-overlay";
 
     await step("Test number formatting", async () => {
       // Test number formatting
       await expect(input).toBeVisible();
-      await expect(input).toHaveValue("12.300");
+      await expect(await canvas.findByTestId(FORMATTED_OVERLAY)).toHaveTextContent("12.300");
     });
 
     await step("Test focus removes formatting", async () => {
       // Test focus removes formatting
       await userEvent.click(input);
-      await expect(input).toHaveValue("12300");
+      await expect(input).toHaveFocus();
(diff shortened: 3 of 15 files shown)


jorisleker:
Good improvement I think; works as expected for normal interactions.

One thing I noticed, is that it affects copy/pasting the entire table, because both the input's value and the overlay are copied (but only when pasting to a tool that does not support rich text/formatting). Not sure whether this is a valid use case / worth fixing; data from data entry forms is not supposed to be copied in this manner. 

Example when copy pasting to a textarea:
```
Toegelaten kiezers en uitgebrachte stemmen
B1-3.1 en 3.2
Veld	Geteld aantal	Omschrijving
A	
15
15
Stempassen
B	
1555
1.555
Volmachtbewijzen

250000000
250.000.000
Totaal toegelaten kiezers
E	
250000000
250.000.000
Stemmen op kandidaten
F	
Blanco stemmen
G	
Ongeldige stemmen
```

Example when copy pasting to a text editor:
Scherm­afbeelding 2025-07-25 om 11 38 55

oliver3:
> Good improvement I think; works as expected for normal interactions.
> 
> One thing I noticed, is that it affects copy/pasting the entire table, because both the input's value and the overlay are copied (but only when pasting to a tool that does not support rich text/formatting).

Thanks for testing!

The copy/paste was easily fixable with https://developer.mozilla.org/en-US/docs/Web/CSS/user-select so it now only copies the input value in your scenario. Note that this is a low effort fix because as you mention, the use case is quite exotic.

marlonbaeten (approved):
- Tested on Linux Chrome / Firefox.
- Keyboard acces / tabbing and text selection still works as intended

Nice improvement!

praseodym:
@oliver3 this branch has merge conflicts

frontend/src/components/election/NumberOfVotersForm.stories.tsx
```diff
@@ -21,7 +21,7 @@ export const Form: Story = {
   },
   play: async ({ args, canvas, userEvent }) => {
     await expect(canvas.getByText(args.instructions)).toBeInTheDocument();
-    await expect(canvas.getByLabelText("Aantal kiesgerechtigden")).toHaveValue("2.000");
+    await expect(canvas.getByRole("textbox", { name: "Aantal kiesgerechtigden" })).toHaveValue("2000");
```
oliver3, line 24: `getByLabelText()` does not ignore the overlay (which has aria-hidden) but `getByrole("textbox", {name})` does. See also https://github.com/testing-library/dom-testing-library/issues/618

praseodym (approved):
lgtm! It looks like this also fixes #1861, can you confirm that?

jschuurk-kr:
> lgtm! It looks like this also fixes #1861, can you confirm that?

Looks to me like both active input and overlay are --gray-700, i.e. #344054, not --gray-600. While [the design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=8663-32055&m=dev#1359660558) has black for the active input and --gray-600 for the blurred ones.

The overlayed input fields are black, i.e. #000000, according to the "computed" tab, but the devtools also claim the element inherits `color: var(--text-color-body);` from `body`, which is --gray-700?

jschuurk-kr:
> The copy/paste was easily fixable with https://developer.mozilla.org/en-US/docs/Web/CSS/user-select so it now only copies the input value in your scenario. Note that this is a low effort fix because as you mention, the use case is quite exotic.

I'm getting different results, depending on the browser I copy from and where I copy to:
- Firefox does not copy the values. Chromium does.
- Pasting to Sublime Text or LibreOffice as unformatted text includes the values.
- Pasting to LibreOffice (normal paste) does not include the values.

(edit to clarify) "includes the values" means behaving correctly as in: the value is copy-pasted only once.

jschuurk-kr:
Update after looking at this with Oliver.

> Looks to me like both active input and overlay are --gray-700, i.e. #344054, not --gray-600. While [the design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=8663-32055&m=dev#1359660558) has black for the active input and --gray-600 for the blurred ones.

Correction: active input is correctly black. Overlay is --gray-700 instead of --gray-600.

> The overlayed input fields are black, i.e. #000000, according to the "computed" tab, but the devtools also claim the element inherits `color: var(--text-color-body);` from `body`, which is --gray-700?

This is a bug in firefox. The computed tab correctly shows the color coming from in `input > fieldtext`, but the element view does not show the value from body being overridden.

PR #1844, opened 2025-07-25
Improve data entry testability

### LINKED ISSUE #1820: Improve testability of data entry structure and mapping
Improve testability of data entry structure and mapping by making its types more generic, i.e. not dependant on the polling station results struct in the backend. This will save us from rewriting many tests every time we add new fields to the polling station results struct and allows us to easier test functionality like checkboxes and radio buttons that are not currently part of the polling station results struct.

praseodym:
- **Simplify data entry structure types**
  This makes the code more testable because we can use arbirtary types to
  test, not just `PollingStationResults` as used in the backend.
  
  Also removed arbitrary string values for radio subsections, because we
  only need boolean values.
  
  Resolves #1820

- **Use data entry structure in `isFormSectionEmpty`**


Lionqueen94 (approved):
Looks good to me

PR #1845, opened 2025-07-25
Streaming ZIP responses

marlonbaeten:
Merge #1830 first

This adds streaming ZIP support.

- PDF files are sent over a channel when generated
- The newly added `ZipStream` listens to this file channel and adds the files to a zip builder on the fly
- The zip builder writes its resulting bytes via a channel to a streamed (chunked) HTTP response

The advantage is that generated PDF files are not kept in memory, also the resulting ZIP file is not kept in memory, they are directly streamed to the client. This allows generating ZIP files with hundreds of PDF files while only ever keeping a single PDF file in memory.

### DIFF
--- a/backend/src/pdf_gen/embedded/mod.rs
+++ b/backend/src/pdf_gen/embedded/mod.rs
@@ -1,90 +1,132 @@
-mod world;
-
-use std::time::Instant;
-
-use tracing::{debug, info, warn};
-use typst::{diag::SourceDiagnostic, ecow::EcoVec};
+use std::{fmt::Debug, time::Instant};
+use tokio::{sync::mpsc::Sender, task::JoinHandle};
+use tracing::{debug, error, info, warn};
+use typst::{comemo, diag::SourceDiagnostic, ecow::EcoVec};
 use typst_pdf::{PdfOptions, PdfStandard, PdfStandards};
 
 use super::{PdfGenResult, models::PdfModel};
-use crate::APIError;
+use crate::{APIError, pdf_gen::models::PdfFileModel, zip::FileEntry};
 
-pub async fn generate_pdf(model: PdfModel) -> Result<PdfGenResult, APIError> {
+mod world;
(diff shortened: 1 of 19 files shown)


rnijveld (dismissed):

rnijveld (approved):

backend/src/pdf_gen/embedded/mod.rs
```diff
@@ -12,7 +12,12 @@ mod world;
 /// Generates a PDF using the embedded typst library.
 pub async fn generate_pdf(file_model: PdfFileModel) -> Result<PdfGenResult, APIError> {
     Ok(tokio::task::spawn_blocking(move || {
-        compile_pdf(&mut world::PdfWorld::new(), file_model.model)
+        let result = compile_pdf(&mut world::PdfWorld::new(), file_model.model);
+
+        // Evict the cache to free up memory + speed up next compile
+        comemo::evict(0);
```
praseodym, line 18: Why not make this part of `compile_pdf`?

marlonbaeten: If an error occurs anywhere in `compile_pdf` we still should clear the cache (an thus prevent memory leaks). This way we will always call evict.

marlonbaeten:
Also in this PR:

- Fix memory leak / performance issues in typst by clearing the cache after each PDF generation
- Set correct status when generating committee session
- Correctly handle errors curing PDF generation
- Remove unnecessary download file javascript code (previously used to fix the MSW)
- Created `PdfFileModel` to keep track of file name when creating a PDF

praseodym (approved):
lgtm! Creating a ZIP with 1000 polling station PDFs takes about a minute now (~60ms per PDF) and the user gets feedback through the browser's download panel.

I did notice that cancelling the download in the browser (or closing the browser entirely) does not cancel the PDF generation, but that's not a blocker.

PR #1849, opened 2025-07-28
Add DSO/CSO selection page

### LINKED ISSUE #1811: DSO/CSO selection frontend
- CSO or DSO selection screen ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12222-131106&t=apV5L65tdTg6KHp4-4)), note that DSO should be disabled for now

If the elections details screen ([Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=9529-61657&m=dev), #1714) is already implemented by the time this issue is implemented, the DSO/CSO choice should also be shown there.

cikzh:
Closes #1811 

Adds page to select the counting method. Note that DSO is not supported, so CSO is selected by default, DSO is disabled, and nothing is saved to the backend yet. 

Also updated the tests to include the page and specifically wrote a test that should fail once we un-disable the DSO option, as a reminder we should update the test there.

You can test this by going through the election creation process.


bramtweedegolf:
Looks good! Small issue: the page differs on two small points from the design: 

Screenshot_2025-07-28_15-48-53

marlonbaeten:
For other reviewers: not that the polling station numbers are sparse.

PR #1856, opened 2025-07-29
Handle abort download by client while streaming a zip

marlonbaeten:
When streaming a ZIP of PDF files to a client, the client could cancel the download. This change also stop PDF from being generated when a stream is cancelled.

### DIFF
--- a/backend/src/pdf_gen/external.rs
+++ b/backend/src/pdf_gen/external.rs
@@ -44,7 +44,9 @@ pub fn generate_pdfs(
         }
 
         if let Err(e) = sender.send(None).await {
-            error!("Failed to send finish signal: {e:?}");
+            error!(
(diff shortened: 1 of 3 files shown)


backend/src/pdf_gen/external.rs
```diff
@@ -37,7 +37,7 @@ pub fn generate_pdfs(
             );
 
             if let Err(e) = sender.send(Some((file_name, content.buffer))).await {
-                error!("Failed to send PDF: {e:?}");
+                error!("Failed to send PDF: {e} - the client might have closed the connection");
```
praseodym, line 40: I don't get this error but `ERROR abacus::pdf_gen::embedded: Error generating PDF: ChannelClosed`

praseodym (approved):
lgtm, generation is cancelled when cancelling the download

PR #1859, opened 2025-07-29
Perform air gap detection using async (tokio) primitives

marlonbaeten:
Use tokio's `TcpStream::connect` and `lookup_host` to perform air gap detection on the server side.

This simplifies the code, since we do not need a special channel to sent updates from a thread to the async context. We can directly log an audit log event from the detection task.

We also have explicit timeouts for the checks, which, in the case of the DNS lookup, previously was implicit or dictated by the OS.

### DIFF
--- a/backend/src/airgap/detect.rs
+++ b/backend/src/airgap/detect.rs
@@ -4,8 +4,8 @@ use std::{
     sync::{Arc, RwLock, atomic::AtomicBool},
     time::{Duration, Instant},
 };
-use tokio::time::timeout;
-use tracing::{error, info, trace, warn};
+use tokio::{task::JoinSet, time::timeout};
+use tracing::{debug, error, info, trace, warn};
 
 use crate::audit_log::AuditEvent;
 
@@ -19,7 +19,7 @@ pub struct AirgapDetection {
 
 const SECURE_PORT: u16 = 443; // Default secure port for HTTPS
 
-const CONNECT_TIMEOUT: Duration = Duration::from_secs(5); // Timeout for TCP connections
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(10); // Timeout for TCP connections
 
 const IPV4: [Ipv4Addr; 2] = [
     Ipv4Addr::new(104, 26, 1, 225), // Cloudflare (informatiebeveiligingsdienst.nl)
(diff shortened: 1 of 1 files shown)


backend/src/airgap/detect.rs
```diff
@@ -59,138 +53,92 @@ impl AirgapDetection {
...
         }
-
-        self.set_last_check();
     }
 
     /// Detects if the system is in an airgap by attempting to connect to IPv4, IPv6 addresses and performing DNS lookups
-    fn is_connected(&self) -> bool {
-        // attempt to connect to known IPv4 addresses over TCP
-        let tcp_ipv4_connection_success = IPV4
+    async fn is_connected(&self) -> bool {
```
praseodym, line 115: Maybe it's easier to spawn all connection/lookup tasks in parallel, e.g. using a [JoinSet](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html)?

marlonbaeten: Sure - see [a88def8](https://github.com/kiesraad/abacus/pull/1859/commits/a88def8e2352f327c81fc5d7e0f7d85888220492)

marlonbaeten:
> It seems like the connection handling has bugs, even without an internet connection Abacus now logs a successful TCP connection:
> 
> ```
> 2025-07-31T07:18:23.167040Z  INFO abacus: Airgap detection is enabled, starting airgap detection task...
> 2025-07-31T07:18:23.167339Z  WARN abacus::airgap::detect: TCP connection to IP address 104.26.1.225 successful
> 2025-07-31T07:18:23.167454Z ERROR abacus::airgap::detect: Airgap violation detected, abacus is connected to the internet!
> ```

Sorry about that, the timeout function apparently wraps the result in a new result, fixed in [a88def8](https://github.com/kiesraad/abacus/pull/1859/commits/a88def8e2352f327c81fc5d7e0f7d85888220492)
Unfortunately this is hard to unit test...

praseodym:
> > It seems like the connection handling has bugs, even without an internet connection Abacus now logs a successful TCP connection:
> > ```
> > 2025-07-31T07:18:23.167040Z  INFO abacus: Airgap detection is enabled, starting airgap detection task...
> > 2025-07-31T07:18:23.167339Z  WARN abacus::airgap::detect: TCP connection to IP address 104.26.1.225 successful
> > 2025-07-31T07:18:23.167454Z ERROR abacus::airgap::detect: Airgap violation detected, abacus is connected to the internet!
> > ```
> 
> Sorry about that, the timeout function apparently wraps the result in a new result, fixed in [a88def8](https://github.com/kiesraad/abacus/pull/1859/commits/a88def8e2352f327c81fc5d7e0f7d85888220492) Unfortunately this is hard to unit test...

Can confirm it's fixed!

PR #1860, opened 2025-07-29
Add number of voters and counting method to backend for EML import

### LINKED ISSUE #1773: Add number of voters and counting method to backend for EML_NL import
Depends on #1714

rnijveld:
Fixes #1773

### DIFF
--- a/backend/src/election/api.rs
+++ b/backend/src/election/api.rs
@@ -18,6 +18,7 @@ use crate::{
     audit_log::{AuditEvent, AuditService},
     authentication::{Admin, User},
     committee_session::{CommitteeSession, CommitteeSessionCreateRequest},
+    election::VoteCountingMethod,
     eml::{EML110, EML230, EMLDocument, EMLImportError, EmlHash, RedactedEmlHash},
     polling_station::PollingStation,
 };
@@ -172,6 +173,10 @@ pub struct ElectionAndCandidateDefinitionValidateRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[schema(value_type = Option<String>, nullable = false)]
     polling_station_data: Option<String>,
+
+    counting_method: Option<VoteCountingMethod>,
+
+    number_of_voters: Option<u32>,
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
@@ -223,6 +228,11 @@ pub async fn election_import_validate(
         election = EML230::from_str(&data)?.add_candidate_lists(election)?;
     }
 
+    // update counting method if available
+    if let Some(cm) = edu.counting_method {
+        election.counting_method = cm;
+    }
+
     // parse and validate polling stations, and update number of voters
     let polling_stations;
     let mut number_of_voters = 0;
@@ -231,6 +241,11 @@ pub async fn election_import_validate(
         number_of_voters = EML110::from_str(&data)?.get_number_of_voters()?;
     } else {
         polling_stations = None;
+    }
+
+    // override number of voters if provided
+    if let Some(nov) = edu.number_of_voters {
+        number_of_voters = nov;
     }
 
     Ok(Json(ElectionDefinitionValidateResponse {
@@ -250,6 +265,8 @@ pub struct ElectionAndCandidatesDefinitionImportRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[schema(value_type = Option<String>, nullable = false)]
     polling_station_data: Option<String>,
+    counting_method: VoteCountingMethod,
+    number_of_voters: u32,
 }
 
 /// Uploads election definition, validates it, saves it to the database, and returns the created election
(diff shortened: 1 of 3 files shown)


backend/src/election/api.rs
```diff
@@ -172,6 +173,10 @@ pub struct ElectionAndCandidateDefinitionValidateRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[schema(value_type = Option<String>, nullable = false)]
     polling_station_data: Option<String>,
+
+    counting_method: Option<VoteCountingMethod>,
+
+    number_of_voters: Option<u32>,
```
praseodym: Please add `#[serde(skip_serializing_if = "Option::is_none")]` to both fields

backend/src/election/api.rs
```diff
@@ -250,6 +265,8 @@ pub struct ElectionAndCandidatesDefinitionImportRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[schema(value_type = Option<String>, nullable = false)]
     polling_station_data: Option<String>,
+    counting_method: VoteCountingMethod,
+    number_of_voters: u32,
```
praseodym: And also add `#[serde(skip_serializing_if = "Option::is_none")]` here

praseodym (dismissed):
This PR has test failures

backend/src/election/api.rs
bramtweedegolf: Fixed in https://github.com/kiesraad/abacus/pull/1860/commits/4171bbf88c108fbf9ecc231ebfb57d484b60bdc7

backend/src/election/api.rs
bramtweedegolf: Fixed in https://github.com/kiesraad/abacus/pull/1860/commits/4171bbf88c108fbf9ecc231ebfb57d484b60bdc7

bramtweedegolf:
> This PR has test failures

Fixed in https://github.com/kiesraad/abacus/pull/1860/commits/4171bbf88c108fbf9ecc231ebfb57d484b60bdc7

bramtweedegolf (dismissed):
Can I still approve this PR even though I made some changes?

jschuurk-kr:
> Can I still approve this PR even though I made some changes?

Depends on the scope of your changes, so it's a bit of a judgment call.

And this one can go either way, I think. Considering the size and complexity of the changes, I'd lean towards yes. Comparing the original commit and your commit, I'd lean towards no.

backend/src/election/api.rs
```diff
@@ -250,6 +268,11 @@ pub struct ElectionAndCandidatesDefinitionImportRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[schema(value_type = Option<String>, nullable = false)]
     polling_station_data: Option<String>,
+    #[schema(value_type = Option<VoteCountingMethod>, nullable = false)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    counting_method: Option<VoteCountingMethod>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    number_of_voters: Option<u32>,
```
praseodym, line 277: This also needs `#[schema(nullable = false)]` to prevent the `| null` in the OpenAPI schema

backend/src/election/api.rs
```diff
@@ -172,6 +173,13 @@ pub struct ElectionAndCandidateDefinitionValidateRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[schema(value_type = Option<String>, nullable = false)]
     polling_station_data: Option<String>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = Option<VoteCountingMethod>, nullable = false)]
+    counting_method: Option<VoteCountingMethod>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    number_of_voters: Option<u32>,
```
praseodym, line 183: Another `#[schema(nullable = false)]` here

backend/src/election/api.rs
rnijveld: I'm not quite sure why the `skip_serializing_if` annotations were added in this struct? It is never serialized to the frontend anyway? I also don't get the `nullable = false` annotations, the fields are nullable. And it should also work just fine, any of these fields can be null from the APIs point of view, as long as they are not null as soon as you do your final submit, but that is using a separate struct with non-nullable fields.

praseodym: It does end up in the OpenAPI spec and we don't want to have TypeScript types that can be `undefined` or `null`, but only `undefined`.

bramtweedegolf: Fixed in https://github.com/kiesraad/abacus/pull/1860/commits/fad4da692dbc46ab675899a161695eccd5a4fe85

backend/src/election/api.rs
bramtweedegolf: Fixed in https://github.com/kiesraad/abacus/pull/1860/commits/fad4da692dbc46ab675899a161695eccd5a4fe85

cikzh (approved):

PR #1864, opened 2025-07-29
Add committee session details page

Lionqueen94:
## Description
- Moved dateTime related utils from `format.ts` to separate file `dateTime.ts`
- Added date and time validators and date conversion function to `dateTime.ts`
- Renamed util file `format.ts` to `number.ts`
- Added some missing bad request responses
- Added CommitteeSessionDetailsPage
- Added redirect to CommitteeSessionDetailsPage if details are not filled in when going to the ReportPage
- Added custom copy to CommitteeSessionDetailsPage if redirect from ReportPage happened


Lionqueen94:
@praseodym I've added frontend validation for the date and time fields now, shall I implement the same validation in the backend before saving the data?

praseodym:
> @praseodym I've added frontend validation for the date and time fields now, shall I implement the same validation in the backend before saving the data?

I think you can use a [`chrono::NaiveDate`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDate.html) or [`chrono::NaiveDateTime`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDateTime.html) and rely on serde to throw an error when the date is invalid.

frontend/src/utils/number.test.ts
```diff
@@ -0,0 +1,55 @@
+import { describe, expect, test } from "vitest";
+
+import { deformatNumber, formatNumber, validateNumberString } from "./number";
+
+describe("Format util", () => {
```
oliver3:
```suggestion
describe("Number util", () => {
```

Lionqueen94: Fixed in f5cee31a863cdf317477a8f815ff76e93dc1c77f

frontend/src/i18n/format.tsx
```diff
@@ -10,7 +10,7 @@ interface Element {
   children: AST;
 }
 
-export const DEFAULT_ALLOWED_TAGS = ["ul", "li", "p", "strong", "code", "h2", "h3", "h4"];
```
marlonbaeten, line 13: I think the `<em>` tag is preferred above the `<i>` tag.

frontend/src/features/election_management/components/report/ElectionReportPage.tsx
```diff
@@ -14,17 +14,22 @@ import {
...
+import cls from "../ElectionManagement.module.css";
 
 export function ElectionReportPage() {
   const { committeeSession, election } = useElection();
   const client = useApiClient();
   const navigate = useNavigate();
   const [changeStatusError, setChangeStatusError] = useState<AnyApiError | null>(null);
 
+  // Redirect to update details page if committee session details have not been filled in
+  if (committeeSession.location === "" || committeeSession.start_date === "" || committeeSession.start_time === "") {
```
marlonbaeten, line 29: Could `start_date` also be `undefined` or `null`?
Maybe we should handle these cases in `formatFullDateWithoutTimezone`, like returning '-' if there was an error parsing the date.

Currently it wil return the string "Invalid Date" in my browser:

```ts
(new Date('tijd voor frietjes')).toLocaleTimeString("nl-NL", {
    hour: "numeric",
    minute: "numeric",
    day: "numeric",
    month: "long",
    year: "numeric",
  });
  ```

marlonbaeten (changes requested):

frontend/src/features/election_management/components/report/ElectionReportPage.tsx
Lionqueen94: They cannot be undefined or null because in the backend they are saved as an empty string on creation of the committee session. However there was one empty string missing in this query so this is fixed in ce88807f9f948572e4a183565adb06229a7393ac
I added validation to both front- and backend for the date and time, so even when tampering with the request on saving the details, the invalid date should not be shown because the request will be rejected and the invalid date and time not saved.
A future improvement can definitely be to add error handling to the format date utils

frontend/src/i18n/format.tsx
Lionqueen94: Changed in ce88807f9f948572e4a183565adb06229a7393ac

marlonbaeten (approved):

Issue #1740, opened 2025-07-08
Add data entry section B1-1 'Extra onderzoek'


praseodym:
Add data entry section B1-1 'Extra onderzoek'

https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=8663-33012&t=coOg182cFYjjAsKM-4

PR #1866, opened 2025-07-29
Add `deny_unknown_fields` to structs

praseodym:
I noticed when implementing #1740 that having missing properties doesn't throw an error in the backend, which can lead to bugs. An example of those bugs can be seen in the changes to `backend/tests` in this PR.

This PR adds [`#[serde(deny_unknown_fields)]`](https://serde.rs/container-attrs.html#deny_unknown_fields) to most structs (except EML_NL structs) so that an error is thrown when unknown properties are encountered in the JSON data during deserialization. The EML_NL structs are skipped because the standard has some optional fields that we don't use.


Lionqueen94 (dismissed):

praseodym:
In discussion with @oliver3 I've added the same attribute to many other structs and enums (can't hurt)

backend/src/pdf_gen/models/model_na_31_2.rs
```diff
@@ -9,6 +9,7 @@ use crate::{
 };
 
 #[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
```
oliver3, line 12: Should `deny_unknown_fields` also be added to `ModelNa31_2Bijlage1Input` below?

praseodym: Yes, done in 239b86283b089d639ffb633428795649056aabf8

Lionqueen94 (approved):

Issue #1862, opened 2025-07-29
Data entry components should use gray colour


praseodym:
Most data entry components (at least radio and checkboxes) should be using `--Gray-900` instead of black.

See https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=8663-33012&m=dev#1359686683

Issue #1861, opened 2025-07-29
Input grid non-focused rows should have different input colour


praseodym:
In the input grid, all non-focused rows should use `--Gray-600` as the colour for the input field text.

See https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=8663-32055&m=dev#1359660558

oliver3:
Also add the input color to the `reset.css` to make the color black for input fields explicit and browser-independent. See comment https://github.com/kiesraad/abacus/pull/1842#issuecomment-3135424146

PR #1873, opened 2025-07-30
Adjust gray colors according to design

oliver3:
- Fixes https://github.com/kiesraad/abacus/issues/1861 which can be seen on data entry and in storybook
- Fixes https://github.com/kiesraad/abacus/issues/1862 which can for now only be seen in storybook


praseodym:
> github-merge-queue bot removed this pull request from the merge queue due to failed status checks

Created #1878 because this isn't the first time I've seen that test fail.

PR #1874, opened 2025-07-30
Update polling station API for coordinator file import

bramtweedegolf:
To keep this PR small, it contains only API / backend changes;
- the frontend will be implemented in https://github.com/kiesraad/abacus/issues/1632
- adding an audit event will be implemented in https://github.com/kiesraad/abacus/issues/1915.

This PR adds two API endpoints for polling stations:
-  /api/elections/{election_id}/polling_stations/validate-import
-  /api/elections/{election_id}/polling_stations/import

These endpoints will be used when the coordinator imports a file with polling stations for an existing election. The flow is similar to the step in the election creation flow.

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -1338,6 +1338,155 @@
         }
       }
     },
+    "/api/elections/{election_id}/polling_stations/import": {
+      "post": {
+        "summary": "Import file with Polling Stations",
+        "operationId": "polling_station_import",
+        "parameters": [
+          {
+            "name": "election_id",
+            "in": "path",
+            "description": "Election database id",
+            "required": true,
+            "schema": {
+              "type": "integer",
+              "format": "int32",
+              "minimum": 0
+            }
+          }
+        ],
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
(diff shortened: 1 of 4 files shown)


cikzh (dismissed):
LGTM. No tests, but they'll be added in a subsequent PR as you mentioned in the description

frontend/src/i18n/locales/nl/log.json
praseodym, line 1: This PR is only for backend so the changes in this file shouldn't be part of this PR

praseodym:
Also why not implement the backend tests in this PR?

frontend/src/i18n/locales/nl/log.json
cikzh: Wouldn't the TS compiler complain that there are missing translations?

praseodym: These new translations aren't used in this PR right?

praseodym: Never mind, I see that these are used for the audit log.

bramtweedegolf:
> Also why not implement the backend tests in this PR?

Good point, backend tests added

Issue #1702, opened 2025-07-02
Improve the NavBarPgObj


bramtweedegolf:
#### Make the NavBarPgObj role-aware
Different roles have access to different nav bars. The page object should reflect that.

#### Create a base page object with the navbar and have other page objects inherit from it
The NavBarPgObj is used inconsistently. There's the `DataEntryBasePage`, with all data entry pages inheriting from it. Then there are four page objects that use it directly. And as of [PR #1667](https://github.com/kiesraad/abacus/pull/1667) it's also used in a test. 

I'd like to have a base page that has the NavBarPgObj as attribute and then all other pages can inherit from it. That way the structure of the pages in our application matches the structure of our page objects. (If there's a better way to achieve that goal, let's do that. Also considering the two parts of this issue might not be entirely compatible.)

jschuurk-kr:
Starting to think that "Create a base page object with the navbar and have other page objects inherit from it" is not a good idea. The actual issues to address for the NavBar page object, are:
1. different roles have different nav bars and that should be reflected in the page object(s);
2. the page object is instantiated in different ways, this should be made consistent.

One way to achieve (2) is through a base page object as mentioned in the description. The problem I see with that solution is that it complicates things considering (1) and that the only advantage is to adhere to the principle that the structure of the page objects matches the structure of the pages. Adhering to that principle doesn't bring enough practical benefits, though, compared to having separate page objects for the nav bars and instantiating those on their own when needed.

praseodym:
"Composition over inheritance" rearing its head again 🙂

marlonbaeten:
@jschuurk-kr like this: https://github.com/kiesraad/abacus/compare/test-base-page-object-with-navbar ?

jschuurk-kr:
> [@jschuurk-kr](https://github.com/jschuurk-kr) like this: https://github.com/kiesraad/abacus/compare/test-base-page-object-with-navbar ?

@marlonbaeten That's the inheritance I had in mind, yes, but I suspect that combining this change with making the Navbar role-aware is going to get quite messy. Hence my comment above that it's probably better to have separate page objects for the nav bars and instantiating those on their own when needed.

jschuurk-kr:
> "Composition over inheritance" rearing its head again 🙂

what makes sense in Selenium/Python turns out not to always make sense in Playwright/TypeScript...

marlonbaeten:
Ok, not sure what to change then. Maybe we should discuss this in person.

PR #1901, opened 2025-08-04
Improve NavBar page object

marlonbaeten:
Fixes [#1702](https://github.com/kiesraad/abacus/issues/1702)

- This introduces the `TypistNavBar` and the `AdminNavBar` for these specific roles
- The nav bar is always used as a separate page object  in tests (not as a property of other page objects)


bramtweedegolf (dismissed):

cikzh (approved):
Nice improvement, makes it more readable IMO

frontend/e2e-tests/tests/authentication.e2e.ts
```diff
@@ -1,4 +1,5 @@
 import { expect } from "@playwright/test";
+import { NavBar } from "e2e-tests/page-objects/nav_bar/NavBarPgObj";
```
jschuurk-kr: Am wondering if it's better to always use the AdminNavBar or the TypistNavBar and never the NavBar. Last time I checked there's no way to enforce that in TypeScript though. What do you think?

marlonbaeten: While authenticating we do not "know" the role yet. So for be the generic variant makes sense here. Also, this follows the structure of our `NavBar` component, which will render the same `#navbar-username` and `#navbar-role` fields independent of the role.

jschuurk-kr: We do know the role because of the test data. The "login happy path" logs in with username "admin1". The "first login" uses the `newTypist` fixture. And I'm not convinced our page objects should follow our component structure, since page objects should not care about anything deeper than the DOM.

marlonbaeten: Fair enough - see a31e6e6

bramtweedegolf (approved):

PR #1903, opened 2025-08-04
Add new audit event type

bramtweedegolf:
A new type AuditEvent is added, 'PollingStationsImported', which stores the following details:

- election id
- imported filename
- number of imported polling stations


backend/src/audit_log/audit_event.rs
```diff
@@ -84,6 +84,14 @@ pub struct PollingStationDetails {
     pub polling_station_locality: String,
 }
 
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, ToSchema)]
+#[serde(rename_all = "camelCase")]
```
oliver3: Why the camelCase for this struct specifically?

marlonbaeten: I think this is to adhere to Javascript naming conventions for attribute names.

cikzh (approved):

backend/src/audit_log/audit_event.rs
rnijveld: I think this attribute should probably be removed as discussed during the standup this morning (to keep consistency with the rest of the API)?

bramtweedegolf: @rnijveld All audit events currently use camel case like this. So it seems slightly more consistent to me to keep this in camel case for now, until we refactor the entire audit log (and maybe other parts) to stop using camel case, which is beyond the scope of this PR. Do you agree, or would you still change this field now in preparation of the refactor?

oliver3: I've created https://github.com/kiesraad/abacus/pull/1931 to refactor the audit log api

Lionqueen94: Fixed!

Lionqueen94 (approved):

cikzh (approved):

PR #1910, opened 2025-08-04
Fix Alert styling

### LINKED ISSUE #1869: Space below form alerts missing
There are some pages where the `FormAlert` is used inside of the `<main>` where there is some space missing below the alert. We should resolve this in a reusable way.

Image

Image

### LINKED ISSUE #1888: Text in Alert is now not wide enough
See in the screenshot:
Image

stacktraceghost:
- Resolves #1869
  -  Spacing was missing when using `FormLayout.Alert`, added a margin-bottom to that element

- Resolves #1888
  - Adjusted max-width to 40rem (according to design)
  - Width got overridden `formLayout p`, so the `width: 100%` makes sure we use our own max-width

### DIFF
--- a/frontend/src/components/ui/Form/FormLayout.module.css
+++ b/frontend/src/components/ui/Form/FormLayout.module.css
@@ -52,6 +52,7 @@
 .formAlert {
   overflow: hidden;
   margin-top: -2.5rem;
+  margin-bottom: 2rem;
   margin-left: -5rem;
   margin-right: -5rem;
 }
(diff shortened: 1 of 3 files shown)


frontend/src/components/ui/Form/FormLayout.module.css
```diff
@@ -52,6 +52,7 @@
 .formAlert {
   overflow: hidden;
   margin-top: -2.5rem;
+  margin-bottom: 2rem;
   margin-left: -5rem;
   margin-right: -5rem;
```
Lionqueen94:
```suggestion
  margin: -2.5rem -5rem 2rem -5rem;
```

stacktraceghost:
> Nice fixes! I think some alerts in the design have width 39rem. Maybe @jorisleker can decide if it should be 39rem or 40rem for all alerts?

Thanks for the review! This has been adjusted to 40rem in Figma to make it equal to the feedback component.

Issue #1899, opened 2025-08-04
Create frontend components for create first admin flow


marlonbaeten:
See design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12900-291403&t=VmoTFDiE8qMOpopr-11

Issue #1892, opened 2025-08-04
Create design for the "create admin" form


marlonbaeten:
Some explanation on how initialisation works.
Should contain fields for the username and the password.

[Design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12900-291403&t=VmoTFDiE8qMOpopr-11)

PR #1911, opened 2025-08-04
Initialisation state plumbing

### LINKED ISSUE #1893: Create environment variable to show/hide the "dev" page

### LINKED ISSUE #1894: Add a new API endpoint that checks whether the application is initialised

### LINKED ISSUE #1896: Navigate to create admin form when the application is not initialised

marlonbaeten:
Fixes #1893 
Fixes #1894 
Fixes #1896

This PR adds and endpoint that wil return whether the application is initialised or not. We define that the application is initialised if there is at least one user that has logged in at least once.

This PR also introduces an environment variable that wil either enable of disable the "/dev" page. The "/dev" page is enabled by default.

Flow:

- If the "/dev" page is not enabled and a user opens Abacus for the first time, they will be redirected to the login component
- The login components always calls the "is initialised" endpoint
- If the application is not yet initialised the user will be redirected to the initialisation flow (see #1892 and #1899) - which is a page showing "TODO" for now

How to test:

- Clear the database and startup abacus. The login page should redirect to the "TODO" page (initialisation flow)
- Startup abacus with fixtures (`--seed-data`) and navigate to the login page. This should **not** redirect to the "TODO" page.

FYI we use status code 418 to indicate that the application is not yet initialised, and thereby logically and spiritually identical to a teapot. The choice for this status code is non-negotiable.


backend/src/authentication/user.rs
```diff
@@ -366,6 +366,21 @@ pub async fn list(conn: impl DbConnLike<'_>) -> Result<Vec<User>, Error> {
     Ok(users)
 }
 
+/// Fetch the first admin user ever created
+pub async fn get_active_user_count(conn: impl DbConnLike<'_>) -> Result<u32, AuthenticationError> {
+    let row = sqlx::query!(
+        r#"
+        SELECT COUNT(*) AS "count: u32"
```
rnijveld: Do we really need the active user count? For the current endpoint `SELECT 1 FROM users WHERE last_activity_at IS NOT NULL LIMIT 1` with a `fetch_optional().await?.unwrap_or(0)` would also work and prevent any unnecessary work of scanning the users table (not that the users table is expected to be particularly large though.

marlonbaeten: See [8015bdf](https://github.com/kiesraad/abacus/pull/1911/commits/8015bdff34bfb8d5adba97f5c68a3ab8d4ae232c)

rnijveld (approved):

Issue #1863, opened 2025-07-29
Body font size should be 1rem (16px)


praseodym:
Body font size should be 1rem (16px) instead of the 1.125rem (18px) that we currently use. Some components should still use 1.125rem, but that shouldn't be the default.

PR #1916, opened 2025-08-05
Fix table font sizes

oliver3:
- Resolves https://github.com/kiesraad/abacus/issues/1863 for tables
- Best to review per commit:
  - Change the default font for `Table.NumberCell` from bold to normal, and with the number font. Update usages accordingly
  - Table font size is inherited and therefore body font size, except for the row header/first column which should have large font
  - Add table variant `information` that does not have a larger font for the row header or first column

Tables in the application should now be styled according to designs and how we want it to be.

- Polling stations tables on [list page](https://e8522086.kiesraad-abacus.pages.dev/elections/1/polling-stations) and import page
- [Users table](https://e8522086.kiesraad-abacus.pages.dev/users)
- [Logs table](https://e8522086.kiesraad-abacus.pages.dev/logs)
- [Elections table](https://e8522086.kiesraad-abacus.pages.dev/elections)
- [Election management tables](https://e8522086.kiesraad-abacus.pages.dev/elections/1)
- [Election status table](https://e8522086.kiesraad-abacus.pages.dev/elections/1/status)
- [Resolve differences table](https://e8522086.kiesraad-abacus.pages.dev/elections/1/status/3/resolve-differences)
- Apportionment pages with tables ([separate tables on storybook](https://e8522086.kiesraad-abacus.pages.dev/storybook/?path=/story/features-apportionment-components-apportionment--default-apportionment-table))

### DIFF
--- a/frontend/src/components/ui/Table/Table.module.css
+++ b/frontend/src/components/ui/Table/Table.module.css
@@ -103,7 +103,7 @@
 }
 
 /* Row header and first cell in the row large font, using `:where()` to give it lowest specificity */
-:where(.table th[scope="row"]),
-:where(.table td:first-child) {
+:where(.table.defaultVariant th[scope="row"]),
+:where(.table.defaultVariant td:first-child) {
   font-size: var(--font-size-lg);
 }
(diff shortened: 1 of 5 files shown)


Lionqueen94 (changes requested):
The table on the ElectionHomePage should also get variant="information" I think :)

Lionqueen94 (approved):

Issue #1880, opened 2025-07-31
Prevent user from deleting own account


praseodym:
Prevent user from deleting their own account, so that there is always at least one account present.

- Backend: throw error when user is trying to delete their own account
- Frontend: hide delete button for own account

PR #1921, opened 2025-08-05
Prevent deleting the last admin

marlonbaeten:
Currently, in Abacus, an administrator can delete any user. There is a risk of the administrator accidentally deleting all administrators and thereby "bricking" Abacus.  [Murphy's law](https://en.wikipedia.org/wiki/Murphy%27s_law) teaches us that this will happen.

This PR prevents deleting the last administrator. Note that we only count administrators that have logged in at least once. The user fixtures are updated to be similar to active user accounts, by setting the "last_activity_at".


praseodym:
We also have #1880 to prevent a user from deleting their own account, I think that's a more simple solution that fixes the same "bricking" issue.

marlonbaeten:
> We also have #1880 to prevent a user from deleting their own account, I think that's a more simple solution that fixes the same "bricking" issue.

Check - I will implement that issue.

marlonbaeten:
Implementing prevent delete own account

PR #1923, opened 2025-08-05
Prevent a user from deleting their own account

### LINKED ISSUE #1880: Prevent user from deleting own account

marlonbaeten:
Fixes #1880

Prevent an admin from deleting their own account and thereby being locked out of Abacus.

This also prevents Abacus from ever returning to the "not initialised" state after the first admin is successfully logged in.

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -4819,6 +4819,7 @@
           "InvalidVoteCandidate",
           "InvalidVoteGroup",
           "InvalidXml",
+          "OwnAccountCannotBeDeleted",
           "PasswordRejection",
           "PdfGenerationError",
           "PollingStationRepeated",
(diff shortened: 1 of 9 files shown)


marlonbaeten:
> `OwnAccountCannotBeDeleted` isn't returned from the endpoint in backend, the test also fails

Test-driven development ;-) See [c5a3eb9](https://github.com/kiesraad/abacus/pull/1923/commits/c5a3eb92dc8202518ff76a19b59477527d07ef41)

praseodym (approved):

cikzh (approved):
Tested by verifying the delete button is indeed hidden in the UI. Haven't tested by manually querying the backend, but that integration test covers that. Looks good!

PR #1928, opened 2025-08-05
Implement modal for 409 CommitteeSessionPaused error

Lionqueen94:

### DIFF
--- a/frontend/src/components/committee_session/CommitteeSessionStatus.module.css
+++ b/frontend/src/components/committee_session/CommitteeSessionStatus.module.css
@@ -1,0 +1,5 @@
+.committee_session_status {
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
+}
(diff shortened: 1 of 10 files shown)


oliver3 (dismissed):

jschuurk-kr:
Minor thing, probably not worth fixing.
When the session is pauzed, if a typist clicks on the election on the elections page (so page was loaded before session was pauzed), the typist does end up on the data entry home page, with the modal being shown. Arguably the modal should be shown on the elections page without navigating to the data entry home page. If we were to fix this, the modal button "Naar startscherm" should be renamed to "Ok" or something, because the typist already is on the startscherm.

jschuurk-kr:
If data entry is pauzed and the typist returns to the elections overview page, when data entry is unpauzed again, the typist needs to refresh the page through the browser to be able to get to the data entry home page. It would be nicer if there was a button to refresh the status of the elections. cc @jorisleker

oliver3:
Maybe refreshing the elections overview page every x seconds, like we did [for the elections status page](https://github.com/kiesraad/abacus/pull/1924) would alleviate both issues @jschuurk-kr  @jorisleker ?

PR #1905, opened 2025-08-04
Updated translation for polling station

lama-d3v:
Resolves #1870

Changes:
- updated translation for polling station
- fixed styling for correct spacing in form sections

PR #1929, opened 2025-08-06
Form layout fixes

oliver3:
Note: PR #1905 has to be merged first, then check if base branch has become `main`.

---

- updated user forms to adhere to current design on Figma (https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=7673-5132&m=dev and the designs below) 
- made sure the account setup page also looks good when zooming out - aka footer with button sticks to the bottom
main:
image
this PR:
image

## Question
- Forms that use a FormLayout now get a default 2.5rem gap between its children (generally FormLayout.Section and FormLayout.Controls elements). @jorisleker the number of voters form design has a gap of 1.5rem between the form content and the button, in this PR this is by default 2.5rem. Is this fine?

## How to test
- Check the forms for adding/updating user
- Check the forms for adding/updating polling station
- Check the form for updating committee session details
- Check the form for updating the number of voters
- Check the form for login
- Check the form for account setup (and zoom out)


Lionqueen94:
> Tested all forms/pages mentioned in the PR, Found a couple of issues. See screenshots:
> 
>     * Alerts now have h2's, which they should not
> 
>     * On the account setup page, the success and error messages overlap the content of the page, instead of being positioned at the top of the content section

Fixed the Alert issues and changed the h2 into a strong tag

jorisleker (dismissed):

frontend/src/components/ui/AppLayout/AppLayout.module.css
```diff
@@ -83,6 +70,7 @@
     padding: 2rem 5rem;
     font-size: var(--font-size-sm);
     line-height: 1.25rem;
+    max-height: 5rem;
```
lama-d3v:
```suggestion
    padding: 0 5rem;
    font-size: var(--font-size-sm);
    line-height: 1.25rem;
```
Fixes the weird height difference with Figma

marlonbaeten:
I'm concerned that the selectors on role "strong" might turn out to be not unique enough, since any new "strong" element on the page might cause a collision. Is this conscious choice we have to fix once new "strong" elements are introduced or something we can tackle in this PR?

frontend/src/app/AuthorizationDialog.test.tsx
```diff
@@ -74,8 +74,8 @@ describe("AuthorizationDialog", () => {
       </TestUserProvider>,
     );
 
-    const loginSuccess = screen.getByRole("alert");
-    expect(await within(loginSuccess).findByRole("heading", { name: "Je bent automatisch uitgelogd" })).toBeVisible();
```
lama-d3v, line 78: Is there a specific reason for deleting `.toBeVisible()` (twice)?

frontend/src/components/election/NumberOfVotersForm.tsx
```diff
@@ -26,11 +26,10 @@ export function NumberOfVotersForm({ defaultValue, instructions, hint, button, o
 
   return (
     <>
-      <Form onSubmit={handleSubmit}>
+      <Form title={t("election_management.how_many_voters")} onSubmit={handleSubmit}>
```
lama-d3v: The empty tag (`<>`) could perhaps be replaced entirely by the Form tag (`<Form title={t("election_management.how_many_voters")} onSubmit={handleSubmit}>`).

Lionqueen94:
> I'm concerned that the selectors on role "strong" might turn out to be not unique enough, since any new "strong" element on the page might cause a collision. Is this conscious choice we have to fix once new "strong" elements are introduced or something we can tackle in this PR?

They are generally selecting within the alert, and always specifically on the text expected. Do you have a (set of) selector(s) you're worried about that might nog be unique enough?

marlonbaeten:
> They are generally selecting within the alert, and always specifically on the text expected. Do you have a (set of) selector(s) you're worried about that might nog be unique enough?

Ah I see - yes that makes sense. Some forms possibly have multiple alerts, but filtering on their contents prevents any collisions. :+1:

I have updated my PR to include the changes to the Alerts: eb0dc1f017e181b2f41ffe0738845246c54dfade

frontend/src/components/ui/Form/FormLayout.module.css
```diff
@@ -1,6 +1,4 @@
 .formLayout {
-  margin-bottom: var(--space-md);
-
   h1,
   h2,
```
lama-d3v, line 3: Are there any conflicts with the width because the title has been semantically changed (legend <> h2)? In this css file, for both tags (legend and h2) different widths are configured.

frontend/src/components/ui/Form/FormLayout.module.css
```diff
@@ -18,11 +16,9 @@
...
-  margin-bottom: var(--space-md);
-
-  &:not(:first-child) {
-    margin-top: var(--space-xl);
-  }
+  display: flex;
+  flex-direction: column;
+  gap: var(--space-md-lg);
 
   h2.formSectionTitle {
```
lama-d3v: The formSectionTitle has been changed to h3 in `frontend/src/components/ui/Form/FormLayout.tsx`. Shouldn't this be removed (probably isn't used anymore)

marlonbaeten (dismissed):

stacktraceghost (dismissed):

frontend/src/components/ui/Form/FormLayout.tsx
```diff
@@ -17,10 +17,10 @@ export function FormLayout({ children, disabled }: FormLayoutProps) {
   );
 }
 
-function FormSection({ children, title }: { children: React.ReactNode; title?: string }) {
+function FormSection({ children, title }: { children: React.ReactNode; title?: string | React.ReactElement }) {
```
lama-d3v, line 20: Should ` | React.ReactElement ` be really added (and why)? 
```suggestion
function FormSection({ children, title }: { children: React.ReactNode; title?: string | React.ReactElement }) {
```

frontend/src/features/election_management/components/update/CommitteeSessionDetailsPage.tsx
```diff
@@ -112,62 +112,63 @@ export function CommitteeSessionDetailsPage() {
       <header>
         <section>
           <h1>
-            {t("election_management.custom_committee_session_details", {
-              sessionLabel: sessionLabel,
-            })}
+            {/* TODO: Change to conditional GSB/HSB/CSB when implemented */}
+            {t("GSB")} {election.location}
```
lama-d3v, line 116: Maybe add "GSB" as an election type (e.g.) in election?

(also specified for frontend/src/features/election_management/components/update/NumberOfVotersPage.tsx)

frontend/src/features/users/components/update/UserDelete.tsx
```diff
@@ -36,8 +36,8 @@ export function UserDelete({ user, onDeleted, onError }: UserDeleteProps) {
   const deleting = requestState.status === "loading";
 
   return (
-    <>
-      <Button type="button" variant="tertiary-destructive" leftIcon={<IconTrash />} onClick={toggleModal}>
+    <div className="mt-md">
```
lama-d3v, line 39: Should we use margin-top here?

frontend/src/components/ui/Form/FormLayout.tsx
Lionqueen94: Yes this is needed because the title content can be a ReactElement when there is not just a string but also styling (`<strong>` tags for example). Translate then returns a ReactElement instead of a string.

frontend/src/features/election_management/components/update/CommitteeSessionDetailsPage.tsx
Lionqueen94: This will be done in the future when we will add CSB functionality, that's why it's currently added like this instead.

frontend/src/features/users/components/update/UserDelete.tsx
Lionqueen94: Yes because the delete button is not part of the Controls but separate below it, so this was needed

frontend/src/app/AuthorizationDialog.test.tsx
Lionqueen94: I thought it would not be needed if you use toHaveTextContent, but that might be wrong. @jschuurk-kr You probably know all about this. Should I also keep toBeVisible in the test?

frontend/src/components/ui/Form/FormLayout.module.css
Lionqueen94: I checked and this applies to the legend of the choiceList (which is not an <h>), because the title <legend> is not located inside the formLayout but in the form itself. So as far as I can see this does not give conflicts

frontend/src/components/ui/Form/FormLayout.module.css
Lionqueen94: Removed in 2db131bc472f4a379839e716f9504bb2d7a3d3a2

frontend/src/components/election/NumberOfVotersForm.tsx
Lionqueen94: Changed in 2db131bc472f4a379839e716f9504bb2d7a3d3a2

lama-d3v (approved):

frontend/src/app/AuthorizationDialog.test.tsx
jschuurk-kr: Short answer: we should keep the toBeVisible.

__Long answer__
You don't need to check visibility before performing an action on an element. Both React Testing Library's [user-event](https://testing-library.com/docs/user-event/intro#differences-from-fireevent) and [Playwright](https://playwright.dev/docs/actionability#introduction) do some visibility and interactibility checks.

You do need to check visibility when you're only doing asserts. Neither jest-dom's [toHaveTextContent](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohavetextcontent) (the library used by React Testing Library for this), nor Playwright's [toContainText](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-contain-text) check for visiblity in addition to checking the text.

There's also the question if the locator checks for visiblity. In Playwright the answer is a definitive no, because Playwright only executes the locator code when you want to do something with the (to be) located element(s). React Testing Library does query for the element when you do a getBy/queryBy/findBy, e.g. at `const loginSuccess = screen.getByRole("alert");`. Based on this page about [RTL queries](https://testing-library.com/docs/queries/about#types-of-queries), however, I would say it only queries for presence in the DOM, not for visibility. (Because how else would you be able to check for present but invisible element?)

Lionqueen94: Re-added in 8de554b1b27d8157374161b1563d00ab0670a5f0

Issue #1930, opened 2025-08-06
Replace party names with fictional party names in our test data


jschuurk-kr:
We should have a clear distinction between the test data we generate/edit ourselves and production data we use for testing. The former should be completely fictional, the latter we should not edit.

However, some of our test data contains the names of actual political groups, i.e. the files in backend/src/eml/tests. These names need to be replaced with fictional names and any failing tests need to be fixed. Reason for this is that we want to avoid any chance of people going through our tests and asking why we always test certain scenarios with certain existing political groups.

PR #1933, opened 2025-08-06
Use election import endpoint for e2e tests

### LINKED ISSUE #1793: Use election import for Playwright e2e tests
The Playwright e2e tests currently use the development `POST /api/election` endpoint to create test elections:
https://github.com/kiesraad/abacus/blob/39363409806684d0f61930c4e9ddd994e7a141ec/frontend/e2e-tests/fixtures.ts#L88-L98
https://github.com/kiesraad/abacus/blob/39363409806684d0f61930c4e9ddd994e7a141ec/backend/src/election/api.rs#L119-L159

Now that we have a election import API that imports EML_NL, we can use that API to create the instead.

stacktraceghost:
Resolves #1793 

Changes in this PR:
- Use election import endpoint for e2e fixture
  - Use existing EMLs for this purpose
- Remove dev election_create endpoint from backend
- Adjust e2e tests to reflect data in existing EMLs

Note: The existing EMLs make use of actual names of political groups. A ticket has been made (#1930 ) to adjust these to fictional names.


stacktraceghost:
The backend is still using the `POST /api/elections` endpoint in test `test_election_apportionment_error_apportionment_not_available_no_polling_stations`. 

Edit: Fixed.

backend/tests/election_integration_test.rs
```diff
@@ -70,63 +68,7 @@ async fn test_election_details_works(pool: SqlitePool) {
...
-        .unwrap();
-
-    // Ensure the response is what we expect
-    assert_eq!(response.status(), StatusCode::CREATED);
-    let body: ElectionWithPoliticalGroups = response.json().await.unwrap();
-    assert_eq!(body.name, "Test Election");
-}
-
-#[test(sqlx::test(fixtures(path = "../fixtures", scripts("users"))))]
-#[cfg(feature = "dev-database")]
```
stacktraceghost, line 128: It seems like this feature attribute is not necessary here. Please check

backend/README.md
```diff
@@ -51,7 +51,7 @@ Use `cargo clippy --all-targets --all-features -- -D warnings` to lint the proje
 
 ### Testing
 
-Use `cargo test` to run the tests. The tests are also run in a GitHub Actions workflow.
+Use `cargo test --features embed-typst` to run the tests. The tests are also run in a GitHub Actions workflow.
```
praseodym: Using the `embed-typst` feature shouldn't be needed as long as you have [Typst](https://typst.app/) installed on your machine. It's faster that way because you can avoid compiling Typst as part of Abacus.

stacktraceghost: Ah perfect, that works and is indeed a lot faster. Thanks, I've reverted the readme change.

backend/tests/election_integration_test.rs
praseodym: Removing it is fine, it was there because the `/api/elections` endpoint was only enabled when the `dev-database` feature was enabled.

backend/README.md
jschuurk-kr: Might be useful to refer to the explanation in the "Running" section about this in the "Testing" section, so the next person doesn't need to rediscover this.

Lionqueen94 (approved):

oliver3 (approved):

PR #1949, opened 2025-08-11
Add audit log for imported polling stations for admin

bramtweedegolf:
When the admin creates a new election with polling stations, create an audit log event for the import.
If an election is created without polling stations, do not create the event.

See also: https://github.com/kiesraad/abacus/pull/1950
Closes: https://github.com/kiesraad/abacus/issues/1632

### DIFF
--- a/backend/src/election/api.rs
+++ b/backend/src/election/api.rs
@@ -182,6 +182,10 @@ pub struct ElectionAndCandidateDefinitionValidateRequest {
     #[schema(nullable = false)]
     #[serde(skip_serializing_if = "Option::is_none")]
     number_of_voters: Option<u32>,
+
+    #[schema(nullable = false)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    polling_station_file_name: Option<String>,
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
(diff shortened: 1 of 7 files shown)


backend/src/election/api.rs
```diff
@@ -278,6 +278,9 @@ pub struct ElectionAndCandidatesDefinitionImportRequest {
     #[schema(nullable = false)]
     #[serde(skip_serializing_if = "Option::is_none")]
     number_of_voters: Option<u32>,
+    #[schema(nullable = false)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    file_name: Option<String>,
```
rnijveld: I would change this field to polling_station_file_name just to be sure not to confuse it with any of the other files uploaded.

backend/src/election/api.rs
```diff
@@ -312,7 +315,13 @@ pub async fn election_import(
     // Process polling stations
     let mut polling_places = None;
     let mut number_of_voters = 0;
+
     if let Some(polling_station_data) = edu.polling_station_data {
+        // If polling stations are submitted, file name must be also
```
rnijveld, line 329: I would also add the same file name check to the validate endpoint, just to keep the logic synchronized between the two.

rnijveld (changes requested):

backend/src/election/api.rs
bramtweedegolf: Updated in https://github.com/kiesraad/abacus/pull/1949/commits/d685a5e3a269780996a378fb526b06b74de1024e

backend/src/election/api.rs
bramtweedegolf: Updated in https://github.com/kiesraad/abacus/pull/1949/commits/d685a5e3a269780996a378fb526b06b74de1024e

rnijveld (approved):

cikzh (approved):

PR #1952, opened 2025-08-11
Simplify ZIP streaming

marlonbaeten:
- Use separate reader and writer side for ZIP responses, based on tokio duplex buffer.
- This makes write_stream.rs completely obsolete
- Explicitly start tokio task to generate PDF / ZIP files
- Added unit tests to ZIP response


backend/src/fixtures.rs
```diff
@@ -21,7 +21,7 @@ macro_rules! load_fixtures {
 ///
 /// This list should be updated manually when a new fixture is added that
 /// needs to be loaded when seeding fixtures.
-const FIXTURES: &[Fixture] = load_fixtures!(["election_1", "users"]);
+const FIXTURES: &[Fixture] = load_fixtures!(["election_1", "election_2", "users"]);
```
Lionqueen94: I'm not sure this change should have been committed

marlonbaeten: I want to blame my editor - but it is me who is to blame.

cikzh (approved):

PR #1955, opened 2025-08-11
Create link button component

bramtweedegolf:
Move the link button from the admin election creation polling station import to a component, so it can be reused in the coordinator polling station import.

Closes https://github.com/kiesraad/abacus/issues/1965

### DIFF
--- a/frontend/src/styles/typography.css
+++ b/frontend/src/styles/typography.css
@@ -89,14 +89,6 @@ section.md {
   }
 }
 
-/* Link styling */
-.link {
-  all: unset;
-  text-decoration: underline;
-  cursor: pointer;
-  color: var(--link-default);
-}
-
 a {
   color: var(--link-default);
 


cikzh (approved):
Looks good. Maybe a rename would be nice, since we also have a `ButtonLink`, referring to the `Link` component in `react-router`. The current name confuses me a little, because it suggest the button performs some kind of linking, which isn't necessarily true when this button will be reused in other places.  I think this shouldn't block merging the PR though.

bramtweedegolf:
> Looks good. Maybe a rename would be nice, since we also have a `ButtonLink`, referring to the `Link` component in `react-router`. The current name confuses me a little, because it suggest the button performs some kind of linking, which isn't necessarily true when this button will be reused in other places. I think this shouldn't block merging the PR though.

Good point! Its a button that look like a simple link. What would be a better name? Maybe AnchorButton or something similar?

Lionqueen94:
> > Looks good. Maybe a rename would be nice, since we also have a `ButtonLink`, referring to the `Link` component in `react-router`. The current name confuses me a little, because it suggest the button performs some kind of linking, which isn't necessarily true when this button will be reused in other places. I think this shouldn't block merging the PR though.
> 
> Good point! Its a button that look like a simple link. What would be a better name? Maybe AnchorButton or something similar?

On the ElectionStatusPage I use:
```
function getLink() {
    if (committeeSession.status === "data_entry_not_started") {
      return (
        <button
          key="start"
          className="link"
          onClick={() => {
            handleStatusChange("data_entry_in_progress");
          }}
        >
          {t("election_status.start")}
        </button>
      );
    } else if (committeeSession.status === "data_entry_in_progress") {
      return (
        <button
          key="pause"
          className="link"
          onClick={() => {
            togglePauseModal();
          }}
        >
          {t("election_status.pause")}
        </button>
      );
    } else if (committeeSession.status === "data_entry_paused") {
      return (
        <button
          key="resume"
          className="link"
          onClick={() => {
            handleStatusChange("data_entry_in_progress");
          }}
        >
          {t("election_status.resume")}
        </button>
      );
    }
  }
```
For buttons that look like links, it might be nice to also use the new component here? The styling is also already handled here ("link"). Maybe we should even add this type of button as a variant of the `Button` component. We can call it Button.Linkstyling or Button.Underlined or something?

oliver3:
I would suggest making it a variant of the already existing `<Button>`

bramtweedegolf:
Good point, a variant of the already existing Button is a better idea. I removed the extra component and added the variant "underlined" to the Button component. How does this look?

We could use the same variant for the ElectionStatusPage, but the design seems to be slightly different:

Screenshot_2025-08-18_09-58-26

Screenshot_2025-08-18_09-54-45

oliver3:
Could you please also add it to the [Storybook](https://8727a5b3.kiesraad-abacus.pages.dev/storybook/?path=/story/components-ui-button--buttons) ?

Lionqueen94:
> Good point, a variant of the already existing Button is a better idea. I removed the extra component and added the variant "underlined" to the Button component. How does this look?
> 
> We could use the same variant for the ElectionStatusPage, but the design seems to be slightly different:
> Screenshot_2025-08-18_09-58-26 Screenshot_2025-08-18_09-54-45

I'd say make the import flow one underlined as well for now and let's ask Joris on Figma if all should be underlined or not

bramtweedegolf:
@oliver3 @Lionqueen94 I added the variant to storybook, replaced the import flow buttons and added a comment on figma for Joris.

Lionqueen94:
> @oliver3 @Lionqueen94 I added the variant to storybook, replaced the import flow buttons and added a comment on figma for Joris.

You should be able to remove the `link` styling from `typography.css` now as well, since it was only used for the links on the ElectionStatusPage

bramtweedegolf:
> > @oliver3 @Lionqueen94 I added the variant to storybook, replaced the import flow buttons and added a comment on figma for Joris.
> 
> You should be able to remove the `link` styling from `typography.css` now as well, since it was only used for the links on the ElectionStatusPage

Check, removed in https://github.com/kiesraad/abacus/pull/1955/commits/1e46b6fdc496e80caf6646bf00a92d482bcdaaf9

Issue #1937, opened 2025-08-07
Fix UI debt for election import flow


Lionqueen94:
- Import election page gap between button and text should be 1.5rem
- Error state for form fields when entering wrong hash on check election definition page is missing and hashInput should also turn red instead of blue (does work when leaving fields empty)
- Also check further pages with design and adjust where needed
- Add Shift+Enter keyboard hints

bramtweedegolf:
@Lionqueen94: Most are the changes were no problem, but I encountered a few problems with the FormAlerts:

1) The FormAlerts have negative margins:  `margin: -2.5rem -5rem 2rem -5rem;`

<img width="1572" height="1284" alt="Image" src="https://github.com/user-attachments/assets/81c8ca46-7d55-406e-a6b1-125bb1cf78eb" />

2) The FormAlert have an inside div with a margin. Together with the FormSection gap, this creates too much space:

<img width="1542" height="858" alt="Image" src="https://github.com/user-attachments/assets/7bf217e2-9533-42f9-81bf-0c612834e9cb" />

3) According to the [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=10065-121926&t=b3cXALR6IqEFyoYd-1) the top FormAlert should be wider than the form. But maybe we can change that in the design?

Lionqueen94:
@bramtweedegolf Ah yes in this case FormLayout.Alert messes up the alignment, so I'd indeed not use those. 
What you can do is remove the margin-bottom from the inline Alert in de `Alert.module.css` file:
```
.inlineAlert {
  border: 1px solid;
  border-radius: var(--rounded-default);
  padding: 1.5rem;
  position: relative;
```
For the other alert you can use `margin={"mb-0"}` instead of `"mb-md"`

@jorisleker Is it fine if the alerts in the import flow are not the width of the screen but just a bit wider than the form like so:
<img width="820" height="874" alt="Image" src="https://github.com/user-attachments/assets/4ef83f50-d0d6-4723-be31-92a65fc1274a" />

PR #1957, opened 2025-08-11
Add Number of Voters and Counting Method to backend

cikzh:
Implements #1773, [#1633](https://github.com/kiesraad/abacus/issues/1633)

- Adds number of voters page
Added checks in e2e test, including the Check And Save page
- Adds saving counting method to backend
Saving is now hardcoded in the frontend, until we add support for DSO

- Does not iterate on the design which will be implemented in #1937 

@oliver3 added you as a reviewer specifically for the changes in the NumberOfVotersForm component

### DIFF
--- a/frontend/src/features/election_create/components/NumberOfVoters.tsx
+++ b/frontend/src/features/election_create/components/NumberOfVoters.tsx
@@ -30,6 +30,8 @@ export function NumberOfVoters() {
     }
   }
 
+  const showHint = !state.isNumberOfVotersUserEdited && state.numberOfVoters !== undefined;
+
   return (
     <section className="md">
       <NumberOfVotersForm
@@ -40,7 +42,7 @@ export function NumberOfVoters() {
         })}
         button="Volgende"
         onSubmit={(value) => void handleSubmit(value)}
-        hint={state.isNumberOfVotersUserEdited ? undefined : t("election.number_of_voters.hint")}
+        hint={showHint ? t("election.number_of_voters.hint") : undefined}
         error={error}
       />
     </section>


frontend/src/features/election_create/components/CheckAndSave.tsx
```diff
@@ -17,7 +18,7 @@ export function CheckAndSave() {
   const { create } = useCrud<ElectionAndCandidatesDefinitionImportRequest>({ create: path });
 
   // if no election, election data or candidate data is found in the state, go back to the beginning
```
Lionqueen94:
```suggestion
  // if no election, election data, candidate data or counting method is found in the state, go back to the beginning
```

frontend/src/features/election_create/components/NumberOfVoters.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { state, dispatch } = useElectionCreateContext();
+  const [error, setError] = useState<string | undefined>();
+  const navigate = useNavigate();
+
+  // if no data was stored, navigate back to beginning
+  if (!state.election) {
+    return <Navigate to="/elections/create" />;
+  }
+
+  async function handleSubmit(numberOfVoters: number | undefined) {
```
oliver3: No need for undefined here, it can only be a number

frontend/src/features/election_create/components/NumberOfVoters.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <section className="md">
+      <NumberOfVotersForm
+        defaultValue={state.numberOfVoters}
+        instructions={t("election_management.enter_number_of_voters", {
+          name: state.election.name,
+          location: state.election.location,
+        })}
+        button="Volgende"
+        onSubmit={(value) => void handleSubmit(value)}
+        hint={state.numberOfVoters ? t("election.number_of_voters.hint") : undefined}
```
oliver3: Why the conditional hint?

cikzh: The hint is only shown when number_of_voters is imported (and thus auto filled) from the election definition

oliver3: In that case: if you go back to this page after filling it in yourself and proceeding, you will get this message too..

frontend/src/features/election_create/components/NumberOfVoters.tsx
cikzh: Not at my computer, so I can't test, but an empty input field would yield undefined, doesn't it?

frontend/src/features/election_create/components/NumberOfVoters.tsx
Lionqueen94: I'd say give the hint as a prop to the form when it should be shown?

cikzh: > In that case: if you go back to this page after filling it in yourself and proceeding, you will get this message too..

That's true, good catch!

frontend/src/features/election_create/components/NumberOfVoters.tsx
cikzh: `undefined` is indeed unnecessary. The empty numeric input will submit `0`. Updated type and removed check in [00fbe30](https://github.com/kiesraad/abacus/pull/1957/commits/00fbe30dea81649fa36c4ac37bccd234d1285f43)

frontend/src/features/election_create/components/NumberOfVoters.tsx
cikzh: - The e2e test now covers this case in [39dfc7e](https://github.com/kiesraad/abacus/pull/1957/commits/39dfc7e9b88c5560927b1b61ae4f0550eee3406a)
- Implemented the fix in [43e8c2d](https://github.com/kiesraad/abacus/pull/1957/commits/43e8c2d39def10cec9cb91b5ffe8c99c7582aa6c)

PR #1795, opened 2025-07-17
Remove fields from data entry

praseodym:
- **Use translated strings in data entry structure**
  This simplifies the code and the ability to write tests for subsection
  components.
  

- **Remove fields from data entry**
  Remove the recounted section, recounted voters counts and voter card
  counts from data entry, because these fields are no longer present in
  the new model reports.
  
  The following validation rules related to these fields have also been
  removed:
  - F.101: recounted field not filled in
  - F.203: voters recounts incorrect total
  - W.204: large difference between voters recount and votes cast
  - W.207: voters recount and votes cast exceeding eligible voters
  - W.209: identical voters recount and votes counts

Resolves #1739

PR #1958, opened 2025-08-11
Data entry e2e test cleanup

oliver3:
A separate PR in preparation of https://github.com/kiesraad/abacus/issues/1741 to cleanup the e2e tests after adding extra investigation section

### DIFF
--- a/frontend/e2e-tests/tests/data-entry/data-entry.valid.model.e2e.ts
+++ b/frontend/e2e-tests/tests/data-entry/data-entry.valid.model.e2e.ts
@@ -3,12 +3,15 @@ import { createTestModel } from "@xstate/graph";
 import { AbortInputModal } from "e2e-tests/page-objects/data_entry/AbortInputModalPgObj";
 import { DataEntryHomePage } from "e2e-tests/page-objects/data_entry/DataEntryHomePgObj";
 import { DifferencesPage } from "e2e-tests/page-objects/data_entry/DifferencesPgObj";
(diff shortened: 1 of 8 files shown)


frontend/e2e-tests/tests/data-entry/data-entry.valid.model.e2e.ts
jschuurk-kr, line 1: The state `extraInvestigationPageEmpty` is not the result of action on a state, so the state and its action (the `SUBMIT`) can be removed. I assume the same is the case for the other two model-based tests.

jschuurk-kr: Same applies to `voterVotesPageCached`, so looks like the test was broken before your changes.

Let's solve in a new issue? Because that will require walking through all the states and actions.

jschuurk-kr: looks like it happened in PR https://github.com/kiesraad/abacus/pull/1795

oliver3: New issue sounds like a plan, also to check if we can do a well placed assertion that there are no orphaned states in the machine.

oliver3: Created issue https://github.com/kiesraad/abacus/issues/1960

jschuurk-kr (approved):

Issue #796, opened 2025-01-05
Epic: user flow testing


praseodym:
Write tests that cover a realistic user flow (create first user, create election, data entry).

Based on a PoC (branch [e2e-full-flow](https://github.com/kiesraad/abacus/tree/e2e-full-flow)) by @praseodym we can use Playwright for this, because it should be fast and stable enough. If that turns out not to be the case, we should switch to testing the same flows against the backend through APIs.

Playwright setup discussion: [#1942](https://github.com/kiesraad/abacus/issues/1942)

# Scope
- Create election, 2 polling stations and 2 entries
- Include committee session status changes (by coordinator)

PR #1962, opened 2025-08-12
Proposal: additional webserver without fixtures for e2e

marlonbaeten:
This adds an additional backend when running our e2e tests, which allows us to test abacus without any fixtures loaded. We need this to test the initialization flow.

The disadvantage of this approach is the extra time it takes to startup an additional backend. Also, this backend might not be needed for all test shards.

Any remarks or alternative ideas welcome!


marlonbaeten:
Closing for now in favor of "fixture tests" for [#796](https://github.com/kiesraad/abacus/issues/796)

PR #1967, opened 2025-08-12
Fix candidate list (EML230b) parsing when countries are included

rnijveld:
- Small change in the parsing
- Updated test file to include an example with a country
- Lots of updates to hashes


jschuurk-kr:
Tested with the candidate list I got from OSV. Import successful! :confetti_ball:

Lionqueen94 (approved):

jschuurk-kr (approved):

PR #1968, opened 2025-08-12
Improved logout

marlonbaeten:
Problem statement:

- On some browsers the logout API call takes around 3 seconds, since the browser is instructed to clear all cache using the "Clear-Site-Data" header
- The logout API call was done async from the React rendering, that could cause a state where the frontend navigates to a route with the "old" user still in iets state. The user is only cleared after the logout call succeeds.

Solution:

- This PR removed the "cache" value from the "Clear-Site-Data" header. In my opinion the keeping the browser cache after logout is, in the context of Abacus, not a security issue.
- The logout component now shows a loader until the logout call is finished and then navigates to the login page.
- Any error when logging in will be propagated to the error component.

This also fixes some Playwright tests I was writing for #796


Lionqueen94 (approved):
Works smoothly in Chrome now

jschuurk-kr:
This undoes a tiny part of https://github.com/kiesraad/abacus/pull/1466 and means we longer pass the [oshp-validator](https://github.com/oshp/oshp-validator). Not saying that's not ok, but we might want to document that somewhere. (Also considering the pen test.) cc @praseodym @ring-ring-ring

praseodym:
> This undoes a tiny part of #1466 and means we longer pass the [oshp-validator](https://github.com/oshp/oshp-validator). Not saying that's not ok, but we might want to document that somewhere. (Also considering the pen test.) cc @praseodym @ring-ring-ring

Leaving just cache data would be fine (should just be JavaScript code) but [MDN documentation on `Clear-Site-Data: "cache"`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Clear-Site-Data#cache) mentions that some browsers remove more state which can be useful from a security perspective.

But seeing this is causing issues, I'd prefer being pragmatic and `"cache"` from the header (for now).

marlonbaeten:
So I think the security issue here is that a browser might have cached a page or other resource that should be confidential after logging out.

PR #1969, opened 2025-08-12
User flow testing

marlonbaeten:
Work In progress

- Create "--seed-data" state using Playwright tests


oliver3 (approved):

frontend/e2e-tests/initialise.ts
jschuurk-kr, line 1: This is both a test and a script to set up additional users. So it would make sense to me to split it up into two files.

marlonbaeten: We have to force the order of these tests (you can only create users when the application is initialised) - I'm not sure if we can force Playwright to execute in serial if we split these up in two files.

jschuurk-kr: Two files, each with their own project and give the second a `dependencies` on the first? And arguably move the consts `firstAdmin` and `users` to a file in e2e-tests/test-data, since at least the `firstAdmin` is relevant to both files.

marlonbaeten: Fixed in 497c7055febfa992c5fcef50c21e4c881b87effe

oliver3 (approved):

jschuurk-kr (commented):
Changes to the e2e tests look good to me. Have not reviewed any of the other changes.

PR #1980, opened 2025-08-14
Fixed (read only) data entry layout (including StatusList)

### LINKED ISSUE #1939: Fix Data Entry flow styling
### Data Entry
- Change legend into legend with h2
- Adjust sectionNumber badge to updated design
- Make sure there is a 1.5rem gap between h2 and alert and between alert and inputgrid

Lionqueen94:
Closes #1939

## Description
- Now using `HeadingSubsection` for 2.1 and 2.3 on B1-2 page of data entry
- Added `<h2>` inside `<legend>` in data entry pages and in read only data entry (resolve errors) pages
- Adjusted `SectionNumber` badge and moved it to its own component
- Made sure alignment of all elements on all data entry pages is correct
- Made sure alignment of all elements on all read only data entry pages (resolve errors) is correct
- Renamed `ChoiceList.Title` to `ChoiceList.Legend` to be more explicit
- Moved `CheckboxesSubsection` title into `ChoiceList.Legend` component and changed the fieldset into a div
- Moved `RadioSubsection` title into `ChoiceList.Legend` component and changed the fieldset into a div
- Added `StatusList.Section` to wrap `StatusList.Title` and `StatusList` components and add gaps
- Added `StatusList.Wrapper` to wrap multiple `StatusList.Section` components and add gaps
- Removed all margin and padding from `StatusList` styling

Note: There are 2 css files `DataEntrySection.module.css` and `ReadOnlyDataEntrySection.module.css` with the same content. Should I think of a way to fix this now or do we potentially change this in the future? @praseodym @oliver3 

## Screenshots
Resolve errors:
image

Resolve errors read only subsection:
image

Check and save with errors and warnings:
image

Check and save without errors or warnings
image

Check and save with blocking errors:
Screenshot from 2025-08-14 16-17-37

B2-1 data entry section (now with headings):
Screenshot from 2025-08-14 15-13-24

### DIFF
--- a/frontend/src/components/data_entry/subsections/MessageSubsection.module.css
+++ b/frontend/src/components/data_entry/subsections/MessageSubsection.module.css
@@ -1,3 +1,4 @@
 .message {
   color: var(--gray-900);
+  width: 37.5rem;
 }
--- a/frontend/src/features/resolve_errors/components/ReadOnlyDataEntrySection.module.css
+++ b/frontend/src/features/resolve_errors/components/ReadOnlyDataEntrySection.module.css
@@ -1,0 +1,15 @@
+.titleContainer {
+  margin-bottom: 1.5rem;
+  h2 {
+    display: flex;
+    align-items: center;
+    gap: 1rem;
+    margin-bottom: 0;
+  }
+}
+
+.formContainer {
+  display: flex;
+  flex-direction: column;
+  gap: 1.5rem;
+}
--- a/frontend/src/features/resolve_errors/components/ReadOnlyDataEntrySection.test.tsx
+++ b/frontend/src/features/resolve_errors/components/ReadOnlyDataEntrySection.test.tsx
@@ -25,16 +25,26 @@ describe("ReadOnlyDataEntrySection", () => {
     return render(<ReadOnlyDataEntrySection section={section} data={data} validationResults={validationResults} />);
   };
 
-  test("renders section with title", () => {
+  test("renders section with title", async () => {
     renderComponent();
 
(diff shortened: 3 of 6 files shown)


frontend/src/components/data_entry/subsections/CheckboxesSubsection.tsx
```diff
@@ -23,10 +23,9 @@ export function CheckboxesSubsectionComponent({
     subsection.options.some((option) => errorsAndWarnings?.get(`data.${option.path}`));
 
   return (
-    <fieldset className={cls.container}>
-      {subsection.title && <legend>{subsection.title}</legend>}
-      {subsection.description && <p>{subsection.description}</p>}
+    <div className={cls.container}>
```
praseodym, line 26: This was changed to a `fieldset` in https://github.com/kiesraad/abacus/commit/aa6dd7bf250725cf13579676b901723fa3ce6e72#diff-c64352f95871020a459e990bc404e509aa445ff34472548d1de4912fff014c4eR26 (just two weeks ago) because it makes it easier locate the element in Playwright.

Nesting fieldsets is allowed in the HTML spec so I'm not sure why we'd have to revert that change (and break e2e tests) in this PR.

Lionqueen94: The problem was that the title was located outside the ChoiceList fieldset, I now moved the title into the ChoiceList as ChoiceList.Legend and now the extra fieldset is not needed and the e2e tests pass :+1:

oliver3 (approved):

PR #1981, opened 2025-08-14
Prevent password from showing on first login

### LINKED ISSUE #1975: Password shows briefly on login page
Issue occurred in Firefox on Linux: 

On the Inloggen page, after filling in the Gebruikersnaam and Wachtwoord fields, when you click the Inloggen button, the password is briefly visible in the Wachtwoord field. This occurs both for correct credentials and incorrect credentials.

praseodym:
Prevent password from showing on first login by changing form fields from `disabled` to `readOnly` on load. Fixes #1975.


marlonbaeten (approved):

jschuurk-kr (approved):
Tested on Firefox and Chrome (both Linux). Can no longer reproduce.

Issue #1960, opened 2025-08-12
Verify no unused states in the model based e2e tests (and fix the tests)


oliver3:
From comment https://github.com/kiesraad/abacus/pull/1958#discussion_r2268759595

> The state extraInvestigationPageEmpty is not the result of action on a state, so the state and its action (the SUBMIT) can be removed. I assume the same is the case for the other two model-based tests.

> Same applies to voterVotesPageCached, ..

So the coverage of the model-based tests needs to be checked and fixed where needed. (also see [PR 1795](https://github.com/kiesraad/abacus/pull/1795/files))

It would be nice if we could also verify this in our test so that this cannot happen anymore.

(Maybe also: update missed `pollingStationsPage..` names to `dataEntryHomePage..`)

jschuurk-kr:
Coverage that was removed in [PR 1975](https://github.com/kiesraad/abacus/pull/1795)

__valid__
- voterVotesPageEmpty -> go to previous page -> recountedPageEmpty
- votersVotesPageFilled -> go to previous page -> voterVotesPageCached -> return to VotersVotesPage -> voterVotesPageCached
- votersVotesPageChangedFilled -> go to previous page -> unsavedChangesModalChanged from where you either save or discard the changes and return to the VotersVotesPage (votersVotesPageChangedSubmitted or votersVotesPageSubmitted)

__error__
- votersVotesPageChangedToError -> go to previous page -> unsavedChangesModalChangedToError from where you either save or discard the changes
- VotersVotesPageFilledError -> recountedPageFilledError
- votersVotesPageErrorSubmitted -> recountedPageErrorSubmitted -> return to VotersVotesPage -> votersVotesPageErrorSubmitted
- votersVotesPageCorrectBackToValid -> unsavedChangesModalCorrectBackToValid from where you save the changes

__warning__
- votersVotesPageWarningSubmitted -> go to previous page -> recountedPageWarningSubmitted -> return to VotersrVotesPage -> votersVotesPageWarningSubmitted
- votersVotesPageChangedToWarning -> go to previous page -> unsavedChangesModalChangedToWarning from where you either save or discard the changes

PR #1982, opened 2025-08-14
Restore and improve model based e2e tests

oliver3:
- Resolves https://github.com/kiesraad/abacus/issues/1960
- Add extra assertion to compare machine definition states with initial state plus target states
- Move assertions to helpers
- Add type checking implemented by @stacktraceghost (thanks x 2!)
- Restore states and events removed in https://github.com/kiesraad/abacus/pull/1795 with the following changes:
  - `recountedPage` is replaced with `countingDifferencesPollingStationPage` which is now the previous page
  - Removed `unsavedChangesModalCorrectBackToValid` which was unused it seems?

### DIFF
--- a/frontend/e2e-tests/helpers-utils/xstate-helpers.ts
+++ b/frontend/e2e-tests/helpers-utils/xstate-helpers.ts
@@ -3,43 +3,80 @@ import { expect } from "@playwright/test";
 export type TestStates = Record<string, () => Promise<void>>;
 export type TestEvents = Record<string, () => Promise<void>>;
 
-interface MyMachineConfig {
+export interface MachineDefinition {
   initial: string;
   states: {
     [key: string]: { on?: Record<string, string> };
   };
 }
 
-export function getStatesAndEventsFromMachineDefinition(machineDef: MyMachineConfig) {
-  const machineFromStates: string[] = Object.keys(machineDef.states);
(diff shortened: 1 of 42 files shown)


frontend/e2e-tests/helpers-utils/xstate-helpers.ts
```diff
@@ -1,35 +1,85 @@
+import { expect } from "@playwright/test";
+
 export type TestStates = Record<string, () => Promise<void>>;
 export type TestEvents = Record<string, () => Promise<void>>;
 
-interface MyMachineConfig {
+export interface MachineDefinition {
```
stacktraceghost: No export is necessary for types `TestStates`/`TestEvents`/`MachineDefinition`, since they are not imported elsewhere.

stacktraceghost (approved):

jschuurk-kr:
@oliver3 "unsavedChangesModalCorrectBackToValid" was reached by returning to the previous page from "votersVotesPageCorrectBackToValid", but there was no action leading to "votersVotesPageCorrectBackToValid".

I don't think we need to re-introduce this, since the unsaved changes modal should not care whether you make change from error to valid, or from valid to error, or from waring to ... etc.

jschuurk-kr (approved):

Issue #1972, opened 2025-08-13
Update Na 31-2 report with new data


oliver3:
- checkboxes in polling stations list
- differences
- Resolve comment from https://github.com/kiesraad/abacus/pull/2027#discussion_r2300030681

PR #1994, opened 2025-08-18
Change data entry section B1-3.1+3.2 'Toegelaten kiezers en uitgebrachte stemmen' (voters and votes)

### LINKED ISSUE #1742: Change data entry section B1-3.1+3.2 'Toegelaten kiezers en uitgebrachte stemmen' (voters and votes)
Extend page 'Toegelaten kiezers en uitgebrachte stemmen' (voters and votes) with political group votes totals E.1 to E.n.

https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=8663-32055&t=coOg182cFYjjAsKM-4

stacktraceghost:
Resolves #1742
- Adjusted form fields
- Renamed field `votes_count.votes_cadidates_count` to `votes_count.total_votes_candidates_count`
- Renamed struct `PoliticalGroupVotes` to `PoliticalGroupCandidateVotes` due to name confusion new propery
- Added struct `PoliticalGroupTotalVotes`, added as `votes_count.political_group_total_votes`
  - Adjusted tests accordingly
  - Added implementations for comparison, addition and basic validation, incl. tests.
- Adjusted `InputGrid` according to design, column with field letters are now bold

Notes:
- Typst models use `election.political_groups` iso `input.summary.votes_counts.political_group_total_votes` for the new fields. If necessary, this could be adjusted in #1972

### DIFF
--- a/frontend/src/testing/test-utils.ts
+++ b/frontend/src/testing/test-utils.ts
@@ -83,15 +83,13 @@ export async function userTypeInputsArray<T extends Record<string, string | numb
   user: UserEvent,
   inputArray: T[],
   fieldPath: string,
-  valueKey: keyof T,
-  indexKey?: keyof T,
+  key: keyof T,
 ) {
-  for (const [i, item] of Object.entries(inputArray)) {
-    const index = indexKey && Number.isInteger(item[indexKey]) ? Number(item[indexKey]) - 1 : i;
-    const input = await screen.findByTestId(`${fieldPath}[${index}].${valueKey.toString()}`);
+  for (const [index, item] of inputArray.entries()) {
+    const input = await screen.findByTestId(`${fieldPath}[${index}].${key.toString()}`);
     await user.clear(input);
-    await user.type(input, String(item[valueKey]));
-    expect(input).toHaveValue(String(item[valueKey]));
+    await user.type(input, String(item[key]));
+    expect(input).toHaveValue(String(item[key]));
   }
 }
 
--- a/frontend/src/features/data_entry/components/voters_and_votes/VotersAndVotesForm.test.tsx
+++ b/frontend/src/features/data_entry/components/voters_and_votes/VotersAndVotesForm.test.tsx
@@ -252,7 +252,6 @@ describe("Test VotersAndVotesForm", () => {
         political_group_total_votes,
         "data.votes_counts.political_group_total_votes",
         "total",
-        "number",
       );
 
       const spy = vi.spyOn(global, "fetch");


backend/src/data_entry/validation.rs
```diff
@@ -574,6 +581,34 @@ impl Validate for VotesCounts {
...
+
+        // check each political group
+        for (i, pgv) in self.iter().enumerate() {
+            let number = pgv.number;
+            if number as usize != i + 1 {
+                return Err(DataError::new(
+                    "political group numbers are not consecutive",
+                ));
+            }
+        }
```
praseodym: Error could be slightly better:
```suggestion
        // check if the list of political group total votes has the correct length
        if election.political_groups.len() != self.len() {
            return Err(DataError::new(
                "list of political group total votes does not have correct length",
            ));
        }

        // check each political group total votes
        for (i, pgv) in self.iter().enumerate() {
            let number = pgv.number;
            if number as usize != i + 1 {
                return Err(DataError::new(
                    "political group total votes numbers are not consecutive",
                ));
            }
        }
```
(We're checking if the submitted polling station result has the same number of political groups as the election, not the other way around)

praseodym (dismissed):
lgtm, and thanks for creating separate commits for the field and struct renames!

See comment for a minor suggestion for the validation error message (which should only ever be visible during development).

backend/src/data_entry/validation.rs
stacktraceghost: Agreed! Thanks

backend/src/summary/mod.rs
```diff
@@ -486,11 +596,32 @@ mod tests {
...
+            .political_group_votes
+            .push(ps1_result.political_group_votes[1].clone());
+        ps2_result
+            .political_group_votes
+            .push(ps2_result.political_group_votes[1].clone());
         let totals = ElectionSummary::from_results(
             &election,
             &[(ps[0].clone(), ps1_result), (ps[1].clone(), ps2_result)],
         );
+        dbg!(&totals);
```
Lionqueen94:
```suggestion
```
This debug statement can be removed

frontend/src/features/resolve_differences/testing/polling-station-results.ts
```diff
@@ -20,7 +20,11 @@ export function pollingStationResultsMockData(first: boolean): PollingStationRes
       total_admitted_voters_count: 42,
     },
     votes_counts: {
-      votes_candidates_count: first ? 42 : 44,
+      political_group_total_votes: [
+        { number: 1, total: first ? 1512 : 1481 },
+        { number: 2, total: 2 },
+      ],
+      total_votes_candidates_count: first ? 42 : 44,
```
Lionqueen94: This is quite an high difference between political_group_total_votes and total_votes_candidates_count 😅

stacktraceghost: Good catch, this was based on the existing election group votes. I've adjusted the test in https://github.com/kiesraad/abacus/pull/1994/commits/f40021e1f7ab99ca933b911e3cf9a6b11ce8ef9b

backend/src/summary/mod.rs
stacktraceghost: Thanks for spotting this, I've found two more and removed them in https://github.com/kiesraad/abacus/pull/1994/commits/f40021e1f7ab99ca933b911e3cf9a6b11ce8ef9b

frontend/src/testing/test-utils.ts
```diff
@@ -79,6 +79,22 @@ export async function userTypeInputs(user: UserEvent, inputs: { [key: string]: s
...
+) {
+  for (const [i, item] of Object.entries(inputArray)) {
+    const index = indexKey && Number.isInteger(item[indexKey]) ? Number(item[indexKey]) - 1 : i;
+    const input = await screen.findByTestId(`${fieldPath}[${index}].${valueKey.toString()}`);
+    await user.clear(input);
+    await user.type(input, String(item[valueKey]));
+    expect(input).toHaveValue(String(item[valueKey]));
+  }
+}
+
```
oliver3: It took me just a bit too long to see what this did. Might I suggest a simpler variant that also works?

```suggestion
export async function userTypeInputsArray<T extends Record<string, string | number>>(
  user: UserEvent,
  inputArray: T[],
  fieldPath: string,
  key: keyof T,
) {
  for (const [index, item] of inputArray.entries()) {
    const input = await screen.findByTestId(`${fieldPath}[${index}].${key.toString()}`);
    await user.clear(input);
    await user.type(input, String(item[key]));
    expect(input).toHaveValue(String(item[key]));
  }
}
```

(NB: usage site has to be changed then too)

oliver3 (commented):
Looks good! One suggestion to simplify a test helper function

frontend/src/testing/test-utils.ts
stacktraceghost: Good point. It was built to follow the numbers of the objects to ensure the right order. However, there is already a check for consecutive numbering in election groups. I've committed your suggestion and adjusted the usage site in https://github.com/kiesraad/abacus/pull/1994/commits/4b966eb9e11f90723e222b4a8cd5b299999f113f

praseodym (approved):

PR #2008, opened 2025-08-19
End-to-end test for happy path

### LINKED ISSUE #1991: Happy flow Abacus usage

marlonbaeten:
Fixes #1991

Adds an end-to-end test for the "Happy path" usage of Abacus:

- Login as admin and create an Election and an additional polling station
- Login as coordinator and fill the session details and start the session
- Login as typist and perform the 1st data entry
- Login as the second typist and perform the 2nd data entry
- Repeat for the second polling station
- Login as coordinator, finish data entry and download the resulting PDF

Test by running the e2e tests.

### DIFF
--- a/frontend/e2e-tests/page-objects/authentication/LoginPgObj.ts
+++ b/frontend/e2e-tests/page-objects/authentication/LoginPgObj.ts
@@ -12,4 +12,10 @@ export class LoginPgObj {
     this.loginBtn = page.getByRole("button", { name: "Inloggen" });
     this.alert = page.getByRole("alert");
   }
+
+  async login(username: string, password: string) {
+    await this.username.fill(username);
+    await this.password.fill(password);
+    await this.loginBtn.click();
+  }
 }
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -18,9 +18,9 @@
     "test": "vitest",
     "test:e2e": "cross-env LOCAL_CI=true RUST_LOG=tower_http-warning playwright test",
     "test:e2e-dev": "cross-env RUST_LOG=tower_http-warning playwright test",
-    "test:e2e-debug": "cross-env DEBUG_DEVELOPMENT=true playwright test --debug --no-deps",
+    "test:e2e-debug": "cross-env DEBUG_DEVELOPMENT=true playwright test --project firefox --debug --no-deps",
```
jschuurk-kr, line 21: What's the reason for defaulting to firefox here?

praseodym (approved):

praseodym (approved):

frontend/package.json
marlonbaeten: Could have been chrome - but I got the impression that there are a lot of firefox users in our team. I do think that we only want to debug tests in a single browser since you want to have fast iterations when using this command.

jschuurk-kr: Makes sense. Easier to default to a browser then have to always remember to add the project flag.

Lionqueen94 (approved):

PR #2015, opened 2025-08-20
Fix bug and styling for election import flow

bramtweedegolf:
Fixed:
- ~Bug: Error state for form fields when entering wrong hash on check election definition page is missing and hashInput should also turn red instead of blue (does work when leaving fields empty)~
- Import election page gap between button and text should be 1.5rem
- Add FormLayout to each Form
- Move title to Form title
- Add all form content into a FormLayout.Section for each form
- Add button to FormLayout.Controls in the FormLayout for each form
- Also compare all pages with design and adjust more where needed
- Check / update Shift+Enter keyboard hints

Fixes https://github.com/kiesraad/abacus/issues/1937

### DIFF
--- a/frontend/src/features/election_create/components/CheckHash.tsx
+++ b/frontend/src/features/election_create/components/CheckHash.tsx
@@ -1,4 +1,4 @@
-import { FormEvent, ReactNode, useEffect, useState } from "react";
+import { FormEvent, ReactNode, useState } from "react";
 
 import { Alert } from "@/components/ui/Alert/Alert";
 import { Button } from "@/components/ui/Button/Button";
@@ -61,32 +61,19 @@ export function CheckHash({ date, title, header, description, redactedHash, erro
           newStub.error = t("election.check_eml.check_hash.hint");
           stubsAreValid = false;
         } else {
-          if (error) {
-            newStub.error = t("error.api_error.InvalidHash");
-          }
           completeHash[stub.index] = value;
         }
         setStubs(newStubs);
       }
     }
 
-    // Only allow submit when a field has been focussed or blurred
+    // Only allow form submit when a field has been focussed or blurred
     // and both values are the correct length and type
     if (changed && stubsAreValid) {
       onSubmit(completeHash);
     }
     setChanged(false);
   }
-
-  // If there is an error, add error to stubs
-  useEffect(() => {
-    setStubs((prevStubs) =>
-      prevStubs.map((stub) => {
-        stub.error = error ? t("error.api_error.InvalidHash") : "";
-        return stub;
-      }),
-    );
-  }, [error]);
 
   return (
     <section className="md">
@@ -134,7 +121,6 @@ export function CheckHash({ date, title, header, description, redactedHash, erro
               />
             ))}
           </FormLayout.Section>
-
           <FormLayout.Controls>
             <Button type="submit">{t("next")}</Button>
             <KeyboardKeys keys={[KeyboardKey.Shift, KeyboardKey.Enter]} />
--- a/frontend/src/i18n/locales/nl/error.json
+++ b/frontend/src/i18n/locales/nl/error.json
@@ -21,7 +21,7 @@
     "InternalServerError": "Er is een interne fout opgetreden",
     "InvalidCommitteeSessionStatus": "De zitting is niet in de juiste status voor deze actie",
     "InvalidData": "De invoer is niet geldig",
-    "InvalidHash": "Code komt niet overeen",
+    "InvalidHash": "De hash is niet geldig",
     "InvalidJson": "De JSON is niet geldig",
     "InvalidPassword": "Het opgegeven wachtwoord is onjuist",
     "InvalidPoliticalGroup": "De politieke partij is niet geldig",


frontend/src/components/ui/InputField/InputField.module.css
```diff
@@ -22,6 +22,10 @@
       width: 10rem;
     }
 
+    &:global(.narrow) ~ span:global(.error) {
```
oliver3: Could you use `.full-field-with-narrow-input` ?

frontend/src/features/election_create/components/UploadPollingStationDefinition.tsx
```diff
@@ -92,73 +94,76 @@ export function UploadPollingStationDefinition() {
...
+                </>
+              )}
+            </FormLayout.Section>
+          </FormLayout>
+        </Form>
+
+        <p className="mt-lg">
           <Button onClick={() => void next()}>{t("next")}</Button>
-        </div>
+        </p>
```
Lionqueen94: This can be moved into a FormLayout.Controls as well I believe. While adding type="button" to the Button component to not trigger a submit.

frontend/src/features/election_create/components/UploadPollingStationDefinition.tsx
```diff
@@ -83,48 +85,53 @@ export function UploadPollingStationDefinition() {
...
-        <Button variant="underlined" size="md" onClick={() => void skip()}>
-          {t("election.polling_stations.skip_step")}
-        </Button>
-      </p>
+            <FormLayout.Controls>
+              <Button type="button" variant="underlined" size="md" onClick={() => void skip()}>
+                {t("election.polling_stations.skip_step")}
+              </Button>
+            </FormLayout.Controls>
+          </FormLayout.Section>
```
Lionqueen94, line 132:
```suggestion
          </FormLayout.Section>
          <FormLayout.Controls>
            <Button type="button" variant="underlined" size="md" onClick={() => void skip()}>
              {t("election.polling_stations.skip_step")}
            </Button>
          </FormLayout.Controls>
```

frontend/src/features/polling_stations/components/PollingStationImportPage.tsx
```diff
@@ -94,50 +96,54 @@ export function PollingStationImportPage() {
...
+          )}
+
+          <p>{t("polling_station.import_instructions")}</p>
+
+          <FormLayout.Controls>
+            <FileInput id="upload-eml" onChange={(e) => void onFileChange(e)}>
+              {t("select_file")}
+            </FileInput>
+          </FormLayout.Controls>
+        </FormLayout.Section>
```
Lionqueen94, line 115:
```suggestion
        </FormLayout.Section>
        <FormLayout.Controls>
          <FileInput id="upload-eml" onChange={(e) => void onFileChange(e)}>
            {t("select_file")}
          </FileInput>
        </FormLayout.Controls>
```

frontend/src/components/ui/InputField/InputField.module.css
bramtweedegolf: Thanks, I overlooked that one. Updated in https://github.com/kiesraad/abacus/pull/2015/commits/d0415d467084b7e6103573ec86c60410ff2dc750

frontend/src/features/election_create/components/UploadPollingStationDefinition.tsx
bramtweedegolf: Thanks! Updated in https://github.com/kiesraad/abacus/pull/2015/commits/d0415d467084b7e6103573ec86c60410ff2dc750

frontend/src/features/election_create/components/UploadPollingStationDefinition.tsx
bramtweedegolf: This change  would result in slightly too much margin above the skip button  (2 rem, should be 1.5 according to the [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=1455-13075&m=dev).) 

Screenshot_2025-08-25_11-56-37

frontend/src/features/polling_stations/components/PollingStationImportPage.tsx
bramtweedegolf: Same here, margin would be 2 rem, but should be 1.5 rem according to the [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=11045-148078&m=dev)

Screenshot_2025-08-25_12-09-38

frontend/src/features/election_create/components/UploadPollingStationDefinition.tsx
Lionqueen94: Then you don't need to use the Controls at all I suppose, but it's fine like this

frontend/src/features/election_create/components/CheckHash.tsx
```diff
@@ -34,8 +47,12 @@ export function CheckHash({ date, title, header, description, redactedHash, erro
     const stub = newStubs[stubIndex];
     if (stub) {
       stub.selected = highlight;
+      stub.error = "";
     }
     setStubs(newStubs);
+    if (!highlight) {
```
oliver3: This causes the alert on top to disappear on focus/blur. @Lionqueen94 and I prefer to keep the alert, and ideally also the input field color and message, until submit.

bramtweedegolf: Updated in https://github.com/kiesraad/abacus/pull/2015/commits/6c276d792d5228d975ffce69a034b5a6cdb7426c. 

The alert and the input field color and message remain until submit.
To prevent resubmitting when nothing has changed, we update a 'changed' var on focus or blur of the input fields.

frontend/src/features/election_create/components/ElectionCreate.test.tsx
```diff
@@ -273,6 +273,8 @@ describe("Election create pages", () => {
   test("It shows an error on invalid input", async () => {
     overrideOnce("post", "/api/elections/import/validate", 200, electionValidateResponse(newElectionMockData));
 
+    vi.spyOn(console, "error").mockImplementation(() => {});
```
oliver3: Mocking of console error is not needed and can hide certain issues. Please remove it again.

frontend/src/features/election_create/components/CheckHash.tsx
```diff
@@ -69,19 +57,21 @@ export function CheckHash({
...
           completeHash[stub.index] = value;
         }
         setStubs(newStubs);
       }
     });
 
-    // Only submit when there are no errors
-    if (stubs.every((stub) => stub.error === "")) {
+    // Only allow submit when a field has been focussed or blurred
+    if (changed) {
```
oliver3: Previously the frontend did not allow a hash which was too short to be submitted, now there still is a check in the frontend which shows an error, but the submit still goes through. You have seen this because you had to change the test to make it pass again. Why was this changed?

frontend/src/features/election_create/components/ElectionCreate.test.tsx
bramtweedegolf: Removed in https://github.com/kiesraad/abacus/pull/2015/commits/d9548c472c023542799ef10ea08700aade1c7d30

frontend/src/features/election_create/components/CheckHash.tsx
bramtweedegolf: Thanks for spotting, fixed in https://github.com/kiesraad/abacus/pull/2015/commits/d9548c472c023542799ef10ea08700aade1c7d30

oliver3: Also removed the `overrideOnce` to let it fail might this code break again in https://github.com/kiesraad/abacus/pull/2015/commits/aba859c7512627bb3c2f2dcde7c018a2898109a5

Lionqueen94 (changes requested):
When the one code is entered correctly and the other one is not, the first field still shows an error:
image

Lionqueen94:
> When the one code is entered correctly and the other one is not, the first field still shows an error: image

Since the validate endpoint only returns that the hash is invalid, I removed marking the fields as invalid and showing an error for each field. This can be confusing for the user if only one of the codes entered is invalid. See 804b59d9625e9e01c6ac4a0b58356fd39a92b259

Lionqueen94 (approved):

PR #2039, opened 2025-08-25
Let the `dev` script wait for the backend to start

marlonbaeten:
Small addition to `frontend/scripts/check_server.js` so that this script actively waits for the backend to start before starting the frontend.


praseodym (approved):
I also wouldn't mind dropping this script entirely.

PR #2058, opened 2025-08-26
Move polling stations under committee sessions

### LINKED ISSUE #2001: Polling station model restructure
Adjust data structure so that polling stations are linked to committee sessions instead. 

`polling_stations.id_prev_session` is a new field that refers to the `polling_stations.id` of the previous committee session, since polling stations will be copied from one to another. 

Possible ER diagram where polling stations are linked to committee sessions instead:
```mermaid
erDiagram
      elections {
          INTEGER id PK
          TEXT name
          TEXT counting_method
          TEXT election_id
          TEXT location
          TEXT domain_id
          TEXT category
          INTEGER number_of_seats
          TEXT election_date
          TEXT nomination_date
          TEXT political_groups
      }

      committee_sessions {
          INTEGER id PK
          INTEGER number
          INTEGER election_id FK
          TEXT location
          TEXT start_date
          TEXT start_time
          TEXT status
          INTEGER number_of_voters
      }

      polling_stations {
          INTEGER id PK
          INTEGER committee_session_id FK
          INTEGER id_prev_session FK
          TEXT name
          INTEGER number
          INTEGER number_of_voters
          TEXT polling_station_type
          TEXT address
          TEXT postal_code
          TEXT locality
      }

      polling_station_data_entries {
          INTEGER polling_station_id FK
          TEXT state
          TEXT updated_at
      }

      polling_station_results {
          INTEGER polling_station_id FK
          TEXT data
          TEXT created_at
      }

      elections ||--o{ committee_sessions : "has"
      committee_sessions ||--o{ polling_stations : "has"
      polling_stations ||--o{ polling_stations : "has parent"
      polling_stations ||--o{ polling_station_data_entries : "has"
      polling_stations ||--o{ polling_station_results : "has"
```

rnijveld:
Closes #2001


backend/fixtures/election_5.sql
```diff
@@ -308,6 +308,7 @@ INSERT INTO committee_sessions (id, number, election_id, status, location, start
 VALUES (5, 1, 5, 'data_entry_finished', 'Grote Stad', '2026-03-19', '09:15', 2000),
        (6, 2, 5, 'data_entry_in_progress', '', '', '', 2000);
 
-INSERT INTO polling_stations (id, election_id, name, number, number_of_voters, polling_station_type, address,
+INSERT INTO polling_stations (id, committee_session_id, id_prev_session, name, number, number_of_voters, polling_station_type, address,
                               postal_code, locality)
-VALUES (8, 5, 'Testgebouw', 41, NULL, 'FixedLocation', 'Testweg 3', '1234 QA', 'Grote Stad');
+VALUES (9, 5, NULL, 'Testgebouw', 41, NULL, 'FixedLocation', 'Testweg 3', '1234 QA', 'Grote Stad'),
+(8, 6, 9, 'Testgebouw', 41, NULL, 'FixedLocation', 'Testweg 3', '1234 QA', 'Grote Stad');
```
Lionqueen94, line 314: The polling station ids here should be switched right? You first create the polling station with id 9 and then the one with id 8

rnijveld: I decided to switch it around so I don't have to fix test ids all over the place. Also a good test that we're not using id to sort :)

bramtweedegolf (dismissed):

backend/src/polling_station/repository.rs
```diff
@@ -37,50 +46,91 @@ pub async fn get(conn: impl DbConnLike<'_>, id: u32) -> Result<PollingStation, s
...
+        JOIN committee_sessions AS c ON c.id = p.committee_session_id
+        WHERE p.id = $1
         "#,
         id
     )
     .fetch_one(conn)
     .await
 }
 
+pub async fn get_by_previous_id(
```
oliver3, line 75: In what situation would you use `get_by_previous_id()`? Anything else than for a test and if so could the test be rewritten so this is maybe not needed?

backend/src/committee_session/api.rs
```diff
@@ -61,29 +61,31 @@ pub async fn committee_session_create(
...
-    if committee_session.status == CommitteeSessionStatus::DataEntryFinished {
-        let committee_session = crate::committee_session::repository::create(&pool, {
+    let current_committee_session =
+        crate::committee_session::repository::get_election_committee_session(
+            &pool,
+            request.election_id,
+        )
+        .await?;
+    if current_committee_session.status == CommitteeSessionStatus::DataEntryFinished {
+        // retrieve the polling stations from the last committee session
```
oliver3: I don't understand this comment, is it still correct?

rnijveld: Nope, it is an old comment, removed it in 7c4f82d0e4034c05fbd5b1f7b521d0e2ba00fdcf

backend/src/polling_station/repository.rs
rnijveld: I do think that the test that uses this function is still useful. This comment triggered me to look at that test one more time and I actually updated that test and found a small bug where a user could still claim a polling station for the id in the previous committee session. I've also made this function itself a `#[cfg(test)]` function as it currently isn't being used anywhere outside of tests.

oliver3: Nice find :) and adding `#[cfg(test)]` makes it more clear at least to me, thanks.

Lionqueen94 (approved):

PR #2045, opened 2025-08-25
Frontend for add investigation in new committee session

marlonbaeten:
Fixes #1716
Fixes #1998

- Select polling station for investigation [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12603-406219&m=dev)
- Add investigation reason [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12640-6952&m=dev)
- Add print corrigendum page [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12660-64441&m=dev)

Note that these are just the frontend components, not the backend functionality or interactions with any new backend functionality.

To test:

- Go to https://69721eb269d86968c0cff8fe0c66c479964605d1.abacus-test.nl/dev and login as a coordinator
- Go to the path `/elections/1/investigations` and navigate to the 4 newly created components / pages.

Issue #1902, opened 2025-08-04
Backend: Add polling stations list to committee session


Lionqueen94:
- Pre-fill for first committee session with all polling stations
- Endpoint for adding polling stations to subsequent committee session

PR #2059, opened 2025-08-26
Frontend for investigations findings page

### LINKED ISSUE #1999: Frontend: Add findings page

marlonbaeten:
Merge #2045 first

Fixes #1999

Adds and investigation findings page:

image


frontend/src/features/investigations/components/InvestigationFindings.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    if (error) {
+      return;
+    }
+
+    // TODO: Handle form submission
+    // const correctedResults = correctedResultsChoice === "yes";
+    // console.log({ findings, correctedResults });
+    void navigate("../../../");
```
praseodym, line 46: This relative navigation path is not ideal and cause issues

praseodym (dismissed):

frontend/src/features/investigations/components/InvestigationFindings.tsx
Lionqueen94: I'll add it to the backend issue for adding the endpoint that it should be changed when adding the form submission handling :+1: (see #1902)

jschuurk-kr:
@marlonbaeten Shift+Enter does not submit on "Aanleiding en opdracht van het centraal stembureau". I think it shoud? cc @jorisleker

(edit): Never mind. Shift + Enter does submit when the input field does not have focus. Makes sense and is consistent with other input field in the flow to not submit when the input field has focus, because of conventions around Enter and Shift+Enter when using text input fields.

frontend/src/features/investigations/components/InvestigationPrintCorrigendum.tsx
```diff
@@ -1,59 +1,60 @@
...
-              <li>{tx("investigations.print_corrigendum_instructions.download_and_print")}</li>
-              <li>{tx("investigations.print_corrigendum_instructions.print_recommendation")}</li>
+              <li>{tx("investigations.print_corrigendum.download_and_print")}</li>
+              <li>{tx("investigations.print_corrigendum.print_recommendation")}</li>
             </ul>
             <DownloadButton
               icon="download"
               href="#" // TODO insert download URL
-              title={t("investigations.download_corrigendum_link", { number: pollingStationId })}
+              title={t("investigations.print_corrigendum.download_corrigendum_link", { number: pollingStationId })}
```
jschuurk-kr: This should be the pollginStationNumber, not the Id:

Image

Lionqueen94: Good find! Fixed in 485bf1e5d3a786c2c569fdc54b2d00e6b59b482f

PR #2060, opened 2025-08-26
Link report page to committee session and prevent resuming previous session

Lionqueen94:
- Removed get election committee sessions endpoint and added committee sessions to get election endpoint
- Changed ReportPage URL to include committee session id
- Only show resume committee session button when on ReportPage for current committee session

Note: To not inflate this PR too much, I separated out issue #2061 to generate the telbestand and proces-verbaal once and save them in the database before serving to the frontend. So currently the generated files are not matching the committee session the ReportPage is opened for.


backend/src/election/api.rs
```diff
@@ -99,10 +99,14 @@ pub async fn election_details(
 ) -> Result<Json<ElectionDetailsResponse>, APIError> {
     let election = crate::election::repository::get(&pool, id).await?;
     let polling_stations = crate::polling_station::repository::list(&pool, id).await?;
-    let committee_session =
+    let current_committee_session =
         crate::committee_session::repository::get_election_committee_session(&pool, id).await?;
+    let committee_sessions =
+        crate::committee_session::repository::get_election_committee_session_list(&pool, id)
```
praseodym: Can we combine this into a single database operation?

frontend/src/features/election_management/routes.tsx
```diff
@@ -21,12 +21,17 @@ export const electionManagementRoutes: RouteObject[] = [
         Component: FinishDataEntryPage,
       },
       {
-        path: "download",
-        element: __API_MSW__ ? (
-          <NotAvailableInMock title={`${t("election.title.finish_data_entry")} - Abacus`} />
-        ) : (
-          <ElectionReportPage />
-        ),
+        path: "committee-session/:sessionNumber/download",
```
praseodym: What is the reason for using the committee session number instead of its id?

Lionqueen94: Good point, changed to id in 44e25ae92dad5cea22720fb475ad496393587d60

backend/src/election/api.rs
Lionqueen94: Done in 44e25ae92dad5cea22720fb475ad496393587d60

praseodym (approved):

bramtweedegolf (approved):
Looks good!

praseodym (approved):

PR #2067, opened 2025-08-27
Update test election generation tool

### LINKED ISSUE #2065: Votes in gen-test-election are always zero
Running e.g. `cargo run --bin gen-test-election -- -rs --political-groups 10 --with-data-entry` will have 0 votes in the created polling station results.

praseodym:
- **Update test election generation tool**
  - Fixes #2065
  - Validate created results to catch errors
  - Add random choices for `extra_investigation` and
    `counting_differences_polling_station`
  

- **Make `DifferencesCounts` match paper report**

### DIFF
--- a/backend/src/bin/gen-test-election.rs
+++ b/backend/src/bin/gen-test-election.rs
@@ -512,10 +512,7 @@ fn generate_polling_station_results(
             // 3% weight - yes to investigation, no to recount
             3,
             ExtraInvestigation {
-                extra_investigation_other_reason: YesNo {
-                    yes: true,
-                    no: false,
-                },
+                extra_investigation_other_reason: YesNo::yes(),
                 ballots_recounted_extra_investigation: YesNo::no(),
             },
         ),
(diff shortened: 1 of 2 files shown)


backend/src/bin/gen-test-election.rs
```diff
@@ -466,9 +489,73 @@ fn generate_polling_station_results(
...
+                extra_investigation_other_reason: YesNo::no(),
+                ballots_recounted_extra_investigation: YesNo::no(),
+            },
+        ),
+        (
+            // 3% weight - yes to investigation, no to recount
+            3,
+            ExtraInvestigation {
+                extra_investigation_other_reason: YesNo {
+                    yes: true,
```
oliver3: You could use the new `yes()` here too

praseodym: Thanks, I added `yes()` after I wrote that code. Fixed in cb7242ed952634a8ebfdfefe9118b899e3a83e7e.

oliver3 (approved):
Good idea to have validation in there!

PR #2082, opened 2025-09-01
Fixed bug committee session status remains created when importing polling stations

### LINKED ISSUE #2081: Bug: Committee session status remains "Created" when importing polling stations
Found by @marjoleintamis

Lionqueen94:
Closes #2081 

Committee session status is now set to "DataEntryNotStarted" when importing polling stations. 
e2e and backend checks have been added.

### DIFF
--- a/frontend/e2e-tests/page-objects/nav_bar/AdminNavBarPgObj.ts
+++ b/frontend/e2e-tests/page-objects/nav_bar/AdminNavBarPgObj.ts
@@ -16,6 +16,6 @@ export class AdminNavBar {
   }
 
   getElectionBreadcrumb(electionName: string) {
-    return this.navigation.getByRole("link", { name: new RegExp(electionName) });
+    return this.navigation.getByRole("link", { name: electionName });
   }
 }


marjoleintamis:
Tested it with the same xmls that I used before, and it works!

oliver3 (dismissed):

jschuurk-kr (changes requested):
If I create an election with polling stations through the initialisation flow, I get two committee sessions with the same number and election_id in the database:
```
sqlite> select * from committee_sessions;
              id = 1
          number = 1
     election_id = 1
        location = 
      start_date = 
      start_time = 
          status = created
number_of_voters = 612694

              id = 2
          number = 1
     election_id = 1
        location = 
      start_date = 
      start_time = 
          status = data_entry_not_started
number_of_voters = 612694
```

Thought it might be because of React strict mode, but it's reproducible with memory-serve enabled.

jschuurk-kr:
@Lionqueen94 The same issue still exists when importing the polling stations after creating an election. I've extended an existing e2e test for this scenario, hence the pipeline now failing.

I'm surprised fixing this in one place did not fix it in the other. (I checked and the issue also exists on main.) So that suggests the code could use some refactoring.

Lionqueen94:
> @Lionqueen94 The same issue still exists when importing the polling stations after creating an election. I've extended an existing e2e test for this scenario, hence the pipeline now failing.
> 
> I'm surprised fixing this in one place did not fix it in the other. (I checked and the issue also exists on main.) So that suggests the code could use some refactoring.

Good one, apparently the code was duplicated for the polling station import. I now added a separate function that is called in both endpoints and includes the audit logging and committee session status change

backend/src/polling_station/mod.rs
```diff
@@ -312,6 +311,49 @@ pub struct PollingStationsRequest {
...
+            committee_session.id,
+            CommitteeSessionStatus::DataEntryNotStarted,
+            pool.clone(),
+            audit_service,
+        )
+        .await?;
+    }
+
+    Ok(polling_stations)
+}
```
praseodym, line 355: Maybe good to execute all this in the same database transaction?

Lionqueen94: I don't think I know how to change this. Can you show me how to do this?

praseodym: I'll create a separate PR because it's not insignificant work

praseodym (approved):

PR #2084, opened 2025-09-01
Configure `testIdAttribute` for Storybook tests

praseodym:
This allows `getByTestId` to work in Storybook interaction tests just as in Vitest and Playwright.


jschuurk-kr (approved):

praseodym:
@oliver3 if the tests didn't fail, were the `data-testid` attributes even used?

oliver3:
> @oliver3 if the tests didn't fail, were the `data-testid` attributes even used?

Yes, their values were the same as the id, so it transitioned quietly into using the `id` instead.

PR #2087, opened 2025-09-01
Updated validation feedback texts

Lionqueen94:
- Removed W.3xx warnings and renamed W.205 to W.204 to make sure the warnings are incremental
- Removed validation for W.3xx warnings and related tests
- Corrected all feedback copy
- Updated all tests to new feedback copy
- Removed tests for imagined warnings

See [Storybook](https://98f9faf8.kiesraad-abacus.pages.dev/storybook/?path=/story/components-ui-feedback--coordinator) for the Feedback Component with updated copy


frontend/src/features/resolve_errors/components/ReadOnlyDataEntrySection.test.tsx
```diff
@@ -68,20 +68,18 @@ describe("ReadOnlyDataEntrySection", () => {
     expect(screen.queryByText("F.303")).not.toBeInTheDocument();
   });
 
-  test("shows warning feedback for on correct section", () => {
+  test("shows warning feedback for correct section", () => {
     const validationResults: ValidationResults = {
       errors: [],
       warnings: [
         validationResultMockData.W201, // blank_votes_count warning
-        validationResultMockData.W301, // differences_counts warning (should not show)
```
stacktraceghost, line 76: We could test a different warning here to see if it does not show

Lionqueen94: The problem is we only have warnings for this page (which will show) and the general W.001 one that can show on any page. Any suggestions on how to go about this?

stacktraceghost: Ah I see. We could (inline) place a custom validationResult here? For example:
```
      warnings: [
        validationResultMockData.W201,
        {
            fields: ["data.differences_counts.more_ballots_count"],
            code: "W999",
         }
     ]
          
```

Lionqueen94: The code is validated against type `ValidationResultCode` :') so making up a code also is not allowed

stacktraceghost: `"W999" as ValidationResultCode` works or we could use an existing code, as long as the field is pointing towards a field on another page it should not show.

Lionqueen94: I re-added some checks for (imagined) warnings in 63185c873ffd1dae013ddaa3a18ce77f871f622f

frontend/src/features/data_entry/components/differences/DifferenceForm.test.tsx
```diff
@@ -370,77 +370,32 @@ describe("Test DifferencesForm", () => {
   });
 
   describe("DifferencesForm warnings", () => {
-    test("clicking next without accepting warning results in alert shown and then accept warning", async () => {
+    test("Imagined warning on this form", async () => {
```
oliver3: If I understood correctly, this test is here to test warnings for this section but there are no warnings for the current models in this screen, just like the `Imagined warning on this form` test in `CandidatesVotesForm.test.tsx`. 

I think we should keep the tests in these tests (DifferenceForm, CandidatesVotesForm etc) to test for real world scenarios because it is all rendered by DataEntrySection etc and rendering warning should be tested there or implicitly already in other sections. 

What do you think @praseodym

praseodym: Please remove this test and other similar tests covering generic functionality (i.e. not specific to the differences form section). It doesn't make sense to have these tests at this level, they are remnants from before the `DataEntrySection` was written.

Lionqueen94: Removed in 5d7646c14736c553530b60685a46b3244b4081a6

oliver3 (approved):

frontend/src/features/data_entry/components/candidates_votes/CandidatesVotesForm.test.tsx
```diff
@@ -519,120 +517,4 @@ describe("Test CandidatesVotesForm", () => {
       expectFieldsToNotHaveIcon(expectedValidFieldIds);
     });
   });
-
-  describe("CandidatesVotesForm warnings", () => {
-    test("Imagined warning on this form", async () => {
```
lama-d3v, line 524: What's the reason for deleting these tests?

praseodym: See https://github.com/kiesraad/abacus/pull/2087#discussion_r2316275609

Issue #2088, opened 2025-09-02
Signing out during election creation does not sign out


jschuurk-kr:
Steps to reproduce:
1. Log in as admin.
2. Start adding a new election (/elections/create).
3. Upload the election definition, complete the hash, click "Volgende".
4. You're now on the page to upload the candidates list.
5. Click "Afmelden" to sign out.
6. Click "Verkiezing niet opslaan" in the modal.

Result:
- User is not signed out.
- No API calls in Network-tab of the dev tools.
- User is returned to the first page of election creation (upload election definition).

PR #2035, opened 2025-08-22
Fix flaky test 'OverviewPage > Renders elections for typist'

Lionqueen94:
Flaky test should be fixed now and elections still get refetched when clicking "Naar startscherm" on the CommitteeSessionPaused modal.
- Removed `ElectionListProvider` in favor of a get request in the `OverviewPage` directly
- Also added the get request for elections to `DevHomePage`
- Also added an every 30s refetch for `OverviewPage` (resolves #2044)

PR #2091, opened 2025-09-02
Fix signout does not work during election creation bug

bramtweedegolf:
The dispatch({ type: "RESET" }) lead to a navigation, which blocks the navigation from the blocker.proceed().

But as it turns out, explicitly clearing the state is not necessary because navigating away from the election create will clear the state in any case. So the dispatch can just be removed; the behavior of the back and forward buttons remains unchanged.

Note: I intended to add a test to cover this case, but logging out during a test breaks subsequent/parallel tests.

Fixes https://github.com/kiesraad/abacus/issues/2088

### DIFF
--- a/frontend/src/features/election_create/components/AbortModal.tsx
+++ b/frontend/src/features/election_create/components/AbortModal.tsx
@@ -8,7 +8,7 @@ import { t, tx } from "@/i18n/translate";
 import { useElectionCreateContext } from "../hooks/useElectionCreateContext";
 
 export function AbortModal() {
-  const { state, dispatch } = useElectionCreateContext();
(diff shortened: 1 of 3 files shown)


praseodym (approved):
Slightly related to https://github.com/kiesraad/abacus/pull/2035/commits/25e9b5199f81e491e014db817f08172949139606 :)

jschuurk-kr:
Tested the fix and logging out now works while in the election creation wizards.

@bramtweedegolf For the e2e test, we could add a Playwright fixture that creates a new user. Am not sure we should, though. And it does make me wonder if there are other parts of the application where we might have a similar bug.

Issue #2089, opened 2025-09-02
Implement GSB/CSB selection page for election create


bramtweedegolf:
Implement GSB/CSB selection page for election create

Only GSB is possible for now,  the CSB checkbox will be disabled
and the GSB checkbox enabled by default.

Design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=10471-70987&t=iGed9kCuinL3HWhx-11

PR #2092, opened 2025-09-02
GSB/CSB selection page

bramtweedegolf:
Implemented the GSB/CSB. Note that the CSB selection is disabled for now, and the selection is not stored or posted to the backend. 

Design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=14634-17569&m=dev

Fixes https://github.com/kiesraad/abacus/issues/2089

### DIFF
--- a/frontend/src/features/election_create/components/PollingStationRole.tsx
+++ b/frontend/src/features/election_create/components/PollingStationRole.tsx
@@ -63,7 +63,9 @@ export function PollingStationRole() {
             </ChoiceList>
           </FormLayout.Section>
 
-          <Button type="submit">{t("next")}</Button>
+          <FormLayout.Controls>
+            <Button type="submit">{t("next")}</Button>
+          </FormLayout.Controls>
         </FormLayout>
       </Form>
     </section>
--- a/frontend/src/features/election_create/components/PollingStationRole.tsx
+++ b/frontend/src/features/election_create/components/PollingStationRole.tsx
@@ -1,0 +1,76 @@
+import { FormEvent } from "react";
+import { Navigate, useNavigate } from "react-router";
+
+import { Button } from "@/components/ui/Button/Button";
+import { ChoiceList } from "@/components/ui/CheckboxAndRadio/ChoiceList";
+import { Form } from "@/components/ui/Form/Form";
+import { FormLayout } from "@/components/ui/Form/FormLayout";
+import { KeyboardKeys } from "@/components/ui/KeyboardKeys/KeyboardKeys";
+import { t } from "@/i18n/translate";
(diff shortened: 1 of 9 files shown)


jschuurk-kr:
Tested the page. Found an issue and a non-issue.

Issue: The page says you can use shift+enter to submit, but that only works after you put focus on the checkboxes. Suggested solution: remove the hint about shift+enter. The similar "Type stemopneming" page already does this.

Non-issue: The design does not have "(Niet beschikbaar in deze versie)" in bold. However, a similar message is bold on the "Type stemopneming"-page, so I agree with it being bold. I left a comment on the design about this.

frontend/src/features/election_create/components/PollingStationRole.tsx
```diff
@@ -0,0 +1,76 @@
...
+                }
+                checked={false}
+                disabled
+              ></ChoiceList.Checkbox>
+            </ChoiceList>
+          </FormLayout.Section>
+
+          <FormLayout.Controls>
+            <Button type="submit">{t("next")}</Button>
+            <KeyboardKeys keys={[KeyboardKey.Shift, KeyboardKey.Enter]} />
```
jschuurk-kr: The page says you can use shift+enter to submit, but that only works after you put focus on the checkboxes. Suggested solution: remove the hint about shift+enter. The similar "Type stemopneming" page already does this.

jschuurk-kr: done in 7e88862f

frontend/src/features/election_create/components/PollingStationRole.tsx
```diff
@@ -0,0 +1,71 @@
...
+                    <b>{t("election.polling_station_type.csb_not_supported")}</b>)
+                  </span>
+                }
+                checked={false}
+                disabled
+              ></ChoiceList.Checkbox>
+            </ChoiceList>
+          </FormLayout.Section>
+
+          <Button type="submit">{t("next")}</Button>
```
Lionqueen94:
```suggestion
          <FormLayout.Controls>
            <Button type="submit">{t("next")}</Button>
          </FormLayout.Controls>    
```
For the layout we need the Controls component around the button

jschuurk-kr: I removed the FormLayout.Controls to make it consistent with CountingMethodType. And with a single element inside the Controls component does not make a difference. However, in UploadPollingStationDefinition we do have two FormLayout.Controls with only a single Button in them.

I added the FormLayout.Controls to PollingStationRole and CountingMethodType in 979d1251 to make the components consistent.

PR #2094, opened 2025-09-02
Rework of PollingStationResults to an enum

### LINKED ISSUE #2047: Backend: Rework `PollingStationResults` to a single variant enumeration
Rework to an enum with a single variant containing the existing data of `PollingStationResults`

Each model should correspond with a specific variant of the PollingStationResults, i.e. `PollingStationResults` currently implements Na 31-2.

rnijveld:
- Fixes #2047
- Backend:
  - `PollingStationResults` is now an enum containing variants for the different models, starting with only `CSOFirstSessionResults` for now. Serialization will add the field `model` to distinguish the variants in the frontend.
  - Update rest of the code for this change
- Frontend:
  - Use the `model` field to determine the data entry structure
  - Update data entry reducer to allow not knowing the data entry structure when initializing

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -5908,7 +5908,7 @@
             "description": "Results for centrally counted (CSO) elections, first election committee session.\nThis contains the data entry values from Model Na 31-2 Bijlage 2."
           }
         ],
-        "description": "PollingStationResults contains the results for a polling station.\n\nThe exact type of results depends on the election counting method and\nwhether this is the first or any subsequent data entry session. Based on\nthis and of four different models can apply"
+        "description": "PollingStationResults contains the results for a polling station.\n\nThe exact type of results depends on the election counting method and\nwhether this is the first or any subsequent data entry session. Based on\nthis, any of four different models can apply"
       },
       "PollingStationType": {
         "type": "string",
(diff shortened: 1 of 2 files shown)


oliver3 (approved):

backend/src/data_entry/structs.rs
```diff
@@ -47,7 +47,43 @@ pub struct PollingStationResultsEntry {
     pub created_at: DateTime<Utc>,
 }
 
-/// PollingStationResults, following the fields in Model Na 31-2 Bijlage 2.
+/// PollingStationResults contains the results for a polling station.
+///
+/// The exact type of results depends on the election counting method and
+/// whether this is the first or any subsequent data entry session. Based on
+/// this and of four different models can apply
```
praseodym:
```suggestion
/// this, any of four different models can apply
```

oliver3 (approved):

backend/src/error.rs
```diff
@@ -93,6 +93,7 @@ pub enum APIError {
     InvalidData(DataError),
     InvalidHeaderValue,
     InvalidHashError,
+    InvalidResultsType,
```
praseodym, line 96: It looks like this was added but is not used?

oliver3: WIll confirm with @rnijveld, resolving this comment for now. It looks like `InvalidData` was used instead in `list_entries_with_polling_stations_first_session()`.

oliver3 (approved):

PR #2095, opened 2025-09-03
Implement validation rules: Political groups

### LINKED ISSUE #2019: Implement validation rules: Political groups
Implement all CSO validation rules (#1611) that are used on political groups ("Lijst {n}") data entry page, according to the Validatieregels en plausibiliteitschecks voor invoer tellingen document.
- Some rules may have the same error code but a new rule, so all rules will need to be checked.
- Make sure that they are also added to the individual section unit tests.
- Please add appropriate docstrings to the enum `ValidationResultCode`

stacktraceghost:
Resolves #2019

Changes:
- F401 became F402 (with addition that total number of votes on a list is not empty or 0)
- Added F401/F403
- Added structured tests per validation rule
- Added tests for validation rules spanning multiple structs: W203/F403
- Added `context` support to `ValidationResult`
  - This allows the backend to pass information which can be used in an error message. In this case, `{ political_group_number: "2" }` which is shown in the error as `E.2`
  - The openapi generator for TS has been adjusted to support `Record<string, x>`. This way a Rust `HashMap<string, string>` turns into a TypeScript `Record<string, string>`.
- Added React Router `<Link>` support to i18n 
  - Reason: a normal hyperlink reloads the entire page
  - Implemented dynamically, now it is possible to use `<Link to="../path">Link text</Link>` in translations
- Adjusted Storybook for `Feedback` component to support `ValidationResult[]` instead of `ValidationResultCode[]`

⚠️ Note: for [F.403](https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/validatieregels-plausibiliteitschecks-tellingen.md#cso--f403-totaal-aantal-stemmen-op-een-lijst-komt-niet-overeen-met-het-lijsttotaal-van-corresponderende-ex) only the total of the list gets highlighted. If we also highlight the field E.x on the voters and votes page, the error automatically gets shown there as well. I've created a separate ticket for this: https://github.com/kiesraad/abacus/issues/2096

Please test:
- The validation rules F401/F402/F403 in [validatieregels-plausibiliteitschecks-tellingen.md](https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/validatieregels-plausibiliteitschecks-tellingen.md)
- Verify that the expected fields are marked
- Verify they are shown on the political list page
  - On the correct position (on top or below the list in case of F401)
  - Link is working
  - Context works (E.x)
- Verify that the coordinator sees the warnings/errors when the typist accepts them

### DIFF
--- a/backend/src/data_entry/validation.rs
+++ b/backend/src/data_entry/validation.rs
@@ -1903,6 +1903,22 @@ mod tests {
         assert_eq!(result1.warnings.len(), 0);
     }
 
+    #[test]
+    fn test_count_err_out_of_range() {
+        let mut validation_results = ValidationResults::default();
+        let count = 1_000_000_000;
+
+        let result = count.validate(
+            &election_fixture(&[]),
+            &polling_station_fixture(None),
+            &mut validation_results,
+            &"".into(),
+        );
+
+        assert!(result.is_err());
+        assert!(result.unwrap_err().message.eq("count out of range"),);
+    }
+
     /// Tests the above_percentage_threshold function with various input combinations.
     #[test]
     fn test_above_percentage_threshold() {


backend/src/data_entry/validation.rs
```diff
@@ -2018,134 +2230,6 @@ mod tests {
...
-                number: 2,
-                total: 1_000_000_000, // out of range
-                candidate_votes: vec![
-                    CandidateVotes {
-                        number: 1,
-                        votes: 0,
-                    },
-                    CandidateVotes {
-                        number: 2,
-                        votes: 1_000_000_000, // out of range
```
oliver3, line 2021: I cannot find any test checking for this scenario anymore?

backend/src/data_entry/validation.rs
```diff
@@ -42,6 +42,9 @@ impl ValidationResults {
 pub struct ValidationResult {
     pub fields: Vec<String>,
     pub code: ValidationResultCode,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(nullable = false)]
+    pub context: Option<HashMap<String, String>>,
```
praseodym: I'd prefer a simple context struct or just a `pub political_group_number: Option<PGNumber>` field instead of a generic solution that is only used for one thing.

praseodym: In the future we can look into refactoring the validation results code, e.g. by creating an `enum ValidationResult` with all context as enum fields.

frontend/src/i18n/format.tsx
```diff
@@ -1,16 +1,22 @@
...
 interface Element {
   tag: string;
+  attributes?: Record<string, string>;
   children: AST;
 }
 
-export const DEFAULT_ALLOWED_TAGS = ["ul", "li", "p", "strong", "em", "code", "h2", "h3", "h4"];
+export const DEFAULT_ALLOWED_TAGS = ["ul", "li", "p", "strong", "em", "code", "h2", "h3", "h4", "Link"];
+
+const DEFAULT_RENDER_CALLBACKS: Record<string, RenderCallback> = {
```
oliver3: Would the names `RenderFunction` and `DEFAULT_RENDER_FUNCTIONS` be better?

backend/src/data_entry/validation.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2095/commits/032d5f0fe71a2eb697b2c83bed2738a6d0968a7b

frontend/src/i18n/format.tsx
stacktraceghost: Renamed in https://github.com/kiesraad/abacus/pull/2095/commits/95018d740b98a21d3f6d6653ed82618a4fed99bf

backend/src/data_entry/validation.rs
stacktraceghost: Added test in https://github.com/kiesraad/abacus/pull/2095/commits/e6eac835e0efaf131f394dbdbd5487b75c15a0af

I am only testing the implemented validation function here. We could check every/some props that use `Count` like some old tests did, but just like any other validation, it will only work when called.

Issue #2051, opened 2025-08-26
Frontend: add data entry with originals column


rnijveld:
- Frontend renders an additional 'original value' column
- Only if the data in the results differs from the original data: show value in second column
- When submitting: submit all values, not just those that changed

PR #2098, opened 2025-09-04
Prepare input grid for adding corrigendum column

oliver3:
- Refactor input grid to be able to add an extra column for https://github.com/kiesraad/abacus/issues/2051
  - Classes instead of index-based styling
  - CSS modules
- Add props to [storybook](https://755187bd.kiesraad-abacus.pages.dev/storybook/?path=/story/components-ui-inputgrid--default-grid) to see that the styling is not affected

### DIFF
--- a/frontend/src/components/ui/InputGrid/InputGrid.tsx
+++ b/frontend/src/components/ui/InputGrid/InputGrid.tsx
@@ -21,9 +21,13 @@ export function InputGrid({ zebra, children }: InputGridProps) {
   );
 }
 
-InputGrid.Header = ({ children }: { children: InputGridRowCells }) => (
+InputGrid.Header = ({ field, value, title }: { field: string; value: string; title: string }) => (
   <thead>
-    <tr>{children}</tr>
+    <tr>
+      <th className={cls.field}>{field}</th>
+      <th className={cls.value}>{value}</th>
+      <th className={cls.title}>{title}</th>
+    </tr>
   </thead>
 );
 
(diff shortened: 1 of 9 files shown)


frontend/src/components/ui/InputGrid/InputGrid.tsx
```diff
@@ -1,25 +1,27 @@
-import * as React from "react";
+import { ReactNode, useRef } from "react";
```
praseodym, line 1: Just wondering, why the change in import?

Lionqueen94:
I think the "Kandidaat" header column in Storybook is not aligned correctly (should be left aligned like the body content right?):
image

frontend/src/components/ui/InputGrid/InputGrid.tsx
oliver3: I thought we usually did it like this and it is more in line with other imports, but maybe it is just personal preference.

oliver3:
> I think the "Kandidaat" header column in Storybook is not aligned correctly

Oh yes forgot to add the appropriate classes for the header, thanks! Changed the `<InputGrid.Header>` to be able to do that in https://github.com/kiesraad/abacus/pull/2098/commits/1f4f3d585bc3b1d49a1c3f759ba5bf146e7f8927 ([storybook](https://755187bd.kiesraad-abacus.pages.dev/storybook/?path=/story/components-ui-inputgrid--default-grid))

Lionqueen94 (approved):

praseodym (approved):

PR #2099, opened 2025-09-04
Generate telbestand and proces verbaal once and save to database

Lionqueen94:


backend/src/report/api.rs
```diff
@@ -248,23 +302,21 @@ async fn election_download_pdf_results(
     ),
     params(
         ("election_id" = u32, description = "Election database id"),
+        ("committee_session_id" = u32, description = "Committee session database id"),
     ),
 )]
 async fn election_download_xml_results(
```
Lionqueen94: @praseodym We don't actually use this endpoint in the application, shall I remove it? Or do we potentially need it in the future for downloading just an EML file?

praseodym: Let's remove it

backend/src/data_entry/repository.rs
```diff
@@ -180,10 +180,10 @@ pub async fn make_definitive(
     Ok(())
 }
 
-/// Get a list of polling station results for an election
+/// Get a list of polling station results for a committee session
 pub async fn list_entries(
     conn: impl DbConnLike<'_>,
-    election_id: u32,
+    committee_session_id: u32,
```
rnijveld, line 186: I'm not really sure why this change was needed? I believe the subquery basically does what is now done at the handler level, but without doing it in a transaction or within the same query.

rnijveld:
I was unaware that for subsequent committee sessions we need to generate multiple PDFs (we both need a Na 14-2 and P 2a), so I guess this is something that we may need to adjust, but maybe we should do that in a follow-up PR because we need to write code to connect those to the rest of the process still anyway, and to not make this one too large?

jschuurk-kr:
> I was unaware that for subsequent committee sessions we need to generate multiple PDFs (we both need a Na 14-2 and P 2a), [...]

For future reference: https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/input-output-bestanden.md

backend/src/report/api.rs
Lionqueen94: Done in 47831841e999aea06fff0aa082a3c0a2e105a368

backend/src/data_entry/repository.rs
Lionqueen94: You could technically already have started a new session before downloading the files. Meaning by default the results of the current session (hence, no results) would end up in the files when you do download them for the first session. We could prevent this by not allowing to create a new session if the files have not been saved in the database for the current session. What do you think?

rnijveld: Conclusion: this breaks next committee sessions, but at the same time that is already not functional yet, and this will get fixed in future PRs anyway, so for now accept it the way it is!

rnijveld (approved):

Issue #1775, opened 2025-07-14
Reimplement tests using Storybook


praseodym:
#1753 removed some Vitest tests using the Ladle stories, we might want to re-implement some of those tests using [Storybook interaction tests](https://storybook.js.org/docs/writing-tests/interaction-testing).

- [x] FormField
- [x] NavBar: no storybook tests needed in addition to `NavBar.test.tsx`
- [x] Alert
- [x] CheckBoxAndRadio: no storybook tests needed in addition to `CheckboxAndRadio.test.tsx`
- [x] ChoiceList
- [x] FileInput
- [x] Form (no tests were removed in #1753)
- [x] FormLayout (no tests were removed in #1753)
- [x] IconButton
- [x] Table (some tests were removed in #1753)
- [x] FormKeyboardNavigation (no tests were removed in #1753): no storybook tests needed in addition to `useFormKeyboardNavigation.test.tsx`.
- [x] ElectionStatus (some tests were removed in #1753)

jschuurk-kr:
For Table the big question is if we can get our custom toHaveTableContent() matcher to work with storybook's vitest. Regardless, we'll need a regular vitest test to be able to test that a LinkRow navigates to the correct url when clicked.

Same applies to ElectionStatus: toHaveTableContent() would be really helpful there.

@oliver3 something you could help me with?

PR #2103, opened 2025-09-05
FormLayout storybook tests

jschuurk-kr:
relates to #1775 

### Scope
- add Alert to FarmLayout stories
- add DisabledFormLayout to FormLayout stories

The only tests I could come up with are for the `disabled` parameter of FormLayout and the `title` parameter of FormSection. I don't think those tests add enough value to write them. (They didn't exist prior to PR #1753 either.)

### DIFF
--- a/frontend/src/components/ui/Form/FormLayout.stories.tsx
+++ b/frontend/src/components/ui/Form/FormLayout.stories.tsx
@@ -1,5 +1,6 @@
 import type { Meta, StoryFn } from "@storybook/react-vite";
 
+import { Alert } from "../Alert/Alert";
 import { Button } from "../Button/Button";
 import { ChoiceList } from "../CheckboxAndRadio/ChoiceList";
 import { InputField } from "../InputField/InputField";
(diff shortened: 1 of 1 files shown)


frontend/src/components/ui/Form/FormLayout.stories.tsx
```diff
@@ -8,6 +9,12 @@ import { FormLayout } from "./FormLayout";
 export const DefaultFormLayout: StoryFn = () => (
   <div>
     <FormLayout>
+      <FormLayout.Alert>
+        <Alert type="success">
+          <h2>It works!</h2>
+          <p>This is a success alert.</p>
+        </Alert>
+      </FormLayout.Alert>
```
Lionqueen94: Normally the alert is added just outside the `FormLayout` to prevent extra margin on the bottom (which does not make sense with the name `FormLayout.Alert` yes)
image

jschuurk-kr: done in d3a1e612

praseodym (approved):

oliver3 (approved):

PR #2104, opened 2025-09-05
Fixed showing filename instead of "Geen bestand gekozen" on file upload

### LINKED ISSUE #2093: After uploading an invalid file, the label next to the button says "geen bestand gekozen" instead of the filename.
If you upload an invalid file, e.g. during election creation, the text next to the "Bestand kiezen"-button should show the name of the invalid file. It currently says "Geen bestand gekozen".

Note that if a valid file is uploaded, we don't show the "Bestand kiezen"-button, so in that scenario this is not an issue.

Design:
Image

Latest release build:
Image

Lionqueen94:
Closes #2093

### DIFF
--- a/frontend/src/components/ui/FileInput/FileInput.module.css
+++ b/frontend/src/components/ui/FileInput/FileInput.module.css
@@ -14,6 +14,7 @@ label.selectedFile {
   align-content: center;
   padding-left: 1rem;
   color: var(--gray-600);
+  overflow-wrap: anywhere;
 }
 
 .fileInputBtn {


jschuurk-kr:
Found small styling issue in case of long file names:
image

Or file names that can't we wrapped:
image

jorisleker:
Prefer to have it behave this way:
image

See [Figma for details on spacing/line-height ](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=14751-6767&t=DnkWOwEyBoKVwF2o-11) etc

bramtweedegolf (approved):

bramtweedegolf:
@jorisleker The updated version looks like this, is that acceptable?

zero
one

Filenames that can't normally be wrapped, are wrapped anyway:

three

cikzh (approved):

oliver3 (approved):

Issue #2111, opened 2025-09-08
If election does not match the election in the uploaded polling station list, show a warning.


bramtweedegolf:
If election does not match the election in the uploaded polling station list, show a warning.

PR #2114, opened 2025-09-09
Add warning when uploading polling stations with mismatching election id

bramtweedegolf:
While creating a new election, uploadind a polling station file with a different election id should show a warning.

The warning currently looks like this:
Screenshot_2025-09-09_09-18-09

To test:

- Create a new election
- Upload a new election and candidates
- For the polling station, upload eml110b_not_matching_election_id.eml.xml
- The warning should be present

Closes https://github.com/kiesraad/abacus/issues/2111

### DIFF
--- a/frontend/e2e-tests/helpers-utils/e2e-test-browser-helpers.ts
+++ b/frontend/e2e-tests/helpers-utils/e2e-test-browser-helpers.ts
@@ -103,5 +103,6 @@ export async function uploadPollingStations(page: Page, eml = eml110b) {
 
   const checkDefinitionPage = new CheckPollingStationDefinitionPgObj(page);
   await expect(checkDefinitionPage.header).toBeVisible();
+  await expect(checkDefinitionPage.warning).toBeHidden();
   await checkDefinitionPage.next.click();
 }
--- a/frontend/e2e-tests/page-objects/election/create/CheckPollingStationDefinitionPgObj.ts
+++ b/frontend/e2e-tests/page-objects/election/create/CheckPollingStationDefinitionPgObj.ts
@@ -6,6 +6,7 @@ export class CheckPollingStationDefinitionPgObj {
   readonly next: Locator;
   readonly table: Locator;
   readonly stations: Locator;
+  readonly warning: Locator;
 
   constructor(protected readonly page: Page) {
     this.header = page.getByRole("heading", { level: 2, name: "Controleer stembureaus" });
@@ -13,5 +14,7 @@ export class CheckPollingStationDefinitionPgObj {
     this.next = page.getByRole("button", { name: "Volgende" });
     this.table = page.getByTestId("overview");
     this.stations = this.table.locator("tbody").getByRole("row");
+    this.showMore = page.getByRole("button", { name: /Toon alle \d+ stembureaus/ });
+    this.warning = page.getByRole("strong").filter({ hasText: "Afwijkende verkiezing" });
   }
 }
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -4831,6 +4831,9 @@
             "format": "int32",
             "minimum": 0
           },
+          "polling_station_definition_matches_election": {
+            "type": "boolean"
+          },
           "polling_stations": {
             "type": "array",
             "items": {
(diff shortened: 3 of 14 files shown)


frontend/e2e-tests/helpers-utils/e2e-test-browser-helpers.ts
```diff
@@ -95,13 +95,20 @@ export async function uploadCandidatesAndInputHash(page: Page) {
...
   await expect(uploadElectionDefinitionPage.main).toContainText(eml.filename);
 
   const checkDefinitionPage = new CheckPollingStationDefinitionPgObj(page);
   await expect(checkDefinitionPage.header).toBeVisible();
+
+  if (!warningShouldBeShown) {
+    await expect(checkDefinitionPage.warning).toBeHidden();
+  } else {
+    await expect(checkDefinitionPage.warning).toBeVisible();
+  }
```
jschuurk-kr: I'm not convinced we need to test this warning in an e2e test. An RTL test should be sufficient.

Also, if we test it, the assertion should not be in a helper function, but in the test itself. (Any expects in helper functions are there to ensure the helper is working. We might still have a few too many, though.)

jschuurk-kr:
@jorisleker @praseodym Considering the varied ways in which a municipality manages their polling station EML files, I'm not sure we should care if the `Id` of the `ElectionIdentifier` of the polling station EML is the same as the one in the election definition. What do you think?

frontend/e2e-tests/page-objects/election/create/CheckPollingStationDefinitionPgObj.ts
```diff
@@ -6,12 +6,15 @@ export class CheckPollingStationDefinitionPgObj {
...
   readonly stations: Locator;
+  readonly warning: Locator;
 
   constructor(protected readonly page: Page) {
     this.header = page.getByRole("heading", { level: 2, name: "Controleer stembureaus" });
     this.showMore = page.getByRole("button", { name: /Toon alle \d+ stembureaus/ });
     this.next = page.getByRole("button", { name: "Volgende" });
     this.table = page.getByTestId("overview");
     this.stations = this.table.locator("tbody").getByRole("row");
+    this.showMore = page.getByRole("button", { name: /Toon alle \d+ stembureaus/ });
```
jschuurk-kr:
```suggestion
```

already exists as second attribute of the class

jorisleker:
> @jorisleker @praseodym Considering the varied ways in which a municipality manages their polling station EML files, I'm not sure we should care if the `Id` of the `ElectionIdentifier` of the polling station EML is the same as the one in the election definition. What do you think?

Yes I agree. In a future release, we might consider adding an alert if the electionidentifier does not match. Something like "Het lijkt erop dat dit bestand met stembureaus is gemaakt voor een andere verkiezing". But never blocking, but there are two valid usecases:
- multiple elecitons at the same time, with same set of polling stations. Does not make sense to prevent using the same file twice.
- no changes in polling stations since previous election (smaller municipalities). Same: should not prevent using the same file.

jschuurk-kr:
> Yes I agree. In a future release, we might consider adding an alert if the electionidentifier does not match.

@ring-ring-ring said he would check how the election identifier of the polling stations file gets filled by municipalities.

frontend/e2e-tests/page-objects/election/create/CheckPollingStationDefinitionPgObj.ts
bramtweedegolf: Removed duplicate in https://github.com/kiesraad/abacus/pull/2114/commits/6ae4d2c73b842113515d0ab6bf1098e4ab13108a

frontend/e2e-tests/helpers-utils/e2e-test-browser-helpers.ts
bramtweedegolf: Agreed, the e2e tests add nothing to the RTL test; removed in the latest commits.

ring-ring-ring:
> > Yes I agree. In a future release, we might consider adding an alert if the electionidentifier does not match.
> 
> @ring-ring-ring said he would check how the election identifier of the polling stations file gets filled by municipalities.

In the polling station files I have the election identifier has been included. Since it does not block the flow (warning only) my preference would be to merge the PR and keep an eye on this function during user tests.

cikzh (approved):

PR #2116, opened 2025-09-09
Fix Model N 10-2 for unknown polling station type

praseodym:
Fix Model N 10-2 for unknown polling station type (`undefined` in JSON)


oliver3 (approved):

praseodym:
> Should we update a polling station in election_1.sql to only have minimal data, i.e. only id, committee_session_id, name, number? Because right now there is no real difference between polling stations "Studio The Rules" and "Positivo Zaal".

Changed in b14c58d5!

PR #2120, opened 2025-09-09
Coordinator validation feedback on resolve error overview page

stacktraceghost:
Resolves #2073

- Updated styling according to design
- Fixed `feedback.json`: some keys were called `action` instead of `actions`
- Added a small test to make sure the feedback structure is validated

PR #2123, opened 2025-09-10
Bugfix: StickyNav behaviour

stacktraceghost:
As noticed [here](https://github.com/kiesraad/abacus/pull/2120#issuecomment-3273848104), the StickyNav menu moves upwards when you reach the end of the page (A).

Next to that:
- B: The political group list doesn't have a minimum height
- C: The menu can get cut off (make your viewport 300px height and scroll fully down on a data entry page)
image

This PR fixes issues A + B by:
- Not allowing a negative value to the max-height calculation (`--sticky-nav-top-offset`)
- Adding padding when `<StickyNav>` itself starts to move out of screen (because it is part of `<main>`)
- Setting a min-height for the scrollable ProgressList section

This PR does not fix C.

### DIFF
--- a/frontend/src/components/ui/AppLayout/StickyNav.tsx
+++ b/frontend/src/components/ui/AppLayout/StickyNav.tsx
@@ -14,7 +14,8 @@ export function StickyNav({ children }: StickyNavProps) {
     if (nav) {
       const updateTopOffset = () => {
         const navTop = nav.getBoundingClientRect().top;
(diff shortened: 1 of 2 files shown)


stacktraceghost:
// Edit: this has been solved by the more simple approach. 

image

Found a bug when the window is very narrow and the menu turns to position fixed.

stacktraceghost:
Implemented a more simple approach.

PR #2132, opened 2025-09-11
Implemented design of ReportPage

### LINKED ISSUE #1974: Frontend: Implement ElectionReportPage as designed and linked to committee session

Lionqueen94:
Closes #1974 

## Description
- Corrected fill of some icons and changed checkVerified icon to match design
- Set download link color to default blue also when `:visited`
- Implemented ReportPage as designed
- Adjusted tests and removed pdf rendering e2e test
- See #2061 for the justification of removing the pdf preview (until we implement the watermark)

image


frontend/e2e-tests/tests/pdf-rendering.e2e.ts
praseodym, line 1: Why was the e2e test removed?

Lionqueen94: Because the button is removed. Once we implement the pdf preview functionality we can re-add this test. See #2061 for the justification for this (it was discussed with Joris).

PR #2133, opened 2025-09-12
Update `sigrid.yaml`

praseodym:
- Move MockTest component to features/dev directory
- Update `sigrid.yaml`

### DIFF
--- a/sigrid.yaml
+++ b/sigrid.yaml
@@ -71,9 +71,6 @@ components:
       - ".*/backend/tests/.*"
 
   # Frontend src/components
-  - name: "(Frontend) App"
-    include:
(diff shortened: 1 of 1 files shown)


sigrid.yaml
```diff
@@ -58,26 +70,48 @@ components:
     include:
       - ".*/backend/tests/.*"
 
-  # Frontend components
+  # Frontend src/components
   - name: "(Frontend) App"
     include:
       - ".*/frontend/src/app/.*"
```
Lionqueen94: Might be more logical to move this down to line 165 so that all components are together

praseodym: Thanks, done in 4ea68f5

PR #2134, opened 2025-09-12
Backend API for adding investigations to committee session

### LINKED ISSUE #1902: Backend: Add polling stations list to committee session

praseodym:
Add a backend API for adding investigations to committee session, and use this API in the existing frontend screens. Closes #1902.

### DIFF
--- a/backend/src/committee_session/structs.rs
+++ b/backend/src/committee_session/structs.rs
@@ -100,35 +100,3 @@ pub struct CommitteeSessionFilesUpdateRequest {
     pub results_eml: Option<u32>,
     pub results_pdf: Option<u32>,
 }
-
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, ToSchema)]
-#[serde(deny_unknown_fields)]
-pub struct PollingStationInvestigation {
-    pub polling_station_id: u32,
-    pub reason: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    #[schema(nullable = false)]
-    pub findings: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    #[schema(nullable = false)]
-    pub corrected_results: Option<bool>,
-}
-
-impl IntoResponse for PollingStationInvestigation {
-    fn into_response(self) -> Response {
-        Json(self).into_response()
-    }
-}
-
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, ToSchema)]
-#[serde(deny_unknown_fields)]
-pub struct PollingStationInvestigationCreateRequest {
-    pub reason: String,
-}
-
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, ToSchema)]
-#[serde(deny_unknown_fields)]
-pub struct PollingStationInvestigationConcludeRequest {
-    pub findings: String,
-    pub corrected_results: bool,
-}
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -2727,7 +2727,7 @@
     },
     "/api/polling_stations/{polling_station_id}/investigations": {
       "put": {
-        "summary": "Conclude an investigation for a certain polling station",
+        "summary": "Conclude an investigation for a polling station",
         "operationId": "committee_session_investigation_conclude",
         "parameters": [
           {
@@ -2809,7 +2809,7 @@
         }
       },
       "post": {
-        "summary": "Create an investigation for a certain polling station",
+        "summary": "Create an investigation for a polling station",
         "operationId": "committee_session_investigation_create",
         "parameters": [
           {
--- a/backend/src/audit_log/audit_event.rs
+++ b/backend/src/audit_log/audit_event.rs
@@ -5,7 +5,7 @@ use strum::VariantNames;
 use utoipa::ToSchema;
 
 use super::AuditEventLevel;
-use crate::{ErrorResponse, committee_session::PollingStationInvestigation, error::ErrorReference};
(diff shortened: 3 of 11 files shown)


backend/src/committee_session/structs.rs
```diff
@@ -100,3 +100,44 @@ pub struct CommitteeSessionFilesUpdateRequest {
...
+#[serde(deny_unknown_fields)]
+pub struct PollingStationInvestigationCreateRequest {
+    pub polling_station_id: u32,
+    pub reason: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(nullable = false)]
+    pub findings: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(nullable = false)]
+    pub corrected_results: Option<bool>,
```
Lionqueen94: This is not needed for the create investigation I believe. You only need the polling station and the reason.

frontend/src/features/investigations/routes.ts
```diff
@@ -2,9 +2,9 @@ import { RouteObject } from "react-router";
 
 import { AddInvestigationLayout } from "./components/AddInvestigationLayout";
 import { AddInvestigationPage } from "./components/AddInvestigationPage.tsx";
-import { InvestigationFindings } from "./components/InvestigationFindings";
+import { InvestigationFindingsPage } from "./components/InvestigationFindingsPage.tsx";
 import { InvestigationPrintCorrigendumPage } from "./components/InvestigationPrintCorrigendumPage";
-import { InvestigationReason } from "./components/InvestigationReason";
+import { InvestigationReasonPage } from "./components/InvestigationReasonPage.tsx";
 import { InvestigationsLayout } from "./components/InvestigationsLayout";
 import { InvestigationsOverviewPage } from "./components/InvestigationsOverviewPage.tsx";
```
Lionqueen94:
```suggestion
import { InvestigationReasonPage } from "./components/InvestigationReasonPage";
import { InvestigationsLayout } from "./components/InvestigationsLayout";
import { InvestigationsOverviewPage } from "./components/InvestigationsOverviewPage";
```

frontend/src/features/investigations/routes.ts
```diff
@@ -2,9 +2,9 @@ import { RouteObject } from "react-router";
 
 import { AddInvestigationLayout } from "./components/AddInvestigationLayout";
 import { AddInvestigationPage } from "./components/AddInvestigationPage.tsx";
```
Lionqueen94:
```suggestion
import { AddInvestigationPage } from "./components/AddInvestigationPage";
```

frontend/src/features/investigations/routes.ts
```diff
@@ -2,9 +2,9 @@ import { RouteObject } from "react-router";
...
-import { InvestigationFindings } from "./components/InvestigationFindings";
+import { InvestigationFindingsPage } from "./components/InvestigationFindingsPage.tsx";
```
Lionqueen94:
```suggestion
import { InvestigationFindingsPage } from "./components/InvestigationFindingsPage";
```

backend/src/committee_session/structs.rs
praseodym: Fixed in 7e77718!

backend/migrations/10_polling_station_investigations.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE polling_station_investigations
+(
+    id                   INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    polling_station_id   INTEGER                           NOT NULL,
+    committee_session_id INTEGER                           NOT NULL,
```
rnijveld: I'm wondering if this is smart to do, given that a polling station is always ever in a single committee session anyway? I'm also not seeing usage in the `PollingStationInvestigation` struct either? Anywhere it is required it should be easily fixed with a simple join.

backend/src/committee_session/structs.rs
```diff
@@ -100,3 +100,38 @@ pub struct CommitteeSessionFilesUpdateRequest {
...
+}
+
+impl IntoResponse for PollingStationInvestigation {
+    fn into_response(self) -> Response {
+        Json(self).into_response()
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, ToSchema, Type, FromRow)]
```
rnijveld: I don't think this, nor the `PollingStationInvestigationConcludeRequest` need `sqlx::FromRow` in their derives, I would not expect these to be queried from the database. The `sqlx::Type` derives also seem weird to me, as they aren't column values either.

rnijveld (changes requested):

backend/migrations/10_polling_station_investigations.sql
praseodym: Agreed, I'll change this

backend/src/committee_session/structs.rs
praseodym: Fixed in f72ff356d4aa860f2760ff34aa7b1e31cc0181db

backend/migrations/10_polling_station_investigations.sql
praseodym: Fixed in 4c41847827892d2d63c240e762fda550ed272c7e, and since it didn't make sense to keep the endpoint in the committee session module, moved it to a dedicated module in 612dbade56defc03f7384ec2495f674d68633020

Lionqueen94 (approved):

Issue #2055, opened 2025-08-26
Frontend: keep zero visible during data entry with originals column


rnijveld:
Zero in this view means that the value was corrected to zero, and has a different meaning than 'no entry', which means that the original value should be kept.

PR #2149, opened 2025-09-15
Show and save data entry with corrections if previousResults is set

oliver3:
- Resolves https://github.com/kiesraad/abacus/issues/2051
- Resolves https://github.com/kiesraad/abacus/issues/2055
- Add `previousResults` from claim response to data entry reducer state
- Determine and pass on `previousValues` to `InputGridSubsectionComponent`
- Let `InputGridSubsectionComponent` show and save the corrections based on the `previousValues` (saved to backend as consolidated results)
- (extra) do not hard-code `INITIAL_FORM_SECTION_ID` but take the first section of the dataEntryStructure instead

To test:
- Finish a complete first committee session (you could use `full-flow.e2e.ts`)
- (Coordinator) On the election details page, prepare a new session (Nieuwe zitting voorbereiden)
- (Coordinator) On the investigations overview page (Aangevraagde onderzoeken), use the temporary Start button to start data entry
- (Typist) Do the data entry for a next session

Note:
At this moment, the backend claim endpoint returns an empty polling station results as an initial data entry, which causes the input grids to start off with all zeros as corrected values. There will be a future PR where the initial data entry is returned with the previous results for the next session data entry, which will make sure the initial data entry is as expected.

### DIFF
--- a/frontend/src/components/data_entry/subsections/InputGridSubsection.tsx
+++ b/frontend/src/components/data_entry/subsections/InputGridSubsection.tsx
@@ -44,8 +44,8 @@ export function InputGridSubsectionComponent({
     <InputGrid id={id} zebra={subsection.zebra}>
       <InputGrid.Header
         field={subsection.headers[0]}
-        previous={previousValues && t("data_entry.previous_value")}
-        value={subsection.headers[1]}
+        previous={previousValues && t("data_entry.original")}
+        value={previousValues ? t("data_entry.corrected") : subsection.headers[1]}
         title={subsection.headers[2]}
       />
       <InputGrid.Body>
--- a/frontend/src/components/data_entry/subsections/InputGridSubsection.tsx
+++ b/frontend/src/components/data_entry/subsections/InputGridSubsection.tsx
@@ -2,6 +2,7 @@ import { InputGrid } from "@/components/ui/InputGrid/InputGrid";
 import { InputGridRow } from "@/components/ui/InputGrid/InputGridRow";
 import { t } from "@/i18n/translate";
 import { InputGridSubsection, InputGridSubsectionRow, SectionValues } from "@/types/types";
+import { correctedValue, determineCorrections } from "@/utils/dataEntryMapping";
 
 export interface InputGridSubsectionProps {
   id?: string;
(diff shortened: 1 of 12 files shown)


stacktraceghost (approved):

praseodym (changes requested):
- The new value input field is emptied when its value matches the previous value. This means that a value cannot be easily corrected to a value that is prefixed by the original value, e.g. correcting `20` to `200` or `2000`. See this [screen recording](https://github.com/user-attachments/assets/0494161a-327f-4020-b6fa-8d593121b06c) where I type `200` first, then `21`, then `2000`.

- When making corrections, the current values column should have the title "Nieuw geteld aantal" according to the design.

oliver3:
> * The new value input field is emptied when its value matches the previous value. This means that a value cannot be easily corrected to a value that is prefixed by the original value, e.g. correcting `20` to `200` or `2000`. See this [screen recording](https://github.com/user-attachments/assets/0494161a-327f-4020-b6fa-8d593121b06c) where I type `200` first, then `21`, then `2000`.

Fixed in https://github.com/kiesraad/abacus/pull/2149/commits/05d1a9cfda0c11004734ad5a6c5a9f1a03e75874

>     * When making corrections, the current values column should have the title "Nieuw geteld aantal" according to the design.

Updated in Figma to "Oorspronkelijk" and "Gecorrigeerd" and implemented in https://github.com/kiesraad/abacus/pull/2149/commits/05d1a9cfda0c11004734ad5a6c5a9f1a03e75874 and the test in https://github.com/kiesraad/abacus/pull/2149/commits/66f7e8de99b2245729c1263708809f8323fe32fc

frontend/src/components/data_entry/subsections/InputGridSubsection.tsx
```diff
@@ -27,12 +30,22 @@ export function InputGridSubsectionComponent({
...
+  }
+
   return (
     <InputGrid id={id} zebra={subsection.zebra}>
       <InputGrid.Header
         field={subsection.headers[0]}
-        previous={previousValues && t("data_entry.previous_value")}
-        value={subsection.headers[1]}
+        previous={previousValues && t("data_entry.original")}
+        value={previousValues ? t("data_entry.corrected") : subsection.headers[1]}
```
praseodym, line 48: I think it makes more sense to set the correct header values in the data entry structure

praseodym (approved):

stacktraceghost (approved):

frontend/src/components/data_entry/subsections/InputGridSubsection.tsx
oliver3: I'll check again when we have a different model for the CSO next session 👍

Issue #2141, opened 2025-09-15
Parse the number of voters from the polling station list EML (110b)


rnijveld:
Split from #1926

In 110b, the `PollingStation` element is defined as having text (currently stored as the token field inside of that struct), but this is supposed to be interpreted as containing the number of voters if it is a positive integer value. We need to parse this, and during import we need to store this in the number_of_voters for that polling station.

PR #2150, opened 2025-09-16
Parse the number of voters from the polling station list eml 110b

bramtweedegolf:
In 110b, the PollingStation element is defined as having text (currently stored as the token field inside of that struct), but this is supposed to be interpreted as containing the number of voters if it is a positive integer value.

Closes https://github.com/kiesraad/abacus/issues/2141


cikzh (dismissed):

backend/src/eml/eml_110.rs
```diff
@@ -446,10 +448,31 @@ pub struct PhysicalLocation {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct PollingStation {
-    #[serde(rename = "$text")]
-    token: String,
     #[serde(rename = "@Id")]
     id: String,
+    #[serde(deserialize_with = "deserialize_number_of_voters", rename = "$text")]
+    number_of_voters: Option<u64>,
```
rnijveld: Maybe just make this a u32 to not have to do the try_into when trying to put the data into a PollingStation?

bramtweedegolf: The try_into is used when we convert from crate::polling_station::PollingStation, where number_of_voters is i64. So we would need the try_from anyway. Should we make it a i64? (Or maybe I'm not understanding the question correctly?)

bramtweedegolf: Discussed this issue; I'll make it a i64

praseodym (approved):

PR #2152, opened 2025-09-16
Add CSONextSession variant for storing Na 14-2 bijlage 1 results

### LINKED ISSUE #2048: Backend: Add variant to `PollingStationResults` for data entry for Na 14-2
Depends on #2047

- Update validation rules to be shared whenever the rules are the same between both variants of the results
- Allow generating EML510 from both variants of PollingStationResults

rnijveld:
Not ready for review yet

Resolves #2048 

- Implements a new `CSONextSession` variant for `PollingStationResults` that implements the Na 14-2 bijlage 1 results entry.
- Adds a new `CommonPollingStationResults` struct that implements the common fields for any variant of the results, which is now returned for the previous results in the claim endpoint
- Updates ElectionSummary to make use of the PollingStationResults, thus allowing usage of either variant in EML and PDF generation

### DIFF
--- a/backend/src/data_entry/repository.rs
+++ b/backend/src/data_entry/repository.rs
@@ -1,6 +1,6 @@
 use sqlx::{Connection, SqliteConnection, query, query_as, types::Json};
 
-use super::{CSOFirstSessionResults, PollingStationDataEntry, status::DataEntryStatus};
+use super::{PollingStationDataEntry, status::DataEntryStatus};
 use crate::{
     data_entry::{ElectionStatusResponseEntry, PollingStationResults},
     polling_station::PollingStation,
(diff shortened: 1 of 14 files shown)


backend/src/data_entry/repository.rs
stacktraceghost, line 1: There are some complicated queries in this file (e.g. `list_entries_for_committee_session`/`most_recent_results_for_polling_station`). 

Could we add some tests to ensure that we are receiving what we are expecting? The indirect tests give coverage, however I am unsure about the quality of that coverage (in terms of these functions). If we directly test these functions, it will be more clear. Could be done in a separate PR.

stacktraceghost:
Found an interesting bug
1. Go to my [temporary test environment](https://model-processen-benoemd.abacus-test.nl/) 
2. Log in as coordinator
3. Go to [resolve errors of the 2nd committee session](https://model-processen-benoemd.abacus-test.nl/elections/1/status/7/resolve-errors)
4. If you start on "Fouten en waarschuwingen" and then click on "Aantal kiezers en stemmen" and then on "Lijst 1 - Lijst Hekking" or vice versa, the numbers disappear

stacktraceghost (dismissed):

backend/src/data_entry/repository.rs
praseodym: Not a priority for now

oliver3 (dismissed):

Issue #2140, opened 2025-09-15
Update XML for 510 (counts results file)


rnijveld:
Split from #1926 

- [x] `Cast` element should contain kiesgerechtigden (Z), not (D)
- [x] For `ReportingUnitVotes`, the cast field must be the number of voters per polling station, and must always be an integer (set to 0 if not available)
- [x] The old uncounted votes reasons should be removed from the EML, they are no longer in the new models: `<UncountedVotes ReasonCode="meegenomen stembiljetten">0</UncountedVotes> <UncountedVotes ReasonCode="te weinig uitgereikte stembiljetten">0</UncountedVotes> <UncountedVotes ReasonCode="te veel uitgereikte stembiljetten">0</UncountedVotes> <UncountedVotes ReasonCode="andere verklaring">0</UncountedVotes> <UncountedVotes ReasonCode="geen verklaring">0</UncountedVotes>`
- [x] Implement this issue from EML in our implementation: https://github.com/kiesraad/EML_NL/issues/3

PR #2153, opened 2025-09-16
Update xml for 510 counts results file

bramtweedegolf:
Updates XML and parsing for 510

- Updated Cast value
- Old UncountedVotes reasons are removed
- Added optional ReportingUnitInvestigations to ReportingUnitVotes

Closes https://github.com/kiesraad/abacus/issues/2140

### DIFF
--- a/backend/src/eml/eml_510.rs
+++ b/backend/src/eml/eml_510.rs
@@ -674,7 +674,7 @@ mod tests {
     }
 
     #[test]
-    fn test_eml510d_with_investigations() {
+    fn test_eml510b_with_investigations() {
         let data = include_str!("./tests/eml510b_with_investigations.eml.xml");
         let doc = EML510::from_str(data).unwrap();
 
--- a/backend/src/eml/eml_510.rs
+++ b/backend/src/eml/eml_510.rs
@@ -166,29 +166,50 @@ impl TotalVotes {
                 ),
             ],
             uncounted_votes: vec![
-                UncountedVotes::new(
-                    UncountedVotesReason::PollCard,
-                    summary.voters_counts.poll_card_count as u64,
-                ),
-                UncountedVotes::new(
-                    UncountedVotesReason::ProxyCertificate,
-                    summary.voters_counts.proxy_certificate_count as u64,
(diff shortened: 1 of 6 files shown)


backend/src/eml/eml_510.rs
```diff
@@ -558,4 +658,50 @@ mod tests {
...
+        assert_eq!(doc.count.election.contests[0].reporting_unit_votes.len(), 1);
+        assert!(
+            doc.count.election.contests[0].reporting_unit_votes[0]
+                .reporting_unit_investigations
+                .is_none()
+        );
+    }
+
+    #[test]
+    fn test_eml510d_with_investigations() {
```
Lionqueen94:
```suggestion
    fn test_eml510b_with_investigations() {
```

backend/src/eml/eml_510.rs
```diff
@@ -203,6 +230,72 @@ pub struct ReportingUnitVotes {
     rejected_votes: Vec<RejectedVotes>,
     #[serde(default)]
     uncounted_votes: Vec<UncountedVotes>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    reporting_unit_investigations: Option<ReportingUnitInvestigations>,
+}
+
+/// When results were investigated by the GSB
+/// return the relevant investigations.
+fn create_investigations_from_results(
```
praseodym, line 239: This is very similar to this code: https://github.com/kiesraad/abacus/blob/f673fec9c87f932b9b36196503471cf2d45a330d/backend/src/summary/mod.rs#L212-L266

Maybe we should create a function to share that business logic across the summary and eml code?

bramtweedegolf: Maybe something like this? https://github.com/kiesraad/abacus/pull/2153/commits/e2632fd5625637fa59e25975acf8931de2b5a07c
This does not reused that much code, but is more clear and readable imo.

praseodym: Yes, that's better!

praseodym (approved):

PR #2155, opened 2025-09-17
update and dedupe frontend dependencies

jschuurk-kr:
### Scope

Ran [`npm dedupe`](https://docs.npmjs.com/cli/v11/commands/npm-dedupe) which "Searches the local package tree and attempts to simplify the overall structure by moving dependencies further up the tree, where they can be more effectively shared by multiple dependent packages."

Ran an `npm update` after that. And another `npm dedupe`, but that didn't result in any changes.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -4495,9 +4495,9 @@
       }
     },
     "node_modules/debug": {
-      "version": "4.4.1",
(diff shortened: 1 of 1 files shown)


praseodym (approved):
`npm audit` reports 0 vulnerabilities so I think we're good

PR #2167, opened 2025-09-19
Delete investigation

### LINKED ISSUE #2156: Backend: Delete investigation from committee session
- Endpoint to delete investigation from committee session
- Add audit logging
- Change committee session status back to `created` when deleting last investigation

### LINKED ISSUE #2157: Frontend: Delete investigation from committee session
- Add delete button + modal according to flow in [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=15087-45696&m=dev)
- Show alert in overview after deletion

Lionqueen94:
Closes #2156 
Closes #2157 

## Description
- Add delete investigation endpoint
- Add delete investigation button and modal
- Delete data entry and result for a polling station of which the investigation is deleted or the corrected results are changed from true to false
- Moved frontend investigation form tests to page tests to make sure all tests are in one file and no separate testfile is needed just for the delete

### DIFF
--- a/frontend/src/features/investigations/components/InvestigationDelete.tsx
+++ b/frontend/src/features/investigations/components/InvestigationDelete.tsx
@@ -4,7 +4,6 @@ import { AnyApiError, isSuccess } from "@/api/ApiResult";
 import { useCrud } from "@/api/useCrud";
 import { IconTrash } from "@/components/generated/icons";
 import { Button } from "@/components/ui/Button/Button";
-import { Loader } from "@/components/ui/Loader/Loader";
 import { Modal } from "@/components/ui/Modal/Modal";
 import { useElection } from "@/hooks/election/useElection";
 import { useElectionStatus } from "@/hooks/election/useElectionStatus";
@@ -21,18 +20,12 @@ export function InvestigationDelete({ pollingStation, onDeleted, onError }: Inve
   const [showModal, setShowModal] = useState(false);
   const { refetch } = useElection(pollingStation.id);
   const electionStatuses = useElectionStatus();
-  const pollingStationStatus = electionStatuses.statuses.find(
-    (status) => status.polling_station_id === pollingStation.id,
-  );
+  const status = electionStatuses.statuses.find((status) => status.polling_station_id === pollingStation.id);
   const path = `/api/polling_stations/${pollingStation.id}/investigation`;
   const { remove, requestState } = useCrud({ remove: path });
 
   function toggleModal() {
     setShowModal(!showModal);
-  }
-
-  if (!pollingStationStatus) {
-    return <Loader />;
   }
 
   function handleDelete() {
@@ -57,7 +50,7 @@ export function InvestigationDelete({ pollingStation, onDeleted, onError }: Inve
       {showModal && (
         <Modal title={`${t("investigations.delete_investigation")}?`} onClose={toggleModal}>
           <p>{t("investigations.delete_are_you_sure")}</p>
-          {pollingStationStatus.status !== "first_entry_not_started" && <p>{tx("investigations.delete_data_entry")}</p>}
+          {status && status.status !== "first_entry_not_started" && <p>{tx("investigations.delete_data_entry")}</p>}
           <nav>
             <Button
               leftIcon={<IconTrash />}


backend/src/investigation/repository.rs
```diff
@@ -98,6 +98,25 @@ pub async fn get_polling_station_investigation(
...
+
+    let rows_affected = query!(
+        r#"DELETE FROM polling_station_investigations WHERE polling_station_id = ?"#,
+        polling_station_id,
+    )
+    .execute(&mut *tx)
+    .await?
+    .rows_affected();
+
+    tx.commit().await?;
```
praseodym: No need to start and commit a transaction here, it's just one query

Lionqueen94: Good one, removed in 212fd3436c03fd7e8c5f6a7550b07ec9ad8ad24c

backend/src/investigation/api.rs
```diff
@@ -206,6 +209,59 @@ async fn polling_station_investigation_update(
...
+    _user: Coordinator,
+    State(pool): State<SqlitePool>,
+    audit_service: AuditService,
+    CurrentSessionPollingStationId(polling_station_id): CurrentSessionPollingStationId,
+) -> Result<StatusCode, APIError> {
+    let mut tx = pool.begin_immediate().await?;
+    let investigation = get_polling_station_investigation(&mut tx, polling_station_id).await?;
+    let polling_station =
+        crate::polling_station::repository::get(&mut tx, polling_station_id).await?;
+    delete_polling_station_investigation(&mut tx, polling_station_id).await?;
```
rnijveld: We should also delete any related polling station results and polling station data entries to make sure that we don't leave any old results in there.

Lionqueen94: Implemented and tested in 08eadc01c08f5b37e703c9e12ef1b08cfe778d6e

stacktraceghost (approved):

backend/src/investigation/api.rs
```diff
@@ -201,11 +206,85 @@ async fn polling_station_investigation_update(
...
+    audit_service: AuditService,
+    Path(polling_station_id): Path<u32>,
+) -> Result<StatusCode, APIError> {
+    let mut tx = pool.begin_immediate().await?;
+    let investigation = get_polling_station_investigation(&mut tx, polling_station_id).await?;
+    let polling_station =
+        crate::polling_station::repository::get(&mut tx, polling_station_id).await?;
+
+    // Delete investigation and potential data entry and results linked to the polling station
+    delete_polling_station_investigation(&mut tx, polling_station_id).await?;
```
praseodym: Shouldn't you check the return value here, and throw 404 when nothing was deleted?

Lionqueen94: I already check if the investigation exists, but I can also check if the delete was successful yes. We don't currently do this everywhere though so this might be good to improve in the future for other endpoints

marlonbaeten: I can make a separate PR to add this to all delete endpoints if you want (after merging this).

Lionqueen94: I made a change in c4bc0a483008a8d76ec81869297124cc0a684ece not sure if this is the best way

frontend/src/features/investigations/components/InvestigationDelete.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export function InvestigationDelete({ pollingStation, onDeleted, onError }: InvestigationDeleteModalProps) {
+  const [showModal, setShowModal] = useState(false);
+  const { refetch } = useElection(pollingStation.id);
+  const electionStatuses = useElectionStatus();
+  const pollingStationStatus = electionStatuses.statuses.find(
+    (status) => status.polling_station_id === pollingStation.id,
+  );
+  const path = `/api/polling_stations/${pollingStation.id}/investigation`;
+  const { remove, requestState } = useCrud<PollingStationInvestigation>({ remove: path });
```
Lionqueen94: Is this line correct here? With a delete I don't expect an investigation back from the endpoint :thinking: @oliver3

oliver3: Hm no, I see it does not even return a boolean which some delete endpoints do, so you can remove the generic I'd say: `useCrud({ remove: path });`

Lionqueen94: Fixed in 419c2072b5747cb199b1cb87cd345b508c831832

stacktraceghost (approved):

PR #2176, opened 2025-09-23
Fix response types for polling station investigation endpoints

oliver3:
The `useCrud` generic was used incorrectly and the response type was not exposed in the API so could not be used.


frontend/src/features/investigations/components/InvestigationReason.tsx
```diff
@@ -27,8 +28,8 @@ export function InvestigationReason({ pollingStationId }: InvestigationReasonPro
   const { pushMessage } = useMessages();
   const [nonEmptyError, setNonEmptyError] = useState(false);
   const path = `/api/polling_stations/${pollingStationId}/investigation`;
-  const { create } = useCrud<PollingStationInvestigationCreateRequest>(path);
-  const { update } = useCrud<PollingStationInvestigationUpdateRequest>(path);
+  const { create } = useCrud<PollingStationInvestigation>(path);
+  const { update } = useCrud<PollingStationInvestigation>(path);
```
Lionqueen94:
```suggestion
  const { create, update } = useCrud<PollingStationInvestigation>(path);
```

PR #2180, opened 2025-09-23
Coordinator can manage typists only

### LINKED ISSUE #2177: Limit user management for coordinator
Coordinator can
- see all users in user management list
- can try to add an user, but receives error when submitting
- gets an unhandled error when editing one

According to the [authorisation matrix](https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/autorisatiematrix.md) only a `Beheerder` should be able to manage users in the `must have` phase. 

**Reproduction steps:**
1. Log in as coordinator
2. Go to status overview of data entry
3. Open the menu and choose user management
  3a. Coordinator should not see *all* users
4. Click on a user
  4a. 403 error is not handled properly, page shows `Cannot read properties of undefined (reading 'username')` 

**Required changes**
- Remove `Gebruikers` from coordinator menu
- Don't allow coordinator to retrieve users
- Handle 403 forbidden on user view page

marlonbaeten:
Fixes #2177

This allows the coordinator to manage users (get, list, create, update, delete) but only for typists.

Previously the coordinator could list *all* users, and clicking add or edit resulted in an error.

This also splits the authentication API endpoints from the user management API endpoints in separate files.

To test:

- Manage users as an admin
- Manage users as a coordinator and see that you can only create/update/delete typists


jschuurk-kr:
@marlonbaeten Am guessing the answer is "no", but would this PR be a good opportunity to address an issue I have with the `/user` API?
Currently the `/user` endpoint is used both for actions by a user on their own behalf, e.g. logout, and for actions by an admin or coordinator on the users in the application, e.g. listing all users. I think the endpoint for the second set of actions should be `/users`.

I can imagine that changing this will derail this PR too much, but I figured if I mention it, we might address it one day. :-)

marlonbaeten:
> @marlonbaeten Am guessing the answer is "no", but would this PR be a good opportunity to address an issue I have with the `/user` API? Currently the `/user` endpoint is used both for actions by a user on their own behalf, e.g. logout, and for actions by an admin or coordinator on the users in the application, e.g. listing all users. I think the endpoint for the second set of actions should be `/users`.
> 
> I can imagine that changing this will derail this PR too much, but I figured if I mention it, we might address it one day. :-)

Well I did split up the endpoints into separate routers, so this might the the time.

Edit: see https://github.com/kiesraad/abacus/pull/2180/commits/a482bc2854305ff39d6aea6025eb6a1f8b4151ad

stacktraceghost:
Functionally I've noticed:
- No dev link to user management for coordinator
- `fatal-api-errors` are not handled in FE for a 403/404
- Coordinators cannot delete typists

I will push some changes for this

stacktraceghost (approved):

PR #2184, opened 2025-09-24
Update Election Status overview for next committee session

### LINKED ISSUE #2002: Update Election Status overview for next committee session
For a next committee session, only show polling stations that have an investigation with corrected results

NB: related to #2151, make sure that we implement the backend change in one of the two issues.

Changes to Status overview design that need to be implemented when not first committee session:
https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=15091-45835&m=dev

### LINKED ISSUE #2129: Adjust showing finish data entry alert for next committee session
Depends on #2151 

Show finish data entry alert for next committee session if investigations with corrected results and those polling stations have 2 definitive data entries

stacktraceghost:
Resolves #2002
Resolves #2129 

- `data_entry::repository::statuses`
  - now uses `committee_session_id` iso `election_id`
  - returns all polling station statuses for first committee session
  - returns only polling stations with investigations that have corrected results for next committee session
- FE `useElectionStatus` now filters polling stations based on which are available within `statuses`. This way the status counts/progress is only based on polling stations relevant for that committee session
- FE: Show different text in next committee session if no polling stations are shown in status overview
- FE: Show different button in overview for next session (to investigations iso polling stations)
- Adjusted fixture `election_5_with_results` so it does not trigger F304 ("Vergelijk D&H" no checkboxes checked in this case)
- Storybook: adjusted `ElectionStatus`  story and added next committee session with/without polling stations, testing the differences compared to the first one.
- Added tests
  - Apportionment: next committee session without and with polling stations that have investigations with corrected results
  - Data entry API: tests to validate what statuses are returned for first/next committee session
  - Moved `create_investigation`/`update_investigation` to `tests/shared/mod.rs` 

  
[Temporary test environment](https://toegewezen-permanente-hoofdstembureau.abacus-test.nl/)
1. Log in as coordinator
2. [View an election with first session](https://toegewezen-permanente-hoofdstembureau.abacus-test.nl/elections/1/status)
3. [View an election with second session (with investigations, one of which has a corrected result)](https://toegewezen-permanente-hoofdstembureau.abacus-test.nl/elections/5/status)
4. [View an election with second session (no investigations)](https://toegewezen-permanente-hoofdstembureau.abacus-test.nl/elections/6/status)

**To check/assumption**
If during a next committee session a polling station gets added, it will not be visible in the status overview until an investigation is created with corrected results. Does this make sense?


jorisleker:
> **To check/assumption**
> If during a next committee session a polling station gets added, it will not be visible in the status overview until an investigation is created with corrected results. Does this make sense?

Yes, this is how it was intended. It is only relevant in the status overview when question 'are there corrected results' in the investigation is answered with 'yes'.

frontend/src/features/election_status/hooks/useElectionStatus.ts
```diff
@@ -63,6 +63,10 @@ export function useElectionStatus(
 ): ElectionStatusData {
   const { getName } = useUsers();
 
+  // Only include polling stations that have a status
+  // This filters out polling stations that are not relevant to a committee session
+  pollingStations.filter((ps) => statuses.some((s) => s.polling_station_id === ps.id));
```
oliver3: The `filter()` does not modify the array. So maybe this code is not needed?

backend/src/data_entry/repository.rs
```diff
@@ -114,13 +116,14 @@ pub async fn statuses(
...
-                SELECT MAX(c2.number)
-                FROM committee_sessions AS c2
-                WHERE c2.election_id = c.election_id
-            )
+            LEFT JOIN polling_station_investigations AS psi ON psi.polling_station_id = p.id
+            WHERE c.id = $1 AND
+                CASE
+                    WHEN c.number > 1 THEN psi.corrected_results = TRUE
+                    ELSE TRUE
+                END
```
oliver3:
```suggestion
            WHERE c.id = $1 AND (c.number = 0 OR psi.corrected_results)
```
could be better, not sure

stacktraceghost:
> > **To check/assumption**
> > If during a next committee session a polling station gets added, it will not be visible in the status overview until an investigation is created with corrected results. Does this make sense?
> 
> Yes, this is how it was intended. It is only relevant in the status overview when question 'are there corrected results' in the investigation is answered with 'yes'.

For documentation, the following has been discussed:

**If the following scenario happens:**
In committee session 2:
1. Add an investigation with corrected results to polling station A which already existed in committee session 1
2. Add a **new** polling station (B) which did **not** exist in committee session 1 and either [don't create an investigation with corrected results]/[create an investigation without corrected results]

Now only polling station A will be shown in the status overview of committee session 2. 

If you then complete the data entries for polling station A and finalize the committee session, polling station B will have no data entries. This results into errors later in the process (like creating documents).

**Proposed solution:**
- When adding a polling station in a next committee session, automatically add an investigation with `investigation.corrected_results=true`
- Deleting this investigation should result in the deletion of the polling station
- Adjusting `investigation.corrected_results` to anything other than `true` should not be possible

We have decided to split this off into two new issues:
1. Must-have: when finishing data entry for a committee session, check if each polling station has finalized data entries » https://github.com/kiesraad/abacus/issues/2188
2. Should-have: implement proposed solution above » https://github.com/kiesraad/abacus/issues/2187

backend/tests/apportionment_integration_test.rs
```diff
@@ -203,6 +205,59 @@ async fn test_election_apportionment_works_for_19_or_more_seats(pool: SqlitePool
...
+    let response = reqwest::Client::new()
+        .post(&url)
+        .header("cookie", coordinator_cookie)
+        .send()
+        .await
+        .unwrap();
+
+    // Ensure the response is what we expect
+    assert_eq!(response.status(), StatusCode::OK);
+    let body: ElectionApportionmentResponse = response.json().await.unwrap();
```
oliver3, line 223: We shouldn't use structs in the integration tests but JSON, as described in https://github.com/kiesraad/abacus/blob/main/documentatie/ontwikkelproces/test-tooling.md#integratietests
.. but I see that the rest of this integration test is not according to this guideline yet, so you can ignore this comment.

frontend/src/features/election_status/components/ElectionStatus.test.tsx
```diff
@@ -61,62 +60,19 @@ describe("ElectionStatus", () => {
...
-    ]);
-    const errorsAndWarningsRows = within(table).getAllByRole("row");
-    // Click on row of polling station with data entries with differences
-    errorsAndWarningsRows[1]!.click();
-    expect(navigate).not.toHaveBeenCalled();
-    // Click on row of polling station with data entry with errors
-    errorsAndWarningsRows[2]!.click();
-    expect(navigate).not.toHaveBeenCalled();
+      expect(navigate).not.toHaveBeenCalled();
+    }
```
oliver3, line 76: Why did you remove the verification of the table's content?

itsamaik: It is still being checked above on line 57. This change merges two almost exactly similar tests to prevent code duplication.

backend/tests/apportionment_integration_test.rs
stacktraceghost: Other integration tests in the codebase seem to have the same issue too. The documentation is not fully clear if no structs should be for the request or also the response (I assume so). Should we make a separate task for this?

frontend/src/features/election_status/components/ElectionStatus.test.tsx
stacktraceghost: It is still being checked above on line 57. This change merges two almost exactly similar tests to prevent code duplication.

backend/src/data_entry/repository.rs
stacktraceghost: Agreed! Although I will use `c.number = 1`

stacktraceghost: Fixed in https://github.com/kiesraad/abacus/pull/2184/commits/8a70efcc36eed90cdac1c9977740258458136ff8

frontend/src/features/election_status/hooks/useElectionStatus.ts
stacktraceghost: I overlooked this one, thanks! Removed in https://github.com/kiesraad/abacus/pull/2184/commits/8a70efcc36eed90cdac1c9977740258458136ff8

oliver3 (dismissed):

backend/tests/apportionment_integration_test.rs
stacktraceghost: Issue added: https://github.com/kiesraad/abacus/issues/2191

backend/src/investigation/mod.rs
praseodym, line 1: The visibility here was on purpose:
- `repository` submodule is private, only accessible through what is re-exported. So if you need a function from the repository in another module, re-export it instead of making the entire `repository` submodule public.
- The re-exports are `pub(crate)`, because [integration tests should be using JSON](https://github.com/kiesraad/abacus/blob/main/documentatie/ontwikkelproces/test-tooling.md#integratietests) instead of directly using structs from the abacus crate.

backend/tests/apportionment_integration_test.rs
```diff
@@ -203,6 +205,59 @@ async fn test_election_apportionment_works_for_19_or_more_seats(pool: SqlitePool
...
+    assert_eq!(body.seat_assignment.seats, 23);
+    assert_eq!(body.seat_assignment.quota, Fraction::new(1200, 23));
+    assert_eq!(body.seat_assignment.steps.len(), 4);
+    let total_seats = get_total_seats_from_apportionment_result(&body.seat_assignment);
+    assert_eq!(total_seats, vec![12, 6, 1, 2, 2]);
+}
+
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("users", "election_5_with_results"))))]
+async fn test_election_apportionment_error_apportionment_not_available_until_data_entries_finalised_second_committee_session(
```
praseodym: Please shorten this test name a bit :)

backend/src/investigation/mod.rs
praseodym: Fixed in 7f69ac6738b7a37eddb6e34be3808c61d8c43b1c

backend/tests/apportionment_integration_test.rs
praseodym: Fixed in 08df86645a54d07903807bd68f120b55e1ee5178

praseodym (dismissed):

backend/src/data_entry/repository.rs
```diff
@@ -181,13 +183,10 @@ pub async fn statuses(
             FROM polling_stations AS p
             LEFT JOIN committee_sessions AS c ON c.id = p.committee_session_id
             LEFT JOIN polling_station_data_entries AS de ON de.polling_station_id = p.id
-            WHERE c.election_id = $1 AND c.number = (
-                SELECT MAX(c2.number)
-                FROM committee_sessions AS c2
-                WHERE c2.election_id = c.election_id
-            )
+            LEFT JOIN polling_station_investigations AS psi ON psi.polling_station_id = p.id
+            WHERE c.id = $1 AND (c.number = 1 OR psi.corrected_results)
```
rnijveld:
```suggestion
            WHERE c.id = $1 AND (c.number = 1 OR psi.corrected_results IS NOT NULL)
```

I would not rely on the specific type juggling that sqlite does here.

stacktraceghost: This would not be correct, since we are checking if `corrected_results` is true. I would propose `psi.corrected_results = 1` in that case. 

Note that `corrected_results` does not contain results, but is an integer that we use for a boolean value (or null for not answered).

stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2184/commits/1d37692337a0efa7c2a5ddc397605bdfaacfab2b

rnijveld (approved):

PR #2195, opened 2025-09-29
Dependabot: also update indirect dependencies

rnijveld:


oliver3:
If I understand correctly, `all` does the direct dependencies, and one level lower, but still not all? Not sure if that changes your suggestion though. 

From https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#dependency-type-allow
> All explicitly defined dependencies. For bundler, pip, composer, cargo, gomod, also the dependencies of direct dependencies.

Edit: might just be imprecise wording in the documentation above..

praseodym:
Dependabot [said it created a pull request](https://github.com/kiesraad/abacus/actions/runs/18091017634) but I don't see it :(

praseodym:
Re-runnning: ~~https://github.com/kiesraad/abacus/actions/runs/18091017634~~ https://github.com/kiesraad/abacus/network/updates/21168368/jobs

praseodym:
https://github.com/kiesraad/abacus/pull/2198

PR #2206, opened 2025-09-29
Remove apportionment API

praseodym:
Remove apportionment API and its integration test, because the GSB version of Abacus does not need to do apportionment.


oliver3:
I would rather remove this feature completely for now:
- this functionality cannot be reached so it will "rot" over time
- every PR that touches this has some extra small effort, and cannot be tested well
- this PR already removes some code that has to be restored
- this code can raise questions in reviews etc

praseodym:
> I would rather remove this feature completely for now:
> 
>     * this functionality cannot be reached so it will "rot" over time
> 
>     * every PR that touches this has some extra small effort, and cannot be tested well
> 
>     * this PR already removes some code that has to be restored
> 
>     * this code can raise questions in reviews etc

I agree, closing in favour of #2208

PR #2210, opened 2025-09-30
Only start data entry for investigations if corrected_results=true

### LINKED ISSUE #2151: Only start data entry for investigations "Is er een gecorrigeerde uitslag?" yes
- Update data entry home to only show polling stations that are under investigation and have "Is er een gecorrigeerde uitslag?" set to yes
- Update claim endpoint to check for the same

Implement by updating the status endpoint to return only polling stations that comply with this condition.

stacktraceghost:
Resolves #2151 

Part of this issue was already resolved by https://github.com/kiesraad/abacus/pull/2184:
- Statuses endpoint only returns polling stations which have an investigation with `corrected_results=true`

This PR adds validation to the data_entry API endpoints:
- If `committee_session.number > 1`, check if there is an investigation with `corrected_results=true`
- Added this validation to endpoints: `claim, save, delete, finalise, get_errors, resolve_errors, get_differences, resolve_differences`
- Deduplicated data/validation logic by adding `validate_and_get_data` and centralizing that for these endpoints.
- Added tests for claiming data entry


backend/src/data_entry/api.rs
```diff
@@ -81,23 +82,74 @@ pub fn router() -> OpenApiRouter<AppState> {
         .routes(routes!(election_status))
 }
 
-async fn get_polling_station_election_and_committee_session_id(
+async fn validate_and_get_data(
     polling_station_id: u32,
+    user: &User,
     conn: &mut SqliteConnection,
```
oliver3: Not sure why but I expected `conn` to be the first argument

backend/src/data_entry/api.rs
```diff
@@ -1181,6 +1159,71 @@ mod tests {
...
+            742,
+            "Test".into(),
+            Some("Test".into()),
+            Some(false),
+        )
+        .await
+        .unwrap();
+
+        let response = claim(pool.clone(), 742, EntryNumber::FirstEntry).await;
+        assert_eq!(response.status(), StatusCode::CONFLICT);
```
oliver3, line 1172: Maybe assert the `ErrorReference` as well here?

backend/src/data_entry/api.rs
```diff
@@ -1977,6 +2066,18 @@ mod tests {
             .await
             .unwrap();
         add_results(&pool, 722, 702, &election.political_groups, false).await;
+
+        // Add investigation with corrected_results to be able to claim the polling station
+        insert_test_investigation(
+            &mut pool.acquire().await.unwrap(),
+            742,
+            "Test".into(),
+            Some("Test".into()),
```
oliver3: Wondering if we should hard-code reason and findings in the `insert_test_investigation()`

oliver3 (commented):
Three small suggestions that you can also dismiss :)

oliver3:
image

When you try to start data entry for an investigation with corrected_results set to No, you will get this error, is that ok @jorisleker  ?

oliver3:
When an investigation has been added, it seems from the data entry homepage that you can start data entry for that polling station, but when starting you will get an error `Dit kan nu niet` `409 De zitting is niet in de juiste status voor deze actie`

jorisleker:
@oliver3 You are right, the feedback "Geen stembureau gevonden met nummer 41" is not correct. We should only show this when the polling station does not exist.

When the polling station exists, but no investigation has been added, or the findings for the investigations have not been filled in, or the investigation has new results set to 'no', we could update the feedback to:
"Stembureau 41 kan nu niet ingevoerd te worden". 

You should not be able to start a data entry in these situations.

oliver3:
The frontend reacts a bit funny to an investigation that changes from corrected_results Yes to No while doing data entry, I suggest that we look at that in a separate issue.
image

backend/src/data_entry/api.rs
```diff
@@ -81,23 +82,74 @@ pub fn router() -> OpenApiRouter<AppState> {
...
+
+    let data_entry_status = crate::data_entry::repository::get_or_default(
+        conn,
+        polling_station_id,
+        committee_session.id,
+    )
+    .await?;
+
+    // Validate polling station
+    if committee_session.number > 1 {
```
lama-d3v:
```suggestion
    if committee_session.is_next_session() {
```

lama-d3v: I did not test this code, but it feels like that when its the first committee session, this function could still return OK(). In my opinion, this isn't correct.

backend/src/data_entry/api.rs
```diff
@@ -81,23 +82,74 @@ pub fn router() -> OpenApiRouter<AppState> {
...
+        Role::Coordinator => {
+            if committee_session.status != CommitteeSessionStatus::DataEntryInProgress
+                && committee_session.status != CommitteeSessionStatus::DataEntryPaused
+            {
+                return Err(APIError::CommitteeSession(
+                    CommitteeSessionError::InvalidCommitteeSessionStatus,
+                ));
+            }
+        }
+        _ => {}
```
lama-d3v: Maybe trigger an error if for some reason a different role is trying to execute this code?

backend/src/data_entry/api.rs
```diff
@@ -1138,6 +1103,19 @@ mod tests {
         assert_eq!(response.status(), StatusCode::OK);
     }
 
+    #[test(sqlx::test(fixtures(path = "../../fixtures", scripts("election_2"))))]
+    async fn test_claim_data_entry_ok(pool: SqlitePool) {
+        let response = claim(pool.clone(), 1, EntryNumber::FirstEntry).await;
+        assert_eq!(response.status(), StatusCode::OK);
+
+        // Check that row was created
+        let row_count = query!("SELECT COUNT(*) AS count FROM polling_station_data_entries")
```
lama-d3v: I think this query can be replaced with the function `data_entry_exists()` from `repository.rs`

lama-d3v: Same for line 1168 and 1195

frontend/src/features/data_entry/components/DataEntryProvider.tsx
```diff
@@ -25,12 +25,17 @@ export function DataEntryProvider({ election, pollingStationId, entryNumber, chi
...
-        void navigate(`/elections/${election.id}/data-entry#invalid-action-${pollingStationId}`);
+      switch (stateAndActions.error.reference) {
+        case "DataEntryAlreadyClaimed":
+          void navigate(`/elections/${election.id}/data-entry#data-entry-claimed-${pollingStationId}`);
+          break;
+        case "DataEntryAlreadyFinalised":
+          void navigate(`/elections/${election.id}/data-entry#data-entry-finalised-${pollingStationId}`);
+          break;
+        case "InvalidStateTransition":
+        case "DataEntryNotAllowed":
```
lama-d3v, line 36: Perhaps add a default case here as well?

backend/src/data_entry/api.rs
stacktraceghost: Ah did not notice the `is_next_session()` method, thanks.

The first committee session should pass here, because data entry should be possible then (no requirement based on investigation).

backend/src/data_entry/api.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2210/commits/ac7ca98a9f81b48102366bc065f82caa7db42cd7

backend/src/data_entry/api.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2210/commits/3c28466d67e3c267acdb7c82620c1c1fa166643b (~R1173-1175)

backend/src/data_entry/api.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2210/commits/9c0a53fa10984f45b92420f045f1af801a3a00da

stacktraceghost:
> image
> When you try to start data entry for an investigation with corrected_results set to No, you will get this error, is that ok @jorisleker ?

Adjusted in https://github.com/kiesraad/abacus/pull/2210/commits/75545c306f2022df0bc242c16062821867bba8ba. It checks if there was a status for the chosen polling station in the `/api/elections/5/status` response. That response only contains statuses for polling stations that are relevant in the current session. If there is no status given, we assume that data entry is not allowed.

Also simplified `PollingStationChoiceForm.tsx‎` and grouped the i18n based on the statuses.

backend/src/data_entry/api.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2210/commits/04ee7a1b6e37b7e38008aef07f4c8de820dd9448

backend/src/data_entry/api.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2210/commits/3c28466d67e3c267acdb7c82620c1c1fa166643b

backend/src/data_entry/api.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2210/commits/ac7ca98a9f81b48102366bc065f82caa7db42cd7 (~R138-140)

frontend/src/features/data_entry/components/DataEntryProvider.tsx
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2210/commits/676607387d2723045c3a70d6f6789e4029e69503

Edit: reverted in https://github.com/kiesraad/abacus/pull/2210/commits/748f1b54d6768d7257c8caecc21cccdd2f065a95
Changing this logic results in another error not being shown and breaking e2e tests. My changes only convert the ifs to switch and adds one case. I'd prefer not to make this PR bigger.

praseodym (approved):

oliver3:
> The frontend reacts a bit funny to an investigation that changes from corrected_results Yes to No while doing data entry, ...

Added separate issue https://github.com/kiesraad/abacus/issues/2245

PR #2232, opened 2025-09-30
Generate test election from the dev page

marlonbaeten:
Create a test election from the `/dev` page:

image

Note that:

- we do not yet have the the code to generate a second committee session / investigations / corrections
- the code to generate an election was moved from the `gen-test-election.rs` file to the `test_data_gen` module

To test: use abacus-test.nl


backend/Cargo.toml
```diff
@@ -64,6 +64,7 @@ cookie = { version = "0.18.1", features = ["percent-encode"] }
 async_zip = { version = "0.0.18", features = ["tokio", "deflate", "chrono"] }
 strum = { version = "0.27", features = ["derive"] }
 tokio-util = { version = "0.7.16", features = ["io"] }
+rand_chacha = "0.9.0"
```
praseodym: Why change from `rand` to `rand_chacha`?

marlonbaeten: `rand::rng()` is not `Send` which is a requirement for an axum handler.

praseodym: I see! Can you put `rand_chacha` behind the `dev-database` feature in `Cargo.toml` too?

praseodym: Alternatively it looks like `let mut rng = StdRng::from_rng(&mut rand::rng());` works too and avoids an extra dependency (see [StdRng](https://docs.rs/rand/latest/rand/rngs/struct.StdRng.html))

marlonbaeten: Nice!

marlonbaeten: `StdRng` is a wrapper around [rand_chacha](https://docs.rs/rand_chacha/0.9.0/rand_chacha/index.html), but using `StdRng` prevents an additional dep.

marlonbaeten: https://github.com/kiesraad/abacus/pull/2232/commits/0be95bc3d02095015fe1662b97c66eae16ef1932

praseodym (approved):

stacktraceghost (approved):

PR #2238, opened 2025-10-06
Typst template smoke tests with data edge cases

marlonbaeten:
Edge case tests for our typst templates.

These tests are created to find errors in the typst templates by generating PDF files with various input values.

The input values are semi-randomly constructed with the intent to find errors in the handling of edge-cases (empty strings, None values, empty arrays, missing data, etc.)

Note that these tests do not validate that the correct data is rendered in the PDF.

Two errors were found and fixed:

- `backend/templates/model-na-14-2.typ` contained an undefined parameter (on the current `main`)
- Na 31-2 fails if the start date/time of the committee session is not set. I have added a check in the code that will verify the date/time is set before generating the PDF.

### DIFF
--- a/backend/src/report/api.rs
+++ b/backend/src/report/api.rs
@@ -136,7 +136,11 @@ async fn generate_and_save_files(
     let mut conn = pool.acquire().await?;
     let committee_session =
         crate::committee_session::repository::get(&mut conn, committee_session_id).await?;
-    if committee_session.status != CommitteeSessionStatus::DataEntryFinished {
+
+    // Only generate files if the committee session is finished and has a start date time
+    if committee_session.status != CommitteeSessionStatus::DataEntryFinished
+        || committee_session.start_date_time.is_none()
+    {
         return Err(APIError::CommitteeSession(
             CommitteeSessionError::InvalidCommitteeSessionStatus,
         ));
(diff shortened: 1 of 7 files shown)


backend/src/report/api.rs
```diff
@@ -136,7 +136,11 @@ async fn generate_and_save_files(
     let mut conn = pool.acquire().await?;
     let committee_session =
         crate::committee_session::repository::get(&mut conn, committee_session_id).await?;
-    if committee_session.status != CommitteeSessionStatus::DataEntryFinished {
+
+    // Only generate files if the committee session is finished and has a start date time
```
lama-d3v, line 140: does this check need to be added to the typst tests as well?

praseodym (approved):

backend/src/report/api.rs
marlonbaeten: We tests the typst models with all possible values from the data structure: https://github.com/kiesraad/abacus/pull/2238/files#diff-cdf0cab3d9c666e0d5500c1463b1c6a0f12d6659215773b2557d7bb0967f84b8R216

These tests make sure our models will be robust in accepting different inputs, we do not want to do data validation within typst, that should be done in our Rust code.

Issue #2219, opened 2025-09-30
EML imports


oliver3:
Explore EML imports
with text editor and the application frontend
to discover robustness against exotic files

Timebox 90 minutes

cikzh:
Working on #2110 turned into an exploratory test session, so I've self-assigned this one

cikzh:
#### Tested on abacus-test.nl with commit: 29bf7a93ff50c15a4ef89790c4b2e17750d2053d
1. As administrator, navigated to /elections -> Verkiezing toevoegen
2. Uploaded a binary file (I used a 59.7MB tarball)
3. Get either "Abacus is stuk" screen or and inline alert, the latter being the intended behavior.
4. Repeat steps 2 and 3

#### Upon diving in further
Tested locally with a fresh frontend build and `cargo run --features memory-serve -- -rs`
- HTTP response as observed in Firefox devtools when behavior is as expected:
```http
POST /api/elections/import/validate HTTP/1.1
```
- HTTP response as observed in Firefox devtools when frontend expected HTTP 413 response:
```http
POST /api/elections/import/validate undefined
```
- In both cases the backend logs is the following:
```
INFO request{method=POST uri=/api/elections/import/validate version=HTTP/1.1}: tower_http::trace::on_response: finished processing request latency=7 ms status=413
```

#### Some experiments in the backend
We're using `axum::extract::DefaultBodyLimit` to limit incoming request bodies. Adding `tower_http::limit::RequestBodyLimit` doesn't change the behavior. 
Instead of returning 413 immediately when checking the `Content-Length` header, the backend seems to read through the whole body, which is unexpected to me, since the docs state it limits based on  the `Content-Length` header.
Checking the devtools We do actually provide the `Content-Length` header, which I observed in the devtools our request does provide.

#### Observing the traffic with Wireshark
<img width="2512" height="502" alt="Image" src="https://github.com/user-attachments/assets/5688778d-b80c-42ca-bb7f-e73a484bad07" />

We can see from the Wireshark capture that the HTTP 413 response and TCP reset are sent before the HTTP request is processed. 

### Conclusion
It is understandable that the backend resets the connection on excessive payloads. What I don't understand after these tests is why the behavior is nondeterministic.

cikzh:
Next test: go through the Election create flow with files that individually are within upload limits, but exceed the limit together

cikzh:
#### Preparation
- Prepared valid, but large EML files. 4.9MB each
[...]

PR #2240, opened 2025-10-06
Frontend file size check

### LINKED ISSUE #2110: Importing a large file (e.g. election definition) results in "Abacus is stuk" when running locally
Fixing this can wait until after the epics for corrected results have been completed.

---

When you import a large file (e.g. election definition, 26 MB is large enough to trigger the error) while running locally, you get the "Abacus is stuk" page with error message "NetworkError when attempting to fetch resource., POST /api/elections/import/validate". The network tab in the dev tools shows an NS_ERROR_NET_RESET in the Transferred column. I don't see any errors or warnings in the backend logs. The audit log does have an entry that says "API fout: Het bestand is groter dan toegestaan."

To reproduce this, run the application locally with `--features memory-serve` as this behavior is not reproducible on abacus-test.nl.

Proposed solution is to have the front-end check the size of the upload, because the backend already does the only thing it can do in case of a large file: drop the connection.

### LINKED ISSUE #2268: Tune upload file size limits
- Determine useful file size limits and adjust them for these cases:
1. Election definition upload
2. Candidate list upload
3. Polling station import (election create)
4. Polling station import (polling station CRUD)

- Adjust the backend maximum payload size, by summing case 1-3

cikzh:
- Fixes #2110
- Fixes #2268
- Relates to https://github.com/kiesraad/abacus/issues/2219

Scope:
- Adds checks to file uploads in the frontend.
- Adds 110a test file that is valid, but too large
- Set the frontend upload limit to 5MB (let me know if it's an appropriate limit)

Testing tips:
- Try uploading an extremely large file (>50MB). Notice that instead a long 'pause', while waiting for the file to be uploaded and rejected by the backend, the error will be shown immediately
- Upload fields are in Election Create and Polling Station Import pages

### DIFF
--- a/frontend/e2e-tests/page-objects/election/create/UploadCandidateDefinitionPgObj.ts
+++ b/frontend/e2e-tests/page-objects/election/create/UploadCandidateDefinitionPgObj.ts
@@ -2,13 +2,13 @@ import { type Locator, type Page } from "@playwright/test";
 
 export class UploadCandidateDefinitionPgObj {
   readonly header: Locator;
-  readonly error: Locator;
+  readonly invalidFileAlert: Locator;
   readonly upload: Locator;
   readonly main: Locator;
 
   constructor(protected readonly page: Page) {
     this.header = page.getByRole("heading", { level: 2, name: "Importeer kandidatenlijsten" });
-    this.error = page.getByRole("strong").filter({ hasText: "Ongeldige kandidatenlijsten" });
+    this.invalidFileAlert = page.getByRole("alert").filter({ hasText: "Ongeldige kandidatenlijsten" });
     this.upload = page.getByRole("button", { name: "Bestand kiezen" });
     this.main = page.getByRole("main");
   }
@@ -18,9 +18,5 @@ export class UploadCandidateDefinitionPgObj {
     await this.page.getByText("Bestand kiezen").click();
     const fileChooser = await fileChooserPromise;
     await fileChooser.setFiles(path);
-  }
-
-  fileTooLargeError(filename: string) {
-    return this.page.getByRole("strong").filter({ hasText: filename });
   }
 }
--- a/frontend/e2e-tests/tests/election-create.e2e.ts
+++ b/frontend/e2e-tests/tests/election-create.e2e.ts
@@ -191,7 +191,7 @@ test.describe("Election creation", () => {
     const uploadElectionDefinitionPage = new UploadElectionDefinitionPgObj(page);
     await expect(uploadElectionDefinitionPage.header).toBeVisible();
     await uploadElectionDefinitionPage.uploadFile(eml110b.path);
-    await expect(uploadElectionDefinitionPage.error).toBeVisible();
+    await expect(uploadElectionDefinitionPage.invalidFileAlert).toBeVisible();
   });
 
   test("it fails on incorrect hash for candidate list", async ({ page }) => {
@@ -227,8 +227,10 @@ test.describe("Election creation", () => {
     const uploadElectionDefinitionPage = new UploadElectionDefinitionPgObj(page);
     await expect(uploadElectionDefinitionPage.header).toBeVisible();
     await uploadElectionDefinitionPage.uploadFile(eml110aTooLargeTestFile.path);
-    await expect(uploadElectionDefinitionPage.error).toBeVisible();
-    await expect(uploadElectionDefinitionPage.fileTooLargeError(eml110aTooLargeTestFile.filename)).toBeVisible();
+    await expect(uploadElectionDefinitionPage.invalidFileAlert).toBeVisible();
+    await expect(uploadElectionDefinitionPage.invalidFileAlert).toContainText(
+      `Het bestand ${eml110aTooLargeTestFile.filename} is te groot. Kies een bestand van maximaal 5 Megabyte.`,
+    );
   });
 
   test("it fails on valid, but incorrect file for candidate list", async ({ page }) => {
@@ -248,7 +250,7 @@ test.describe("Election creation", () => {
     const uploadCandidateDefinitionPage = new UploadCandidateDefinitionPgObj(page);
     await expect(uploadCandidateDefinitionPage.header).toBeVisible();
     await uploadCandidateDefinitionPage.uploadFile(eml110b.path);
-    await expect(uploadCandidateDefinitionPage.error).toBeVisible();
+    await expect(uploadCandidateDefinitionPage.invalidFileAlert).toBeVisible();
   });
 
   test("it fails on too large file for candidate list", async ({ page, eml230bTooLargeTestFile }) => {
@@ -268,8 +270,10 @@ test.describe("Election creation", () => {
     const uploadCandidateDefinitionPage = new UploadCandidateDefinitionPgObj(page);
     await expect(uploadCandidateDefinitionPage.header).toBeVisible();
     await uploadCandidateDefinitionPage.uploadFile(eml230bTooLargeTestFile.path);
-    await expect(uploadCandidateDefinitionPage.error).toBeVisible();
-    await expect(uploadCandidateDefinitionPage.fileTooLargeError(eml230bTooLargeTestFile.filename)).toBeVisible();
+    await expect(uploadCandidateDefinitionPage.invalidFileAlert).toBeVisible();
+    await expect(uploadCandidateDefinitionPage.invalidFileAlert).toContainText(
+      `Het bestand ${eml230bTooLargeTestFile.filename} is te groot. Kies een bestand van maximaal 5 Megabyte.`,
+    );
   });
 
   test("warning modal close button should stay on page", async ({ page }) => {
@@ -541,7 +545,7 @@ test.describe("Election creation", () => {
     await expect(uploadElectionDefinitionPage.header).toBeVisible();
     await uploadElectionDefinitionPage.uploadFile(eml110a.path);
     await expect(uploadElectionDefinitionPage.main).toContainText(eml110a.filename);
-    await expect(uploadElectionDefinitionPage.error).toBeVisible();
+    await expect(uploadElectionDefinitionPage.invalidFileAlert).toBeVisible();
   });
 
   test("it fails on too large file for polling stations", async ({ page, eml110bTooLargeTestFile }) => {
@@ -558,13 +562,15 @@ test.describe("Election creation", () => {
     await pollingStationRolePage.next.click();
 
     await uploadCandidatesAndInputHash(page);
-    //
+
     // Polling stations page
-    const uploadElectionDefinitionPage = new UploadPollingStationDefinitionPgObj(page);
-    await expect(uploadElectionDefinitionPage.header).toBeVisible();
-    await uploadElectionDefinitionPage.uploadFile(eml110bTooLargeTestFile.path);
-    await expect(uploadElectionDefinitionPage.error).toBeVisible();
-    await expect(uploadElectionDefinitionPage.fileTooLargeError(eml110bTooLargeTestFile.filename)).toBeVisible();
+    const uploadPollingStationDefinitionPage = new UploadPollingStationDefinitionPgObj(page);
+    await expect(uploadPollingStationDefinitionPage.header).toBeVisible();
+    await uploadPollingStationDefinitionPage.uploadFile(eml110bTooLargeTestFile.path);
+    await expect(uploadPollingStationDefinitionPage.invalidFileAlert).toBeVisible();
+    await expect(uploadPollingStationDefinitionPage.invalidFileAlert).toContainText(
+      `Het bestand ${eml110bTooLargeTestFile.filename} is te groot. Kies een bestand van maximaal 5 Megabyte.`,
+    );
   });
 
   test("show more button should show full list of polling stations", async ({ page }) => {
--- a/frontend/e2e-tests/tests/polling-station-import.e2e.ts
+++ b/frontend/e2e-tests/tests/polling-station-import.e2e.ts
@@ -24,7 +24,7 @@ test.describe("Polling station import", () => {
 
     // Select incorrect file
     await importPage.uploadFile(eml110a.path);
-    await expect(importPage.error).toContainText("Ongeldig stembureaubestand");
+    await expect(importPage.invalidFileAlert).toContainText("Ongeldig stembureaubestand");
 
     // Select correct file
     await importPage.uploadFile(eml110b.path);
@@ -53,7 +53,9 @@ test.describe("Polling station import", () => {
 
     // Select file that's too large
     await importPage.uploadFile(eml110bTooLargeTestFile.path);
-    await expect(importPage.error).toBeVisible();
-    await expect(importPage.fileTooLargeError(eml110bTooLargeTestFile.filename)).toBeVisible();
+    await expect(importPage.invalidFileAlert).toBeVisible();
+    await expect(importPage.invalidFileAlert).toContainText(
+      `Het bestand ${eml110bTooLargeTestFile.filename} is te groot. Kies een bestand van maximaal 5 Megabyte.`,
+    );
   });
 });
--- a/frontend/e2e-tests/page-objects/election/create/UploadElectionDefinitionPgObj.ts
+++ b/frontend/e2e-tests/page-objects/election/create/UploadElectionDefinitionPgObj.ts
@@ -2,13 +2,13 @@ import { type Locator, type Page } from "@playwright/test";
 
 export class UploadElectionDefinitionPgObj {
   readonly header: Locator;
-  readonly error: Locator;
+  readonly invalidFileAlert: Locator;
   readonly upload: Locator;
   readonly main: Locator;
 
   constructor(protected readonly page: Page) {
     this.header = page.getByRole("heading", { level: 2, name: "Importeer verkiezingsdefinitie" });
-    this.error = page.getByRole("strong").filter({ hasText: "Ongeldige verkiezingsdefinitie" });
+    this.invalidFileAlert = page.getByRole("alert").filter({ hasText: "Ongeldige verkiezingsdefinitie" });
     this.upload = page.getByRole("button", { name: "Bestand kiezen" });
     this.main = page.getByRole("main");
   }
@@ -18,9 +18,5 @@ export class UploadElectionDefinitionPgObj {
     await this.page.getByText("Bestand kiezen").click();
     const fileChooser = await fileChooserPromise;
     await fileChooser.setFiles(path);
-  }
-
-  fileTooLargeError(filename: string) {
-    return this.page.getByRole("strong").filter({ hasText: filename });
   }
 }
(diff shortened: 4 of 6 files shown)


stacktraceghost (approved):

frontend/src/utils/fileUpload.ts
praseodym, line 1: I feel like this util is doing very little and could just be inlined wherever its being used. That would also allow the max file sizes to be changed per component.

frontend/e2e-tests/tests/election-create.e2e.ts
```diff
@@ -219,6 +219,17 @@ test.describe("Election creation", () => {
...
 
+  test("it fails on valid, but too large file", async ({ page }) => {
+    await page.goto("/elections");
+    const overviewPage = new ElectionsOverviewPgObj(page);
+    await overviewPage.create.click();
+
+    const uploadElectionDefinitionPage = new UploadElectionDefinitionPgObj(page);
+    await expect(uploadElectionDefinitionPage.header).toBeVisible();
+    await uploadElectionDefinitionPage.uploadFile(page, eml110a_too_large.path);
+    await uploadElectionDefinitionPage.fileTooLargeError(page, eml110a_too_large.filename);
```
oliver3: For me it is not clear now that the test ends in asserting that the error is present on the page. I would prefer to have the expectation here and that the Page Object contains the locator only.

frontend/e2e-tests/page-objects/election/create/UploadCandidateDefinitionPgObj.ts
```diff
@@ -19,4 +19,9 @@ export class UploadCandidateDefinitionPgObj {
     const fileChooser = await fileChooserPromise;
     await fileChooser.setFiles(path);
   }
+
+  async fileTooLargeError(page: Page, filename: string) {
```
oliver3: No need for the page argument, the page object already has a `this.page`. Please correct the `uploadFile()` as well.

frontend/src/utils/fileUpload.ts
```diff
@@ -0,0 +1,17 @@
+import { ReactElement } from "react";
+
+import { tx } from "@/i18n/translate";
+
+// Maximum file upload size in Megabytes
+export const MAX_FILE_UPLOAD_SIZE_MB = 5;
```
oliver3: Could you put in a comment how this 5 MB relates the `MAX_BODY_SIZE_MB` from the backend, so that it is clear for future devs that it is not an error that these numbers are not the same.

cikzh: ~~I think the whole situation around file uploads deserve some thorough documentation. There are a lot of moving parts and implicit intricacies with the uploads and their reported sizes and they apply to the application globally. Do you, or @marjoleintamis have suggestions on where and how to document this? I would be nice if this documentation will be discovered when we're going to implement new features that use an upload feature.~~

As we decided to move the limits from `utils` to the individual components, and to make the backend limit the sum of the three limits in the election create flow, I think it makes most sense to explain this in a comment in the backend. What do you think @oliver3?

frontend/src/utils/fileUpload.ts
cikzh: Being able to change the size per component is nice to have indeed, I'll change it :+1:

frontend/e2e-tests/page-objects/election/create/UploadCandidateDefinitionPgObj.ts
cikzh: Fixed in [ddaf7fc](https://github.com/kiesraad/abacus/pull/2240/commits/ddaf7fc65a1c8d445c8fa4776e42b741ed3cb865)

frontend/src/utils/fileUpload.ts
cikzh: Fixed in [e118020](https://github.com/kiesraad/abacus/pull/2240/commits/e1180201ee0de8b80b4abd987a2ef93e65170271)

frontend/src/features/election_create/components/UploadCandidatesDefinition.tsx
```diff
@@ -29,6 +32,20 @@ export function UploadCandidatesDefinition() {
...
+            "file_too_large",
+            {},
+            {
+              filename: currentFile.name,
+              max_size: `${MAX_FILE_UPLOAD_SIZE_MB}`,
+            },
+          ),
+        );
+        return;
+      }
```
Lionqueen94: It seems like the setError call is not considered covered by codecov in each component :thinking:

frontend/e2e-tests/tests/election-create.e2e.ts
cikzh: Changed in [e413845](https://github.com/kiesraad/abacus/pull/2240/commits/e413845d8f4417cd008614cb9570f3054504fc83)

backend/src/eml/tests/eml110a_invalid_file_size.eml.xml
praseodym, line 1: We should think about a better solution that doesn't require us to commit big test files to the repository /cc @marlonbaeten

frontend/src/utils/fileUpload.ts
oliver3: Yes that sounds like a plan!

oliver3:
If possible, please don't rebase and force push once you've submitted something for review. Github cannot help you anymore with what the newest changes are, and reviewers that checked out the branch to review and test have to remember to reset or clone the branch anew.

backend/src/eml/tests/eml110a_invalid_file_size.eml.xml
cikzh: Maybe we could script something that creates the file on the fly

praseodym: Yes I think that'd be good

cikzh: Discussed this with @marlonbaeten, ended up with a solution where the e2e test calls a feature-flag-gated endpoint that generates the EML files and temporarily saves it to disk.
The endpoint accepts a path parameter that indicates how big the resulting EML must be in bytes, so changing the size in the future is trivial.

praseodym:
@cikzh can you collaborate with @marlonbaeten on how to make this work for release builds?

cikzh:
@praseodym 

@marlonbaeten will add the feature to the release build once this PR is merged

praseodym:
> @praseodym
> 
> @marlonbaeten will add the feature to the release build once this PR is merged

But then we'll be shipping test code in our release build?

cikzh:
> But then we'll be shipping test code in our release build?

Sorry, I meant to say: to the release build that is used for e2e tests and abacus-test.nl, not the actual release build. 

See also: https://github.com/kiesraad/abacus/pull/2321 where the split happens

frontend/src/features/election_create/components/UploadCandidatesDefinition.tsx
cikzh: Added back tests in [63c44de](https://github.com/kiesraad/abacus/pull/2240/commits/63c44de02d4ac51df741d0b1c3d28ea4a77635f0)

marlonbaeten (dismissed):

backend/src/eml/eml_110.rs
```diff
@@ -355,7 +355,7 @@ pub struct EventIdentifier {}
 #[serde(rename_all = "PascalCase")]
 pub struct Election {
     election_identifier: ElectionIdentifier,
-    contest: Contest,
+    pub(crate) contest: Contest,
```
oliver3: It feels a bit random which fields are made `pub(crate)` and initially I could not find why because it tested and compiled fine without them. I realised later that it is only used for feature `e2e-helpers`.

Can we maybe do something similar to what we already do in `eml_230.rs`, but then with `#[cfg(feature = "e2e-helpers")]` to make this more obvious? 

```rust
impl EML230 {
    #[cfg(test)]
    pub fn affiliations(&self) -> &[Affiliation] {
        &self.candidate_list.election.contest.affiliations
    }
}
```

frontend/e2e-tests/tests/election-create.e2e.ts
```diff
@@ -235,8 +265,38 @@ test.describe("Election creation", () => {
...
+    await uploadCandidateDefinitionPage.uploadFile(eml110b.path);
+    await expect(uploadCandidateDefinitionPage.error).toBeVisible();
+  });
+
+  test("it fails on too large file for candidate list", async ({ page, request }) => {
+    // Generate EML via backend helper, save it to a file and upload it
+    const fileSize = 5 * 1024 * 1024; // 5MB
+    const response = await request.get(`/api/generate_eml230b/${fileSize}`);
+    const content = await response.text();
+    await promises.writeFile("../backend/src/eml/tests/eml230b_invalid_file_size.eml.xml", content, "utf8");
```
oliver3: These kind of calling-the-backend setups should go to the `fixtures.ts`. 

Added benefit is that the unlink/cleanup can go there as well and does not leave files in the repo if the test fails.

frontend/e2e-tests/tests/election-create.e2e.ts
```diff
@@ -21,9 +21,18 @@ import { AdminNavBar } from "e2e-tests/page-objects/nav_bar/AdminNavBarPgObj";
 import { PollingStationImportPgObj } from "e2e-tests/page-objects/polling_station/PollingStationImportPgObj";
 import { PollingStationListEmptyPgObj } from "e2e-tests/page-objects/polling_station/PollingStationListEmptyPgObj";
 import { PollingStationListPgObj } from "e2e-tests/page-objects/polling_station/PollingStationListPgObj";
+import { promises } from "fs";
```
oliver3: Usually imported like `import { unlink, writeFile } from "node:fs/promises";`

backend/src/eml/eml_110.rs
cikzh: This is a great idea, I'll implement it

.github/workflows/release.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y musl-tools
         if: matrix.target.os == 'ubuntu-latest'
       - name: Run tests
-        run: cargo --locked test --release --no-default-features --features memory-serve,embed-typst,airgap-detection --target=${{ matrix.target.name }}
+        run: cargo --locked test --release --no-default-features --features memory-serve,embed-typst,e2e-helpers,airgap-detection --target=${{ matrix.target.name }}
```
praseodym: I don't like the idea of creating release builds with test helper code

oliver3: Oh missed that.. I agree this should not be compiled into the release build

marlonbaeten: This feature should be added on line 84, not 81. The non-airgap version is used for the e2e tests and abacus-test. The "official" pre-release should not contain this feature.

frontend/e2e-tests/tests/polling-station-import.e2e.ts
```diff
@@ -23,16 +24,46 @@ test.describe("Polling station import", () => {
...
     await expect(listPage.alert).toContainText(/Er zijn \d+ stembureaus geïmporteerd/);
   });
+
+  test("As coordinator import a polling stations definition that's too large", async ({
+    page,
+    request,
+    emptyElection,
+  }) => {
+    // Generate EML via backend helper, save it to a file and upload it
+    const fileSize = 5 * 1024 * 1024; // 5MB
```
oliver3: This should use the fixture

oliver3: Fixed in https://github.com/kiesraad/abacus/pull/2240/commits/5690d9beac4c4b8192e13a99a8de9689b7e2f27e

.github/workflows/release.yml
oliver3: Should we rename that build, e.g. `abacus-${{ matrix.target.os }}-e2e-build` ?

frontend/e2e-tests/tests/election-create.e2e.ts
oliver3: Fixed in https://github.com/kiesraad/abacus/pull/2240/commits/703771df7f90caa6f079582f2de10f3a8f94034c

frontend/e2e-tests/test-data/eml-files.ts
```diff
@@ -25,11 +30,21 @@ export const eml110a = {
   ],
 };
 
+export const eml110a_too_large = {
+  filename: "eml110a_invalid_file_size.eml.xml",
+  path: "../backend/src/eml/tests/eml110a_invalid_file_size.eml.xml",
+};
+
```
jschuurk-kr: It doesn't make sense to me to have the generated files listed in this way. They look the same as the other files, but you can't use them directly, you have to call the fixture. So I think the path needs to be part of the fixture. (Same for the filename, which is used in the test.)

oliver3: Agreed, and while discussing this we realised that there is a possibility that one test could be run for e.g. two browsers in parallel, so the filenames should also be unique.

frontend/e2e-tests/tests/election-create.e2e.ts
```diff
@@ -219,6 +219,18 @@ test.describe("Election creation", () => {
     await expect(checkCandidateDefinitionPage.error).toBeVisible();
   });
 
+  test("it fails on valid, but too large file", async ({ page, eml110aTooLargeTestFile }) => {
```
jschuurk-kr, line 222: Am having the same problem as the pipeline, the error does not appear:

Image

Running the test with `npm run test:e2e -- -g "it fails on valid, but too large file" --project firefox`.

```
~/workspace/abacus/backend/src/eml/tests (frontend-file-size-check *%=) $ ls -lh eml110a_invalid_file_size.eml.xml 
-rw-rw-r-- 1 joep joep 4.6K Oct 29 12:37 eml110a_invalid_file_size.eml.xml
```

backend/src/eml/test_api.rs
```diff
@@ -0,0 +1,144 @@
...
+"#;
+/// Router for the e2e helper endpoints
+pub fn router() -> OpenApiRouter<AppState> {
+    OpenApiRouter::default()
+        .routes(routes!(generate_eml110a))
+        .routes(routes!(generate_eml110b))
+        .routes(routes!(generate_eml230b))
+}
+
+#[utoipa::path(
```
jschuurk-kr, line 67: We do have the utoipa-annotation, but I don't see the endpoints to generate the EMLs in our openapi.json.

backend/src/eml/test_api.rs
```diff
@@ -0,0 +1,144 @@
...
+    path = "/api/generate_eml110a/{min_size_bytes}",
+    responses(
+        (status = 200, description = "EML generated"),
+        (status = 500, description = "Internal server error", body = ErrorResponse),
+    ),
+)]
+async fn generate_eml110a(
+    Path(min_size_bytes): Path<usize>,
+) -> Result<(StatusCode, String), APIError> {
+    Ok((StatusCode::OK, eml110a(min_size_bytes)))
```
jschuurk-kr: Doesn't seem to matter what `min_size_bytes` I provide in the url (tried with 2, 1024 and 1073741824), I always get the same xml with length 103 lines back. Which is shorter than the file that's used as a strarting point, because the `<kr:ElectionTree>` is not in the generated file.

frontend/e2e-tests/page-objects/election/create/UploadCandidateDefinitionPgObj.ts
```diff
@@ -13,10 +13,14 @@ export class UploadCandidateDefinitionPgObj {
...
-    await page.getByText("Bestand kiezen").click();
+  async uploadFile(path: string) {
+    const fileChooserPromise = this.page.waitForEvent("filechooser");
+    await this.page.getByText("Bestand kiezen").click();
     const fileChooser = await fileChooserPromise;
     await fileChooser.setFiles(path);
   }
+
+  fileTooLargeError(filename: string) {
+    return this.page.getByRole("strong").filter({ hasText: filename });
```
jschuurk-kr: Would be better to check within an element with role alert.

jschuurk-kr: And to check error message to mention the file size.

frontend/e2e-tests/test-data/eml-files.ts
oliver3: Resolved in https://github.com/kiesraad/abacus/pull/2240/commits/818adae3fff8ef6951f88499dc1a35907fe67080

frontend/e2e-tests/tests/election-create.e2e.ts
oliver3: Issue caused by `clone()` in https://github.com/kiesraad/abacus/pull/2240/commits/818adae3fff8ef6951f88499dc1a35907fe67080 which results in files that do not have the large size as requested.

backend/src/eml/test_api.rs
oliver3: Added to OpenAPI in commit https://github.com/kiesraad/abacus/pull/2240/commits/236e8c27c9e004fb6754ddeab504eb1d9ed2851c

backend/src/eml/test_api.rs
oliver3: This is actually the root cause of "[Am having the same problem as the pipeline, the error does not appear" above](https://github.com/kiesraad/abacus/pull/2240#discussion_r2472677819).

frontend/e2e-tests/page-objects/election/create/UploadCandidateDefinitionPgObj.ts
oliver3: Fixed in https://github.com/kiesraad/abacus/pull/2240/commits/5a770edff6039285fd0df82c7fd4c063c9af4a75

backend/src/eml/test_api.rs
praseodym: 1. I don't agree that this should be exposed in the OpenAPI, since this is test functionality that will now also be included in the production TypeScript client
2. If we do want to include it everywhere, we should make `e2e-helpers` a default feature instead of adding `--features e2e-helpers` all over the place

jschuurk-kr: > I don't agree that this should be exposed in the OpenAPI, since this is test functionality that will now also be included in the production TypeScript client

Doesn't the same then apply to the endpoint to generate test elections? That was the reason to add these three to the OpenAPI, so that it lists all endpoints. (Where else would you be able to get that overview?) So maybe we need to add the ability to filter out endpoints in the TypeScript client generation?

oliver3:
After discussion we found that generating the EML in the backend is much too complex for what it is worth. An alternative has been implemented in PR https://github.com/kiesraad/abacus/pull/2365

PR #2241, opened 2025-10-07
Improve audit log filter UX

marlonbaeten:
The audit log filter has become a bit messy:

image

This PR tries to improve the UX by:
- Sorting filter options
- Changing the order of the filter elements

Since the list of event types has grown to a large list, I moved this to the bottom.
I noticed that I use the "since" a lot, so I moved that to the top.

image


stacktraceghost (approved):

lama-d3v:
I get an error when filtering on `FileDeleted` http://localhost:3000/logs?event=FileDeleted when testing locally.
image

praseodym:
> I get an error when filtering on `FileDeleted` http://localhost:3000/logs?event=FileDeleted when testing locally. 

Do you happen to have an adblocker active that blocks this request?

PR #2243, opened 2025-10-07
More uniform usage of the API methods

marlonbaeten:
- Fix throwing errors inside a promise
- Fix API specification for `polling_station_validate_import`
- Use API types where possible
- Use the `useCrud` hook where possible
- Added the `isLoading` in `useCrud` for better readability
- Remove unnecessary state for API errors


frontend/src/api/useCrud.ts
```diff
@@ -10,86 +10,71 @@ export type ApiRequestIdleState = {
...
+};
 
 // Call the api and return the current status of the request, optionally throws an error when the request fails
-export function useCrud<Response>(path: ApiPaths): UseCrudReturn<Response> {
+export function useCrud<Response>({ throwAllErrors = false, ...paths }: UseCrudOptions): UseCrudReturn<Response> {
   const client = useApiClient();
   const [requestState, setRequestState] = useState<ApiRequestIdleState | ApiRequestState<Response>>({ status: "idle" });
-  const paths = typeof path === "string" ? { get: path, create: path, update: path, remove: path } : path;
 
   // throw fatal errors
```
lama-d3v:
```suggestion
  // throw all errors
```

frontend/src/features/account/components/AccountSetupForm.tsx
```diff
@@ -54,14 +54,12 @@ export function AccountSetupForm({ user, onSaved }: AccountSetupFormProps) {
...
-          });
-        } else {
-          setApiError(result);
-        }
+      } else if (result instanceof ApiError && result.reference === "PasswordRejection") {
+        setValidationErrors({
+          password: t("account.password_rules"),
+        });
+      } else if (result instanceof ApiError) {
+        setApiError(result);
```
lama-d3v, line 62: is it correct that there is no else statement here anymore?

frontend/src/features/election_create/components/CheckAndSave.tsx
```diff
@@ -30,14 +25,16 @@ export function CheckAndSave() {
...
 
-    if (isSuccess(response)) {
-      // set navigation state to prevent blocking
-      await navigate("/elections", { state: { success: true } });
-    } else if (isError(response)) {
-      throw new Error();
-    }
+  // if no election, election data, candidate data or counting method is found in the state, go back to the beginning
+  if (!state.election || !state.electionDefinitionData || !state.candidateDefinitionData || !state.countingMethod) {
+    return <Navigate to="/elections/create" />;
```
lama-d3v, line 37: Shouldn't this be checked/executed before the handleSubmit (these states are required for this create request)?

frontend/src/api/useCrud.ts
marlonbaeten: Updated the comment in [25f0fc5](https://github.com/kiesraad/abacus/pull/2243/commits/25f0fc5eb48f4e7a70261ec86b002415ea042df6)

frontend/src/features/account/components/AccountSetupForm.tsx
marlonbaeten: Yes, in fact the extra condition is not nessesary, we are always dealing with an ApiError here, removed the condition in [25f0fc5](https://github.com/kiesraad/abacus/pull/2243/commits/25f0fc5eb48f4e7a70261ec86b002415ea042df6)

frontend/src/features/election_create/components/CheckAndSave.tsx
marlonbaeten: The `handleSubmit` is just a function definition. It's best practice to keept the hooks and definitions/assignments at the top of your function/component  and the JSX/render logic at the bottom.

frontend/src/features/users/components/update/UserUpdateForm.tsx
```diff
@@ -60,27 +57,25 @@ export function UserUpdateForm({ user, onSaved, onAbort }: UserUpdateFormProps)
           onSaved(result.data);
         } else if (result instanceof ApiError && result.reference === "PasswordRejection") {
           setValidationErrors({ temp_password: t("error.api_error.PasswordRejection") });
-        } else {
-          setError(result);
+        } else if (result instanceof ApiError) {
+          setError(result.reference);
```
oliver3, line 61: Wouldn't errors that are not an `ApiError` now become not dealt with? So FatalApiError | NetworkError | NotFoundError
Edit: While reading previous comments and answers I see that this will be caught somewhere else, but still TypeScript does not agree. Can we fix this?

frontend/src/api/useCrud.ts
```diff
@@ -10,86 +10,71 @@ export type ApiRequestIdleState = {
...
 
     return handleApiResult(result, setRequestState, controller);
   };
 
   // Update an existing resource
   const update = async (requestBody: object, controller?: AbortController) => {
-    if (!paths.update) {
-      throw new Error("No get path provided for update request");
+    if (!paths.updatePath) {
+      throw new Error("No update path provided for update request");
```
oliver3, line 52: So these errors are not caught if I am not mistaken?

frontend/src/features/users/components/update/UserUpdateForm.tsx
oliver3: Looked a bit how this could be done and the only way I can think of is throwing fatal errors in `handleApiResult` so that they will "disappear" intentionally and returning `ApiResult<T, ApiError>`, because the fatal errors in the requestState will be the same error and they will be thrown and handled by the error boundary.

```ts
// Update request state or throw an error based on the result of an API call
export function handleApiResult<T>(
  result: ApiResult<T>,
  setRequestState: (state: ApiRequestState<T>) => void,
  controller?: AbortController,
): ApiResult<T, ApiError> {
  // Do not update state if the request was aborted (mainly caused by an unmounted component)
  const aborted = controller instanceof AbortController && controller.signal.aborted;
  if (!aborted) {
    setRequestState(fromApiResult(result));
  }

  if (isFatalError(result)) {
    throw result;
  }

  return result;
}
```

marlonbaeten: So `FatalApiError | NetworkError | NotFoundError` will be thrown by `useCrud`. Throwing in `handleApiResult` will introduce a new problem: there will be uncaught errors, and our test suite will fail on those. So the current solution is a bit more verbose, but overall a nice improvement I think.

frontend/src/api/useCrud.ts
marlonbaeten: These errors are there for our test suite! They occur if a method is called for which no path was provided. In that case the test suite will fail on an uncaught error. But developers probably discover these kind of errors earlier on.

frontend/src/features/users/components/update/UserUpdateForm.tsx
oliver3: Hm ok I did not realise that would cause failing tests. I still think returning errors that you should ignore in the component is confusing but I do not know how to do that then. And yes this PR is already an improvement.

oliver3:
please merge main into branches instead of rebasing them, because after a rebase GitHub loses track of which changes were already reviewed.

It also causes issues if a reviewer updates their local checkout without realising the force push.

frontend/src/features/investigations/components/InvestigationDelete.tsx
```diff
@@ -13,34 +13,29 @@ import { PollingStation } from "@/types/generated/openapi";
...
     setShowModal(!showModal);
   }
 
   function handleDelete() {
-    void remove().then(async (result) => {
+    void remove().then((result) => {
       if (isSuccess(result)) {
-        await refetch();
+        void refetch();
         onDeleted(pollingStation);
```
oliver3: Sorry for the nabrander, but it does not wait for `refetch()` anymore before calling `onDelete()`

PR #2256, opened 2025-10-09
Avoid writing file data to audit log

### LINKED ISSUE #2255: Store filename instead of file content as the audit log event for FileCreated
When logging `FileCreated` events to the audit log, we also store the contents of the file in the database. This causes performance issues when generating documents and after that when viewing the audit log. 

Maybe this is also the cause of the speed issues mentioned in https://github.com/kiesraad/abacus/issues/2252

It would be sufficient to store the filename instead.

Image

praseodym:
Instead of the complete file, write only the file size to the audit log.

Fixes #2255

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -5643,19 +5643,11 @@
         "type": "object",
         "required": [
           "file_id",
-          "file_data",
           "file_name",
-          "file_mime_type"
+          "file_mime_type",
+          "file_size_bytes"
         ],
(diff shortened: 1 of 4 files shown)


oliver3:
Maybe we can now also show these Details in the audit log modal?

praseodym:
> Maybe we can now also show these Details in the audit log modal?

I think this is broken for many new audit log events. I've created #2257.

PR #2271, opened 2025-10-14
Bind sessions to user agent and client IP

marlonbaeten:
As the title says: this stores the user agent header and client IP address in the session table. When authenticating a request the client IP and user agent header must match.

To test:
- Login and verify you are logged in

Bonus points:
- Change your user agent header or IP and verify you cannot access you existing session.

Note that abacus-test is behind a reverse proxy - in the specific case of abacus-test the client IP will always be the same.

### DIFF
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -55,10 +55,24 @@ const config: PlaywrightTestConfig = defineConfig({
       use: { ...devices["Desktop Chrome"], channel: "chromium" },
     },
     {
-      name: "setup-test-users",
+      name: "setup-test-users-chrome",
       workers: 1,
       testMatch: /setup-test-users\.ts/,
       use: { ...devices["Desktop Chrome"], channel: "chromium" },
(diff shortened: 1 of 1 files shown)


praseodym:
I'd expect to see some tests for when the IP address and/or user agent changes

marlonbaeten:
> I'd expect to see some tests for when the IP address and/or user agent changes

See https://github.com/kiesraad/abacus/pull/2271/files#diff-8bce0e54608ad7b489041c51937b494397ae71406fa3f314d7b07432dd3a92a7R639

praseodym (approved):

cikzh (approved):
Went for the bonus points.

Tested in Firefox:

1. Logged in as admin, with default FF user agent
2. Went to about:config, added `general.useragent.override: "NetscapeNavigator"`
3. Noticed that any backend call results in login screen (received 401 from backend)
4. Verified via DevTools that the user agent was indeed overridden
5. Changed user agent back to default, which restores functionality, which is as intended

PR #2273, opened 2025-10-14
Update contributors.yml

ring-ring-ring:
add roheryn5 to the contributors list


contributors.yml
```diff
@@ -19,3 +19,4 @@
 - Hilbrand
 - lama-d3v
 - stacktraceghost
+- roheryn5
```
jschuurk-kr:
```suggestion
- Roheryn5
```

Issue #2187, opened 2025-09-24
Adding polling station to next committee session should open investigation


stacktraceghost:
Based on discussion https://github.com/kiesraad/abacus/pull/2184#issuecomment-3328335118

**Problem/scenario**
In committee session 2:
1. Add an investigation with corrected results to polling station A which already existed in committee session 1
2. Add a **new** polling station (B) which did **not** exist in committee session 1 and either [don't create an investigation with corrected results]/[create an investigation without corrected results]

Now only polling station A will be shown in the status overview of committee session 2. 

If you then complete the data entries for polling station A and finalize the committee session, polling station B will have no data entries. This results into errors later in the process (like creating documents).

**Proposed solution:**
- When adding a polling station in a next committee session, automatically add an investigation with `investigation.corrected_results=true`
- Deleting this investigation should result in the deletion of the polling station
- Adjusting `investigation.corrected_results` to anything other than `true` should not be possible

praseodym:
Resolved with a different solution in #2276

Issue #1812, opened 2025-07-21
Epic: coordinator read-only data entry view with delete option


praseodym:
Epic kick-off: 16 October 2025 13:00 @Lionqueen94 @stacktraceghost @praseodym @jschuurk-kr @lama-d3v @marjoleintamis @ring-ring-ring

# Context

# Architecture

# Scope

Not in this scope: when first and second data entry are filled in, only the second data entry will be shown to the coordinator, not both data entries

- [x] Double check whether the functionality `Volledig ingevoerde tellingen SB's bekijken` in [authorisation matrix](https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/autorisatiematrix.md) is still up to date for administrator and typist. And does this have to be added to this epic as well (administrator shouldn't be able to see/use the delete data entry button)?

- [Design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=1635-58883&t=apV5L65tdTg6KHp4-4)

- Change [delete polling station not possible text](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12640-36204&m=dev) and add link to readonly view

- [Use case "De coördinator GSB en de invoerders voeren alle gegevens in de applicatie in (vlieger)"](https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/gsb-invoer-eerste-zitting.md#de-co%C3%B6rdinator-gsb-en-de-invoerders-voeren-alle-gegevens-in-de-applicatie-in-vlieger)

PR #2276, opened 2025-10-15
Check if committee session can actually be finished

### LINKED ISSUE #2188: Check if committee session can actually be finished
Based on discussion https://github.com/kiesraad/abacus/pull/2184#issuecomment-3328335118

Currently it is possible to finish the data entry for a committee session when not all data entries have been finalized (through the API). 

There should be a backend check if all (relevant) polling stations of the current committee session have their data entries finalized. 
- For the first committee session all polling stations are relevent. 
- For the next committee session:
  - Polling stations with an investigation that has `corrected_results=true`
  - ⚠️ But also newly added polling stations without investigation or with `investigation.corrected_results=false`. Since they do not have any previous data entries

In the future, the last case should be caught by https://github.com/kiesraad/abacus/issues/2187, but the check should still be done.

stacktraceghost:
Resolves #2188

**Background**
It is allowed to add new polling stations in next committee sessions. New polling stations do not have results yet, but the system does not warn about that and allows data entry to be finished regardless. This causes problems down the line.

To solve this, this PR adds:
- A warning when all investigations are concluded/complete, however an investigation is missing for a new polling station
- Block the state transition to finished if not all polling stations have results when they should

Note https://github.com/kiesraad/abacus/issues/2187: in the future we will automatically create an investigation with `corrected_results=true` for a new polling station in a next committee session.  -- However, this might not be necessary anymore now we have the warning seen below.

**Changes**
image

- Added a warning to investigation overview page
- Added a redirect for `FinishDataEntryPage` to investigations overview page if investigations are missing
- Blocked state transition to `CommitteeSessionStatus::DataEntryFinished` if results are missing
- Blocked saving/generation of report when results are missing
- Disables corrected results choice for investigations of new polling stations and defaults it to `Ja`. Backend check is also included and tests for FE+BE.

To ensure that we have all results, I have adjusted the CTE (Common Table Expression) in the `data_entry::repository`. `list_results_for_committee_session` looks for the last results of a polling station by recursively going backwards through committee sessions. It did not keep investigations. with `corrected_results=true` and no results in mind. In that case, it would continue looking for older results. The query has been adjusted to stop looking backwards when there is an investigation with `corrected_results=true` and expects results then. 

Furthermore, `most_recent_results_for_polling_station` was the same query but based on one polling station. To lower maintenance cost/duplication, this function will now use `list_results_for_committee_session` and only search for the given `polling_station_id`. It will also throw an error if no result is found, since we always expect a result when calling this query. 

- `data_entry` repository
  - Renamed `list_entries_for_committee_session` -> `list_results_for_committee_session` since it is retrieving results
  - `list_results_for_committee_session` stops looking backwards for a polling station if results are expected due to an investigation with `corrected_results=true`. This is an extra safeguard so it will not return old results if results are missing on an investigation where results were expected (`corrected_results=true`). 
  - Added tests for `list_results_for_committee_session` (CTE query)
  - `most_recent_results_for_polling_station` now uses `list_entries_for_committee_session` to get results, since the query was the same
  - `most_recent_results_for_polling_station` will now throw an error if no results are found.

Misc:
- Added tests for the `useInvestigations` hook
- Updated `election_7_four_sessions` and removed `election_8_four_sessions_with_results`
  - `election_8` was the same as `election_7`, but with investigations, entries and results. Moved these to 7
  - Adjusted `proxy_certificate_count` in results to be able to differentiate between results in tests
  - Updated related tests
  
Also see [update](https://github.com/kiesraad/abacus/pull/2276#issuecomment-3412240860)

### DIFF
--- a/frontend/src/features/election_management/components/report/FinishDataEntryPage.tsx
+++ b/frontend/src/features/election_management/components/report/FinishDataEntryPage.tsx
@@ -25,13 +25,12 @@ export function FinishDataEntryPage() {
   const { update } = useCrud({ updatePath, throwAllErrors: true });
 
   // Check if all investigations are handled and none are missing
-  const investigationsComplete =
-    investigations.length > 0 &&
-    investigations.length === handledInvestigations.length &&
-    missingInvestigations.length === 0;
+  // Note: There are no investigations in the first committee session
+  const incompleteInvestigations =
+    investigations.length !== handledInvestigations.length || missingInvestigations.length > 0;
 
   useEffect(() => {
-    if (!investigationsComplete) {
+    if (incompleteInvestigations) {
       void navigate(`/elections/${election.id}/investigations`);
     }
 
@@ -39,7 +38,7 @@ export function FinishDataEntryPage() {
     if (currentCommitteeSession.status === "data_entry_finished") {
       void navigate(`/elections/${election.id}/report/committee-session/${currentCommitteeSession.id}/download`);
     }
-  }, [currentCommitteeSession, election, investigationsComplete, navigate]);
+  }, [currentCommitteeSession, election, incompleteInvestigations, navigate]);
 
   function handleFinish() {
     const body: COMMITTEE_SESSION_STATUS_CHANGE_REQUEST_BODY = { status: "data_entry_finished" };
--- a/frontend/src/features/election_management/components/report/FinishDataEntryPage.test.tsx
+++ b/frontend/src/features/election_management/components/report/FinishDataEntryPage.test.tsx
@@ -217,6 +217,16 @@ describe("FinishDataEntryPage", () => {
     });
   });
 
+  test("Do not redirect to investigations overview in first committee session", async () => {
+    overrideOnce("get", "/api/elections/1", 200, getElectionMockData());
+
+    await renderPage(1);
+
+    await waitFor(() => {
+      expect(navigate).not.toHaveBeenCalledWith("/elections/1/investigations");
+    });
+  });
+
   test("Redirect to report download page when committee session status is already data_entry_finished", async () => {
     overrideOnce("get", "/api/elections/1", 200, getElectionMockData({}, { status: "data_entry_finished" }));
 


stacktraceghost:
Back to draft, because I've found a case which bypasses the warning.

If you add an investigation for a new polling station, but set `corrected_results=false`, it will assume that no investigations is missing thus you could finish the data entry. Doing so, will then result in an error.

**Proposed solution**
Add a validation to the conclude investigation step: `IF ps.id_prev_session == NULL AND corrected_results = false` » show error, don't save.

jorisleker:
> Back to draft, because I've found a case which bypasses the warning.
> 
> If you add an investigation for a new polling station, but set `corrected_results=false`, it will assume that no investigations is missing thus you could finish the data entry. Doing so, will then result in an error.
> 
> **Proposed solution** Add a validation to the conclude investigation step: `IF ps.id_prev_session == NULL AND corrected_results = false` » show error, don't save.

I think we should prevent running into this altogether. When editing an investigation for a polling station added in the current session, we should default 'corrected_results' to 'Yes' and disable the 'No' option. See [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=15800-6190&t=lIzdIkXXmECLcIxR-11)

I know this approach differs from the way we handle entering data by _typists_ (there we force them to copy over any mistakes made in the paper reports), but this data entry is being made by a _coordinator_, someone we might expect to be able to correct the paper report in case an error was made and 'No' was checked there by mistake (or report the error to a member of the GSB and have them correct it). Besides that, we know for sure the report is made in the current committee session, and can still be corrected until the committee session is formally finished.

stacktraceghost:
> > Back to draft, because I've found a case which bypasses the warning.
> > If you add an investigation for a new polling station, but set `corrected_results=false`, it will assume that no investigations is missing thus you could finish the data entry. Doing so, will then result in an error.
> > **Proposed solution** Add a validation to the conclude investigation step: `IF ps.id_prev_session == NULL AND corrected_results = false` » show error, don't save.
> 
> I think we should prevent running into this altogether. When editing an investigation for a polling station added in the current session, we should default 'corrected_results' to 'Yes' and disable the 'No' option. See [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=15800-6190&t=lIzdIkXXmECLcIxR-11)
> 
> I know this approach differs from the way we handle entering data by _typists_ (there we force them to copy over any mistakes made in the paper reports), but this data entry is being made by a _coordinator_, someone we might expect to be able to correct the paper report in case an error was made and 'No' was checked there by mistake (or report the error to a member of the GSB and have them correct it). Besides that, we know for sure the report is made in the current committee session, and can still be corrected until the committee session is formally finished.

Fixed in https://github.com/kiesraad/abacus/pull/2276/commits/eefa3c7c5a7edb6774edf29fd138f51c1b7a674b

backend/src/data_entry/repository.rs
```diff
@@ -405,3 +393,365 @@ pub async fn insert_test_result(
...
+        assert_eq!(results[2].0.id, 741);
+        assert_eq!(
+            results[2].1.voters_counts().total_admitted_voters_count,
+            297
+        );
+    }
+
+    /// Test list_results_for_committee_session with 4th session, new polling station with investigation, corrected results=true and results don't exist
+    #[test(sqlx::test(fixtures(path = "../../fixtures", scripts("election_7_four_sessions"))))]
+    async fn test_list_results_next_new_polling_station_investigation_with_corrected_results_incomplete(
```
Lionqueen94: This test seems to test the same as the test on line 600

frontend/src/features/investigations/components/InvestigationsOverviewPage.tsx
```diff
@@ -54,26 +55,44 @@ export function InvestigationsOverviewPage() {
...
+                finishDataEntry();
+              }}
+              disabled={isLoading}
+            >
+              {t("election.title.finish_data_entry")}
+            </Button>
+          </Alert>
+        )
       ) : (
         <Messages />
```
Lionqueen94: Don't we want to show the messages if there is a missingInvestigations warning or the all investigations finished alert? It might still be nice to know your last change was saved? Although 2 alerts above one another is a little ugly.

frontend/src/utils/strings.test.ts
```diff
@@ -67,4 +67,15 @@ describe("Strings util", () => {
...
     }
   });
+
+  test.each([
+    [[], "and", ""],
+    [["A"], "and", "A"],
+    [["A", "B"], "and", "A and B"],
+    [["A", "B", "C"], "and", "A, B and C"],
+    [["A", "B", "C"], "or", "A, B or C"],
+    [["A ", " B", " C"], "or", "A ,  B or  C"],
```
Lionqueen94, line 77: don't we want to remove excess whitespace?

backend/tests/investigation_integration_test.rs
```diff
@@ -666,6 +667,84 @@ async fn test_can_only_update_current_session(pool: SqlitePool) {
...
+        })),
+    )
+    .await;
+
+    assert_eq!(conclude_response.status(), StatusCode::CONFLICT);
+    let conclude_body: Value = conclude_response.json().await.unwrap();
+    assert_eq!(
+        conclude_body["reference"],
+        "InvestigationRequiresCorrectedResults"
+    );
```
Lionqueen94: please rename `conclude_response` and `conclude_body` to `update_response` and `update_body` for more clarity

backend/src/data_entry/repository.rs
stacktraceghost: They are slightly different:
- Line 600: Investigation which is not concluded yet (`corrected_results=null`)
- Line 627: Investigation which has `corrected_results=true` and actually has results
- Line 684: Investigation which has `corrected_results=true` and results are missing (should never happen, but we're testing if this method would then fail)

frontend/src/features/investigations/components/InvestigationsOverviewPage.tsx
stacktraceghost: Good catch! I had staged this change locally already, waiting for a response if we'd like to show these one-time messages above or below the other alerts.

frontend/src/utils/strings.test.ts
stacktraceghost: I've considered this and thought it wasn't the responsibility of this function (hence the test). But maybe it is a good idea to do it anyways.

backend/tests/investigation_integration_test.rs
stacktraceghost: Good catch! That indeed has the wrong name, I'll adjust it. Same around line 691.

frontend/src/utils/strings.test.ts
Lionqueen94: Good point, I was thinking the same, it's also fine to leave it as is

backend/src/data_entry/repository.rs
Lionqueen94: But the test in line 600 still adds Some(false) and not None for the corrected_results, or am I reading it wrong?
Edit: this test name: `test_list_results_next_new_polling_station_investigation_no_corrected_results` so not with not concluded investigation

stacktraceghost: Ah you are right, I wrote the test with `corrected_results=false` instead of `None`. The result is the same, the reason why I did not test both `false/None` is because the recursion stops whenever it is `true` with results. So I simply tested `true` with+without results and a different answer.

stacktraceghost:
@Lionqueen94 pointed out that the CTE can get quite slow with bigger datasets. As an example, I generated an election with:
- 30 political groups
- 50 candidates per group
- 1000 polling stations

On abacus-test claiming takes 7s vs 93ms before. This is caused by the fact that we go through all the data instead of one polling station. To optimize this, I could pass a polling station ID to `list_results_for_committee_session` and filter on it or return to an almost identical query. However, I've looked a bit deeper and noticed that running the CTE directly on the sqlite database is almost instant, so the query itself is not slow. There seems to be quite some overhead, I'm looking into it. 

// Edit 1: Looks like the JSON deserialization is the culprit here.

backend/tests/investigation_integration_test.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2276/commits/b9f3117874c87f65898951c4874ac82edde5ccb7

frontend/src/features/investigations/components/InvestigationsOverviewPage.tsx
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2276/commits/b9f3117874c87f65898951c4874ac82edde5ccb7

stacktraceghost:
> @Lionqueen94 pointed out that the CTE can get quite slow with bigger datasets. As an example, I generated an election with:
> 
> * 30 political groups
> * 50 candidates per group
> * 1000 polling stations
> 
> On abacus-test claiming takes 7s vs 93ms before. This is caused by the fact that we go through all the data instead of one polling station. To optimize this, I could pass a polling station ID to `list_results_for_committee_session` and filter on it or return to an almost identical query. However, I've looked a bit deeper and noticed that running the CTE directly on the sqlite database is almost instant, so the query itself is not slow. There seems to be quite some overhead, I'm looking into it.
> 
> // Edit 1: Looks like the JSON deserialization is the culprit here.

Fixed in https://github.com/kiesraad/abacus/pull/2276/commits/b9f3117874c87f65898951c4874ac82edde5ccb7

The performance hit came from JSON deserialization, the CTE itself ran almost instantly. To minimize the impact of JSON deserialization, I've moved the CTE into function `fetch_results_for_committee_session`, which does the same as before except:
- It fetches the data as string instead of deserialized JSON
- Indexes polling stations beforehand to avoid unnecessary iterations when matching rows
- ~Not use a transaction (seemed unnecessary?)~ -- Restored in https://github.com/kiesraad/abacus/pull/2276/commits/30c6446e448bb61bb13c146d4b1c00c7f8ab366d

Then there are three pub functions:
- `are_results_complete_for_committee_session`: runs CTE and returns OK of no error occured, does not deserialize JSON
   - Used as check for state transition to `CommitteeSessionStatus::DataEntryFinished`
   - Used as check for `generate_and_save_files` in the report API
- `list_results_for_committee_session`: runs CTE and deserializes JSON (as this function did before)
   - Used to get data in the report API and test_data_gen
-  `most_recent_results_for_polling_station`: runs CTE with specified `polling_station_id` and deserializes that JSON
   - Used for data_entry claim
   - Used for investigation download corrigendum PDF
  
Also fixed the check in `committee_session/status.rs` when transitioning to `CommitteeSessionStatus::DataEntryFinished`. It will now call `are_results_complete_for_committee_session` correctly. This caused multiple tests to fail, because entries are now required before you can finish a session. I've adjusted these tests. Notes:
- One test cannot be fixed until #1812 is done, this has been marked as a todo (`polling_station_integration_test.rs`).
- Some tests in `data_entries_resolve_integration_test.rs` have been removed, since they were testing a state which you can't get to anymore (data entry errors/resolve/differences in a finished session).

praseodym:
> * Not use a transaction (seemed unnecessary?)

The transaction was there to prevent seeing changes written to the database in between the first and second query, so I think it's still good to have this transaction.

stacktraceghost:
> > * Not use a transaction (seemed unnecessary?)
> 
> The transaction was there to prevent seeing changes written to the database in between the first and second query, so I think it's still good to have this transaction.

Restored in https://github.com/kiesraad/abacus/pull/2276/commits/30c6446e448bb61bb13c146d4b1c00c7f8ab366d

backend/src/data_entry/repository.rs
```diff
@@ -275,116 +280,137 @@ pub async fn result_exists(conn: &mut SqliteConnection, id: u32) -> Result<bool,
...
-            WHERE sc.data IS NULL
+            JOIN polling_stations AS ps ON ps.id = sc.id_prev_session
+            LEFT JOIN polling_station_results AS r ON r.polling_station_id = ps.id
+            LEFT JOIN polling_station_investigations AS psi ON psi.polling_station_id = ps.id AND psi.corrected_results = 1
+            WHERE sc.data IS NULL AND sc.investigation IS NULL
         )
         SELECT
             sc.original_id AS "original_id!: u32",
-            sc.data AS "data: Json<PollingStationResults>"
+            sc.data AS "data: String"
```
praseodym: I'm surprised that transferring all this data doesn't incur a noticeable performance hit.

Also, instead of `String`, we could consider using [`serde_json::value::RawValue`](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html).

stacktraceghost:
@rnijveld Pointed out that deserialization in dev-mode is a lot slower than production. After testing, it turned out that deserializing all the results in production mode took 300-500ms. 

He also proposed some further adjustments, which have been discussed in the stand-up today and are implemented now:
1. `are_results_complete_for_committee_session` doesn't run the CTE anymore (which would retrieve all results), but has two queries to check if the given committee session is complete:
   a. If first session: `Ok(all_new_ps_have_data)`  -- All polling stations have results
   b. If next session: `Ok(all_new_ps_have_data && all_investigations_finished)`. -- All polling stations have results and investigations are all concluded and have results if `investigation.corrected_results=true`
2. CTE deserializes results immediately like before
  a. `list_results_for_committee_session` returns all results for given committee session
  b. `most_recent_results_for_polling_station` returns recent results for polling station (without retrieving all results of other polling stations)
  
Note:
- `crate::committee_session::repository` already contained a function `all_investigations_finished`. It was used on the same places where we call `are_results_complete_for_committee_session`, so it has been merged into that function. Tests have been moved and extended.

backend/src/data_entry/repository.rs
stacktraceghost: Due to the choice to adjust the functions ([see comment below](https://github.com/kiesraad/abacus/pull/2276#issuecomment-3421968109)), we deserialize the results directly again.

backend/src/data_entry/api.rs
```diff
@@ -268,8 +267,8 @@ async fn polling_station_data_entry_claim(
     let (polling_station, election, committee_session, state) =
         validate_and_get_data(&mut tx, polling_station_id, &user.0).await?;
 
-    let previous_results = if let Some(id) = polling_station.id_prev_session {
-        most_recent_results_for_polling_station(&mut tx, id).await?
+    let previous_results = if polling_station.id_prev_session.is_some() {
+        Some(most_recent_results_for_polling_station(&mut tx, polling_station_id).await?)
```
oliver3: If you claim data entry that already is in progress, won't you get the current results instead of the previous results?

Edit: I see that the function does in fact retrieve the polling station and uses its `id_prev_session`. Maybe the name could be improved to reflect that, or maybe revert to using the `polling_station.id_prev_session` if possible to be less confusing?

backend/src/investigation/api.rs
```diff
@@ -166,6 +166,16 @@ async fn polling_station_investigation_conclude(
 
     let committee_session = validate_and_get_committee_session(&mut tx, polling_station_id).await?;
 
+    let polling_station =
+        crate::polling_station::repository::get(&mut tx, polling_station_id).await?;
+    if polling_station.id_prev_session.is_none() && !polling_station_investigation.corrected_results
+    {
+        return Err(APIError::Conflict(
+            "Investigation requires corrected results, because it is not part of a previous session".into(),
```
oliver3: This text confused me a little, it should error because it has no corrected results and the polling station is one that has been added for this session right?

(same in the check below)

frontend/src/hooks/election/useInvestigation.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    ];
+
+    overrideOnce("get", "/api/elections/1", 200, getElectionMockData({}, { number: 2 }, investigations));
+    overrideOnce("get", "/api/elections/1/status", 200, { statuses: [] });
+
+    const result = renderUseInvestigations();
+    await waitFor(() => {
+      expect(result.current.investigations).toHaveLength(3);
+
+      expect(result.current.currentInvestigations).toHaveLength(2);
```
oliver3: A bit superfluous to check the length if you also check the (mapped) contents, but 🤷

frontend/src/utils/committeeSession.ts
```diff
@@ -1,16 +1,11 @@
 import { t } from "@/i18n/translate";
 
+const isTranslatableSessionNumber = (num: number): num is 1 | 2 | 3 | 4 | 5 => num >= 1 && num <= 5;
```
oliver3: You can also use `hasTranslation()` to let this code be independent on which numbers are translated in the bundle:

```ts
export function committeeSessionLabel(sessionNumber: number, addArticle = false): string {
  const numberTranslation = `committee_session_status.number.${sessionNumber}`;
  if (hasTranslation(numberTranslation)) {
    return `${addArticle ? `${t("committee_session_status.the")} ` : ""}${t(numberTranslation)} ${t(`committee_session_status.session`).toLowerCase()}`;
  } else {
    return `${t("committee_session_status.session")} ${sessionNumber}`;
  }
}
```

backend/tests/data_entries_resolve_integration_test.rs
```diff
@@ -159,28 +159,6 @@ async fn test_polling_station_data_entry_get_errors(pool: SqlitePool) {
     assert_eq!(res.status(), StatusCode::OK);
 }
 
-#[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_2", "users"))))]
```
oliver3, line 162: Doubting if these integration tests should be removed or that they still have some added value..

frontend/src/features/election_management/components/report/FinishDataEntryPage.tsx
```diff
@@ -18,16 +19,22 @@ import cls from "../ElectionManagement.module.css";
 
 export function FinishDataEntryPage() {
   const { currentCommitteeSession, election, refetch } = useElection();
+  const { missingInvestigations } = useInvestigations();
   const navigate = useNavigate();
   const updatePath: COMMITTEE_SESSION_STATUS_CHANGE_REQUEST_PATH = `/api/committee_sessions/${currentCommitteeSession.id}/status`;
   const { update } = useCrud({ updatePath, throwAllErrors: true });
 
   useEffect(() => {
+    // Redirect to investigations overview if there are missing investigations
```
oliver3: A bit inconsistent perhaps dat you do check and redirect for missing investigations, but not for ongoing investigations

frontend/src/utils/committeeSession.ts
stacktraceghost: Nice! I wasn't aware yet of that typeguard, thanks!

backend/src/investigation/api.rs
praseodym: I interpreted the text as saying exactly that, but if it's confusing we can try a different wording

praseodym (commented):
lgtm if @oliver3's comments are addressed

backend/src/data_entry/api.rs
stacktraceghost: Background info:
In the original version, `most_recent_results_for_polling_station` had its own CTE and looked based on given `polling_station_id`. Now the function uses the single CTE which requires a `committee_session_id` and optional `polling_station_id`. 

So I'd need to find the `committee_session_id` related to the given `polling_station_id`, however `polling_station::repository::get` does not allow to retrieve old polling stations. This is the reason why I am passing the current `polling_station_id`, retrieving that polling station and then the `polling_station.id_prev_session` + previous committee session through `committee_session::repository::get_previous_session`

-- 

So I think renaming would be a good choice here, I think `previous_results_for_polling_station` would be a more suitable name.

backend/src/investigation/api.rs
stacktraceghost: Ah I get why this can be confusing, `it` should refer to the related polling station. Would this be better?
`Investigation requires corrected results, because the polling station is not part of a previous session`

backend/tests/data_entries_resolve_integration_test.rs
stacktraceghost: The reason that I've removed them is because we are unable to transition to `CommitteeSessionStatus::DataEntryFinished` when results are incomplete. So I think we cannot get into that state anymore. If we still want to test it, we will have to create a fixture which would already be in a faulty state.

What do you think?

backend/src/data_entry/api.rs
stacktraceghost: Renamed in https://github.com/kiesraad/abacus/pull/2276/commits/97c04fe430b20cda279b38db907ac06a98d326f7

backend/src/investigation/api.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2276/commits/b0c5c31280a4fe8f6f8b5d64457d131015906cc2

frontend/src/utils/committeeSession.ts
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2276/commits/b0c5c31280a4fe8f6f8b5d64457d131015906cc2

frontend/src/hooks/election/useInvestigation.test.tsx
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2276/commits/b0c5c31280a4fe8f6f8b5d64457d131015906cc2

frontend/src/features/election_management/components/report/FinishDataEntryPage.tsx
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2276/commits/fbf86518a2720d1b017aa99e01960cae1aef7921
Edit: and in https://github.com/kiesraad/abacus/pull/2276/commits/b5fd2608a33dd234f870a420578fa37d1706f045, noticed that I did not keep the first committee session (no investigations) into account here. Added a test for it. 

Bonus bugfix in `useInvestigation`, I noticed that the filter for:
- `currentInvestigations` did not check if `corrected_results === undefined`
- `handledInvestigations` checked on `!inv.corrected_results` (which would be truthy for both `undefined` (no answer) and `false`), adjusted this to `inv.corrected_results === false`

Adjusted the tests to prevent this from happening again

backend/src/data_entry/api.rs
oliver3: Thanks for the background! And the "previous" in the new name makes it imho much clearer 👍

backend/src/investigation/api.rs
oliver3: Ah! Yes that was what apparently I could not parse indeed :)

backend/tests/data_entries_resolve_integration_test.rs
oliver3: I did not realise that the "given" is now a faulty state, thanks for your answer (and I don't think fixtures that are faulty are a good idea).

oliver3 (approved):

PR #2280, opened 2025-10-15
Typst templates: render candidate names in 7pt + truncate long names

### LINKED ISSUE #2080: Rework PDF/typst for Na-14-2 and Na 14-2 Bijlage 1 votes per candidate tables
As discussed in PR #2010 and a short meeting: to unblock progress that PR was merged, but is not yet at the standard we require for the layout of these PDFs. Specifically:

- Names of candidates very quickly result in overflows where names end up on three lines of text, meaning they end up overflowing into the borders of the cells and into the next cells
- The candidate number and the originals columns are too similar, which could cause confusion
- The size of the originals column has to be quite small, resulting in it not being suitable for larger numbers, especially beyond the individual polling station level.

Additional notable changes from a content/design perspective (that have been merged in #2010):
- The title for the section that contains the votes per candidate tables has been moved to its own page to make sure the first political group has just as much space as any subsequent group.
- The description text for filling in these tables is no longer repeated above every table, but only displayed once.

A possible solution would be to switch to a single column layout, that would give more space for all of these things.

marlonbaeten:
Fixes #2080

- Added a truncate function
- Full candidate names are now constructed as a string
- Candidate names are now smaller (7pt) and truncated in vote count tables
- Added some example long names in the Typst input `model-na-31-2-bijlage1`

### DIFF
--- a/frontend/src/features/investigations/components/AddInvestigationLayout.test.tsx
+++ b/frontend/src/features/investigations/components/AddInvestigationLayout.test.tsx
@@ -39,6 +39,9 @@ async function renderPage(section: "reason" | "print-corrigendum" | "findings") 
   await router.navigate(`/elections/1/investigations/1/${section}`);
   rtlRender(<Providers router={router} />);
 
+  // Ensure rendering is complete
+  await screen.findByRole("banner");
+
   return router;
 }
 


stacktraceghost:
[Flaky test](https://github.com/kiesraad/abacus/actions/runs/18527986857/job/52803259726?pr=2280). I will add a commit to fix it.

stacktraceghost (approved):
Smart way of finding the truncation spot in `truncate`, took me a moment to understand.

praseodym (approved):

marlonbaeten:
> Smart way of finding the truncation spot in `truncate`, took me a moment to understand.

I was happy to finally find a legit reason to implement a classic search algorithm :-)

Issue #2245, opened 2025-10-08
Handle DataEntryNotAllowed during data entry


oliver3:
The frontend incorrectly shows an `AbortModal` when an investigation changes from corrected_results Yes to No while doing data entry. It seems that navigation to an error page is blocked by the modal.

We should unblock going to the error page and check if that is feedback enough to the user for this situation.

<img width="590" height="662" alt="image" src="https://github.com/user-attachments/assets/3b7b2d14-37a9-4720-abc3-eefbd03fc139" />

_Originally posted by @oliver3 in https://github.com/kiesraad/abacus/issues/2210#issuecomment-3360070966_

PR #2281, opened 2025-10-15
Fix data entry state error handling and use messages hook

oliver3:
- Fixes https://github.com/kiesraad/abacus/issues/2245 by not blocking navigation when these errors occur, listed in `errors.ts` to define them in one location
- Display appropriate warning alert using `useMessages()` hook and `<Messages />` component
  - Add `MessagesProvider` to `ElectionLayout`
  - Change `DataEntryProvider` prop to PollingStation to get the polling station number for the message
  - Update tests and usages accordingly

To test
- As a coordinator (in an incognito tab), create an investigation for a corrigendum session, with corrected results set to yes
- As a typist, start data entry for that polling station and do not complete it
- As a coordinator (in an incognito tab), remove the investigation or set corrected results to no, and remove the data entry
- As the typist, try to continue data entry, and notice that you've been sent to the data entry home page with an appropriate warning message

### DIFF
--- a/backend/src/data_entry/api.rs
+++ b/backend/src/data_entry/api.rs
@@ -110,12 +110,14 @@ async fn validate_and_get_data(
 
     // Validate polling station
     if committee_session.is_next_session() {
-        let investigation = get_polling_station_investigation(conn, polling_station.id).await?;
-        if investigation.corrected_results != Some(true) {
-            return Err(APIError::Conflict(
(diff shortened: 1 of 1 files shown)


stacktraceghost (changes requested):
Followed the test steps. If I remove the investigation while I am doing the data entry, and then press continue, I am met with a 404 error.

PR #2283, opened 2025-10-15
remove domain id from election homepage heading

jschuurk-kr:
fixes issue found in https://github.com/kiesraad/abacus/issues/2126

### Scope

Remove the domain id in the heading of the election home page:

### Old
image

### New
image

### DIFF
--- a/frontend/src/features/election_management/components/ElectionHomePage.test.tsx
+++ b/frontend/src/features/election_management/components/ElectionHomePage.test.tsx
@@ -37,9 +37,7 @@ const renderPage = async (userRole: Role) => {
     </TestUserProvider>,
   );
   expect(await screen.findByRole("heading", { level: 1, name: "Gemeenteraadsverkiezingen 2026" })).toBeVisible();
-  expect(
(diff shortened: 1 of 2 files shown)


oliver3:
Should this number still be at the elections page? 
image

oliver3 (approved):

jschuurk-kr:
> This is implemented as designed in Figma, if this needs to be changed we should change the Figma design first

@jorisleker Can you update the designs to no longer show the domain id on the election home page? Or is it ok if I make that change?

jschuurk-kr:
> Should this number still be at the elections page?

We only show it to admins and coordinators on that page, not to typists. I think it makes sense to keep it there, until we get a significant amount of feedback that people want it removed.

On the election home page the domain id was in the heading and in the table. The feedback we got was that domain id is not important enough to have in the heading, hence the removal.

jorisleker:
> @jorisleker Can you update the designs to no longer show the domain id on the election home page? Or is it ok if I make that change?

Agreed that this number does not add value for typist, so okay to remove it there!
@jschuurk-kr feel free to update Figma

oliver3:
I believe I've changed all the instances in Figma

PR #2285, opened 2025-10-15
consistent field order in user creation forms

jschuurk-kr:
### Scope
- Consistent ordering of fullname and username in forms
- Use "Kies een gebruikersnaam" on the first admin form
- Use "gebruikersnaam" instead of "invoer" in the field hints 
- Hint text update: "Je kan straks via gebruikersbeheer accounts aanmaken voor andere beheerders."

fixes issue in found https://github.com/kiesraad/abacus/issues/2126

image

image

image

image

### DIFF
--- a/frontend/src/i18n/locales/nl/initialise.json
+++ b/frontend/src/i18n/locales/nl/initialise.json
@@ -2,11 +2,10 @@
   "account_details": "Details van het account",
   "admin_configured": "Account van beheerder ingesteld",
   "create_admin_account": "Account voor beheerder aanmaken",
-  "fullname_hint": "Bijvoorbeeld Karel van Tellingen. Accounts in Abacus zijn persoonsgebonden.",
+  "fullname_hint": "Bijvoorbeeld Karel van Tellingen. Accounts in Abacus zijn persoonsgebonden. Je kan straks via gebruikersbeheer accounts aanmaken voor andere beheerders.",
   "installation_success": "Installatie gelukt",
   "login_to_admin_account": "Inloggen met account van beheerder",
   "login_to_finish_setup": "<p>Abacus is nog niet klaar voor gebruik. Log in met het account van de beheerder om de configuratie af te ronden. Ben je de inloggegevens kwijt?</p><p><link>Stel het account van de beheerder opnieuw in.</link></p>",
-  "other_accounts": "Je kan straks via gebruikersbeheer accounts aanmaken voor andere beheerders.",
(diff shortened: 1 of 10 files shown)


jorisleker:
@jschuurk-kr All for consistency, and thanks for spotting this. 

I would prefer to fix the field ordering the other way around; so put full name on top for creating other users as well. This way the login-credentials are nicely grouped in all cases.

frontend/src/i18n/locales/nl/initialise.json
```diff
@@ -2,10 +2,11 @@
   "account_details": "Details van het account",
   "admin_configured": "Account van beheerder ingesteld",
   "create_admin_account": "Account voor beheerder aanmaken",
-  "fullname_hint": "Bijvoorbeeld Karel van Tellingen. Accounts in Abacus zijn persoonsgebonden. Je kan via gebruikersbeheer accounts aanmaken voor meer dan één beheerder.",
+  "fullname_hint": "Bijvoorbeeld Karel van Tellingen. Accounts in Abacus zijn persoonsgebonden.",
   "installation_success": "Installatie gelukt",
   "login_to_admin_account": "Inloggen met account van beheerder",
   "login_to_finish_setup": "<p>Abacus is nog niet klaar voor gebruik. Log in met het account van de beheerder om de configuratie af te ronden. Ben je de inloggegevens kwijt?</p><p><link>Stel het account van de beheerder opnieuw in.</link></p>",
+  "other_accounts": "Je kan straks via gebruikersbeheer accounts aanmaken voor overige beheerders.",
```
jorisleker:
```suggestion
  "other_accounts": "Je kan straks via gebruikersbeheer accounts aanmaken voor andere beheerders.",
```

jorisleker (changes requested):
small copy suggestion, and prefer to fix the field ordering inconsistency in user management instead of here

jschuurk-kr:
> small copy suggestion, and prefer to fix the field ordering inconsistency in user management instead of here

@jorisleker Can you please update the design to what the end result of this PR should be?

jschuurk-kr:
> I would prefer to fix the field ordering the other way around; so put full name on top for creating other users as well. This way the login-credentials are nicely grouped in all cases.

@jorisleker Should this also be done for the screen where you edit users?
image

Because that would put the editable full name above the non-editable identifier username.

jorisleker:
@jschuurk-kr Updated the user management screens in Figma

Issue #2096, opened 2025-09-03
Validation spanning multiple sections, show on one section (F.403)


stacktraceghost:
Validation rule F.403 (see [validations](https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/validatieregels-plausibiliteitschecks-tellingen.md#cso--f403-totaal-aantal-stemmen-op-een-lijst-komt-niet-overeen-met-het-lijsttotaal-van-corresponderende-ex)) should highlight:
- Section 'Political group list x' -> 'Total'
- Section 'Voters and votes' -> 'E.x'

However when doing so, F.403 gets shown on both sections instead of only at the section of the political group. Therefore, currently we only highlight the total field on the section of the political group. 

To meet the highlight requirement, we need to mark both fields but show the error only on the political group section.

praseodym:
I'm not sure if this needs fixing in this epic (or at all), I think the solution implemented in #2095 is already clear enough.

praseodym:
Discussed in the standup: won't fix, if we get user feedback on this we can revisit it.

PR #2291, opened 2025-10-16
Fixed validation error for political group votes

### LINKED ISSUE #2279: Validation error not shown when counting political group total votes

lama-d3v:
- Fixes #2279 
- Update documentation to reflect the implementation choice as mentioned in closed issue https://github.com/kiesraad/abacus/issues/2096

### DIFF
--- a/backend/src/data_entry/validation.rs
+++ b/backend/src/data_entry/validation.rs
@@ -588,6 +588,8 @@ impl Validate for CommonPollingStationResults {
         )?;
 
         for (i, pgcv) in self.political_group_votes.iter().enumerate() {
+            let pgcv_path = path.field("political_group_votes").index(i);
+
             let pgtv = self
                 .votes_counts
                 .political_group_total_votes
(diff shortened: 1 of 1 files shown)


backend/src/data_entry/validation.rs
```diff
@@ -649,7 +655,7 @@ impl Validate for CommonPollingStationResults {
 
 impl Validate for ExtraInvestigation {
     fn validate(
-        &self,
+        &self,``````````````````````````
```
Lionqueen94:
```suggestion
        &self,
```

backend/src/data_entry/validation.rs
```diff
@@ -638,7 +630,21 @@ impl Validate for CommonPollingStationResults {
 
             if has_error_f401 {
                 validation_results.errors.retain(|vr| {
-                    ![ValidationResultCode::F402, ValidationResultCode::F403].contains(&vr.code)
+                    if vr.code == ValidationResultCode::F402
```
stacktraceghost: I believe this logic will not filter correctly. Correct me if I'm wrong, since I'm just thinking through the code. Example:

Political group number 1 has
- F401
- F402

Political group number 2 has:
- F403

In the current logic, we are looping over political_group_votes (pgcv). In every iteration the following happens:
1. `has_error_f401` will be `true` if there is any F401 error, not specifically for that group.
2. Then it will filter out F402/F403 if the `context.political_group_number === pgcv.number`, which will match all F402/F403s in that group. 

Resulting in group 1 having F401 and group 2 in no errors.

Proposed solution:
```rs
let has_error_f401 = validation_results.errors.iter().any(|x| {
    x.code == ValidationResultCode::F401
        && x.context.as_ref().and_then(|c| c.political_group_number)
            == Some(pgcv.number)
});

if has_error_f401 {
    validation_results.errors.retain(|vr| {
        ![ValidationResultCode::F402, ValidationResultCode::F403].contains(&vr.code)
            || vr.context.as_ref().and_then(|c| c.political_group_number)
                != Some(pgcv.number)
    });
}
```

1. Check if F401 occurs for the current political group number
2. If so, filter out F402/F403 for this political group number
  a. If code is not F402/F403, keep the error
  b. If code is F402/F403, keep the error if the political group number doesn't match

Would be great to add a test for this situation, so we can catch it in the future.

backend/src/data_entry/validation.rs
lama-d3v: Fixed this

backend/src/data_entry/validation.rs
lama-d3v: I added a test for multiple errors in different political groups. Also added filter by political group for `has_error_f401`. I think this covers everything for now.

praseodym (changes requested):
I think we can simplify the validation logic to something like (pseudocode):

```
if evaluate_F401 {
  add_F401
} else {
  if evaluate_F402 {
    add_F402
  }
  if evaluate_F403 {
    add_F403
  }
}
```
This would require moving where F.402 is evaluated, but I think that makes more sense than the current implementation.

PR #2313, opened 2025-10-17
Updated some logging actions copy

### LINKED ISSUE #2312: Improve some logging actions
- The logging action of updating a user sounds a bit odd: "Gebruikersaccount opgeslagen". Suggestion to change it to "Gebruikersaccount bijgewerkt". -> Question: do we actually want to log UserAccountUpdated and UserUpdated separately or just use one audit event for this?
- Suggestion: shorten the airgap action copy: "Geen verbinding met het internet meer gedetecteerd." and "Verbinding met het internet gedetecteerd. Abacus gaat op slot." are significantly longer than all other actions
- In Abacus we have the action "Invoer geclaimd", it might be a bit easier to understand for users to change the copy to: "Invoer gestart" -> Question: Any thoughts on this?
- We have actions "Onderzoek afgerond" and "Invoer afgerond", do we also want to add "Zitting afgerond"? -> We cannot, since we don't know when this is and this is not in Abacus.

Lionqueen94:
Fixed soms points found in https://github.com/kiesraad/abacus/issues/2225

Closes #2312 

- [x] "Geen verbinding met het internet meer gedetecteerd." and "Verbinding met het internet gedetecteerd. Abacus gaat op slot." are significantly longer than all other actions. Shortened "verbinding met het internet" to "internetverbinding".
- [x] Changed "Gebruikersaccount opgeslagen" to "Eigen account bijgewerkt" and changed "Gebruiker aangemaakt" to "Gebruikersaccount aangemaakt", "Gebruiker bijgewerkt" to "Gebruikersaccount bijgewerkt" and "Gebruiker verwijderd" to "Gebruikersaccount verwijderd".
- [x] In Abacus we have the action "Invoer geclaimd", separated this action into "Invoer gestart" and "Invoer hervat"
- [x] Made sure data entry progress is logged as a percentage -> empty `audit_log` table or reset the whole database to make this work

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -4985,9 +4985,7 @@
             "minimum": 0
           },
           "data_entry_progress": {
-            "type": "integer",
-            "format": "int32",
-            "minimum": 0
+            "type": "string"
           },
           "data_entry_status": {
             "type": "string"
--- a/backend/src/audit_log/audit_event.rs
+++ b/backend/src/audit_log/audit_event.rs
@@ -121,7 +121,7 @@ pub struct DataEntryDetails {
     pub polling_station_id: u32,
     pub committee_session_id: u32,
     pub data_entry_status: String,
-    pub data_entry_progress: u8,
+    pub data_entry_progress: String,
     #[schema(value_type = Option<String>)]
     pub finished_at: Option<DateTime<Utc>>,
     pub first_entry_user_id: Option<u32>,
--- a/backend/src/data_entry/structs.rs
+++ b/backend/src/data_entry/structs.rs
@@ -32,7 +32,7 @@ impl From<PollingStationDataEntry> for DataEntryDetails {
             polling_station_id: value.polling_station_id,
             committee_session_id: value.committee_session_id,
             data_entry_status: state.status_name().to_string(),
-            data_entry_progress: state.get_progress(),
+            data_entry_progress: format!("{}%", state.get_progress()),
             finished_at: state.finished_at().cloned(),
             first_entry_user_id: state.get_first_entry_user_id(),
             second_entry_user_id: state.get_second_entry_user_id(),
--- a/frontend/src/types/generated/openapi.ts
+++ b/frontend/src/types/generated/openapi.ts
@@ -559,7 +559,7 @@ export interface DataEntry {
 
 export interface DataEntryDetails {
   committee_session_id: number;
-  data_entry_progress: number;
+  data_entry_progress: string;
   data_entry_status: string;
   finished_at?: string | null;
   first_entry_user_id?: number | null;


jorisleker:
- "internetverbinding" is fine!
- we don't have an explicit action to close a committee session, do we? Closing a session is implicit, as it only happens when a new session is added. As it is implicit, not sure we should log this because the event will always coincide with opening a next session.

Lionqueen94:
> * "internetverbinding" is fine!
> 
>     * we don't have an explicit action to close a committee session, do we? Closing a session is implicit, as it only happens when a new session is added. As it is implicit, not sure we should log this because the event will always coincide with opening a next session.

Yes we do, we change the committee session status to "DataEntryFinished" when clicking "Invoerfase afronden" on this page: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=193-38500&m=dev But I guess that the session is not fully done then, since they still have to sign the documents of course

jorisleker:
> Yes we do, we change the committee session status to "DataEntryFinished" when clicking "Invoerfase afronden" on this page: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=193-38500&m=dev But I guess that the session is not fully done then, since they still have to sign the documents of course

"Invoerfase" is just one of the states of the committee session. It would make sense to log this event, but then let's make it explicit and log something like "Invoerfase afgerond". 

Do we log the other state transitions as well?

Lionqueen94:
> > Yes we do, we change the committee session status to "DataEntryFinished" when clicking "Invoerfase afronden" on this page: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=193-38500&m=dev But I guess that the session is not fully done then, since they still have to sign the documents of course
> 
> "Invoerfase" is just one of the states of the committee session. It would make sense to log this event, but then let's make it explicit and log something like "Invoerfase afgerond".
> 
> Do we log the other state transitions as well?

No we do not log any of the state transitions, that might be a nice one to add. "Invoerfase gestart", "Invoerfase gepauzeerd" en "Invoerfase afgerond"?

Do you have any thoughts on this @jorisleker :
In Abacus we have the action "Invoer geclaimd", it might be a bit easier to understand for users to change the copy to: "Invoer gestart"

jorisleker:
> No we do not log any of the state transitions, that might be a nice one to add. "Invoerfase gestart", "Invoerfase gepauzeerd" en "Invoerfase afgerond"?

Great idea! 
Should we add the committee session number? "Invoerfase 2e zitting gestart"? Or is that clear from the context?

> Do you have any thoughts on this @jorisleker : In Abacus we have the action "Invoer geclaimd", it might be a bit easier to understand for users to change the copy to: "Invoer gestart"

Is the claim endpoint that triggers this log message only used by the typist when claiming a data entry? In that case I am okay with this copy change.

Lionqueen94:
> > No we do not log any of the state transitions, that might be a nice one to add. "Invoerfase gestart", "Invoerfase gepauzeerd" en "Invoerfase afgerond"?
> 
> Great idea! Should we add the committee session number? "Invoerfase 2e zitting gestart"? Or is that clear from the context?
> 

The log popup will contain the session information, putting it in the title is not possible, because we can't add custom variables to the copy of the log action title.

> > Do you have any thoughts on this @jorisleker : In Abacus we have the action "Invoer geclaimd", it might be a bit easier to understand for users to change the copy to: "Invoer gestart"
> 
> Is the claim endpoint that triggers this log message only used by the typist when claiming a data entry? In that case I am okay with this copy change.

Yes this endpoint is only used by the typist. They claim the data entry, they can save it (this gets logged also) and then claim it again to continue. If I change it the logging will always reflect "Invoer gestart", is this fine, even if it's not the initial start of the data entry?

jorisleker:
> Yes this endpoint is only used by the typist. They claim the data entry, they can save it (this gets logged also) and then claim it again to continue. If I change it the logging will always reflect "Invoer gestart", is this fine, even if it's not the initial start of the data entry?

What does 'claim it again to continue' mean? In which situation does that happen?
If a user pauses data entry, the data entry remains assigned to them, right?

jorisleker:
Discussed on Mattermost:
first claim: "Invoer gestart"
later claims: "Invoer hervat"

stacktraceghost (approved):

PR #2314, opened 2025-10-17
Fixed bug that polling station with a deleted data entry gave an error on deletion

Lionqueen94:
So the problem that was fixed is that a data entry still exists when you click "niet opslaan" or discard the data entries as a coordinator. The remaining data entry has the status "FirstEntryNotStarted". Since there is no ON CASCADE for the deletion of the data entry of a polling station, it then gives an error and you see an error in the frontend.

We have 2 options on how to fix this:
1. Check if a data entry exists and if so and it has the state "FirstEntryNotStarted", delete the data entry and then the polling station
2. When a data entry is deleted through not saving, delete the data entry fully instead of changing the state to "FirstEntryNotStarted". Same when a coordinator deletes both data entries in the solve differences flow or deletes the data entry in the resolve errors flow.

After discussing with @jschuurk-kr and @praseodym I implemented option 2.
Added new tests for resolve errors and deleting a second data entry.
Also removed `serde_json` imports and converted all uses of `Value` and `json` to full paths for more clarity.


backend/tests/polling_station_integration_test.rs
```diff
@@ -55,7 +55,7 @@ async fn import_polling_stations(
     addr: &SocketAddr,
     election_id: u32,
     file_name: &str,
-    polling_stations: serde_json::Value,
+    polling_stations: Value,
```
praseodym: Please don't shorten imports like this, `serde_json::Value` is descriptive but only `Value` is not

Lionqueen94: I corrected this across the application in d317fc75d04aac53daeb556868eb38de827a2780

praseodym: I see you also changed imports of the `json!` macro, but I'd say that one is descriptive by itself.

Lionqueen94: Shall I re-add that as an import or leave it as is now?

praseodym: It's ok to leave it as is now

backend/src/data_entry/api.rs
```diff
@@ -1652,12 +1699,79 @@ mod tests {
         .await;
         assert_eq!(response.status(), StatusCode::OK);
 
+        // Check that the data entry is created
+        let row_count = query!("SELECT COUNT(*) AS count FROM polling_station_data_entries")
```
stacktraceghost: You could use `data_entry_exists` here and below

backend/src/data_entry/api.rs
```diff
@@ -456,15 +456,20 @@ async fn polling_station_data_entry_delete(
...
-        polling_station_id,
-        committee_session.id,
-        &new_state,
-    )
-    .await?;
 
-    let data_entry = get_data_entry(&mut tx, polling_station_id, committee_session.id).await?;
+    let mut data_entry = get_data_entry(&mut tx, polling_station_id, committee_session.id).await?;
+
+    if new_state == DataEntryStatus::FirstEntryNotStarted {
```
stacktraceghost, line 474: Maybe we can add a short comment to this statement and the ones below to explain what we are doing here.

backend/src/data_entry/api.rs
```diff
@@ -1652,12 +1699,79 @@ mod tests {
...
+            pool.clone(),
+            request_body.clone(),
+            1,
+            EntryNumber::SecondEntry,
+        )
+        .await;
+        assert_eq!(response.status(), StatusCode::OK);
+
+        // Check that the data entry is in SecondEntryInProgress state
+        let row = query!("SELECT state AS 'state: sqlx::types::Json<DataEntryStatus>' FROM polling_station_data_entries")
```
stacktraceghost: `get_data_entry` could be used here and below

backend/src/data_entry/api.rs
Lionqueen94: Replaced all instances with `data_entry_exists` in 63185fd949ae144e2418e3b96647734167bf7710

backend/src/data_entry/api.rs
Lionqueen94: Replaced all instances with `get_data_entry` in 63185fd949ae144e2418e3b96647734167bf7710

backend/src/data_entry/api.rs
Lionqueen94: Done in 63185fd949ae144e2418e3b96647734167bf7710

stacktraceghost:
I will check the merge conflicts, since they are caused by merging https://github.com/kiesraad/abacus/pull/2276

praseodym (approved):

Issue #2248, opened 2025-10-08
Generate P2a if there are no new results, generate P2a and Na14-2 + EML if there are new results


ring-ring-ring:
Found during testing; if there is an investigation with findings and no new results, we only generate the P2a. If there are new results, we generate the P2a, 14-2 + EML. 

Usecases: 

Main flow: https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/gsb-tweede-zitting.md#hoofdscenario-en-uitbreidingen-4

Generate only P2a when there are no new results was hidden on the next page: https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/input-output-bestanden.md#na-14-2-dso-en-cso-corrigendum-op-gsb-niveau-voor-volgende-zitting

PR #2319, opened 2025-10-20
Do not generate EML and results for next sessions with no corrections

rnijveld:
Fixes https://github.com/kiesraad/abacus/issues/2248

### DIFF
--- a/backend/src/report/api.rs
+++ b/backend/src/report/api.rs
@@ -485,7 +485,7 @@ async fn election_download_pdf_results(
         generate_and_save_files(&pool, audit_service, committee_session_id).await?;
 
     let pdf_file = pdf_file.ok_or(APIError::BadRequest(
-        "PDF results is not generated".to_string(),
+        "PDF results are not generated".to_string(),
         ErrorReference::PdfGenerationError,
     ))?;
 


bramtweedegolf:
When there are new results after an investigations, generate all new files:

Screenshot_2025-10-20_13-47-40

But when there are no new results, generate only a new P2a file:

Screenshot_2025-10-20_13-50-28

backend/fixtures/election_9_four_sessions_without_corrections.sql
```diff
@@ -0,0 +1,83 @@
+INSERT INTO elections (id, name, counting_method, election_id, location, domain_id, category, number_of_seats, election_date, nomination_date, political_groups)
```
Lionqueen94: Please rename this file for `election_8_four_sessions_without_corrections` and adjust all ids to 8, 801...804, 811...842, 821, 832 so it's consistent with the id of the election and you can load multiple fixtures at once without conflicts. Note: The current election 8 is being removed in an open PR, so there won't be 2 election fixtures with id 8.

stacktraceghost: Would it be possible to use `election_8_four_sessions_with_results` (which [will soon be](https://github.com/kiesraad/abacus/pull/2276) `election_7_four_sessions`) and adjust `investigations.corrected_results` to `false` in your test? Seems excessive to copy the whole fixture for one test.

Lionqueen94: that would be even better indeed

bramtweedegolf: Agreed, that is better; thanks! Changed in https://github.com/kiesraad/abacus/pull/2319/commits/92a9a22ae25d3b3e07583d7b36a945a0d3efa12b

backend/src/report/api.rs
```diff
@@ -458,6 +484,11 @@ async fn election_download_pdf_results(
     let (_, pdf_file, _) =
         generate_and_save_files(&pool, audit_service, committee_session_id).await?;
 
+    let pdf_file = pdf_file.ok_or(APIError::BadRequest(
+        "PDF results is not generated".to_string(),
```
Lionqueen94:
```suggestion
        "PDF results are not generated".to_string(),
```

Lionqueen94 (approved):

stacktraceghost (approved):

PR #2320, opened 2025-10-20
Better airgap instruction, 30s check interval

marlonbaeten:
The current airgap error page instructs the user to disable their internet connection and refresh the page. This could cause confusion since the airgap status is only checked every 60 seconds, so it could take up to 60 seconds before abacus becomes responsive again:

Screenshot From 2025-10-20 10-11-51

This PR changes the check status from 60 seconds to 30 seconds (which is not a problem since we perform the checks concurrently now and have a 10 second timeout).
It adds the line "Wacht 30 seconden" to instruct the user to wait before refreshing the page and access abacus again:

Screenshot From 2025-10-20 10-15-20

### DIFF
--- a/backend/src/airgap/detect.rs
+++ b/backend/src/airgap/detect.rs
@@ -37,7 +37,7 @@ const DOMAINS: [&str; 3] = [
     "surfnet.nl",
 ];
 
-pub const AIRGAP_DETECTION_INTERVAL: u64 = 60; // interval in seconds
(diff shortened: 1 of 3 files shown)


Lionqueen94:
@marlonbaeten Make sure to also update the design

marlonbaeten:
> @marlonbaeten Make sure to also update the design

If someone can give me edit permissions :grin:

oliver3:
Documentation done in https://github.com/kiesraad/abacus/pull/2330, also found while performing ET https://github.com/kiesraad/abacus/issues/2224

PR #2321, opened 2025-10-20
Airgap detection for release builds

marlonbaeten:
Fixes the (pre-)release build.

This PR adds the airgap detection feature for our release builds and splits the frontend / backend release build.

The release frontend build is now done in a more minimal environment: only production dependencies are installed + the frontend is only build on one platform (ubuntu-latest). Splitting the frontend / backend build has some security advantages and makes inspecting the frontend build assets easier.

Overview:

- Fix issue of missing cross-env dependency in de release job
- Enable airgap detection in release builds
- Non-airgap variants are build for all platforms in order to run the playwright tests and for abacus-test.nl
- Separate backend/frontend builds for releases
- Only install npm production dependencies for release build
- Log environment for frontend production build
- The user-agent is now configured in playwright, instead of keeping separate user setup tasks (needed since we bind the user-agent to a session) this fixes running all browser tests at once.

To test: [see the release build](https://github.com/kiesraad/abacus/actions/runs/18676603950) Note that only the last step cannot complete since: Branch "enable-airgap-detection-in-release-builds" is not allowed to deploy to Release due to environment protection rules.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -10,7 +10,6 @@
     "dev": "npm run check-server && vite",
     "dev:msw": "cross-env API_MODE=mock vite",
     "build": "vite build --config vite-prod.config.ts",
-    "build:prod": "vite build --config vite-prod.config.ts",
     "build:msw": "cross-env API_MODE=mock vite build && cp mockServiceWorker.js dist/",
     "lint": "tsc && cross-env ESLINT_ENV=production eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "prettier": "prettier --ignore-unknown --write .",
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -10,7 +10,7 @@
     "dev": "npm run check-server && vite",
     "dev:msw": "cross-env API_MODE=mock vite",
     "build": "vite build --config vite-prod.config.ts",
-    "build:prod": "cross-env SHOW_DEV_PAGE=false vite build --config vite-prod.config.ts",
+    "build:prod": "vite build --config vite-prod.config.ts",
```
praseodym: We don't need this command if we're not setting the `SHOW_DEV_PAGE=false` environment variable

marlonbaeten: I will remove it.

marlonbaeten: See [0ff7dc0](https://github.com/kiesraad/abacus/pull/2321/commits/0ff7dc096feaf5efb8f900498a54b4141102933e)

cikzh (approved):

PR #2329, opened 2025-10-21
Added hash to footer of Na 14-2, improve page headers

### LINKED ISSUE #2323: Model Na 14-2 missing EML_NL hash
The Model Na 14-2 PDF is missing the EML_NL hash in the footer, it should have one just like Model Na 31-2.

marlonbaeten:
Fixes #2323

Added hash to footer of Na 14-2.
Improved settings page header/footer handling.

Previously, we made multiple calls to `#show: doc => conf` to override the current page header when it differed from the current chapter title. However, this also caused the footer and other page options to reset.

This PR removes those calls and explicitly applies the chapter overrides (in current_chapter_title). On the blank page, the page header is cleared and reset on the following page, instead of overriding the full page configuration.

### DIFF
--- a/backend/templates/model-p-2a.typ
+++ b/backend/templates/model-p-2a.typ
@@ -1,4 +1,4 @@
-#import "common/style.typ": conf, document_numbering
+#import "common/style.typ": conf, document_numbering, default_header
 #import "common/scripts.typ": *
 #let input = json("inputs/model-p-2a.json")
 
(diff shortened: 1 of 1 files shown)


cikzh (approved):

backend/templates/model-na-31-2-inlegvel.typ
```diff
@@ -10,7 +10,7 @@
 #let location_name = is_municipality[Gemeente #input.election.domain_id #input.election.location][Openbaar lichaam #input.election.location]
 #let location_type = is_municipality[gemeentelijk stembureau][stembureau voor het openbaar lichaam]
 
-#show: doc => conf(doc, header-right: location_name, footer: [
+#show: doc => conf(doc, headerright: location_name, footer: [
```
Lionqueen94:
```suggestion
#show: doc => conf(doc, header-right: location_name, footer: [
```

Issue #2307, opened 2025-10-17
Session timeout on different pages and different environments


jschuurk-kr:
Explore the session timeout
on different pages (auto-refresh versus not) and environments (local versus abacus-test.nl)
to find any undesired behavior

See below for the findings and discussion that inspired this charter.

---

If you leave the election status page open for long enough, you get the "Abacus is stuk"-page ("NetworkError when attempting to fetch resource., GET /api/elections/6") . @oliver3 and I have seen this once before this week.

<img width="1925" height="903" alt="Image" src="https://github.com/user-attachments/assets/cc858e15-af67-4307-b47f-4244c9c39bbe" />

Tested on abacus-test.nl, https://lid-datum-briefstembewijs.abacus-test.nl, commit 9adc3f9d.

I presume the cause is a combination of the auto-refresh of the page and the session expiring. This should result in a redirect to the login page, but that does not happen.

Lionqueen94:
Also check the layout of the error page, because that footer is way too big

praseodym:
I'd expect to see a HTTP 401 Unauthorized response code after the session expired, but your screenshot is showing a `NetworkError`. Are you sure this does not have some other cause?

jschuurk-kr:
> I'd expect to see a HTTP 401 Unauthorized response code after the session expired, but your screenshot is showing a `NetworkError`. Are you sure this does not have some other cause?

Could be something specific to abacus-test.nl.

I tried to reproduce on a locally running Abacus with memory serve and get a very different result there. The GETs from the auto-refresh cause the session to be extended, so the user remains logged in and on the election status page. Which leads to the question if that's the behavior we want.

jschuurk-kr:
Changed this issue from a bug to an exploratory testing charter, because we need to look into different behavior on different pages (auto-refresh versus not) and different behaviors (local running versus less important abacus-test.nl).

jschuurk-kr:
fix for refetches extending the session: https://github.com/kiesraad/abacus/pull/2332

jschuurk-kr:
Looks like the `NS_BINDING_ABORTED` is not the real issue. The response before that with a size of 0B is. Because if you log in as typist and go to `/elections`, the same problem occurs, with the "Abacus is stuk"-page saying "can't access property "committee_sessions", getElections.data is undefined".

jschuurk-kr:
~~possible fix for the network issue on abacus-test: https://github.com/kiesraad/abacus/pull/2347~~

[...]

PR #2332, opened 2025-10-21
Do not extend the session on refetch request

marlonbaeten:
On pages where we periodically refetch data the user is never logged uit, since the session will be extended indefinitely.

The PR disables extending the session on a refetch by adding a `X-Do-Not-Extend-Session` header to these requests.

found in #2307

### DIFF
--- a/backend/src/authentication/mod.rs
+++ b/backend/src/authentication/mod.rs
@@ -31,7 +31,7 @@ pub const SESSION_COOKIE_NAME: &str = "ABACUS_SESSION";
 /// Only send cookies over a secure (https) connection
 pub const SECURE_COOKIES: bool = false;
 
-/// Do not extend session header, only its existence of this header is checked, not the value
+/// Do not extend session header, only its existence is checked, not the value
 pub const DO_NOT_EXTEND_SESSION_HEADER: &str = "x-do-not-extend-session";
 
 /// Struct used to create a new user


frontend/src/api/ApiClient.ts
```diff
@@ -8,6 +8,7 @@ const HEADER_ACCEPT = "Accept";
 const HEADER_CONTENT_TYPE = "Content-Type";
 
 export const DEFAULT_CANCEL_REASON = "Component unmounted";
+export const DO_NOT_EXTEND_SESSION = { "X-Do-Not-Extend-Session": "true" };
```
jschuurk-kr: If I change the value to "false", the session does not extend. (Arguably no one will ever do that, though.)

backend/src/authentication/middleware.rs
```diff
@@ -66,11 +67,14 @@ pub async fn extend_session(
         return response;
     };
 
+    // check for the do not extend session header
+    let do_not_extend = headers.get(super::DO_NOT_EXTEND_SESSION_HEADER).is_some();
```
oliver3, line 71: Re: "If I change the value to "false", the session does not extend" from @jschuurk-kr 

Something like this seems to work but not sure if this is the most idiomatic code.

```rust
// check for the do not extend session header
let do_not_extend = headers.get(super::DO_NOT_EXTEND_SESSION_HEADER)
    == HeaderValue::from_str("true").ok().as_ref();
```

marlonbaeten: I think checking the existence of the header is more robust than verifying the actual value (what if super smart HTTP proxy converts the casing or something).
I can, for clarity, change the "true" to something else, like the reason "do-not-extend-on-data-refetch".

jschuurk-kr: If it's too hard to agree on a code solution, I'd be fine with solving this through code comments.

backend:  
`// check for the do not extend session header` -> `// check for the existence of the do not extend session header`

frontend:  
add a comment where we define the header in ‎frontend/src/api/ApiClient.ts that the backend only checks for existence, not the value

marlonbaeten: See [9ea643d](https://github.com/kiesraad/abacus/pull/2332/commits/9ea643d814b6d29ede7d3f72034a3b81e3f63dfc)

backend/src/authentication/mod.rs
```diff
@@ -31,6 +31,9 @@ pub const SESSION_COOKIE_NAME: &str = "ABACUS_SESSION";
 /// Only send cookies over a secure (https) connection
 pub const SECURE_COOKIES: bool = false;
 
+/// Do not extend session header, only its existence of this header is checked, not the value
```
stacktraceghost: Small suggestion:
```suggestion
/// Do not extend session header, only the existence of this header is checked, not the value
```
or
```suggestion
/// Do not extend session header, only its existence is checked, not the value
```

stacktraceghost (approved):

jschuurk-kr (approved):

oliver3 (approved):

PR #2339, opened 2025-10-23
Added data entries delete endpoint for coordinator

### LINKED ISSUE #2295: Backend: add delete data entry endpoint for coordinator
The coordinator should be able to delete a data entry, when:
- while a typist is filling in the data entry (status = first or second data entry is in progress)
- the first data entry is done, second data entry has not begun yet
- First and second data entry are finished, and there are no differences between first and second data entry (this would trigger the differences page).

When a committee session is finished, the coordinator must be able to view and delete a data entry. If the data entry is deleted by the coordinator, the status of the committee session has to change from `finished` to `in progress`.

This issue depends on #2290. This needs to be fixed first.

Lionqueen94:
Closes #2295

## Description
- Added endpoint for coordinator to delete the data entry and optional result linked to a polling station
- Added tests
- Removed committee session status change in delete polling station endpoint and moved this to the new endpoint, in accordance with the flowchart
- Removed commented out test for this status change

### DIFF
--- a/frontend/src/i18n/locales/nl/error.json
+++ b/frontend/src/i18n/locales/nl/error.json
@@ -11,7 +11,7 @@
     "DatabaseError": "Er is een fout opgetreden bij het opslaan van de invoer",
     "DataEntryAlreadyClaimed": "Een andere invoerder is bezig met dit stembureau",
     "DataEntryAlreadyFinalised": "De invoer voor dit stembureau is al gedaan",
-    "DataEntryCannotBeDeleted": "De invoer can niet verwijderd worden",
+    "DataEntryCannotBeDeleted": "De invoer kan niet verwijderd worden",
     "DataEntryNotAllowed": "De invoer voor dit stembureau is niet toegestaan",
     "EmlImportError": "Het EML bestand is niet geldig",
     "EntryNotFound": "Niet gevonden",


backend/src/data_entry/api.rs
```diff
@@ -617,6 +622,81 @@ pub struct DataEntryGetErrorsResponse {
...
+    audit_service: AuditService,
+) -> Result<StatusCode, APIError> {
+    let mut tx = pool.begin_immediate().await?;
+
+    let polling_station =
+        crate::polling_station::repository::get(&mut tx, polling_station_id).await?;
+    let committee_session =
+        crate::committee_session::repository::get(&mut tx, polling_station.committee_session_id)
+            .await?;
+    crate::election::repository::get(&mut tx, committee_session.election_id).await?;
```
stacktraceghost: Is this necessary? If the election doesn't exist, the `committee_session` (and underlying `polling_station`'s) will neither.

backend/src/data_entry/api.rs
```diff
@@ -617,6 +622,81 @@ pub struct DataEntryGetErrorsResponse {
...
+    let data_entry = get_data_entry(&mut tx, polling_station_id, committee_session.id).await?;
+
+    if data_entry.state.status_name() == DataEntryStatusName::FirstEntryHasErrors
+        || data_entry.state.status_name() == DataEntryStatusName::EntriesDifferent
+    {
+        tx.rollback().await?;
+
+        Err(APIError::Conflict(
+            "Data entry cannot be deleted.".to_string(),
+            ErrorReference::InvalidDataEntryStatus,
```
stacktraceghost: The status itself is not necessarily invalid, just for the delete action or the data entry itself, depending on how you read it. Maybe we can name this differently?

backend/src/data_entry/api.rs
Lionqueen94: No it's not indeed, removed in 1e327fc395f67c1ed619b8f53af19e176771867f

backend/src/data_entry/api.rs
Lionqueen94: True, changed to DataEntryCannotBeDeleted in 1e327fc395f67c1ed619b8f53af19e176771867f

backend/tests/polling_station_integration_test.rs
```diff
@@ -620,11 +620,3 @@ async fn test_finished_to_in_progress_on_update(pool: SqlitePool) {
     })
     .await;
 }
-
-// Not possible until we can delete polling stations with results
```
praseodym, line 624: Can't we enable this test now?

Lionqueen94: Well it's not needed, I test the committee session status change in the delete data entries tests. And the delete polling station does not call the delete data entries. The coordinator has to delete the data entries/result first before being allowed to delete the polling station. So this test won't pass, because the data entries need to be deleted first.

stacktraceghost: Following the [committee session state diagram](https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/committee-session-state.md), it actually does not define that we should go back to in progress when deleting a (non-last) polling station. However, I can imagine new documents need to be generated?

Lionqueen94: Generating new documents is a good point, will that be needed? :thinking:

praseodym: When results are deleted, the existing documents are no longer valid

praseodym: `change_committee_session_status` should take care of deleting the documents, right?

Lionqueen94: Yes and the documents get deleted in the change status function, so no need to specifically check anything for the polling station deletion

Issue #2186, opened 2025-09-24
Check CommitteeSessionStatus state transitions with all necessary preconditions


oliver3:
Check based on https://github.com/kiesraad/abacus/pull/2276

Code review if all the state transitions check for the necessary preconditions

stacktraceghost:
Found an issue: it is possible to change the status of a previous committee session.

Not state transitions, but it's also possible to do `committee_session_update` and `committee_session_number_of_voters_change` on a previous committee session.

^^ these issues are fixed in https://github.com/kiesraad/abacus/pull/2337

Lionqueen94:
There is one status change which is incorrectly implemented, which is DataEntryFinished to Created. It is fixed in https://github.com/kiesraad/abacus/pull/2340

PR #2340, opened 2025-10-24
Fix committee session status change inconsistency

Lionqueen94:
Fix finding found in https://github.com/kiesraad/abacus/issues/2186

### DIFF
--- a/backend/tests/investigation_integration_test.rs
+++ b/backend/tests/investigation_integration_test.rs
@@ -925,3 +925,49 @@ async fn test_finished_to_in_progress_on_delete_non_last(pool: SqlitePool) {
     let addr = serve_api(pool).await;
     check_finished_to_in_progress_on(&addr, true, || delete_investigation(&addr, 741)).await;
 }
+
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_7_four_sessions", "users"))))]
+async fn test_finished_to_created_on_delete_last(pool: SqlitePool) {
+    let addr = serve_api(pool).await;
+    let cookie = shared::coordinator_login(&addr).await;
+    let election_id = 7;
+    let committee_session_id = 704;
+
+    assert_eq!(
+        shared::create_investigation(&addr, 741).await.status(),
+        StatusCode::CREATED
+    );
+
+    assert_eq!(
+        conclude_investigation(
+            &addr,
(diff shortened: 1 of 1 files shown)


backend/src/committee_session/status.rs
```diff
@@ -137,10 +137,8 @@ impl CommitteeSessionStatus {
...
             CommitteeSessionStatus::DataEntryNotStarted
             | CommitteeSessionStatus::DataEntryInProgress
-            | CommitteeSessionStatus::DataEntryPaused => Ok(CommitteeSessionStatus::Created),
-            CommitteeSessionStatus::DataEntryFinished => {
-                Err(CommitteeSessionError::InvalidStatusTransition)
-            }
+            | CommitteeSessionStatus::DataEntryPaused
+            | CommitteeSessionStatus::DataEntryFinished => Ok(CommitteeSessionStatus::Created),
         }
     }
```
praseodym, line 143: ~~From what I can tell `prepare_data_entry` is only ever called when the existing status is `CommitteeSessionStatus::Created`:~~

https://github.com/kiesraad/abacus/blob/1771f092b0a7d6e665ac33af7849c7d0feda4e65/backend/src/committee_session/status.rs#L60

~~So I'm not sure what effect this change has, can you explain the change a bit more?~~

Edit: the `change_committee_session_status` function is very confusing but I guess it works

Lionqueen94: Correct, but since the flowchart states that we should be able to go back from `DataEntryFinished` to `Created`, it should also be reflected here I thought

praseodym: No it works differently from how I thought it did: the `status` in the `match status` is the argument given to `change_committee_session_status`, and the committee session could be any state at that point.

Lionqueen94: Yes true, it's a bit confusing indeed

stacktraceghost: Should we add a test for this in `investigation_integration_test.rs`? I know it is one of the many state change combinations, but in this case an action would've failed (deleting last investigation), reproducible on main.

Lionqueen94: Added a test for this in a066edbe3f79186914d69e406f08be75f68d0ece

PR #2347, opened 2025-10-27
Handle aborted and empty requests

marlonbaeten:
If a (re-)fetch fails due to the request being aborted by the client (browser) there were scenario's our code previously could not differentiate between a generic NetworkError, like a broken network connection, and the case where the client aborts / cancels the request. 

This PR introduces a `AbortedError` that is created if a request was aborted by the client. This error is never assigned to the request state in the `handleApiResult` method, a thereby ignored in data re-fetches for instance. This should be ignored since request are aborted when components unmount or when a browser decides to abort a request for any reason.

This PR also improves the status codes our API sends back on success:
- 204 If the request was successful but there is no response body
- 201 if an entity was created
- 200 on a succesfull response with a body

These response codes are more semantic, but also enables us to differentiate certain failed requests that are empty but do have a 200 response code. This happens for instance when FireFox trottles background requests. Our code returns a AbortedError when an empty 200 response is returned without any content-type=json header.

The ApiClient is split up in a generic client and  an Abacus specific one in order to keep the code understandable and the unit size low.

### DIFF
--- a/frontend/src/api/useInitialApiGet.ts
+++ b/frontend/src/api/useInitialApiGet.ts
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useState } from "react";
 
-import { DEFAULT_CANCEL_REASON, DO_NOT_EXTEND_SESSION } from "./ApiClient";
+import { DEFAULT_CANCEL_REASON } from "./ApiClient";
 import {
   ApiRequestState,
   ApiRequestStateWithoutFatalErrors,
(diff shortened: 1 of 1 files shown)


jschuurk-kr:
Unfortunately this change did not fix it:
image

Tested on https://aanduiding-kieskring-stemming.abacus-test.nl (instance based on the branch). Commit no longer exists, because of an amend by @marlonbaeten.

jschuurk-kr:
Issue still reproducible on commit [cd67c8bf](https://github.com/kiesraad/abacus/commit/cd67c8bf9eb29ad8780c7d8aa2ce35be21455d03) (latest) [20a4f06](https://github.com/kiesraad/abacus/commit/20a4f06) (build)

image

marlonbaeten:
@jschuurk-kr it seems there are 2 issues here:
- The `NS_BINDING_ABORTED` error in firefox when a request is aborted from the client.
- The 0 byte successful response

Your screenshot also shows these as independent requests.

It seems that the empty response is more a browser bug - the fetch should throw an error if the request fails, but I can add some handling to this PR to verify that the response body in not-empty. You will get an error page in that case however.

marlonbaeten:
@jschuurk-kr can you verify that the "empty" response does not contain any response mime-type header?

marlonbaeten:
This PR got too big. I will split this PR up in logical parts.

PR #2348, opened 2025-10-27
Update column totals styling in Na 31-1 and Na 31-2

rnijveld:
Now looks like this:

image


marlonbaeten (approved):
We might have to think of a way to reduce the number of parameters for the votes_table, maybe split it up into multiple functions that use the same underlying methods?
For now - LGTM

PR #2353, opened 2025-10-27
Only perform auto refresh if the tab is active/visible

marlonbaeten:
If abacus is running on a hidden / inactive tab it keeps refreshing certain pages in the background. Sometimes browsers throttle or cancel these requests to save memory / bandwidth.

This PR adds some logic to:
- Disable periodic updates if the tab is inactive or hidden
- Re-fetch the data as soon as the tab becomes active again

To prevent duplication of this mechanism the logic is moved to a separate hook `useLiveData` that handles the AbortController, interval and event listener.

### DIFF
--- a/frontend/src/hooks/useLiveData.ts
+++ b/frontend/src/hooks/useLiveData.ts
@@ -8,7 +8,7 @@ export function useLiveData(refetchFunctions: Array<(abortController: AbortContr
     const abortController = new AbortController();
 
     const refetch = () => {
-      // only refetch if the document is hidden (i.e., user is on another tab)
+      // don't refetch if the document is hidden (i.e., user is on another tab)
       if (!document.hidden) {
         refetchFunctions.forEach((refetchFunction) => {
           void refetchFunction(abortController);


frontend/src/hooks/useLiveData.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { DEFAULT_CANCEL_REASON } from "@/api/ApiClient";
+
+// Hook to refetch live data at regular intervals and on tab visibility change
+export function useLiveData(refetchFunctions: Array<(abortController: AbortController) => Promise<unknown>>) {
+  useEffect(() => {
+    const abortController = new AbortController();
+
+    const refetch = () => {
+      // only refetch if the document is hidden (i.e., user is on another tab)
```
Lionqueen94:
```suggestion
      // don't refetch if the document is hidden (i.e., user is on another tab)
```

jschuurk-kr:
@marlonbaeten Good thing you also have that other PR!

image

PR #2354, opened 2025-10-28
Added AbortedError, split ApiClient

marlonbaeten:
This PR introduces an `AbortedError` that is used to indicate a request was aborted by the client. This makes this scenario more explicit. This also splits up the ApiClient into a generic `BaseApiClient` and a more Abacus specific one `ApiClient`.

The AbortedError makes it possible to better handle throttled or empty requests in a next PR.

### DIFF
--- a/backend/src/committee_session/status.rs
+++ b/backend/src/committee_session/status.rs
@@ -32,7 +32,9 @@ use crate::{
 #[serde(rename_all = "snake_case")]
 #[strum(serialize_all = "snake_case")]
 #[sqlx(rename_all = "snake_case")]
+#[derive(Default)]
 pub enum CommitteeSessionStatus {
+    #[default]
     Created,
     DataEntryNotStarted,
     DataEntryInProgress,
@@ -222,12 +224,6 @@ impl CommitteeSessionStatus {
             }
             CommitteeSessionStatus::DataEntryFinished => Ok(self),
         }
-    }
-}
-
-impl Default for CommitteeSessionStatus {
-    fn default() -> Self {
-        Self::Created
     }
 }
 
--- a/backend/src/data_entry/status.rs
+++ b/backend/src/data_entry/status.rs
@@ -40,7 +40,9 @@ impl From<ValidationResults> for DataEntryTransitionError {
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, ToSchema)]
 #[serde(deny_unknown_fields, tag = "status", content = "state")]
+#[derive(Default)]
 pub enum DataEntryStatus {
+    #[default]
     FirstEntryNotStarted, // First entry has not started yet
     FirstEntryInProgress(FirstEntryInProgress),
     FirstEntryHasErrors(FirstEntryHasErrors),
@@ -684,12 +686,6 @@ impl Display for DataEntryTransitionError {
             }
             DataEntryTransitionError::ValidationError(_) => write!(f, "Validation errors"),
         }
-    }
-}
-
-impl Default for DataEntryStatus {
-    fn default() -> Self {
-        Self::FirstEntryNotStarted
     }
 }
 


frontend/src/api/BaseApiClient.ts
```diff
@@ -0,0 +1,198 @@
...
+      }
+
+      return apiResult;
+    } catch (e: unknown) {
+      if (abortController.signal.aborted) {
+        return new AbortedError();
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+      const message = (e as Error).message || "Network error";
```
oliver3:
```suggestion
      const message = e instanceof Error ? e.message : "Network error";
```

frontend/src/api/BaseApiClient.ts
```diff
@@ -0,0 +1,198 @@
...
+        ApiResponseStatus.ServerError,
+        response.status,
+        `Unexpected response status: ${response.status}`,
+        "InvalidData",
+      );
+    } catch (e) {
+      console.error("Error parsing response", e);
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+      return new NetworkError((e as Error).message || "Network error");
```
oliver3:
```suggestion
      return new NetworkError(e instanceof Error ? e.message : "Network error");
```

frontend/src/api/BaseApiClient.ts
```diff
@@ -0,0 +1,198 @@
...
+
+      const isJson = response.headers.get(HEADER_CONTENT_TYPE) === MIME_JSON;
+
+      // handle the response, and return an error when state there is a non-2xx status
+      const apiResult: ApiResult<T> = isJson
+        ? await this.handleJsonBody<T>(response)
+        : await this.handleEmptyOrNonJsonBody(response);
+
+      // dispatch error events to subscribers
+      if (apiResult instanceof ApiError || apiResult instanceof FatalApiError) {
```
oliver3, line 181: At the moment, only `ApiError` is interesting and actually expected in the `subscribeToApiErrors()`. Should the `FatalApiError` be removed here and in the `ApiErrorEvent` class?

frontend/src/api/ApiResult.ts
```diff
@@ -71,6 +78,7 @@ export class NotFoundError extends FatalError {
   }
 }
 
+export type AnyError = ApiError | FatalApiError | NetworkError | NotFoundError | AbortedError;
```
oliver3, line 81: Why is AbortedError not added to the existing AnyApiError?

praseodym (approved):

frontend/src/api/ApiResult.ts
marlonbaeten: The `useCrud` / `useInitialApiGet` hooks still only expose AnyApiError as requestState, so we need to differentiate between these to types since the AbortedError is "ignored" in these hooks.

frontend/src/api/BaseApiClient.ts
marlonbaeten: Fixed in [14b8b1c](https://github.com/kiesraad/abacus/pull/2354/commits/14b8b1cd40ef346157611c2cbe86bba3b3d5a585)

frontend/src/api/BaseApiClient.ts
marlonbaeten: Fixed in [14b8b1c](https://github.com/kiesraad/abacus/pull/2354/commits/14b8b1cd40ef346157611c2cbe86bba3b3d5a585)

frontend/src/api/BaseApiClient.ts
marlonbaeten: One of the use-cases is the AirgapViolationError, which is a FatalApiError. So the signature of `subscribeToApiErrors` was not correct, fixed in https://github.com/kiesraad/abacus/pull/2354/commits/14b8b1cd40ef346157611c2cbe86bba3b3d5a585

oliver3 (approved):

frontend/src/api/ApiResult.ts
oliver3: I believe that TypeScript cannot make the distinction between `NetworkError` and `AbortedError` because it has the same structure. The following seems to compile just fine. This PR is still an improvement on its own, so good for now!

```typescript
export function returnsAbortedError(): AnyApiError {
  return new AbortedError();
}
```

PR #2361, opened 2025-10-29
Ignore invalid empty responses, fix requestState handling

marlonbaeten:
This checks for unexpected empty 200 responses and converts them to an AbortedError. We observed these when using FireFox and Abacus is being throttled / suspended on a hidden tab. In this case Abacus should not show a full page error but ignore the responsive of the aborted request.

Final change (together with #2354, #2353, #2355) to prevent Abacus from showing a full page error on throttled / aborted requests: #2307

This PR also corrects how requestState is handled in the election overview page and the dev page.

### DIFF
--- a/frontend/src/api/BaseApiClient.ts
+++ b/frontend/src/api/BaseApiClient.ts
@@ -21,8 +21,10 @@ function isErrorResponse(object: unknown): object is ErrorResponse {
   );
 }
 
+type Method = "GET" | "POST" | "PUT" | "DELETE";
+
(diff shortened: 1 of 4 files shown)


stacktraceghost (approved):

marlonbaeten:
Converted to draft because I discovered an edge case I also want to fix...

marlonbaeten:
Okay - I will split the PR....

stacktraceghost (approved):

PR #2366, opened 2025-10-30
Add fallback handler for unmatched `/api` routes

praseodym:
Add explicit 404 fallback handler for unmatched `/api` routes, to prevent the frontend `index.html` from being served on these paths by the `memory-serve` fallback handler.

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -1,12 +1,13 @@
 use std::{error::Error, future::Future, net::SocketAddr, str::FromStr};
 
 use airgap::AirgapDetection;
-#[cfg(feature = "memory-serve")]
-use axum::http::StatusCode;
 use axum::{
     Router,
     extract::{DefaultBodyLimit, FromRef},
+    http::StatusCode,
     middleware,
+    response::IntoResponse,
+    routing::any,
     serve::ListenerExt,
 };
 use hyper::http::{HeaderName, HeaderValue, header};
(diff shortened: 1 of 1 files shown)


rnijveld (approved):

backend/src/lib.rs
```diff
@@ -98,14 +99,24 @@ pub fn openapi_router() -> OpenApiRouter<AppState> {
...
 
     // Serve without OpenAPI documentation if the openapi feature is disabled
     #[cfg(not(feature = "openapi"))]
-    Router::from(router)
+    let router = Router::from(router);
+
+    // Add fallback for unmatched /api/* routes to return 404
+    // This must come before the frontend memory-serve routes
+    async fn api_fallback_handler() -> impl IntoResponse {
+        (StatusCode::NOT_FOUND, "Not Found")
```
marlonbaeten, line 114:
```suggestion
        APIError::NotFound("API endpoint not found".to_string(), error::ErrorReference::EntryNotFound)
```

marlonbaeten (changes requested):
Would be nice to return an error from the API that our frontend can interpret.

backend/src/lib.rs
praseodym: This error will only be returned when some programming/deployment mistake was made, e.g. when calling an API endpoint that is behind a feature that is not enabled. I don't think the frontend needs to be able to catch that.

marlonbaeten (approved):

PR #2371, opened 2025-10-31
Update to Typst 0.14

### LINKED ISSUE #2346: Update to Typst 0.14
https://typst.app/blog/2025/typst-0.14/

praseodym:
Resolves #2346


praseodym:
The Codecov backend CI job consistently fails because the disk gets full. I tried removing the GitHub Actions cache, without success. I'm not sure if the issue is that some test uses a lot of disk space, or that we were always close to running out of disk space and these changes use yet a little bit more space causing us to fill up the disk.

PR #2376, opened 2025-11-03
Free up runner disk space in Codecov backend job

praseodym:
The Codecov backend job fails for PRs #2371 and #2374 because the runner is out of disk space. As suggested by @rnijveld, we can free up some disk space on the runner by removing unused SDKs, so that there is more space to use for our job. To avoid adding an additional dependency, I used a simple shell task.


praseodym:
Initial free disk space: 22G
After freeing up disk space: 57G (+35GB)
Final free disk space after job run: 38G (-19GB)

rnijveld (approved):

PR #2384, opened 2025-11-03
Fix eslint ignores

### LINKED ISSUE #2358: Fix lint ignore blocks where possible
See #2342

We should review all current eslint ignore blocks and improve our code where possible.

marlonbaeten:
Fixes #2358

This PR fixes some low-hanging fruit in our code base where we previously used eslint ignore comments.

In general there were two types of of fixes:

- Perform type checks instead of unsafe type assertions.
- Use startTransition when assigning state in a useEffect

There are also some uses of the useMemo and useCallback hooks that are not necessary (useMemo is only needed for heavy computations).

### DIFF
--- a/frontend/src/features/data_entry/utils/actions.ts
+++ b/frontend/src/features/data_entry/utils/actions.ts
@@ -7,8 +7,9 @@ import {
   PollingStationResults,
   SaveDataEntryResponse,
 } from "@/types/generated/openapi";
-import { FormSectionId, SectionValues } from "@/types/types";
+import { DataEntryResults, FormSectionId, SectionValues } from "@/types/types";
 import { mapSectionValues } from "@/utils/dataEntryMapping";
+import { isRecord } from "@/utils/typeChecks";
 
 import {
   DataEntryDispatch,
@@ -18,6 +19,16 @@ import {
   TemporaryCache,
 } from "../types/types";
 import { calculateDataEntryProgress, getClientState } from "./dataEntryUtils";
+
+function isPollingStationResults(value: DataEntryResults): value is PollingStationResults {
+  if (!isRecord(value)) {
+    return false;
+  }
+
+  const model = value.model;
+
+  return typeof model === "string" && (model === "CSOFirstSession" || model === "CSONextSession");
+}
 
 export function setCache(dispatch: DataEntryDispatch) {
   return (cache: TemporaryCache) => {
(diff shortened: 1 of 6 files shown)


frontend/src/features/data_entry/utils/actions.ts
```diff
@@ -72,6 +83,10 @@ export function onSubmitForm(
       return false;
     }
 
+    if (!isPollingStationResults(state.pollingStationResults)) {
```
stacktraceghost, line 86: Is there a specific reason why `export type DataEntryResults = object;`? If we can type those props, we don't have to type assert afterwards

frontend/src/features/data_entry_home/components/PollingStationChoiceForm.tsx
```diff
@@ -62,12 +61,13 @@ export function PollingStationChoiceForm({ anotherEntry }: PollingStationChoiceF
     setCurrentPollingStation(pollingStation);
   }, USER_INPUT_DEBOUNCE);
 
-  // eslint-disable-next-line react-hooks/use-memo
-  useMemo(() => {
-    const parsedInt = parseIntUserInput(pollingStationNumber);
+  // set polling station number and trigger debounced lookup
+  const updatePollingStationNumber = (n: string) => {
+    setPollingStationNumber(n);
```
stacktraceghost, line 66: We could save a number instead of a string in `pollingStationNumber`, saves another `parseIntUserInput` down the line

marlonbaeten: `pollingStationNumber` is the current input field value which should always be a string, parsing while a user types causes all kinds of unwanted behavior.

stacktraceghost: Ah missed that it is constantly updating, makes sense!

frontend/src/features/data_entry/utils/actions.ts
marlonbaeten: I'm not sure why this choice was made, @praseodym maybe?
I think we could make this type more specific, our openapi.ts does define the `PollingStationResults` type for instance...

Either way - we probably want to do that change in a separate PR.

marlonbaeten: Looking at  the code, It might be really hard to make the type more specific, see https://github.com/kiesraad/abacus/blob/c8653482a3ceef3d75bf4c94c0c65df8edd2fb27/frontend/src/utils/dataEntryMapping.test.ts#L787 for instance....

marlonbaeten: I think it models arbitrary form data.

stacktraceghost: In `utils/reducer.ts` it assigns `action.data` to `pollingStationResults`, which is data that was mapped by `mapSectionValues`. So indeed, hard to model at that point. Might be good to investigate later.

stacktraceghost (approved):

PR #2394, opened 2025-11-04
Construct votes table columns and subtotals in Rust, not in Typst

marlonbaeten:
Fixes: #2396

Currently we calculate votes subtotals in our Typst templates. This PR introduces data structures to pre-calculate votes tables. 

When constructing these data structures sanity / integrity checks are performed:
- Make sure all political parties and candidates exist
- Make sure there are not more candidates then space in the columns
- Make sure the sum of the candidate votes / column subtotals match the overall total

A visual diff of all typst models was performed to detect any changes in this PR. Only one change was detected: missing totals in "model-na-14-2-bijlage1", which is an error on main.

To test: try the script at backend/templates/visual-diff.sh, make sure you install the dependencies listed in the source of this script.

### DIFF
--- a/frontend/src/features/data_entry/components/DataEntrySection.tsx
+++ b/frontend/src/features/data_entry/components/DataEntrySection.tsx
@@ -40,9 +40,9 @@ export function DataEntrySection() {
 
   const formId = section.id + "_form";
 
-  const bottomBarType = section.subsections.some((subsection) => subsection.type === "inputGrid")
-    ? "inputGrid"
-    : "form";
+  const lastSubsection = section.subsections[section.subsections.length - 1];
+  const bottomBarType = lastSubsection?.type === "inputGrid" ? "inputGrid" : "form";
+
   const keyboardHintText = section.id.startsWith("political_group_votes_") ? t("candidates_votes.goto_totals") : null;
 
   // Missing totals error for political group votes form
--- a/frontend/src/features/election_overview/components/OverviewPage.tsx
+++ b/frontend/src/features/election_overview/components/OverviewPage.tsx
@@ -27,8 +27,8 @@ function AddFirstElection() {
   return (
     <FormLayout>
       <FormLayout.Section>
-        <h2>{t("election.no_elections_added")}</h2>
-        {tx("election.add_first_election")}
+        <h2 className="mb-0">{t("election.no_elections_added")}</h2>
+        <div>{tx("election.add_first_election")}</div>
       </FormLayout.Section>
       <FormLayout.Controls>
         {isAdministrator && <Button.Link to={"./create"}>{t("election.create")}</Button.Link>}


oliver3:
Related to https://github.com/kiesraad/abacus/issues/2396

praseodym:
Good improvement!

Right now the `CandidatesTables` is passed as an extra field in the Typst template struct, duplicating existing data in a new structure. Can we remove the duplicated data from other fields in the struct, e.g. passing `Election` instead of `ElectionWithPoliticalGroups`?

marlonbaeten:
@praseodym in [49f8694](https://github.com/kiesraad/abacus/pull/2394/commits/49f8694d5d3dff050677795b9abe014ce18249d8) I removed unused templates and removed the duplication in the model inputs.
I have also added more tests to the votes tables data-structures in the last two commits.

backend/templates/visual-diff.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# instruct the user to switch to the feature branch before generating the "new" PDFs
+echo "Please make sure you are on the feature branch to generate the 'new' PDFs."
+read -p "Press Enter to continue..."
+
+typst compile --font-path fonts model-n-10-2.typ model-n-10-2-new.pdf
+typst compile --font-path fonts model-na-14-2-bijlage1.typ model-na-14-2-bijlage1-new.pdf
+typst compile --font-path fonts model-na-14-2.typ model-na-14-2-new.pdf
+typst compile --font-path fonts model-na-31-2-bijlage1.typ model-na-31-2-bijlage1-new.pdf
+typst compile --font-path fonts model-na-31-2-bijlage2.typ model-na-31-2-bijlage2-new.pdf
```
oliver3: It seems that `model-na-31-2-bijlage2` and `model-na-31-2-inlegvel.typ` are also unused at the moment. Should we remove them?

backend/templates/model-na-14-2-bijlage1.typ
```diff
@@ -159,24 +159,16 @@ Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere tell
 
 #pagebreak(weak: true)
 
-#for political_group in input.election.political_groups {
+#for political_group in input.votes_tables {
   votes_table(
     title: [#political_group.number #political_group.name],
     headers: ("Kandidaat", "", "Oorspronkelijk", "Gecorrigeerd"),
     total: none,
```
oliver3: I think I'd prefer `total: political_group.total` to let the input be even more in the lead, and that you can "prove" that the total is set to None by checking if the input model has total set to None

backend/templates/model-na-31-2-bijlage1.typ
```diff
@@ -189,16 +189,12 @@ _(Gebruik het proces-verbaal van het stembureau #sym.arrow.r Tijdens de stemming
 
 == Stemmen per lijst en per kandidaat
 
-#for political_group in input.election.political_groups {
+#for political_group in input.candidates_tables {
   votes_table(
     title: [#political_group.number #political_group.name],
     headers: ("Kandidaat", "", "Stemmen"),
     total: none,
```
oliver3: I think I'd prefer `total: political_group.total` to let the input be even more in the lead, and that you can "prove" that the total is set to None by checking if the input model has total set to None

backend/templates/visual-diff.sh
marlonbaeten: Sure - we can always get them back. See 116b3b5

backend/templates/model-na-14-2-bijlage1.typ
marlonbaeten: Fixed in 116b3b5

backend/templates/model-na-31-2-bijlage1.typ
marlonbaeten: Fixed in 116b3b5

PR #2402, opened 2025-11-05
Added warning icon to status page

### LINKED ISSUE #2351: Frontend: warning icon in status page
Add warning icon in status page for data entries/results with warnings
https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=16044-49306&t=wzhB3sJL9Bp6BeBD-11

lama-d3v:
- Fixes #2351 

Added warning icon to data entry on status page if there are any validation warnings.

Result can be seen [in Storybook](https://volgorde-aanwezige-artikelen.abacus-test.nl/storybook/?path=/story/features-election-status-components-electionstatus--election-status-with-links)

Also update documentation https://github.com/kiesraad/abacus-documentatie/blob/main/src/installeren-en-starten.md

### DIFF
--- a/frontend/src/app/routes.tsx
+++ b/frontend/src/app/routes.tsx
@@ -7,6 +7,7 @@ import { ElectionLayout } from "@/components/layout/ElectionLayout";
 import { ElectionStatusLayout } from "@/components/layout/ElectionStatusLayout";
 import { accountRoutes } from "@/features/account/routes";
 import { dataEntryRoutes } from "@/features/data_entry/routes";
+import { detailRoutes } from "@/features/data_entry_detail/routes";
 import { dataEntryHomeRoutes } from "@/features/data_entry_home/routes";
 import { devRoutes } from "@/features/dev/routes";
 import { electionCreateRoutes } from "@/features/election_create/routes";
@@ -20,7 +21,6 @@ import { pollingStationsRoutes } from "@/features/polling_stations/routes";
 import { resolveDifferencesRoutes } from "@/features/resolve_differences/routes";
 import { usersRoutes } from "@/features/users/routes";
 
-import { detailRoutes } from "../features/data_entry_detail/routes";
 import { RootLayout } from "./RootLayout";
 
 const showDevPage = __SHOW_DEV_PAGE__;
@@ -33,7 +33,7 @@ export const routes: RouteObject[] = [
       {
         index: true,
         path: "/",
-        element: showDevPage ? <Navigate to="/dev" replace /> : <Navigate to="/account/login" replace />,
+        element: showDevPage ? <Navigate to="/dev" replace /> : <Navigate to="/elections" replace />,
         handle: { public: true },
       },
       { path: "account", children: accountRoutes },
--- a/frontend/src/features/data_entry_detail/components/DetailLayout.test.tsx
+++ b/frontend/src/features/data_entry_detail/components/DetailLayout.test.tsx
@@ -1,22 +1,31 @@
 import * as ReactRouter from "react-router";
 
+import { userEvent } from "@testing-library/user-event";
 import { beforeEach, describe, expect, test, vi } from "vitest";
 
+import * as useMessages from "@/hooks/messages/useMessages";
 import { ElectionProvider } from "@/hooks/election/ElectionProvider";
 import { ElectionStatusProvider } from "@/hooks/election/ElectionStatusProvider";
 import { MessagesProvider } from "@/hooks/messages/MessagesProvider";
-import { dataEntryHasWarningsGetMockResponse } from "@/testing/api-mocks/DataEntryMockData";
+import {
+  dataEntryHasWarningsGetMockResponse,
+  dataEntryValidGetMockResponse,
+} from "@/testing/api-mocks/DataEntryMockData";
 import {
   ElectionListRequestHandler,
   ElectionRequestHandler,
   ElectionStatusRequestHandler,
+  PollingStationDataEntriesAndResultDeleteHandler,
   PollingStationDataEntryGetHandler,
 } from "@/testing/api-mocks/RequestHandlers";
 import { overrideOnce, server } from "@/testing/server";
-import { render, screen, waitFor, within } from "@/testing/test-utils";
+import { render, screen, spyOnHandler, waitFor, within } from "@/testing/test-utils";
 import { TestUserProvider } from "@/testing/TestUserProvider";
+import { ErrorResponse } from "@/types/generated/openapi";
 
 import { DetailLayout } from "./DetailLayout";
+
+const navigate = vi.fn();
 
 const renderLayout = () => {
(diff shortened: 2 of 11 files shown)


backend/src/data_entry/status.rs
```diff
@@ -654,6 +674,17 @@ impl DataEntryStatus {
             _ => None,
         }
     }
+
+    /// Returns the timestamp at which point this data entry process was made definitive
```
oliver3: Please remove this comment

backend/src/data_entry/status.rs
```diff
@@ -496,7 +508,13 @@ impl DataEntryStatus {
     }
 
     /// Keep first entry while resolving differences
-    pub fn keep_first_entry(self) -> Result<Self, DataEntryTransitionError> {
+    pub fn keep_first_entry(
+        self,
+        polling_station: &PollingStation,
+        election: &ElectionWithPoliticalGroups,
+    ) -> Result<Self, DataEntryTransitionError> {
+        let validation_results = validate_data_entry_status(&self, polling_station, election)?;
```
Lionqueen94: Is this now called on the second entry or on the first one?

backend/src/data_entry/api.rs
```diff
@@ -991,6 +995,9 @@ pub struct ElectionStatusResponseEntry {
     #[schema(value_type = String)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub finished_at: Option<DateTime<Utc>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(nullable = false)]
+    pub has_warnings: Option<bool>,
```
Lionqueen94: Please add a comment describing what this field does/contains

backend/src/data_entry/status.rs
Lionqueen94: Changed in 3f8594ec94510d35692154099cbf170ce94f9115

backend/src/data_entry/api.rs
Lionqueen94: Added in 3f8594ec94510d35692154099cbf170ce94f9115

backend/src/data_entry/status.rs
Lionqueen94: Changed this function in da35481b8736b29dfa0d632fdca2a5df95ddbe77

oliver3:
I'll migrate the ElectionStatus tests to Storybook tests

Lionqueen94 (approved):

backend/openapi.json
```diff
@@ -5639,6 +5639,10 @@
             "description": "First entry user id",
             "minimum": 0
           },
+          "has_warnings": {
+            "type": "boolean",
+            "description": "Whether the finalised first or second data entry has warnings"
+          },
```
jschuurk-kr: I don't think `has_warnings` covers the meaning of this field sufficiently, which as far as I understand it is: If either the first data entry or the second data entry is finalized AND there are no differences between the two, then `has_warnings` will be in the response (for other statuses it's not in the response) with either a `false` value (no warnings in most recent entry) or `true` (at least one warning in most recent entry).

oliver3:
> If I have a PV with a large number of blank votes and both typists enter this correctly, I get the icon on the status page:
[..] If I then click on the polling station, there is no indication whatsoever what the warning is or to which fields it applies.

Warnings are not shown for the `Definitive` state specifically. Created issue https://github.com/kiesraad/abacus/issues/2448 to fix this in the read only view.

backend/openapi.json
oliver3: Good point, renamed to `finalised_with_warnings` in an attempt to reflect that in https://github.com/kiesraad/abacus/pull/2402/commits/da3fb3ddbed6c74aa9db41f2cfda1ceceb68dbac

backend/src/data_entry/api.rs
```diff
@@ -467,13 +467,15 @@ async fn polling_station_data_entry_delete(
...
-    let (_, _, committee_session, state) =
+    let (polling_station, election, committee_session, state) =
         validate_and_get_data(&mut tx, polling_station_id, &user.0).await?;
 
     let user_id = user.0.id();
     let new_state = match entry_number {
         EntryNumber::FirstEntry => state.delete_first_entry(user_id)?,
-        EntryNumber::SecondEntry => state.delete_second_entry(user_id)?,
+        EntryNumber::SecondEntry => {
+            state.delete_second_entry(user_id, &polling_station, &election)?
```
stacktraceghost, line 477: When a typist is working on the second entry (`DataEntryStatus::SecondEntryInProgress`) and aborts, this method will be called.

- `delete_second_entry` then calls `validate_data_entry_status` to get validation results. 
- `validate_data_entry_status` calls `data_entry_status.validate(...)` and matches `DataEntryStatus::SecondEntryInProgress` and then validates based on the second entry
- `delete_second_entry` then matches `DataEntryStatus::SecondEntryInProgress` and returns `DataEntryStatus::SecondEntryNotStarted` with `finalised_with_warnings: validation_results.has_warnings()` based on the _**second data entry**_ instead of the first one. 

Resulting in a potentially wrong value for `finalised_with_warnings`. 

Lets add a test to prevent this issue from coming back. I would propose a test in which:
- The first data entry has no warnings
- The second data entry is in progress and has warnings
- The second data entry gets aborted
- Assert that `DataEntryStatus::SecondEntryNotStarted` has `finalised_with_warnings: false` (which would be  `true` in the current situation)

backend/src/data_entry/status.rs
```diff
@@ -117,6 +117,8 @@ pub struct SecondEntryNotStarted {
     /// When the first data entry was finalised
     #[schema(value_type = String)]
     pub first_entry_finished_at: DateTime<Utc>,
+    // Whether the first data entry was finalised with warnings
```
stacktraceghost:
```suggestion
    /// Whether the first data entry was finalised with warnings
```

backend/src/data_entry/status.rs
```diff
@@ -170,6 +172,8 @@ pub struct Definitive {
     /// When both data entries were finalised
     #[schema(value_type = String)]
     pub finished_at: DateTime<Utc>,
+    // Whether the second data entry was finalised with warnings
```
stacktraceghost:
```suggestion
    /// Whether the second data entry was finalised with warnings
```

stacktraceghost (changes requested):

backend/src/data_entry/api.rs
oliver3: Awesome catch! Added unit tests for checking all `finalised_with_warnings` resulting from state changes, and then fixing this in https://github.com/kiesraad/abacus/pull/2402/commits/5edab2a8a274e14262c9cb737cd0228f8d2c0002

PR #2403, opened 2025-11-05
Update Teletex character set

praseodym:
- Emoji are not part of the Teletex character set, so we don't need it in the list
- I've sorted the list to match the order in the tables from the [Logisch Ontwerp BRP](https://www.rvig.nl/lo-brp)


marlonbaeten:
The 💩 was used to test our code will render a fallback character. Maybe we can add a separate test for that?

oliver3:
Compared the BRP pdf to the `teletex-test.json` and it seems to correspond with each other!

praseodym:
> The 💩 was used to test our code will render a fallback character. Maybe we can add a separate test for that?

Yeah I'll add a separate test for that, it doesn't make sense to add a random emoji to the teletex set but we should still test the fallback character.

praseodym:
> > The 💩 was used to test our code will render a fallback character. Maybe we can add a separate test for that?
> 
> Yeah I'll add a separate test for that, it doesn't make sense to add a random emoji to the teletex set but we should still test the fallback character.

Added in 1e8f4ff.

PR #2404, opened 2025-11-05
Added delete modal & confirmation alert for data entry for coordinator

### LINKED ISSUE #2297: Frontend: add delete button and modal
[Figma link](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12020-138406&m=dev)

### LINKED ISSUE #2299: Frontend: show data entry is deleted confirmation alert
[Figma design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12020-138886&m=dev)

lama-d3v:
Closes #2297 
Closes #2299

### DIFF
--- a/frontend/src/features/data_entry_detail/components/DetailLayout.test.tsx
+++ b/frontend/src/features/data_entry_detail/components/DetailLayout.test.tsx
@@ -125,11 +125,7 @@ describe("DetailLayout", () => {
     expect(screen.queryByRole("button", { name: "Invoer verwijderen" })).not.toBeInTheDocument();
   });
 
-  test("Redirect to status page when trying to visit detail page when data entry has differences", async () => {
-    // Coordinator is viewing the detail page when second entry has not been finished yet.
-    // Meanwhile, the second typist finishes the second entry which results in different entries.
-    // Next, the coordinator refreshes the detail page and should be redirected to the status page.
-    // Because of the new state (EntriesDifferent) of this data entry, the detail page is not allowed.
+  test("Redirect to status page on error DataEntryGetNotAllowed", async () => {
     vi.spyOn(ReactRouter, "useParams").mockReturnValue({ electionId: "1", pollingStationId: "3" });
 
     overrideOnce("get", "/api/polling_stations/3/data_entries/get", 409, {


frontend/src/i18n/locales/nl/data_entry.json
```diff
@@ -17,6 +17,12 @@
   "continue_with_errors": "Je kan alleen verder als je dit met de coördinator hebt overlegd.",
   "corrected": "Gecorrigeerd",
   "data_entry_not_possible": "Je kan stembureau {nr} niet invoeren",
+  "delete_are_you_sure": "Weet je zeker dat je de invoer voor stembureau {nr} wilt verwijderen? Deze actie kan niet worden teruggedraaid.",
+  "delete_all_are_you_sure": "Weet je zeker dat je de invoer voor stembureau {nr} wilt verwijderen? Let op: zowel de eerste als de tweede invoer wordt verwijderd. Deze actie kan niet worden teruggedraaid.",
+  "delete": "Invoer verwijderen",
+  "delete_data_entry": "Verwijder invoer",
+  "data_entry_deleted_details": "Stembureau {nr} is weer toegevoegd aan de werkvoorraad en kan opnieuw worden ingevoerd.",
+  "data_entry_deleted": "Invoer verwijderd",
```
Lionqueen94: Can you move these to `data_entry_detail.json`?

lama-d3v: Updated this

frontend/src/features/data_entry_detail/components/DetailLayout.tsx
```diff
@@ -41,6 +70,16 @@ export function DetailLayout() {
           <h1>{pollingStation.name}</h1>
           <Badge type={dataEntry.status === "second_entry_not_started" ? "first_entry_finalised" : dataEntry.status} />
         </section>
+        {dataEntry.status !== "first_entry_has_errors" && (
```
Lionqueen94, line 73: Why is this conditional added? The page should not be available for this status anyway right?

lama-d3v: When the first data entry has errors, this page will be used to show all errors and warnings.
In that case you're not allowed to delete the data entry:
image

Lionqueen94: Ah true, you're right!

Lionqueen94 (approved):

frontend/src/features/data_entry_detail/components/DetailLayout.tsx
stacktraceghost: Building on this comment, how about the `EntriesDifferent` status? Or should we even redirect to the resolve_differences route in that case?

lama-d3v: In case of `EntriesDifferent`, this page is not used at all (the resolve differences page is used in that case)

stacktraceghost: True, but lets say we were viewing this page while the status was `SecondDataEntryInProgress` and then the typist is finalising their entry. Now there are differences and the coordinator refreshes this page:
- This page will show warnings/errors? While we intend the coordinator to go to the resolve_differences page from the status page
- This page will show the delete button, which will then result in an error because the backend will not allow to delete with that state

It would be nice to catch this situation as well

lama-d3v: @stacktraceghost this case is also covered

Issue #2222, opened 2025-09-30
Faulty environment


oliver3:
Explore Abacus on a faulty environment
with ??
to discover how Abacus reacts and if data will be lost

--

Network issues
HD on read-only / disk full
memory full

praseodym:
When running Abacus on Windows using a user that does not have privileges to write to the database directory, I noticed that the only mention of this is in the console log:

```
2025-11-06T23:16:57.904409Z ERROR request{method=POST uri=/api/initialise/first-admin version=HTTP/1.1}: abacus::error: Authentication error: Database(Database(SqliteError { code: 8, message: "attempt to write a readonly database" }))
```

I think we should try to detect a readonly database situation directly when starting Abacus, so we can directly exit with an error instead of starting the Abacus web server.

marlonbaeten:
## Target

We want to validate Abacus fails gracefully when resources are unavailable or constrained. These resources include disk space, allocateble memory and network access.

## Resources

- I will use the binary build from our GitHub releases (I should note the exact version)
- Run the release build with a constrained docker image, using the for instance the arguments `--memory` and `--storage-opt`

## Information

- I want to verify Abacus fails with a proper error message on startup if the environment is faulty
- I want to verify Abacus logs errors (both to the audit log if possible and stderr)
- I want to verify the API returns errors when a request failed
- I want to verify the abacus frontend shows informational error messages if the Abacus backend environment is faulty

## Timebox

60 minutes

marlonbaeten:
## Findings

I have used the current main "f3c64f68ccb4429d92a71e4added11ab1420a60d" of abacus with the docker image (release build with memory-serve and embed-typst):

```sh
docker build . -t abacus
```

### 16m memory

```sh
docker run --rm -p 8080:8080 --memory=16m --memory-swap=16m abacus
```

#### Results:

- initial page is shown, but login causes crash
- docker container quits, no error is logged
- frontend:

<img width="494" height="248" alt="Image" src="https://github.com/user-attachments/assets/0acbafb5-19aa-4191-b6cf-bdff0f7e1d26" />

### 32m memory

```sh
docker run --rm -p 8080:8080 --memory=32m --memory-swap=32m abacus
```

#### Results:

- abacus is responsive (login, update users etc.)
- when downloading the report ZIP abacus crashes in the came manner as the 16m test
- last logged message: `INFO abacus::pdf_gen::embedded: Compile took 1303 ms, 0 warnings`

### 64m memory

```sh
[...]

PR #2420, opened 2025-11-10
Improve startup error messages

marlonbaeten:
See https://github.com/kiesraad/abacus/issues/2222#issuecomment-3497495256

Previously an error with the sqlite database (file) was only detected when the first requests to the server are processed. We want the system administrator to be notified of any database problems when Abacus is started.

In this change we write an audit event (`ApplicationStarted`) to the database when Abacus is starting-up. This is also a test whether the database is writable. We map coomon SQLite error codes to user friendly messages to help the system administrator to fix possible problems.

To test: run `cargo run -- --database "sqlite::memory:?mode=ro"` and see the error message:
```
ERROR abacus: Cannot write to the database file "sqlite::memory:?mode=ro". Please check file permissions.
```

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -51,6 +51,8 @@ pub mod zip;
 pub use app_error::AppError;
 pub use error::{APIError, ErrorResponse};
 
+use crate::app_error::{DatabaseErrorWithPath, DatabaseMigrationErrorWithPath};
+
 /// Maximum size of the request body in megabytes.
 pub const MAX_BODY_SIZE_MB: usize = 12;
 
@@ -330,7 +332,7 @@ pub async fn shutdown_signal() {
 /// Log that the application started and thereby check that the database is writeable
 /// This maps common sqlite errors to an AppError
 async fn log_app_started(conn: &mut SqliteConnection, db_path: &str) -> Result<(), AppError> {
-    audit_log::create(
+    Ok(audit_log::create(
         conn,
         &audit_log::AuditEvent::ApplicationStarted,
         None,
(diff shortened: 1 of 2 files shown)


backend/src/lib.rs
```diff
@@ -327,6 +327,34 @@ pub async fn shutdown_signal() {
...
+        None,
+    )
+    .await
+    .map_err(|e| {
+        if let Some(db_error) = e.as_database_error() {
+            let db_path = db_path.to_string();
+            // see https://sqlite.org/rescode.html
+            match db_error.code().unwrap_or_default().as_ref() {
+                "5" => AppError::DatabaseBusy(db_path),
+                "7" => AppError::DatabaseNoMemory(db_path),
```
rnijveld: I'm not sure if OOM situations need to be handled explicitly, the chance of any part of our application to trigger this is pretty equal. On Linux it doesn't even happen. On Windows your system is basically completely unusable already by the time you get OOM situations because it already very aggressively starts swapping to prevent the OOM situation. I would also guess just calling `code()` could just as well fail at this point because that call also needs to do an allocation.

cikzh: What if something like `earlyoom` is installed? Could that trigger the event when it's very tightly tuned?

rnijveld: At least on linux the only way to get malloc to fail is to disable memory overcommit and to build a custom kernel without the oom killer (or to allocate more than 280TiB-ish for a single process). Still: once you reach that state at least from our point there's probably not much we can do other than exit the process as soon as possible (and trying to allocate memory for error strings ourselves does not help that)

marlonbaeten: Removed this case in the latest commit.

Issue #2214, opened 2025-09-30
Committee session state machine


oliver3:
Explore the committee session state machine
using the UI
to discover if a user can reach illegal/undesired states

---

older charter see [comment](https://github.com/kiesraad/abacus/issues/2214#issuecomment-3556897208) as to why

Explore the committee session state machine
using the API and the database
to discover state machine gekke dingen

stacktraceghost:
Just noticed that it is possible to go from any status back to created. I don't think this should be the case.

jschuurk-kr:
Fuzzing is not feasible considering the implementation. So change of plan:
- review unit tests
- exploratory testing through the UI and/or API

Most important risk to explore is if an user can get to an illegal state. Charter updated accordingly.

Lionqueen94:
- Missing validations (`polling_stations.is_empty()`/`investigations.is_empty()`) for going from any state to `Created`
- Checked all status change buttons in the frontend and they are only available in the right state and only for the coordinator

Lionqueen94:
Added issue https://github.com/kiesraad/abacus/issues/2631 to add the missing validations

jschuurk-kr:
Closed, all findings processed.

PR #2422, opened 2025-11-10
Fix fixtures, so they reflect a valid state

cikzh:
Fixes https://github.com/kiesraad/abacus/issues/2400:

- Adds an investigation to `election_5_with_results.sql`, so that it is correctly in the `data_entry_in_progress` state.
- Changes state of `election_7_four_sessions.sql` to `created`
- Updated tests to use either fixture where appropriate

Fixes https://github.com/kiesraad/abacus/issues/2397:

- Added check so we only try  to do a state transition when we are in the first session
- Added missing polling station creation to full-flow e2e test to cover this path

---
Please review thoroughly, especially if the fixtures and test cases reflect real situations

### DIFF
--- a/backend/src/pdf_gen/filter_input.rs
+++ b/backend/src/pdf_gen/filter_input.rs
@@ -4,8 +4,7 @@ use ttf_parser::Face;
 const TOFU: char = '\u{fffd}';
 
 static FONT: OnceLock<Face<'static>> = OnceLock::new();
-static FONT_BYTES: &[u8] =
-    include_bytes!("../../../backend/templates/fonts/DM_Sans/DMSans-Regular.ttf");
+static FONT_BYTES: &[u8] = include_bytes!("../../templates/fonts/DM_Sans/DMSans-Regular.ttf");
 
 fn get_font() -> &'static Face<'static> {
     FONT.get_or_init(|| Face::parse(FONT_BYTES, 0).expect("Failed to parse font"))


backend/tests/committee_session_integration_test.rs
```diff
@@ -80,12 +80,12 @@ async fn test_committee_session_create_current_committee_session_not_finalised(p
...
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_7_four_sessions", "users"))))]
 async fn test_committee_session_delete_ok_status_created(pool: SqlitePool) {
     let addr = serve_api(pool).await;
     let cookie = shared::coordinator_login(&addr).await;
-    let election_id = 5;
-    let committee_session_id = 6;
+    let election_id = 7;
+    let committee_session_id = 704;
 
     shared::change_status_committee_session(
```
stacktraceghost: This is not necessary anymore, since it is already in the created state

backend/tests/committee_session_integration_test.rs
stacktraceghost: Shouldn't this fail? Here we are trying to go from `data_entry_in_progress` -> `created` while we have an investigation added. // Edit, made a note [here](https://github.com/kiesraad/abacus/issues/2214#issuecomment-3521581509)

This test could be shortened, since fixture 5 now already has an investigation, so we could only try to delete the committee session.

Good to check if there are more tests that can be simplified for fixture 5, since no investigations has to be added anymore. I have not seen all tests outside of the changes made.

stacktraceghost: Is already in progress now

stacktraceghost:
Removed from queue, I was still reviewing and found some points. Will post them soon.

backend/tests/investigation_integration_test.rs
```diff
@@ -197,7 +194,7 @@ async fn test_deletion_removes_polling_station_from_status(pool: SqlitePool) {
     let polling_station_id = 9;
 
     let statuses = get_statuses(&addr, &cookie, election_id).await;
-    assert_eq!(statuses.len(), 0);
+    assert_eq!(statuses.len(), 1);
 
     // Add investigation with corrected_results: true
     assert_eq!(
```
stacktraceghost: Not necessary anymore, we don't have to create the investigation and data entry, we can just delete the investigation for polling station 11.

backend/tests/investigation_integration_test.rs
```diff
@@ -834,35 +832,19 @@ where
     Fut: Future<Output = Response>,
 {
     let cookie = shared::coordinator_login(addr).await;
-    let election_id = 7;
-    let committee_session_id = 704;
+    let election_id = 5;
+    let committee_session_id = 6;
 
     if pre_create {
```
stacktraceghost, line 744: `pre_create` could be removed, since fixture 5 now already has an investigation.

backend/tests/committee_session_integration_test.rs
cikzh: Removed in [e82c0bd](https://github.com/kiesraad/abacus/pull/2422/commits/e82c0bd16170a86c6ed6aee79e366b52a2736c07)

backend/tests/investigation_integration_test.rs
cikzh: Simplified the test in [a2b4e0b](https://github.com/kiesraad/abacus/pull/2422/commits/a2b4e0bba4b749293f96720b300e458ad657c596) it's a lot more readable now, thanks :+1:

backend/tests/committee_session_integration_test.rs
cikzh: Removed in [6721f96](https://github.com/kiesraad/abacus/pull/2422/commits/6721f96de8b390ab971dc367a654b0f5a73e41e6)

backend/tests/investigation_integration_test.rs
cikzh: the existing investigation has `corrected_results` set to true, the test expects false, so instead of changing `corrected_results` we might as well keep it i'd say.

backend/tests/committee_session_integration_test.rs
cikzh: Good find, the [flowchart](https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/committee-session-state.md) indeed states it shouldn't be possible.

[f46bf67](https://github.com/kiesraad/abacus/pull/2422/commits/f46bf67fda1c0a2f191b3a6535b572c708978670) simplifies the test. It also expects a different return code (409 instead of 422). Please check if this is correct.

stacktraceghost: It is throwing `409 Conflict` because it is the wrong committee session status (`DataEntryInProgress`). You can only delete when it is `Created`/`DataEntryNotStarted`. So maybe we can use fixture 7, which has status `Created` and add one investigation, then try to delete. 

The issue about the state machine we will discuss outside of the scope of this PR.

cikzh: Changed to fixture 7 in [fd51a50](https://github.com/kiesraad/abacus/pull/2422/commits/fd51a50fa8155b78c0cd8369eec805487d984e4e). I think this also improves the readability of the test, so great suggestion

stacktraceghost:
Pushed fixes related to election_5 tests, it would be great if someone can still check all election_7 tests.

Edit: discussed and will be done after the merge.

Lionqueen94 (approved):

PR #2426, opened 2025-11-11
Split lib.rs into router.rs

marlonbaeten:
Our `lib.rs` in the Rust backend has grown rather large. This PR splits the router specific function up into a router.rs file to improve  readability.

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -12,7 +12,7 @@
   "paths": {
     "/api/account": {
       "get": {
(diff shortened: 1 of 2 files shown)


ring-ring-ring:
Is dit 100% noodzakelijk voor nu, of kunnen we deze laten wachten tot na de GR deadline?

marlonbaeten:
NP to wait with this. Although I'm also confident that these kind of refactors in Rust code can be done without hesitation.
I will close this for now.

Issue #2424, opened 2025-11-11
PDF footer contains local server time, without noting the time zone


oliver3:
<img width="362" height="57" alt="Image" src="https://github.com/user-attachments/assets/82b8b391-bf3c-439b-b898-9f8be96a71e3" />

marlonbaeten:
The date time that is rendered as creation date time in PDF files and EML files is the local time of the server / computer where Abacus is running (on abacus-test.nl this is UTC). In EML files the timezone is explicit (rfc3339).

In order to prevent confusion we have to make a choice here:

- Either always print the UTC time and also mention that the time is in UTC
- Print the local server time and also print the local time zone
- Instruct the maintainers to configure the "correct" timezone on systems where Abacus runs
- Explicitly ask maintainers which timezone Abacus should use

My personal preference is the second option, we should explicitly note the local timezone + instruct system maintainers to configure a correct timezone.

Lionqueen94:
I agree, it should be the local time and there should indeed be an instruction to make sure the time is set correctly on the systems.

jschuurk-kr:
Key part to me is that the date time in the EML is easy to check agains the date time in the pdf. This way it can function as an additional check between the two documents, next to the hash.

That would also be an argument to include the seconds, like OSV does:

<img width="714" height="70" alt="Image" src="https://github.com/user-attachments/assets/a8d52172-67de-4b13-81d3-d16a03425d30" />

PR #2427, opened 2025-11-11
Print creation time zone (offset) and seconds in PDF footer

marlonbaeten:
Possible fix for #2424

This prints the time zone at the creation date time in the footer of PDF files:

Screenshot From 2025-11-11 12-27-21

Note that the time zome is the one of the server.

### DIFF
--- a/backend/src/report/api.rs
+++ b/backend/src/report/api.rs
@@ -28,6 +28,7 @@ use crate::{
         models::{ModelNa14_2Input, ModelNa31_2Input, ModelP2aInput, PdfFileModel, ToPdfFileModel},
     },
     polling_station::structs::PollingStation,
+    report::DEFAULT_DATE_TIME_FORMAT,
     summary::ElectionSummary,
     zip::{ZipResponse, ZipResponseError, slugify_filename},
 };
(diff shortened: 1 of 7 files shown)


backend/src/report/api.rs
```diff
@@ -173,7 +173,10 @@ impl ResultsInput {
 
     fn into_pdf_file_models(self, xml_hash: impl Into<String>) -> Result<PdfModelList, APIError> {
         let hash = xml_hash.into();
-        let creation_date_time = self.creation_date_time.format("%d-%m-%Y %H:%M").to_string();
+        let creation_date_time = self
+            .creation_date_time
+            .format("%d-%m-%Y %H:%M %Z")
```
jschuurk-kr: As commented in the issue, let's also have seconds here, as OSV does.

marlonbaeten:
Added seconds in c06896d28b3c3e1103ff1c66f96058550d155388:

image

backend/src/report/api.rs
marlonbaeten: https://github.com/kiesraad/abacus/pull/2427#issuecomment-3516996942

Lionqueen94 (dismissed):

oliver3:
Maybe update the test strings "04-12-2024 12:08" to reflect the new format as well? (to avoid confusion, wrong assumptions, etc)

We should also update the format in `typst_tests.rs`. Is it possible to make a constant out of this format?

PR #2428, opened 2025-11-11
Only allow changing number of voters in first session and when data entry not started

### LINKED ISSUE #2415: Number of voters change is not reflected in documents, only P2a is generated
Discussed with Arjen, it should not be possible to change the number of voters once the data entry phase has started. And it's not possible to change it in a next session. This needs to be restricted and changed in the use cases.

Lionqueen94:
Closes #2415

## Description
- Added committee session status checks to change voters endpoint
- Added committee session status checks to frontend link to change voters
- Prevent number of voters from being emptied or changed to 0
- Added option for changing number of voters to the documentation
- Allow administrator to also change the number of voters

### DIFF
--- a/backend/tests/committee_session_integration_test.rs
+++ b/backend/tests/committee_session_integration_test.rs
@@ -768,3 +768,75 @@ async fn test_committee_session_number_of_voters_change_not_found(pool: SqlitePo
     // Ensure the response is what we expect
     assert_eq!(response.status(), StatusCode::NOT_FOUND);
 }
+
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_7_four_sessions", "users"))))]
+async fn test_committee_session_investigations_list_works(pool: SqlitePool) {
+    let addr = serve_api(pool).await;
+    let cookie = shared::coordinator_login(&addr).await;
+    let url = format!("http://{addr}/api/elections/7/committee_sessions/702/investigations");
+    let response = reqwest::Client::new()
+        .get(&url)
+        .header("cookie", cookie)
+        .send()
+        .await
+        .unwrap();
+
+    assert_eq!(
+        response.status(),
+        StatusCode::OK,
+        "Unexpected response status"
+    );
+
+    // Validate response and make sure the investigations are from the requested committee session
+    let body: serde_json::Value = response.json().await.unwrap();
+    let investigations = body["investigations"].as_array().unwrap();
+    assert_eq!(investigations.len(), 1);
+    let map = investigations
+        .iter()
+        .map(|inv| inv["polling_station_id"].as_u64().unwrap())
+        .collect::<Vec<u64>>();
+    assert_eq!(map, vec![721]);
+}
+
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_7_four_sessions", "users"))))]
+async fn test_committee_session_investigations_list_works_no_investigations(pool: SqlitePool) {
+    let addr = serve_api(pool).await;
+    let cookie = shared::coordinator_login(&addr).await;
+    let url = format!("http://{addr}/api/elections/7/committee_sessions/704/investigations");
+    let response = reqwest::Client::new()
+        .get(&url)
+        .header("cookie", cookie)
+        .send()
+        .await
+        .unwrap();
+
+    assert_eq!(
+        response.status(),
+        StatusCode::OK,
+        "Unexpected response status"
+    );
+
+    // Validate response and make sure the investigations are empty
+    let body: serde_json::Value = response.json().await.unwrap();
+    let investigations = body["investigations"].as_array().unwrap();
+    assert_eq!(investigations.len(), 0);
+}
+
+#[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_7_four_sessions", "users"))))]
+async fn test_committee_session_investigations_list_fails_election_not_found(pool: SqlitePool) {
+    let addr = serve_api(pool).await;
+    let cookie = shared::coordinator_login(&addr).await;
+    let url = format!("http://{addr}/api/elections/1/committee_sessions/1/investigations");
+    let response = reqwest::Client::new()
+        .get(&url)
+        .header("cookie", cookie)
+        .send()
+        .await
+        .unwrap();
+
+    assert_eq!(
+        response.status(),
+        StatusCode::NOT_FOUND,
+        "Unexpected response status"
+    );
+}
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -234,7 +234,7 @@
     },
     "/api/elections/{election_id}": {
       "get": {
-        "summary": "Get election details including the election's candidate list (political groups),\nits polling stations and the current committee session",
+        "summary": "Get election details including the election's candidate list (political groups),\nits polling stations and the current committee session and its investigations",
         "operationId": "election_details",
         "parameters": [
           {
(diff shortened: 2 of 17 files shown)


jschuurk-kr:
If we don't allow a change to number of voters, we should update our Na14-2
image

jorisleker:
Why can't I change the number of voters after data entry has started? In my opinion, you should be able to update this number until the proces-verbaal has been generated. So as long as data entry has not finished. Stating this because it is not possible to return to the DataEntryNotStarted state after you have started it. So if you forget to update this number by accident, you have no way to correct it anymore

jschuurk-kr:
doublecheck with @arjen-ba and @Lionqueen94 if it should be possible to change number of eligible voters, e.g. because of a mistake when initially entering the number

backend/tests/committee_session_integration_test.rs
```diff
@@ -634,11 +634,30 @@ async fn test_committee_session_status_change_previous_committee_session_fails(p
...
+    pool: SqlitePool,
+) {
     let addr = serve_api(pool).await;
-
-    let url = format!("http://{addr}/api/elections/2/committee_sessions/2/voters");
     let coordinator_cookie = shared::coordinator_login(&addr).await;
+    let election_id = 2;
+    let committee_session_id = 2;
+
+    shared::change_status_committee_session(
```
stacktraceghost: It seems like we are able to go from any status back to created, which does not seem correct.  // Edit, made a note [here](https://github.com/kiesraad/abacus/issues/2214#issuecomment-3521581509)


For this test we could use fixture 6, which is already in a created state.

arjen-ba:
> doublecheck with @arjen-ba and @Lionqueen94 if it should be possible to change number of eligible voters, e.g. because of a mistake when initially entering the number

The number of eligible voters ("kiesgerechtigden" / "Z") is determined on nomination day ("dag van kandidaatstelling"), afterwards additions or subtractions are possible. Also when an incorrect value is entered, it should be possible to adjust. Therefor it should be possible to change "Z" but only until the start of the first session (so before the input of the results starts).

Lionqueen94:
> > doublecheck with @arjen-ba and @Lionqueen94 if it should be possible to change number of eligible voters, e.g. because of a mistake when initially entering the number
> 
> The number of eligible voters ("kiesgerechtigden" / "Z") is determined on nomination day ("dag van kandidaatstelling"), afterwards additions or subtractions are possible. Also when an incorrect value is entered, it should be possible to adjust. Therefor it should be possible to change "Z" but only until the start of the first session (so before the input of the results starts).

@arjen-ba But why do we then have the field in the corrigendum? See https://github.com/kiesraad/abacus/pull/2428#issuecomment-3520539646

arjen-ba:
> > > doublecheck with @arjen-ba and @Lionqueen94 if it should be possible to change number of eligible voters, e.g. because of a mistake when initially entering the number
> > 
> > 
> > The number of eligible voters ("kiesgerechtigden" / "Z") is determined on nomination day ("dag van kandidaatstelling"), afterwards additions or subtractions are possible. Also when an incorrect value is entered, it should be possible to adjust. Therefor it should be possible to change "Z" but only until the start of the first session (so before the input of the results starts).
> 
> @arjen-ba But why do we then have the field in the corrigendum? See [#2428 (comment)](https://github.com/kiesraad/abacus/pull/2428#issuecomment-3520539646)

In an earlier stage the decision has been made that model Na 14-2 will not show both the original value "oorspronkelijk" and the corrected value "gecorrigeerd", this is obsolete / because incorrect. Na 14-2 will only show the original value of "Z", in the last versions of the new models this has been corrected as well. Final versions will be shared on a short notice.

backend/tests/committee_session_integration_test.rs
Lionqueen94: Changed in 602f2028e25160a696fa665686e5586657d51a96

documentatie/use-cases/beheer.md
```diff
@@ -61,7 +61,10 @@ __Uitbreidingen:__
 
 3a. De beheerder slaat het invoeren van stembureaus over:  
 &emsp; 3a1. (tijdens stap 5) De beheerder voert het aantal kiesgerechtigden in de gemeente handmatig in.  
-&emsp; 3a2. (na inrichten applicatie) [De beheerder of coördinator zet de stembureaus in de applicatie.](#de-beheerder-of-coördinator-zet-de-stembureaus-in-de-applicatie-zee)
+&emsp; 3a2. (na inrichten applicatie) [De beheerder of coördinator zet de stembureaus in de applicatie.](#de-beheerder-of-coördinator-gsb-zet-de-stembureaus-in-de-applicatie-zee)
+
+5a. De beheerder stelt vast dat het aantal kiesgerechtigden uit het stembureaubestand (EML 110b) niet klopt:
+&emsp; 5a1. De beheerder corrigeert het aantal kiesgerechtigden in de applicatie.
```
lama-d3v, line 67: Maybe add here that this isn't allowed as soon as the data entry phase has started

Lionqueen94:
@marjoleintamis I see that there is no documentation on updating the number of voters in https://github.com/kiesraad/abacus-documentatie/blob/main/src/instructies-beheerder.md and https://github.com/kiesraad/abacus-documentatie/blob/main/src/instructies-coordinator.md.
Is this on your radar to be added?

marjoleintamis:
@Lionqueen94 yes! I've added it to the PR for the coordinator I'm about to publish and it's also on my todo for the beheerder.

oliver3 (approved):
Looks good!

documentatie/use-cases/gsb-eerste-zitting.md
```diff
@@ -12,33 +12,40 @@ __Hoofdscenario:__
...
 
-8a. Het GSB stelt een probleem vast met het PV:  
-&emsp; 8a1. Het GSB stelt het bezwaar vast.  
-&emsp; 8a2. Het GSB gaat over tot hertelling. (zie uitbreiding 4b)
+4b. De coördinator GSB ziet een fout in het aantal kiesgerechtigden:  
+&emsp; 4b1. De coördinator GSB of de beheerder past het aantal kiesgerechtigden aan.
 
-8b. Er zijn bezwaren en/of bijzonderheden tijdens het voorlezen van het PV:  
-&emsp; 8b1. Het GSB voegt de bijzonderheden en/of bezwaren met de hand toe aan het papieren PV.
+8a. Het GSB stelt verschillen vast d.m.v. het controleprotocol (handmatige controle optellingen software):
```
lama-d3v, line 37: Is de "handmatige controle optellingen" wel software?

frontend/src/features/election_management/components/update/NumberOfVotersPage.tsx
```diff
@@ -14,16 +15,22 @@ import {
...
+    if (numberOfVoters > 0) {
+      setError(undefined);
+      const body: COMMITTEE_SESSION_NUMBER_OF_VOTERS_CHANGE_REQUEST_BODY = { number_of_voters: numberOfVoters };
+      void update(body).then((result) => {
+        if (isSuccess(result)) {
+          void navigate("..");
+        }
+      });
+    } else {
+      setError(t("election.number_of_voters.error"));
```
lama-d3v, line 32: Is error message also triggered when the number of voters is set as 0? Because then this message could be a bit confusing.

frontend/src/features/election_management/components/ElectionInformationTable.tsx
```diff
@@ -61,13 +58,14 @@ export function ElectionInformationTable({
...
         </Table.Row>
-        {isCoordinator ? (
+        {committeeSession.number === 1 &&
+        (committeeSession.status === "created" || committeeSession.status === "data_entry_not_started") ? (
           <Table.LinkRow to={"number-of-voters"}>
             <Table.HeaderCell scope="row" className="normal">
               {t("number_of_voters")}
             </Table.HeaderCell>
             <Table.Cell className="underlined">
-              {numberOfVoters ? formatNumber(numberOfVoters) : t("election_management.still_to_input")}
```
lama-d3v, line 70: Is it correct that this translation is deleted?

Lionqueen94: Yes, because the numberOfVoters cannot be zero

documentatie/use-cases/gsb-eerste-zitting.md
Lionqueen94: No it's not, this is something @jschuurk-kr and I already discussed and will most likely be removed/moved in the future. Not in the scope of this PR

frontend/src/features/election_management/components/update/NumberOfVotersPage.tsx
Lionqueen94: image

This is the error rendered when you enter 0, I think this is fine

documentatie/use-cases/beheer.md
Lionqueen94: @jschuurk-kr What do you think? Should we add this condition to the documentation we added?

jschuurk-kr: It shouldn't go in this use case, because this use case happens before the data entry phase.

We also have the extention "4b. De coördinator GSB ziet een fout in het aantal kiesgerechtigden:"  in the use case "Gemeentelijk stembureau (GSB) stelt uitkomst vast in eerste zitting (wolk)".

To properly get it into the use cases, you would need to find a situation in which someone would want to update the number of eligible voters during data entry, and use that as the starting point for a use case where the application does not allow editing the field. (And ideally, describe what it is the user should do instead.)
 I don't think that that's a good idea. Which means it will only be documented in the use cases in the sense that nowhere it says that it's possible during the data entry phase and it does explicitly say it's possible at two other points in time.

frontend/src/features/election_management/components/update/NumberOfVotersPage.tsx
jschuurk-kr: might be impacted by PR https://github.com/kiesraad/abacus/pull/2464

oliver3 (dismissed):

stacktraceghost (dismissed):

PR #2433, opened 2025-11-12
Generate PDF models command

marlonbaeten:
Small Rust binary (command) to generate the current PDF models using our example data. Once this is merged we can create a github action that generates visual diffs of our models for each pull request.

### DIFF
--- a/backend/src/bin/gen-pdf.rs
+++ b/backend/src/bin/gen-pdf.rs
@@ -1,6 +1,18 @@
 use abacus::pdf_gen::models::{PdfFileModel, PdfModel};
+use clap::Parser;
 use tokio::{fs, process::Command};
 
+/// Generate PDFs for all available models using the input JSON files in
+/// `templates/inputs/` and store them in a subdirectory of `tmp-pdf-gen/`. The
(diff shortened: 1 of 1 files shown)


backend/src/bin/gen-pdf.rs
```diff
@@ -0,0 +1,62 @@
...
+        let input_path = format!("templates/inputs/{}.json", model);
+        let contents = fs::read_to_string(input_path).await?;
+
+        let model = PdfModel::from_name_with_input(model, &contents)?;
+
+        let file_name = format!("{path}/{}.pdf", model.as_model_name());
+
+        let file_model = PdfFileModel::new(file_name.clone(), model);
+
+        let pdf = abacus::pdf_gen::generate_pdf(file_model).await?;
```
lama-d3v: Perhaps print which model is currently generated. That way you can easily see for which model the error occurred for (if an error is triggered somehow)

marlonbaeten: Added file / model name in error output in b6facebf30d635e5e0ef9c09deba0d79c5aa6322

Lionqueen94 (approved):

Issue #2213, opened 2025-09-30
Data entry state machine


oliver3:
Explore the data entry state machine
using the API and the database
to discover state machine gekke dingen

jschuurk-kr:
Closed. See PR #2527 for test results and the fuzzer code. Fuzzer is being added to CI in PR #2616.

PR #2440, opened 2025-11-13
Add fuzzer for data entry state machine

michielp1807:
This PR adds a fuzzer that tests that the data entry system matches the state machine described in the [docs](https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/data-entry-state.md). The fuzzer checks that all possible transitions change the state as expected, or give the expected error. Related to #2213.

The fuzzer covers all possible states and all possible `DataEntryTransitionError` errors, except for the `FirstEntryAlreadyClaimed`, `ValidatorError` and `ValidationError` errors. The `FirstEntryAlreadyClaimed` error is not triggered by the fuzzer for simplicity, but this error is explicitly tested by the `first_entry_in_progress_claim_first_entry_other_user_error` unit test. The `ValidatorError` and `ValidationError` errors are not triggered by the fuzzer, because the fuzzer does not test the API directly. Instead, it only tests the internal functions to focus on verifying the data entry state machine (so this is basically a more complete extension of the unit tests that test individual transitions, which shows that no unexpected errors or transition can happen).

I did not find any real problems, but I did come across one small oddity:
- When in `DataEntryStatus::SecondEntryInProgress`, claiming the first entry gives a `DataEntryTransitionError::Invalid` error instead of `DataEntryTransitionError::FirstEntryAlreadyFinalised`, which I think would be more consistent

### DIFF
--- a/backend/fuzz/README.md
+++ b/backend/fuzz/README.md
@@ -1,0 +1,62 @@
+# Abacus fuzz tests
+
+We use [cargo-fuzz](https://rust-fuzz.github.io/book/cargo-fuzz.html) for fuzzing, below we describe the fuzz tests contained within this crate.
+
+## Data entry status fuzz test
+
+The `data_entry_status` fuzz test tests that the data entry system matches the state machine described in the [docs](https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/data-entry-state.md). The fuzzer checks that all possible transitions change the state as expected, or give the expected error.
+
+This fuzz test covers all possible states and all possible `DataEntryTransitionError` errors, except for the `FirstEntryAlreadyClaimed`, `ValidatorError` and `ValidationError` errors. The `FirstEntryAlreadyClaimed` error is not triggered by the fuzzer for simplicity, but this error is explicitly tested by the `first_entry_in_progress_claim_first_entry_other_user_error` unit test. The `ValidatorError` and `ValidationError` errors are not triggered by the fuzzer because the fuzzer does not test the API directly. Instead, the fuzzer only tests the internal functions to focus on verifying the data entry state machine. This fuzz test can be seen as a more complete extension of the unit tests that test individual transitions, which shows that no unexpected errors or transition can happen.
+
+Run it with the following command (for fuzzing you need the nightly compiler):
+
+```
+cargo +nightly fuzz run data_entry_status
+```
+
+The output of the fuzzer should look something like:
+
+```
+#16384  pulse  cov: 1036 ft: 5295 corp: 482/128Kb lim: 3439 exec/s: 3276 rss: 468Mb
+#16856  REDUCE cov: 1036 ft: 5295 corp: 482/128Kb lim: 3439 exec/s: 3371 rss: 482Mb L: 176/3406 MS: 1 EraseBytes-
+#22839  NEW    cov: 1036 ft: 5305 corp: 483/128Kb lim: 3472 exec/s: 3262 rss: 571Mb L: 620/3406 MS: 4 CrossOver-CrossOver-ChangeASCIIInt-CrossOver-
+```
+
+Here, `NEW` indicates that a new code path has been triggered, and `REDUCE` indicates a previously covered code path has been triggered with a shorter input. The `pulse` is a periodic status update.
+
+If the fuzzer has found a problem, it will exit and print the error. The error message will contain which unexpected transition took place. If after a few minutes the fuzzer has not found any problems we can be fairly certain the data entry system matches the state machine described in the fuzz test.
+
+### Checking the fuzz test coverage
+
+To check the coverage of the fuzzer, we can use the `fuzz coverage` command.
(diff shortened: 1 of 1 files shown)


michielp1807:
Thanks Joep! I've added a README with some additional instructions on how to run the fuzzer and how to check the coverage data afterwards

marlonbaeten (dismissed):
Nice work @michielp1807 In a next iteration we could look at running a (limited) fuzz test on CI.

backend/fuzz/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+```
+$(rustc --print sysroot)/lib/rustlib/$(rustc --print host-tuple)/bin/llvm-cov show \
+ target/$(rustc --print host-tuple)/coverage/$(rustc --print host-tuple)/release/data_entry_status \
+ -instr-profile=fuzz/coverage/data_entry_status/coverage.profdata \
+ -Xdemangler=rustfilt \
+ --format=html \
+ -output-dir=/tmp/coverage \
+ -ignore-filename-regex="\.cargo|\.rustup|fuzz_target"
+```
```
jschuurk-kr, line 60: Great README!

Unfortunately I get stuck at this point, because we use stable by default.

I tried to make it work by running `rustup override set nightly` in the `fuzz` directory and removing the "fuzz" from `-instr-profile`. That resulted in an error:
```
~/tmp/abacus-fuzzer/backend/fuzz (fuzz_data_entry_state_machine =) $ $(rustc --print sysroot)/lib/rustlib/$(rustc --print host-tuple)/bin/llvm-cov show  target/$(rustc --print host-tuple)/coverage/$(rustc --print host-tuple)/release/data_entry_status  -instr-profile=coverage/data_entry_status/coverage.profdata  -Xdemangler=rustfilt  --format=html  -output-dir=/tmp/coverage  -igno
re-filename-regex="\.cargo|\.rustup|fuzz_target"
error: /rustc/3d461af2a23456a2676aadb13b4253c87bdfe28d/library/std/src/sys/thread_local/native/mod.rs: No such file or directory
warning: The file '/rustc/3d461af2a23456a2676aadb13b4253c87bdfe28d/library/std/src/sys/thread_local/native/mod.rs' isn't covered.
```
And unfortunately I don't have the time to experiment further.

marlonbaeten: Can you try running:
```
rustup toolchain install nightly
```
Then this should work:
```
cargo +nightly fuzz run data_entry_status
```
The `+nightly` select the nightly rust version just for this command. No need to use "rustup override".

jschuurk-kr: That part indeed works fine as you describe. I run into issues when trying to output the coverage data with the multi-line command that contains `rustc`.

michielp1807: Indeed, I forgot to add `+nightly` to the `rustc` commands there. Fixed!

jschuurk-kr: Am afraid I'm still getting an error. :-( However, `/tmp/coverage` does contain the coverage overview, so it is working.

In the `backend` folder I run
```
 cargo +nightly fuzz run data_entry_status
```
which I stop after some time with Ctrl+C.

Then I run
```
cargo +nightly fuzz coverage data_entry_status
```
which has as last line of output
```
Coverage data merged and saved in "/home/joep/tmp/abacus-fuzzer/backend/fuzz/coverage/data_entry_status/coverage.profdata".
```
So far so good, I think.

Then I run the command with the `rustc` in it and get the following:
```
~/tmp/abacus-fuzzer/backend (fuzz_data_entry_state_machine =) $ $(rustc +nightly --print sysroot)/lib/rustlib/$(rustc +nightly --print host-tuple)/bin/llvm-cov show \
 target/$(rustc +nightly --print host-tuple)/coverage/$(rustc +nightly --print host-tuple)/release/data_entry_status \
 -instr-profile=fuzz/coverage/data_entry_status/coverage.profdata \
 -Xdemangler=rustfilt \
 --format=html \
 -output-dir=/tmp/coverage \
 -ignore-filename-regex="\.cargo|\.rustup|fuzz_target"
error: /rustc/3d461af2a23456a2676aadb13b4253c87bdfe28d/library/std/src/sys/thread_local/native/mod.rs: No such file or directory
warning: The file '/rustc/3d461af2a23456a2676aadb13b4253c87bdfe28d/library/std/src/sys/thread_local/native/mod.rs' isn't covered.
```

michielp1807: I think this error is just warning that it cannot track a file from Rust's standard library. Not sure why I didn't get this error, but we can prevent this error by ignoring filenames starting with `/rustc`. We already filter out files from `.cargo` and `.rustup` because we are only interested in the coverage within Abacus itself, not within Rust's standard library or other dependencies. I've updated command in the README to also ignore `/rustc`.

michielp1807:
I just noticed the rebase I did earlier today caused some problems in the fuzzer, I will fix this on Tuesday.

PR #2445, opened 2025-11-13
Add workflow that visualizes changes to PDF models

marlonbaeten:
Continuation of #2433

This performs a visual diff of the PDFs files generated from our Typst models using fixture data.

A report table is generated in commented on the PR. If there are differences detected the visuals diffs of the changes are uploaded as an artifact and linked in the PR comment.

Example of the generated comment:

image

Example of a generated diff:

[model-na-14-2-bijlage1.pdf](https://github.com/user-attachments/files/23533196/model-na-14-2-bijlage1.pdf)

image

Also see the comment below.

### DIFF
--- a/.github/workflows/pdf-diff.yml
+++ b/.github/workflows/pdf-diff.yml
@@ -21,7 +21,6 @@ jobs:
     env:
       CARGO_TERM_COLOR: always
       SQLX_OFFLINE: "true"
-      HEAD_REF: ${{ github.head_ref || github.ref_name }}
     steps:
       - uses: actions/checkout@v5
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pdf-diff.yml
jschuurk-kr, line 1: This is really cool!

One comment: I'd prefer the scripts to be in separate files. That might also allow us to use https://github.com/marocchino/sticky-pull-request-comment (used in our CLA workflow) for the comment posting.

marlonbaeten: In general I think preventing additional dependencies is good - but since we already use this dependency; let's also add it in this workflow.

marlonbaeten: See a04d4334074d844d3bc6eb88ac82a30a83285a66

cikzh (approved):
Tested in draft PR (#2454) where I'm actually fixing stuff in Typst templates. Works great!

PR #2446, opened 2025-11-13
Prevent generating invalid/ambiguous dates in typst tests

marlonbaeten:
The `random_naive_date_time` function was able to generate dates that are ambiguous when converting to a timezone. This fixes that issue.

### DIFF
--- a/backend/src/pdf_gen/typst_tests.rs
+++ b/backend/src/pdf_gen/typst_tests.rs
@@ -49,8 +49,9 @@ fn random_value<T: Copy>(rng: &mut impl rand::Rng, variants: &[T]) -> T {
 
 fn random_date_time(rng: &mut impl rand::Rng) -> DateTime<Local> {
     let now = Local::now().timestamp();
-    let ten_years = 10 * 365 * 24 * 60 * 60;
-    let secs = rng.random_range(now..(now + ten_years));
+    let about_ten_years = 10 * 365 * 24 * 60 * 60;
+    let date_range = now - about_ten_years..now + about_ten_years;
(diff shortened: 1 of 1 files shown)


stacktraceghost (dismissed):

backend/src/pdf_gen/typst_tests.rs
```diff
@@ -47,30 +47,22 @@ fn random_value<T: Copy>(rng: &mut impl rand::Rng, variants: &[T]) -> T {
...
-    NaiveDate::from_ymd_opt(
-        rng.random_range(1900..=2100),
-        rng.random_range(1..=12),
-        rng.random_range(1..=28),
-    )
-    .unwrap()
+fn random_date_time(rng: &mut impl rand::Rng) -> DateTime<Local> {
+    let now = Local::now().timestamp();
+    let ten_years = 10 * 365 * 24 * 60 * 60;
+    let secs = rng.random_range(now..(now + ten_years));
```
rnijveld: We should also generate dates in the past because those would normally happen pretty frequently when using abacus

rnijveld (changes requested):

backend/src/pdf_gen/typst_tests.rs
marlonbaeten: I doubt that will add anything in these tests, since the date is simply formatted as a string and then rendered in the template, but since you never known where this function might be reused in the future I have change the range in [97c7e86](https://github.com/kiesraad/abacus/pull/2446/commits/97c7e86b8cc1a25fd026df77fbb29f5ead38276c)

Issue #2448, opened 2025-11-17
Show warnings for definitive results in read only view


oliver3:
Bug: when the status of the data entry is `Definitive`, the `data_entries/get` endpoint does not return validation warnings.

PR #2450, opened 2025-11-17
Show warnings for definitive results in read only view

oliver3:
- Fixes https://github.com/kiesraad/abacus/issues/2448

First commit:
- New trait `ValidateRoot` with default impl for `start_validate()`, and add to `DataEntryStatus` to replace `validate_data_entry_status()` (as proposed by @rnijveld, thanks!)
- Add `ValidateRoot` to `PollingStationResults` and remove creating `FirstEntryInProgress` structs just to validate data entry

Second commit:
- Get and validate results for state `Definitive` for showing warnings in read only view and add tests

### DIFF
--- a/backend/src/data_entry/api.rs
+++ b/backend/src/data_entry/api.rs
@@ -759,15 +759,19 @@ async fn polling_station_data_entry_get(
                 validation_results: ValidationResults::default(),
             }))
         }
-        DataEntryStatus::Definitive(_) => Ok(Json(DataEntryGetResponse {
-            user_id: None,
-            data: get_result(&mut conn, polling_station_id, committee_session.id)
-                .await?
-                .data
-                .0,
-            status: state.status_name(),
-            validation_results: state.start_validate(&polling_station, &election)?,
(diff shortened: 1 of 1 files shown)


backend/src/data_entry/validation.rs
```diff
@@ -226,9 +226,25 @@ pub fn validate_data_entry_status(
     data_entry_status: &DataEntryStatus,
     polling_station: &PollingStation,
     election: &ElectionWithPoliticalGroups,
+) -> Result<ValidationResults, DataError> {
+    validate(data_entry_status, polling_station, election)
```
rnijveld: We should turn these two functions and the validate function below into a trait (i.e. something like `trait ValidatePrimary: Validate { fn start_validate(&self, polling_station, election) }` with a default implementation containing the contents of the validate function below and then just add `impl ValidatePrimary for DateEntryStatus {}` and `impl ValidatePrimary for PollingStationResults {}` (as discussed privately on mattermost)

PR #2452, opened 2025-11-17
Restrict polling station create, update and delete for administrators

### LINKED ISSUE #2430: Administrator should not be allowed to change polling stations once the data entry phase has started
Found by @jschuurk-kr and me during a documentation update

> Precondities:

>    De stembureaus zijn niet tijdens [het inrichten van de applicatie](https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/beheer.md#de-beheerder-richt-de-applicatie-in-voor-gsb-enof-csb-wolk) ingevoerd.
>    De invoerfase van de zitting is nog niet gestart. Na het starten van de invoerfase kan alleen de [coördinator nog wijzigingen in de lijst met stembureaus aanbrengen](https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/gsb-eerste-zitting.md#de-co%C3%B6rdinator-gsb-bewerkt-de-stembureaus-tijdens-de-eerste-zitting-zee).

See: https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/beheer.md#de-beheerder-of-co%C3%B6rdinator-zet-de-stembureaus-in-de-applicatie-zee

Lionqueen94:
Closes #2430

- Added checks for committee session status when admin for create, update and delete polling station endpoints
- Restrict import polling station endpoint to only be allowed when there are no polling stations (same as in the frontend)
- Added backend tests
- Made cookie naming more specific in all backend integration tests
- Conditional rendering for create button and update links when admin on polling station list page
- Redirect admin from update page to list page if committee session status is not created or not started
- Added frontend tests
- Implemented informational alert:
Screenshot from 2025-11-18 15-19-10

### DIFF
--- a/frontend/src/features/polling_stations/components/PollingStationListPage.tsx
+++ b/frontend/src/features/polling_stations/components/PollingStationListPage.tsx
@@ -11,7 +11,7 @@ import { useUserRole } from "@/hooks/user/useUserRole";
 import { t } from "@/i18n/translate";
 
 import { usePollingStationListRequest } from "../hooks/usePollingStationListRequest";
-import { isPollingStationUpdateAllowed } from "../utils/checks";
+import { isPollingStationCreateAndUpdateAllowed } from "../utils/checks";
 
 export function PollingStationListPage() {
   const { isAdministrator, isCoordinator } = useUserRole();
@@ -27,7 +27,11 @@ export function PollingStationListPage() {
   }
 
   const data = requestState.data;
-  const updateAllowed = isPollingStationUpdateAllowed(isCoordinator, isAdministrator, currentCommitteeSession.status);
+  const createAndUpdateAllowed = isPollingStationCreateAndUpdateAllowed(
+    isCoordinator,
+    isAdministrator,
+    currentCommitteeSession.status,
+  );
 
   const labelForPollingStationType = {
     FixedLocation: t("polling_station.type.FixedLocation"),
@@ -46,7 +50,7 @@ export function PollingStationListPage() {
 
       <Messages />
 
-      {!updateAllowed && (
+      {!createAndUpdateAllowed && (
         <Alert type="notify">
           <strong className="heading-md">{t("polling_station.edit_not_allowed_alert.title")}</strong>
           <p>{t("polling_station.edit_not_allowed_alert.description")}</p>
@@ -72,7 +76,7 @@ export function PollingStationListPage() {
           </article>
         ) : (
           <article>
-            {updateAllowed && (
+            {createAndUpdateAllowed && (
               <Toolbar>
                 <Button.Link variant="secondary" size="sm" to="./create">
                   <IconPlus /> {t("polling_station.create")}
@@ -88,7 +92,7 @@ export function PollingStationListPage() {
               </Table.Header>
               <Table.Body className="fs-md">
                 {data.polling_stations.map((station) =>
-                  updateAllowed ? (
+                  createAndUpdateAllowed ? (
                     <Table.LinkRow key={station.id} to={`${station.id}/update`}>
                       <Table.NumberCell>{station.number}</Table.NumberCell>
                       <Table.Cell className="break-word">{station.name}</Table.Cell>
--- a/frontend/src/features/polling_stations/utils/checks.test.ts
+++ b/frontend/src/features/polling_stations/utils/checks.test.ts
@@ -2,29 +2,34 @@ import { describe, expect, test } from "vitest";
 
 import { CommitteeSessionStatus } from "@/types/generated/openapi";
 
-import { isPollingStationUpdateAllowed } from "./checks";
+import { isPollingStationCreateAndUpdateAllowed } from "./checks";
 
-describe("Test isPollingStationUpdateAllowed", () => {
+describe("Test isPollingStationCreateAndUpdateAllowed", () => {
   test.each([
     { isCoordinator: true, isAdministrator: false, status: "created", expected: true },
     { isCoordinator: false, isAdministrator: true, status: "created", expected: true },
+    { isCoordinator: false, isAdministrator: false, status: "created", expected: false },
     { isCoordinator: true, isAdministrator: false, status: "data_entry_not_started", expected: true },
     { isCoordinator: false, isAdministrator: true, status: "data_entry_not_started", expected: true },
+    { isCoordinator: false, isAdministrator: false, status: "data_entry_not_started", expected: false },
     { isCoordinator: true, isAdministrator: false, status: "data_entry_in_progress", expected: true },
     { isCoordinator: false, isAdministrator: true, status: "data_entry_in_progress", expected: false },
+    { isCoordinator: false, isAdministrator: false, status: "data_entry_in_progress", expected: false },
     { isCoordinator: true, isAdministrator: false, status: "data_entry_paused", expected: true },
     { isCoordinator: false, isAdministrator: true, status: "data_entry_paused", expected: false },
+    { isCoordinator: false, isAdministrator: false, status: "data_entry_paused", expected: false },
     { isCoordinator: true, isAdministrator: false, status: "data_entry_finished", expected: true },
     { isCoordinator: false, isAdministrator: true, status: "data_entry_finished", expected: false },
+    { isCoordinator: false, isAdministrator: false, status: "data_entry_finished", expected: false },
   ] satisfies Array<{
     isCoordinator: boolean;
     isAdministrator: boolean;
     status: CommitteeSessionStatus;
     expected: boolean;
   }>)(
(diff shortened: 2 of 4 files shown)


jorisleker:
> TODO: Hide create, delete and update buttons in the frontend? @jorisleker

Okay, here's what to do. Once the data entry has started, if the admin goes to the polling station list page:
- hide the create polling station button
- remove the links to the polling station form + hide the chevron
- if an admin directly navigates to the url of a polling station, redirect to the list page.

If time allows:
- add a blue informational alert at the top of the polling station list page with this text:
> De invoerfase van de eerste zitting is al gestart. Na het starten van de invoerfase kan alleen een coördinator de lijst met stembureaus nog bewerken.

Lionqueen94:
> > TODO: Hide create, delete and update buttons in the frontend? @jorisleker
> 
> Okay, here's what to do. Once the data entry has started, if the admin goes to the polling station list page:
> 
>     * hide the create polling station button
> 
>     * remove the links to the polling station form + hide the chevron
> 
>     * if an admin directly navigates to the url of a polling station, redirect to the list page.
> 
> 
> If time allows:
> 
>     * add a blue informational alert at the top of the polling station list page with this text:
> 
> 
> > De invoerfase van de eerste zitting is al gestart. Na het starten van de invoerfase kan alleen een coördinator de lijst met stembureaus nog bewerken.

Implemented :smile: :
Screenshot from 2025-11-18 15-19-10

backend/src/polling_station/mod.rs
```diff
@@ -263,6 +277,13 @@ async fn polling_station_delete(
     crate::election::repository::get(&mut tx, election_id).await?;
     let committee_session = get_election_committee_session(&mut tx, election_id).await?;
 
+    if !user.is_coordinator()
```
stacktraceghost: API-wise it would still be possible to call `polling_station_import` as admin outside of the given statuses. We could add the check there as well.

To prevent typos and code repetition we could also add a function called for example `validate_access(user, committee_session.status)?`) and call that for every (relevant) function.

Lionqueen94: I moved the validation into a separate function in ed9a917be8490d2eb2cc83b310230f2094ee675e and restricted the import function to only be used when no polling stations exist (as it is only shown then in the frontend as well). This automatically means the session status is created so I think there's no need to explicitly check this as well (as wel cannot make a fixture that is in a later status without polling stations).

frontend/src/features/polling_stations/components/PollingStationListPage.test.tsx
```diff
@@ -64,14 +84,79 @@ describe("PollingStationListPage", () => {
...
+        ["40", "Test kerk", "Vaste locatie"],
+      ]);
+
+      const tableRows = within(table).queryAllByRole("row");
+      await user.click(tableRows[1]!);
+
+      expect(navigate).not.toHaveBeenCalled();
+    });
+
+    test("Shows polling stations without links to update page when committee session status is created", async () => {
```
stacktraceghost: I think this should be with iso without.

You can also do something like:

```ts
  test.each([
    { status: "created", allowed: true },
    { status: "data_entry_not_started", allowed: true },
    { status: "data_entry_in_progress", allowed: false },
    { status: "data_entry_paused", allowed: false },
    { status: "data_entry_finished", allowed: false },
  ] satisfies Array<{ status: CommitteeSessionStatus; allowed: boolean }>)(
    "Polling station update links with status=%s are allowed=%s",
    ({ status, allowed }) => {
      // ...
    },
  );
```

This prevents repetition and tests all statuses. Same for `PollingStationUpdatePage.test.tsx`

Lionqueen94: Good improvement, implemented in dbff063713b053ac7eca745bbb23afacf92fb74e

frontend/src/features/polling_stations/components/PollingStationListPage.tsx
```diff
@@ -23,6 +26,10 @@ export function PollingStationListPage() {
   }
 
   const data = requestState.data;
+  const editAllowed =
+    isCoordinator ||
+    (isAdministrator &&
+      (currentCommitteeSession.status === "created" || currentCommitteeSession.status === "data_entry_not_started"));
```
lama-d3v: I would create a helper/util function for this, so that you can re-use this in the PollingStationUpdatePage component as well.

Lionqueen94: Added in b3c03b5266c5225a811ec6e211f99d9c2d72d24d

frontend/src/features/polling_stations/components/PollingStationListPage.tsx
```diff
@@ -23,6 +27,7 @@ export function PollingStationListPage() {
   }
 
   const data = requestState.data;
+  const updateAllowed = isPollingStationUpdateAllowed(isCoordinator, isAdministrator, currentCommitteeSession.status);
```
lama-d3v: Is it okay that the frontend also determines what the user is allowed to do? Right now you have to maintain this code in both the frontend and backend.
I think it would be nicer to only let the backend decide which actions the user is allowed to execute.
It's probably not feasible to add this for 1.0, but I do think this would be a nice improvement for future development.

lama-d3v: You could add the alert like this here, because `<Messages />` already renders all alerts.

```
  if (updateAllowed) {
    pushMessage({
      type: "notify",
      title: t("polling_station.edit_not_allowed_alert.title"),
      text: t("polling_station.edit_not_allowed_alert.description"),
    });
  }
```

frontend/src/features/polling_stations/components/PollingStationListPage.tsx
```diff
@@ -60,11 +72,13 @@ export function PollingStationListPage() {
           </article>
         ) : (
           <article>
-            <Toolbar>
-              <Button.Link variant="secondary" size="sm" to="./create">
-                <IconPlus /> {t("polling_station.create")}
-              </Button.Link>
-            </Toolbar>
+            {updateAllowed && (
```
lama-d3v: Maybe rename the variable to e.g. `createOrUpdateAllowed`, since this determines whether the user is allowed to create and/or update?

frontend/src/features/polling_stations/utils/checks.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    { isCoordinator: true, isAdministrator: false, status: "created", expected: true },
+    { isCoordinator: false, isAdministrator: true, status: "created", expected: true },
+    { isCoordinator: true, isAdministrator: false, status: "data_entry_not_started", expected: true },
+    { isCoordinator: false, isAdministrator: true, status: "data_entry_not_started", expected: true },
+    { isCoordinator: true, isAdministrator: false, status: "data_entry_in_progress", expected: true },
+    { isCoordinator: false, isAdministrator: true, status: "data_entry_in_progress", expected: false },
+    { isCoordinator: true, isAdministrator: false, status: "data_entry_paused", expected: true },
+    { isCoordinator: false, isAdministrator: true, status: "data_entry_paused", expected: false },
+    { isCoordinator: true, isAdministrator: false, status: "data_entry_finished", expected: true },
+    { isCoordinator: false, isAdministrator: true, status: "data_entry_finished", expected: false },
```
lama-d3v, line 22: Perhaps also add test case for `isCoordinator: false, isAdministrator: false`?

frontend/src/features/polling_stations/components/PollingStationListPage.tsx
Lionqueen94: @praseodym What do you think?

Lionqueen94: This does not work, because you push a message to be rendered on the next page you redirect to. For example when you save the update page, you push a message and it shows on the list page you redirect to

frontend/src/features/polling_stations/components/PollingStationListPage.tsx
Lionqueen94: Done in 4075095beeddae049bb73b71ac25e96a5b489739

frontend/src/features/polling_stations/utils/checks.test.ts
Lionqueen94: Done in 4075095beeddae049bb73b71ac25e96a5b489739

oliver3 (approved):

frontend/src/features/polling_stations/components/PollingStationListPage.tsx
oliver3: I would say this is something for after 1.0

praseodym: Good to think about this for after 1.0, but not feasible to implement such a feature for 1.0

PR #2453, opened 2025-11-17
Handle windows close and shutdown signals

marlonbaeten:
Previously we only listened for an interrupt to gracefully shutdown Abacus on Windows. This adds signal listeners for "close" and "shutdown" on windows and adds the "quit" signal on *unix.

I have painstakingly tested this on a Windows VM.

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -320,17 +320,45 @@ pub async fn shutdown_signal() {
     #[cfg(unix)]
     let terminate = async {
         signal::unix::signal(signal::unix::SignalKind::terminate())
-            .expect("failed to install signal handler")
+            .expect("failed to install terminate signal handler")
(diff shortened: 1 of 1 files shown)


rnijveld (approved):

stacktraceghost:
Tested on Windows with a release build:
- ✅ Running abacus.exe and closing with X -> wal/shm files gone
- ✅ Running abacus.exe and stopping process with ctrl+c -> wal/shm files gone
- ✅ From cmd.exe, running abacus.exe and closing cmd with X -> wal/shm files gone
- ✅ From cmd.exe, running abacus.exe and stopping process with ctrl+c -> wal/shm files gone
- When killing abacus.exe through task manager, the wal/shm files stay there, since it was not a graceful shutdown. This is expected behavior

PR #2474, opened 2025-11-19
Changed order for corrected results answers (#2459)

### LINKED ISSUE #2459: Order of answers to "Is er een gecorrigeerde uitkomst?" is different
Abacus has as answers to "Is er een gecorrigeerde uitkomst?" first "yes" then "no":

Image

On paper it's the other way around:

Image

Design and code should be updated to reflect the order on paper. Copy in Abacus should remain as is.

lama-d3v:
Closes #2459


stacktraceghost (approved):

jorisleker:
> Please also update this on Figma as well https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=12660-52449&m=dev (and the 2 pages next to it)

done

PR #2480, opened 2025-11-19
Do not show delete polling station button when data entry exists

### LINKED ISSUE #2438: Confirm deleting of polling station when data entry or investigation exists
**This is not in scope for 1.0.** because too much work, see first note below.

### Scope

Design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=16743-21506&m=dev

Image

Note:
- this needs some good thinking about a typist starting or discarding data entry while the coordinator is on the page

Lionqueen94:
**First merge #2452 and then rebase onto main**

Closes #2438 

- Replaced delete button with link to data entry in case data entry exists for a polling station

N.B. Separate administrator copy is not needed because they cannot see this page when the data entry has started


Lionqueen94:
Closed for now, to be reopened after the deadline

PR #2481, opened 2025-11-19
OpenAPI security schemes / Swagger roles doc

stacktraceghost:
Adds security requirement information to each endpoint in Swagger: 
- It shows if authentication necessary (lock icon)
- It shows the required roles at the end of the title

image

To validate the documented security requirement, `authorization_test.rs` has been adjusted to test every endpoint to see if the implementation matches the documentation.
- If no `security` attribute has been added, we consider the endpoint to be public. A non-authenticated user and all roles should have access.
- If a `security` attribute is there, we check if the given roles have access and the other ones (including non-authenticated user) do not.

The test also makes sure the given roles actually exist to prevent typos.

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -1116,6 +1116,7 @@
     },
     "/api/elections/{election_id}/download_n_10_2": {
       "get": {
+        "summary": "administrator, coordinator",
         "operationId": "election_download_n_10_2",
         "parameters": [
           {
@@ -1195,7 +1196,15 @@
               }
             }
           }
-        }
+        },
+        "security": [
+          {
+            "cookie_auth": [
+              "administrator",
+              "coordinator"
+            ]
+          }
+        ]
       }
     },
     "/api/elections/{election_id}/download_na_31_2_bijlage1": {
--- a/backend/src/document/api.rs
+++ b/backend/src/document/api.rs
@@ -49,6 +49,7 @@ pub fn router() -> OpenApiRouter<AppState> {
     params(
         ("election_id" = u32, description = "Election database id"),
     ),
+    security(("cookie_auth" = ["administrator", "coordinator"])),
 )]
 async fn election_download_n_10_2(
     _user: AdminOrCoordinator,


backend/src/lib.rs
```diff
@@ -78,11 +85,80 @@ pub struct AppState {
     airgap_detection: AirgapDetection,
 }
 
+pub fn get_scopes_from_operation(operation: &Operation) -> Option<Vec<String>> {
```
marlonbaeten, line 88: `lib.rs` is already way too large. I would suggest move your additions to a separate file.

backend/tests/authorization_test.rs
```diff
@@ -36,33 +50,89 @@ async fn test_route_authorization(pool: SqlitePool) {
...
+                        role
+                    )),
+                    _ => {}
+                }
             }
         }
     }
+
+    if !failures.is_empty() {
+        eprintln!("\n=== Authorization test failures ===");
```
marlonbaeten: It's better to try to wrap this into an assertion, since the error output is handled better. Something like this (untested):
```rust
assert!(failures.is_empty(), "Authorization test failures ({}):\n{}" failures.len(), failures.join("\n"));
```

marlonbaeten (dismissed):

marlonbaeten:
I tested that the tests catch authorization errors and misconfigurations and they do!

backend/src/lib.rs
praseodym: Agreed but I wouldn't do this before the 1.0 release

stacktraceghost: We can pick this up when we continue with https://github.com/kiesraad/abacus/pull/2426

backend/tests/authorization_test.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2481/commits/d72284f16d9a928b8eab6a8e938c41837b02b86e

oliver3 (dismissed):
Nice!

Lionqueen94 (dismissed):

praseodym (approved):

PR #2482, opened 2025-11-19
set VotingMethod value to SVP in EML 110b test files

jschuurk-kr:
As @chrismostert pointed out in [this documentation PR](https://github.com/kiesraad/abacus-documentatie/pull/19#pullrequestreview-3478252627), `VotingMethod` in the EML 110b should have a value. So I updated our EML 110b test files accordingly.


jschuurk-kr:
> It should be SPV, not SVP 😉

:man_facepalming: Time for me to call it a day. Will try again in a [fresh PR](https://github.com/kiesraad/abacus/pull/2483).

Issue #2475, opened 2025-11-19
After logging out, you still get the modal about session expiration


jschuurk-kr:
### Steps to reproduce
1. Log in
2. Log out
3. Wait

### Investigation
@oliver3 looked into this and came to the conclusion that the problem is in useSessionState. User is set to null, but the expiration isn't.

PR #2491, opened 2025-11-20
Do not show "Je wordt bijna uitgelogd" modal after logout

oliver3:
- Fixes https://github.com/kiesraad/abacus/issues/2475
- Check if the user is null when deciding to show the modal


marlonbaeten (changes requested):
I like the introduction of the `clearSession` function. For this specific bug a simple conditional whether the user is set before showing the dialog might have also fixed it. So maybe we also should check whether `user !== null` before showing the dialog.

oliver3:
Fair enough, I'll revert this and do the extra check at the modal itself.

PR #2497, opened 2025-11-20
Provide empty inlegvel Na 31-2 for corrigendum as additional doc in a second session

### LINKED ISSUE #2489: Provide empty Inlegvel Na 31-2 bij corrigendum as additional doc in a second session
If you have a second session, you're expected to add an ["inlegvel"](https://github.com/user-attachments/files/23650262/Na.31-2.inlegvel.pdf) to the PV of the first session, i.e. to the Na 31-2. (Note that there is no "inlegvel" to add to the PV of your second session if you have a third session.)

So on the page with the generated documents for the second session (and only the second session) we should provide a download link to the empty "inlegvel". Empty as in: fill in the "gemeentecode" and "gemeentenaam" in the header, but the body of the document needs to be filled in by hand.

If this is too much work before our deadline, we could down-scope this to only updating the copy of the page to remind users of the "inlegvel". This does increase the risk that users will not actually insert the "inlegvel" into the Na 31-2, though. So preferably we don't down-scope.

While it is possible for Abacus to fill in the "inlegvel" (check the correct box and add the date and time), this is __not in scope__ for this issue.

(implementation of https://github.com/kiesraad/abacus-internal/issues/333)

Lionqueen94:
Closes #2489 

[Model_Na_31_2_Inlegvel.pdf](https://github.com/user-attachments/files/23665726/Model_Na_31_2_Inlegvel.pdf)

Screenshot from 2025-11-20 17-04-13

### DIFF
--- a/frontend/e2e-tests/page-objects/election/ElectionHomePgObj.ts
+++ b/frontend/e2e-tests/page-objects/election/ElectionHomePgObj.ts
@@ -10,6 +10,7 @@ export class ElectionHome {
   readonly pollingStationsRow: Locator;
   readonly downloadBijlage1: Locator;
   readonly downloadN10_2: Locator;
+  readonly downloadInlegvel: Locator;
 
   constructor(protected readonly page: Page) {
     this.header = page.getByRole("heading", { level: 1 });
@@ -21,6 +22,7 @@ export class ElectionHome {
     this.pollingStationsRow = page.getByRole("rowheader", { name: "Stembureaus" });
     this.downloadBijlage1 = page.getByRole("cell", { name: "Na 31-2 Bijlage 1" });
     this.downloadN10_2 = page.getByRole("cell", { name: "N 10-2" });
+    this.downloadInlegvel = page.getByRole("cell", { name: "Na 31-2 Inlegvel" });
   }
 
   getCommitteeSessionCard(number: number) {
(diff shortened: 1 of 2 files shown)


backend/src/document/api.rs
```diff
@@ -188,3 +192,48 @@ async fn election_download_n_10_2(
...
+        (status = 403, description = "Forbidden", body = ErrorResponse),
+        (status = 404, description = "Not found", body = ErrorResponse),
+        (status = 409, description = "Request cannot be completed", body = ErrorResponse),
+        (status = 500, description = "Internal server error", body = ErrorResponse),
+    ),
+    params(
+        ("election_id" = u32, description = "Election database id"),
+    ),
+)]
+async fn election_download_na_31_2_inlegvel(
```
lama-d3v, line 218: download_na_31_2_inlegvel should only be allowed in a second session. Right now this is allowed in all sessions.

Lionqueen94: We didn't restrict the endpoint for the other empty documents either, so I thought it would not be needed for this one as well. @praseodym What do you think? In the frontend the download links are only available in the respective committee sessions

praseodym: It’s fine to not limit it in the backend

PR #2498, opened 2025-11-20
Updated footer

### LINKED ISSUE #2490: Version name + semver bottom right on screen
Currently we show a commit hash on the right bottom corner with a link to Github, which can't be used on an airgapped system.

Image

We would like to show the release version name + number there, example: "Abacus GR26 1.0.0"

Lionqueen94:
Closes #2490 

Showing server, branch and commit only in dev mode:
image


marlonbaeten:
Tested with `SHOW_DEV_PAGE=false npm run dev`

PR #2501, opened 2025-11-21
Adjusted filenames according to what overdrachtsplatform expects

### LINKED ISSUE #2473: Update names of generated files to match what's expected by the "overdrachtsplatform"
The current name of the zip file you download with the results of the first session is "election_result_GR2025_Juinen.zip" (The zip file of a next session has exactly the same name, this will be fixed in https://github.com/kiesraad/abacus/issues/2405.)

For the 1.0 release we should at least have unique names in Dutch for all files you can download. Ideally the file names match what's expected by the "overdrachtsplatform".

Lionqueen94:
Closes #2473

- Adjusted filenames according to what overdrachtsplatform expects
- Added created_at field to files database rows and return this datetime in the zip filename
- Added document naming documentation


backend/src/report/api.rs
```diff
@@ -262,21 +246,48 @@ struct PdfModelList {
...
     use chrono::Datelike;
+    let mut location = election.location.clone();
+    location.retain(|c| c != ' ');
     slugify_filename(&format!(
-        "election_result_{}{}_{}.zip",
+        "Telling_{}{}_{}.zip",
         election.category.to_eml_code(),
         election.election_date.year(),
-        election.location
+        location
```
bramtweedegolf: According to the issue I think the name of the zip file "Telling_GR2026_gemeentenaam.zip" should have the location uncapitalized. (Whereas the EML file within should have the location capitalized)
 
So maybe add a .to_lowercase() here?

Lionqueen94: I made some screenshots of the files generated by OSV that are present in the test files folder of the OSV test team and there the zip with the telling contains a capitalised gemeente naam, see https://github.com/kiesraad/abacus/issues/2473#issuecomment-3562326061

bramtweedegolf (dismissed):

backend/src/report/api.rs
```diff
@@ -262,21 +246,48 @@ struct PdfModelList {
...
-fn zip_filename(election: ElectionWithPoliticalGroups) -> String {
+fn download_zip_filename(
+    election: &ElectionWithPoliticalGroups,
+    committee_session: &CommitteeSession,
+    created_at: DateTime<Local>,
+) -> String {
+    use chrono::Datelike;
+    let location = election.location.clone().to_lowercase();
+    let mut location_without_whitespace = location.clone();
+    location_without_whitespace.retain(|c| c != ' ');
```
marlonbaeten: It looks like
```
        location_without_whitespace,
        location.clone().replace(" ", "-"),
```
duplicates the location name?

Also `slugify_filename` already replaces spaces with `_`.

marlonbaeten: If we need to remove all white-space (instead of replacing spaces to `_`) an approach that does not require a mutable variable would be `.split_whitespace().collect()`.

backend/src/files/mod.rs
```diff
@@ -14,6 +15,8 @@ pub struct File {
     pub data: Vec<u8>,
     pub name: String,
     pub mime_type: String,
+    #[schema(value_type = String)]
+    pub created_at: DateTime<Utc>,
```
praseodym, line 19: Why do we use `DateTime<Local>` everywhere but then `DateTime<Utc>` here?

Lionqueen94: In the database we always save everything in Utc I thought, so to keep it consistent I did that here as well

backend/src/report/api.rs
```diff
@@ -292,28 +303,32 @@ async fn generate_and_save_files(
...
         || committee_session.start_date_time.is_none()
-        || !are_results_complete_for_committee_session(&mut conn, committee_session_id).await?
+        || !are_results_complete_for_committee_session(&mut conn, committee_session.id).await?
     {
         return Err(APIError::CommitteeSession(
             CommitteeSessionError::InvalidCommitteeSessionStatus,
         ));
     }
 
+    let mut created_at = Local::now();
```
marlonbaeten: I think it would be better to generate a `Utc::now()` here (and on line 349).
You could use `.with_timezone(&Local)` to convert the UTC date-time to a local date-time, but only for the places where you really need a Local date-time (like `ResultsInput::new`).

marlonbaeten (changes requested):
See comments.

backend/src/files/mod.rs
marlonbaeten: Yes - the database always uses `DateTime<Utc>` - we only need the `DateTime<Local>` to print a local date in the footer of our Typst models. Also see my comment below.

backend/src/report/api.rs
marlonbaeten: I see in https://github.com/kiesraad/abacus/issues/2473#issuecomment-3562326061 that the location name actually is duplicated....

backend/src/files/mod.rs
Lionqueen94: Made changes in 6591a614a4159a0fcfb75f21bdc2563dbc10fbe1

backend/src/report/api.rs
Lionqueen94: Fixed in 6591a614a4159a0fcfb75f21bdc2563dbc10fbe1

backend/src/report/api.rs
Lionqueen94: yes the location name is twice in the filename in 2 different formats
I made some improvements in 6591a614a4159a0fcfb75f21bdc2563dbc10fbe1

backend/migrations/9_files.sql
```diff
@@ -3,5 +3,6 @@ CREATE TABLE files
     id               INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
     data             BLOB                              NOT NULL,
     name             TEXT                              NOT NULL,
-    mime_type        TEXT                              NOT NULL
+    mime_type        TEXT                              NOT NULL,
+    created_at       TEXT                              NOT NULL
```
lama-d3v, line 7:
```suggestion
    created_at       TEXT                              NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
maybe set the default value here?

lama-d3v (changes requested):

praseodym:
I think adding some brief documentation about what the file structure looks like (e.g. in a docstring or in `documentatie/use-cases/input-output-bestanden.md`) would be useful.

backend/migrations/9_files.sql
Lionqueen94: I didn't do this on purpose, because the value needs to be set by the code. I didn't want to make a mistake and have a fallback value that makes the mistake invisible. The generated files all need the same created date determined by the code.

PR #2504, opened 2025-11-21
Fixed navbar height and button size in finish data entry alert on investigations page

### LINKED ISSUE #2488: Navigation bigger when breadcrumb is present
See screenshot. When the breadcrumb is rendered, the navigation bar is slightly higher.

Image

Also notice the overflow in the error message on very narrow viewports

Lionqueen94:
Closes #2488 

- Fixed navbar height (changed `height: 3rem` to `min-height: 3rem`)
- Fixed error message height (changed `height: 4.5rem` to `min-height: 4.5rem`)
- Adjusted button size in finish data entry alert on investigations page (was "sm", should be "md", matching a similar alert on the election status page)

### DIFF
--- a/frontend/src/features/data_entry_home/components/PollingStationChoice.module.css
+++ b/frontend/src/features/data_entry_home/components/PollingStationChoice.module.css
@@ -24,7 +24,7 @@
 .message {
   padding: 1rem 1.5rem;
   background-color: var(--bg-gray-darker);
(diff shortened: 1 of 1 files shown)


stacktraceghost:
The overflow issue mentioned in the linked issue is still there. I think a `height` has to be changed to `min-height` there.

image

PR #2526, opened 2025-11-25
Add Linux systemd unit file and instructions

### LINKED ISSUE #2512: Create Linux systemd unit file
For Linux deployments we should create a systemd unit file so that Abacus can be run with proper permissions and hardening.

If time allows we can use the systemd unit file as a part of packaging for Linux.

praseodym:
Resolves #2512

### DIFF
--- a/packaging/linux/README.md
+++ b/packaging/linux/README.md
@@ -1,5 +1,10 @@
 # Abacus deployment on Linux with systemd
 
+Prerequisites:
+- A Linux system with systemd
+- The Abacus binary (named `abacus` below), e.g. from the [releases page](https://github.com/kiesraad/abacus/releases)
+- The Abacus systemd unit file (`abacus.service`), from this directory
+
 Copy the Abacus binary and service file to the system, reload systemd, enable the Abacus service to start on system boot and start it right away:
 ```
 sudo cp abacus /usr/local/bin/abacus


marjoleintamis (dismissed):

packaging/linux/README.md
```diff
@@ -0,0 +1,32 @@
+# Abacus deployment on Linux with systemd
+
+Copy the Abacus binary and service file to the system, reload systemd, enable the Abacus service to start on system boot and start it right away:
+```
+sudo cp abacus /usr/local/bin/abacus
+sudo cp abacus.service /etc/systemd/system/abacus.service
```
cikzh, line 11: Do we want to be explicit here about where to find the binary and service file?

For example 
```
sudo cp packaging/linux/abacus.service /etc/systemd/system/abacus.service
```

Tested the instructions and works otherwise

praseodym: The instructions are in the same directory as the unit file, so I thought this would be clear enough.

cikzh: That's fair, but what about the binary? Should we mention they need to be downloaded in the release page? I'm not sure how much hand holding we should do here

cikzh (dismissed):
Open comment is no reason to delay merging in my opinion, so approved. We can change textual details later if we feel inclined to do so

packaging/linux/README.md
praseodym: I've added some pointers in fb28c39

PR #2527, opened 2025-11-25
Add fuzzer for data entry state machine

michielp1807:
This PR adds a fuzzer that tests that the data entry system matches the state machine described in the [docs](https://github.com/kiesraad/abacus/blob/main/documentatie/flowcharts/data-entry-state.md). The fuzzer checks that all possible transitions change the state as expected, or give the expected error. Related to #2213.

The fuzzer covers all possible states and all possible `DataEntryTransitionError` errors, except for the `FirstEntryAlreadyClaimed`, `ValidatorError` and `ValidationError` errors. The `FirstEntryAlreadyClaimed` error is not triggered by the fuzzer for simplicity, but this error is explicitly tested by the `first_entry_in_progress_claim_first_entry_other_user_error` unit test. The `ValidatorError` and `ValidationError` errors are not triggered by the fuzzer, because the fuzzer does not test the API directly. Instead, it only tests the internal functions to focus on verifying the data entry state machine (so this is basically a more complete extension of the unit tests that test individual transitions, which shows that no unexpected errors or transition can happen).

I did not find any real problems, but I did come across one small oddity:
- When in `DataEntryStatus::SecondEntryInProgress`, claiming the first entry gives a `DataEntryTransitionError::Invalid` error instead of `DataEntryTransitionError::FirstEntryAlreadyFinalised`, which I think would be more consistent

This PR replaces #2440, now it's an internal PR so the PDF diff action should no longer fail :)

### DIFF
--- a/backend/fuzz/README.md
+++ b/backend/fuzz/README.md
@@ -15,7 +15,7 @@ rustup install nightly
 cargo +nightly install cargo-fuzz
 ```
 
-Then, run the fuzzer with the following command:
+Then, run the fuzzer with the following command from within the `/backend` folder:
 
 ```
 cargo +nightly fuzz run data_entry_status
@@ -55,7 +55,7 @@ We can use `llvm-cov` to visualize the coverage data. For this, we will need `ru
 cargo +nightly install rustfilt
 ```
 
-Then we can run the following command to output the coverage data to `/tmp/coverage` (thanks [Folkert](https://tweedegolf.nl/en/blog/154/what-is-my-fuzzer-doing#generating-a-coverage-report)!):
+Then we can run the following command from the `/backend` folder to output the coverage data to `/tmp/coverage` (thanks [Folkert](https://tweedegolf.nl/en/blog/154/what-is-my-fuzzer-doing#generating-a-coverage-report)!):
 
 ```
 $(rustc +nightly --print sysroot)/lib/rustlib/$(rustc +nightly --print host-tuple)/bin/llvm-cov show \


backend/fuzz/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+This fuzz test covers all possible states and all possible `DataEntryTransitionError` errors, except for the `FirstEntryAlreadyClaimed`, `ValidatorError` and `ValidationError` errors. The `FirstEntryAlreadyClaimed` error is not triggered by the fuzzer for simplicity, but this error is explicitly tested by the `first_entry_in_progress_claim_first_entry_other_user_error` unit test. The `ValidatorError` and `ValidationError` errors are not triggered by the fuzzer because the fuzzer does not test the API directly. Instead, the fuzzer only tests the internal functions to focus on verifying the data entry state machine. This fuzz test can be seen as a more complete extension of the unit tests that test individual transitions, which shows that no unexpected errors or transition can happen.
+
+Run it with the following command (for fuzzing you need the nightly compiler):
+
+```
+cargo +nightly fuzz run data_entry_status
```
cikzh, line 21: Probably nice to mention that we need to install cargo-fuzz as well

michielp1807: Done in [15466e9](https://github.com/kiesraad/abacus/pull/2527/commits/15466e9087a98a162fefb4d271ec828db2e0e5b5) :+1:

bramtweedegolf (approved):

backend/fuzz/README.md
```diff
@@ -0,0 +1,70 @@
...
+```
+cargo +nightly install rustfilt
+```
+
+Then we can run the following command to output the coverage data to `/tmp/coverage` (thanks [Folkert](https://tweedegolf.nl/en/blog/154/what-is-my-fuzzer-doing#generating-a-coverage-report)!):
+
+```
+$(rustc +nightly --print sysroot)/lib/rustlib/$(rustc +nightly --print host-tuple)/bin/llvm-cov show \
+ target/$(rustc +nightly --print host-tuple)/coverage/$(rustc +nightly --print host-tuple)/release/data_entry_status \
+ -instr-profile=fuzz/coverage/data_entry_status/coverage.profdata \
```
stacktraceghost, line 63:
```suggestion
 -instr-profile=coverage/data_entry_status/coverage.profdata \
```
When running everything from the fuzz folder, this path is not correct. If running from the `/backend`, the target path is not correct. 

Should I have ran all commands from `/backend`? Maybe we should add this to the readme.

bramtweedegolf: Yes, I think all commands should be run from /backend.
That should indeed be added to the README.

michielp1807: It was indeed my expectation to run all commands from `/backend`, but the `cargo fuzz` commands also work from within any of the subfolders. I've now specified in the README that this command should be run from `/backend`.

stacktraceghost: Ah I see you've added it to this command. I falsely assumed that I had to run all commands in `/backend/fuzz` because the folder has its own readme. Maybe we should mention it on the beginning of the readme instead of only at this command. 

Everything works as described when running from `/backend` 👍

michielp1807: Sure, I've now also mentioned for the first cargo fuzz command that it should run from within the `/backend` folder. In practice you can also run the cargo fuzz commands from the fuzz folder and it will work perfectly fine, but it is probably less confusing if you just run everything from `/backend`. I've also briefly looked into if we can change the `llvm-cov show` command to also work from anywhere within the `/backend` folder, but I don't think there is an easy way to do this, and this command is already quite big :smile:

stacktraceghost: Great, thank you!

PR #2523, opened 2025-11-24
Create InnoSetup script

stacktraceghost:
Resolves #2518

This PR adds an InnoSetup file to build a Windows installer for Abacus. It also includes some assets. 

Based on comments/feedback in https://github.com/kiesraad/abacus/issues/1096 and [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=1168-33305&m=dev)

Todo:
- ~Add screenshots~ // See below
- ~Add higher resolution asset~ // [done](https://github.com/kiesraad/abacus/compare/b9fab9e1dd800a36dbe9eaa25d7a0696891dfedb..5045744f2b6d31780f1bc254f6f39861ee6447c0)
- ~Add readme describing how to use this and what to add to the folder to build~ // [done](https://github.com/kiesraad/abacus/pull/2523/commits/f39036e74c8121a27b62b5f6807ebcd3b6aef3ff)
- ~Take a final look at "Keep database" dialog~ // keep as it is

-----

<details>
  <summary>Screenshots install</summary>

  **Install 01: Click to continue**
  

  **Install 02: Installing**

  **Install 03: UAC vc_redist**

  **Install 04: Installing vc_redist**

  **Install 05: UAC netsh firewall**

  **Install 06a: DENIED UAC retry message**

  **Install 06b: ALLOWED UAC success message**

  **Install 07: Finish**

</details>

<details>
  <summary>Desktop shortcuts</summary>

  image
  
  These get removed after uninstall.
</details>

<details>
  <summary>Screenshots uninstall</summary>

  **Uninstall 01: UAC**

  **Uninstall 02: Are you sure**

  **Uninstall 03: Delete database?**
Note: this question only appears of there is a database file (`db.sqlite`).

  **Uninstall 04a: Delete database? -> delete database**

  **Uninstall 04b: Delete database? -> keep**

  **Uninstall 05: UAC netsh firewall**

  **Uninstall 06a: DENIED UAC retry message**

  **Uninstall 07b: ALLOWED UAC success message**

  **Uninstall 08: Uninstalling**

  **Uninstall 09: Done**

</details>

PR #2532, opened 2025-11-26
Create InnoSetup script: signing params

stacktraceghost:
First merge https://github.com/kiesraad/abacus/pull/2523

This PR adds signing params to the setup script with instructions. When activated, it will sign:
- The installer
- The uninstaller
- The given .exe (abacus.exe)

If the given .exe already has a certificate, it will be overwritten (due to the `sign` flag). If the `SignTool` param is commented, it will not be signed (even though the flag is there).


praseodym:
This will need to be rebased onto main after #2523 has been merged

praseodym (approved):

PR #2533, opened 2025-11-26
Create GitHub Actions release workflow

### LINKED ISSUE #2514: Create GitHub Actions release workflow
Our [current release workflow](https://github.com/kiesraad/abacus/blob/c08ec8a17fc08a6b11e5e0b4051cf5cb54ec6de5/.github/workflows/release.yml) creates a alpha tag and prerelease. 

For an actual release, we need to create a tag and build using that tag. We could either create the tag manually and run the workflow on that tag, or create a workflow with an [input parameter](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_callinputs) that can create the tag, build everything and create a prerelease. We can then update the release with signed binaries to make it definitive (uncheck that it is a prerelease).

praseodym:
Resolves #2514

- Change release workflow to use existing tag
- Remove macOS from release workflow
- Create tarball for Linux

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -157,11 +157,18 @@ jobs:
         with:
           pattern: abacus-release-*
           path: release/
-      - name: Rename binary
+      - name: Create Linux tarball
         run: |
-          mv abacus-release-ubuntu-latest/abacus abacus-linux-${{ github.ref_name }}
+          mkdir -p abacus-linux-${{ github.ref_name }}
+          mv abacus-release-ubuntu-latest/abacus abacus-linux-${{ github.ref_name }}/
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,10 +1,9 @@
 name: Release
 
 on:
-  workflow_dispatch:
-  # Run every Sunday at 12:00 UTC
-  schedule:
-    - cron: "0 12 * * 0"
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
```
lama-d3v, line 6: Does this syntax for tags needs to be documented somewhere?

praseodym: I will write some release process documentation later this week!

stacktraceghost (approved):

Issue #2414, opened 2025-11-10
Check fixtures / state


ring-ring-ring:
Nav #2400 andere fixtures nalopen op correcte state

jschuurk-kr:
@stacktraceghost @cikzh Can we close this one, once both PRs #2422 and #2484 have been merged?

stacktraceghost:
> [@stacktraceghost](https://github.com/stacktraceghost) [@cikzh](https://github.com/cikzh) Can we close this one, once both PRs [#2422](https://github.com/kiesraad/abacus/pull/2422) and [#2484](https://github.com/kiesraad/abacus/pull/2484) have been merged?

No, the other fixtures should still be checked if they have a valid committee session state. Those PRs cover fixture 5 and 7.

stacktraceghost:
I've checked the other fixtures and I've found that:
- `committee_session.number_of_votes` is not set for fixture 1, 2 and 3.
  - Since `committee_session.number_of_votes` can be and stay 0, that does not seem like a problem to me.
- The entry/result dates (`updated_at` + JSON) are not realistic in fixture 5 and 7. They happen at the same time as other records or results are before entries. 
  - This is not causing any issues. But I've created a small PR to adjust it: https://github.com/kiesraad/abacus/pull/2562

Otherwise, the committee session statuses look good to me.

Lionqueen94:
> I've checked the other fixtures and I've found that:
> 
>     * `committee_session.number_of_votes` is not set for fixture 1, 2 and 3.
>       
>       * Since `committee_session.number_of_votes` can be and stay 0, that does not seem like a problem to me.
> 
>     * The entry/result dates (`updated_at` + JSON) are not realistic in fixture 5 and 7. They happen at the same time as other records or results are before entries.
>       
>       * This is not causing any issues. But I've created a small PR to adjust it: [Adjust dates in fixtures #2562](https://github.com/kiesraad/abacus/pull/2562)
> 
> 
> Otherwise, the committee session statuses look good to me.

The committee session number of votes cannot be 0 so this should definitely be updated and preferably made impossible to be set to 0. The polling station number of votes can be 0, maybe you confused those?

stacktraceghost:
> ....
> 
> The committee session number of votes cannot be 0 so this should definitely be updated and preferably made impossible to be set to 0. The polling station number of votes can be 0, maybe you confused those?

[...]

PR #2534, opened 2025-11-26
Polling station results for a next session should be CSONextSession

oliver3:
Fix fixture election_5 (which is loaded in abacus-test.nl):
- Polling station results for a next session should be CSONextSession and adjusted data entry sections


praseodym:
Related to #2414

praseodym (approved):

PR #2538, opened 2025-11-26
Fix EML issues

oliver3:
- Fixes EML issues found in https://github.com/kiesraad/abacus/issues/2425

Four separate commits:
- Add `kr:` namespace for `ReportingUnitInvestigations` and `Investigation` tags
- Fix conditionals being accidentally swapped for adding `investigated_other_reason` and `ballots_have_been_recounted`
- TotalVotes Cast should be committee session number of voters
- Show `ReportingUnitInvestigations` only for first committee session, and always show "toegelaten kiezers opnieuw vastgesteld" instead of only when `true`

The reason why we only show `ReportingUnitInvestigations` for the first committee session, is because this code uses the most recent results for a polling station, which could be either a `CSONextSession` (in case of an investigation) or the original `CSOFirstSession` (in case of no investigation or newly added polling station). 

Because in the `CSONextSession` there is no extra investigation section or counting differences, we would be showing the wrong information. Due to the complexity we have decided to not do that at the moment.

### DIFF
--- a/backend/src/eml/eml_510.rs
+++ b/backend/src/eml/eml_510.rs
@@ -43,7 +43,13 @@ impl EML510 {
         let reporting_unit_votes = results
             .iter()
             .map(|(ps, results)| {
-                ReportingUnitVotes::from_polling_station(election, &authority_id, ps, results)
+                ReportingUnitVotes::from_polling_station(
+                    election,
+                    committee_session,
+                    &authority_id,
+                    ps,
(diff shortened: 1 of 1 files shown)


praseodym (approved):
Thanks for creating separate commits!

I've linted the 510b XML file against the EML_NL XSDs and found that they validate (with a valid election identifier, which our test data does not have).

PR #2539, opened 2025-11-26
Confirm deleting of polling station when data entry or investigation exists

### LINKED ISSUE #2438: Confirm deleting of polling station when data entry or investigation exists

Lionqueen94:
Closes #2438

New PR instead of https://github.com/kiesraad/abacus/pull/2480

### Description
- Confirm deleting of polling station when data entry or investigation exists with extra warning
- Delete related data entry, result and/or investigation
- Moved committee session status change from finished to in progress to `delete_data_entry_and_result` and `delete_investigation` functions to match the committee session status change flowchart
- When committee session status is changed from finished to in progress after polling station delete, update or add, show a warning instead of a success message
- Updated committee session status flowchart
- Removed duplicate `DataEntryDeleted` audit logging
- Frontend mockdata change
  - Investigations are only added if committee session number is > 1
  - Tests have been adjusted to make sure the committee session number is correct

### How to test
- Generate 2 elections with data entry and one without
- Finish one election with data entry and create a second session, add a polling station and an investigation
- Check the behaviour when deleting a polling station in each election
- Also check the warning you get when you delete a polling station when the data entry phase has finished
- Finish data entry phase again and update an existing polling station, check that the warning shows
- Finish data entry phase again and add a polling station, check that the warning shows
- Make sure the related data entry, result and/or investigation are indeed gone

### Renderings for different situations
1. Polling station exists in earlier committee session
image

2. Polling station created in the current committee session without investigation and data entry
image

3. Polling station created in the current committee session with investigation and data entry
image

4. Polling station created in the current committee session with investigation
image

5. Polling station created in the current committee session with data entry
image

6. Warning alert when committee session status is changed from finished to in progress on delete polling station
image

7. Warning alert when committee session status is changed from finished to in progress on add polling station
image

8. Warning alert when committee session status is changed from finished to in progress on update polling station
image


frontend/src/testing/api-mocks/ElectionStatusMockData.ts
```diff
@@ -1,51 +1,64 @@
...
+];
+
+export const getElectionStatusMockData = (
+  status: Partial<ElectionStatusResponseEntry> = {},
+): ElectionStatusResponse => {
+  const updatedStatuses = electionStatusesMock;
+  updatedStatuses[0] = {
+    ...electionStatusesMock[0]!,
+    ...status,
+  };
```
praseodym: I think this mutates the shared `electionStatusesMock` array directly, consider using `structuredClone` or `const updatedStatuses = [...electionStatusesMock]`

frontend/src/features/polling_stations/components/PollingStationUpdatePage.tsx
```diff
@@ -116,12 +121,12 @@ export function PollingStationUpdatePage() {
...
-                    <>
+                  {requestState.data.id_prev_session !== undefined ? (
+                    <section className="sm">
                       <strong>{t("polling_station.delete_not_possible.title")}</strong>
-                      <p>{t("polling_station.delete_not_possible.message")}</p>
-                    </>
-                  ) : (
+                      <p>{t("polling_station.delete_not_possible.pre_existing_polling_station")}</p>
+                    </section>
+                  ) : status && status.status === "first_entry_not_started" ? (
```
praseodym:
```suggestion
                  ) : status?.status === "first_entry_not_started" ? (
```

frontend/src/testing/api-mocks/ElectionStatusMockData.ts
Lionqueen94: Changed in 6e2e89729e3d8c80c5461b9e74f4577d7c37a869

frontend/src/features/polling_stations/components/PollingStationUpdatePage.tsx
Lionqueen94: Changed in 6e2e89729e3d8c80c5461b9e74f4577d7c37a869

lama-d3v:
In Figma (and the printscreen in #2438) is suggested that the administrator is able to view the polling station detail page, but unable to delete it.
Does this needs to be implemented?
image

The autorisatiematrix kind of suggests that it could/should be possible: https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/autorisatiematrix.md

And what happens when a typist is starting or discarding data entry while the coordinator is on the page? (as mentioned in #2438)

Lionqueen94:
> In Figma (and the printscreen in #2438) is suggested that the administrator is able to view the polling station detail page, but unable to delete it. Does this needs to be implemented? image
> 
> The autorisatiematrix kind of suggests that it could/should be possible: https://github.com/kiesraad/abacus/blob/main/documentatie/use-cases/autorisatiematrix.md
> 
> And what happens when a typist is starting or discarding data entry while the coordinator is on the page? (as mentioned in #2438)

See my comment on the annotation: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=12640-36438&m=dev#1521348391

When a typist starts a data entry and the coordinator clicks delete, they will get the `PollingStationCannotBeDeleted` error.
When a typist discards a data entry and the coordinator clicks the link to the data entry detail page, they will be redirected to the status overview.

frontend/src/i18n/locales/nl/polling_station.json
```diff
@@ -17,7 +17,8 @@
   },
   "delete_not_possible": {
     "title": "Stembureau verwijderen niet mogelijk",
-    "message": "Er zijn al tellingen ingevoerd."
+    "pre_existing_polling_station": "Er zijn al tellingen ingevoerd in een eerdere zitting.",
+    "delete_existing_data_entry": "Er zijn al tellingen ingevoerd. <link>Verwijder eerst de invoer</link> om dit stembureau te kunnen verwijderen."
```
lama-d3v: This is almost a duplicate of line 39, is that correct? Or should they be the same?

frontend/src/testing/api-mocks/ElectionStatusMockData.ts
```diff
@@ -1,51 +1,64 @@
...
-    },
-    {
-      polling_station_id: 7,
-      status: "second_entry_not_started",
-      first_entry_user_id: 1,
-      first_entry_progress: 100,
-    },
-    { polling_station_id: 8, status: "first_entry_not_started" },
-  ],
+export const electionStatusesMock: ElectionStatusResponseEntry[] = [
```
lama-d3v, line 6: `definitive` and `entries_different` do not match to the election status defined in https://github.com/kiesraad/abacus/blob/f017c54c325c0d8db1b68dd5337316842b39b8d6/frontend/src/features/election_status/components/ElectionStatus.stories.tsx#L15 (also applies to https://github.com/kiesraad/abacus/blob/d4010bfaf916c44a2195c6dd1fa99d1cecac4411/frontend/src/features/data_entry_home/components/PollingStationChoiceForm.test.tsx#L362)

frontend/src/i18n/locales/nl/polling_station.json
Lionqueen94: Link 39 is the content of the error message, there is no link in that message, so that's why the text is slightly different.

stacktraceghost:
Found that I can't delete a polling station after removing its entries. Reproduction:

1. Create a fresh environment on https://abacus-test.nl/ based on this PR
2. Log in as coordinator
3. Go to polling stations of corrigendum, select polling station number 42
4. It says I need to delete the entry first, follow the link and do so
5. Go back to editing the polling station. The delete button shows, press it and confirm
6. A message will appear that you can't delete the polling station because there are entries that first need to be deleted. But those were just deleted.
image


And building on that, if you then:
1. Delete the investigation of polling station number 42
2. Go back to editing the polling station, it says that data entry needs to be removed. 
3. Click on it and you receive an "Abacus is stuk" message

frontend/src/testing/api-mocks/ElectionStatusMockData.ts
Lionqueen94: Improved the mock in ee969ee9e7254b11aad1b49a9ab076c83d34f4b0 and now using it in the story as well to prevent duplication

Lionqueen94:
> Found that I can't delete a polling station after removing its entries. Reproduction:
> 
>     1. Create a fresh environment on https://abacus-test.nl/ based on this PR
> 
>     2. Log in as coordinator
> 
>     3. Go to polling stations of corrigendum, select polling station number 42
> 
>     4. It says I need to delete the entry first, follow the link and do so
> 
>     5. Go back to editing the polling station. The delete button shows, press it and confirm
> 
>     6. A message will appear that you can't delete the polling station because there are entries that first need to be deleted. But those were just deleted.
> 
> 
> image
> 
> And building on that, if you then:
> 
>     1. Delete the investigation of polling station number 42
> 
>     2. Go back to editing the polling station, it says that data entry needs to be removed.
> 
>     3. Click on it and you receive an "Abacus is stuk" message

Fixed and tested in 3e3d97f87e5b7b4e34b73871abe2c1e1eb036ff6

stacktraceghost:
> > Found that I can't delete a polling station after removing its entries. Reproduction:
> > ```
> > 1. Create a fresh environment on https://abacus-test.nl/ based on this PR
> > 
> > 2. Log in as coordinator
> > 
> > 3. Go to polling stations of corrigendum, select polling station number 42
> > 
> > 4. It says I need to delete the entry first, follow the link and do so
> > 
> > 5. Go back to editing the polling station. The delete button shows, press it and confirm
> > 
> > 6. A message will appear that you can't delete the polling station because there are entries that first need to be deleted. But those were just deleted.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > image
> > And building on that, if you then:
> > ```
> > 1. Delete the investigation of polling station number 42
> > 
> > 2. Go back to editing the polling station, it says that data entry needs to be removed.
> > 
> > 3. Click on it and you receive an "Abacus is stuk" message
> > ```
> 
> Fixed and tested in [3e3d97f](https://github.com/kiesraad/abacus/commit/3e3d97f87e5b7b4e34b73871abe2c1e1eb036ff6)

If I follow all the described steps (deleting the data entry and investigation of polling station 42), it will still say that I need to remove the data entry and following that link shows "Abacus is stuk". Maybe we can write a test to follow all those steps.

Out-of-scope for this PR I think, but I also noticed the following: after deleting the data entry, I can't use my browser history back button to get back to the polling station. That might be caused by a redirect on the read-only view page. After deleting the investigation this works, but then I first see an investigation page which is empty (because the investigation doesn't exist).

bramtweedegolf:
When the (first) session is closed and the coordinator edits a polling station, clicking the 'Verwijder eerst de invoer' link leads to a 409 error, even though the warning at the top of the page implies that it should be possible.

I think this is fine, but it would be nice to change the text warning of the warning and maybe also the text at bottom. But this is nitpick, so feel free to ignore it. 

Screenshot 2025-12-02 at 10-18-39 Stembureaus - Abacus

bramtweedegolf:
> > > Found that I can't delete a polling station after removing its entries. Reproduction:
> > > ```
> > > 1. Create a fresh environment on https://abacus-test.nl/ based on this PR
> > > 
> > > 2. Log in as coordinator
> > > 
> > > 3. Go to polling stations of corrigendum, select polling station number 42
> > > 
> > > 4. It says I need to delete the entry first, follow the link and do so
> > > 
> > > 5. Go back to editing the polling station. The delete button shows, press it and confirm
> > > 
> > > 6. A message will appear that you can't delete the polling station because there are entries that first need to be deleted. But those were just deleted.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > image
> > > And building on that, if you then:
> > > ```
> > > 1. Delete the investigation of polling station number 42
> > > 
> > > 2. Go back to editing the polling station, it says that data entry needs to be removed.
> > > 3. Click on it and you receive an "Abacus is stuk" message
> > 
> > 
> > Fixed and tested in [3e3d97f](https://github.com/kiesraad/abacus/commit/3e3d97f87e5b7b4e34b73871abe2c1e1eb036ff6)
> 
> If I follow all the described steps (deleting the data entry and investigation of polling station 42), it will still say that I need to remove the data entry and following that link shows "Abacus is stuk". Maybe we can write a test to follow all those steps.
> 
> Out-of-scope for this PR I think, but I also noticed the following: after deleting the data entry, I can't use my browser history back button to get back to the polling station. That might be caused by a redirect on the read-only view page. After deleting the investigation this works, but then I first see an investigation page which is empty (because the investigation doesn't exist).

The same thing happens when the coordinatior creates a new polling station, after the first session has been closed.

The polling stations that have results for the first session have the correct message:

Screenshot_2025-12-02_11-02-06

But any new polling stations have the message with the link, which leads to an error:

Screenshot_2025-12-02_11-02-13

frontend/src/i18n/locales/nl/polling_station.json
Lionqueen94: This is now outdated because the implementation has changed

Lionqueen94:
> When the (first) session is closed and the coordinator edits a polling station, clicking the 'Verwijder eerst de invoer' link leads to a 409 error, even though the warning at the top of the page implies that it should be possible.
> 
> I think this is fine, but it would be nice to change the text warning of the warning and maybe also the text at bottom. But this is nitpick, so feel free to ignore it.
>
> ...

Good find, definitely not a nitpick! Fixed in ea423de58d8f5d95aeea58edaa26c3d3f773d448

stacktraceghost:
Noticed that now we have modals informing the user what is about to happen, great! When I follow my steps, it works now.

One suggestion: if you go to the finished state for the corrigendum and then remove 42, we could inform the user that this action will put the committee session status back to in progress?

backend/src/investigation/api.rs
```diff
@@ -72,6 +72,25 @@ async fn validate_and_get_committee_session(
     Ok(committee_session)
 }
 
+pub async fn delete_investigation_for_polling_station(
+    conn: &mut SqliteConnection,
+    audit_service: &AuditService,
+    polling_station_id: u32,
+) -> Result<(), APIError> {
+    if let Some(investigation) =
+        delete_polling_station_investigation(conn, polling_station_id).await?
```
stacktraceghost, line 82: Lets make sure you can only delete the investigations of a polling station in the current committee session. Preferably on the repository level.

backend/src/polling_station/mod.rs
```diff
@@ -295,14 +295,10 @@ async fn polling_station_delete(
     let polling_station =
         repository::get_for_election(&mut tx, election_id, polling_station_id).await?;
 
-    if data_entry_exists(&mut tx, polling_station.id).await?
-        || result_exists(&mut tx, polling_station.id).await?
-    {
-        return Err(APIError::Conflict(
-            "Polling station cannot be deleted.".to_string(),
-            ErrorReference::PollingStationCannotBeDeleted,
```
stacktraceghost, line 304: I don't think `PollingStationCannotBeDeleted` is used anymore, so we can remove it from the codebase.

Lionqueen94:
> Noticed that now we have modals informing the user what is about to happen, great! When I follow my steps, it works now.
> 
> One suggestion: if you go to the finished state for the corrigendum and then remove 42, we could inform the user that this action will put the committee session status back to in progress?

@jorisleker What do you think of this suggestion?

backend/src/polling_station/mod.rs
Lionqueen94: Removed in 5b19cb667023bd7fec0b4ba6de573103a3d8af9d

backend/src/investigation/api.rs
stacktraceghost: After discussion we have concluded that the validation already happens in the API function through `validate_and_get_committee_session`. So this extra validation is not necessary.

jorisleker:
> > Noticed that now we have modals informing the user what is about to happen, great! When I follow my steps, it works now.
> > One suggestion: if you go to the finished state for the corrigendum and then remove 42, we could inform the user that this action will put the committee session status back to in progress?
> 
> @jorisleker What do you think of this suggestion?

Clarifying question. If you delete polling station 42 (including investigation and data entries) for a session that was completed, than won't the session remain completed? Or is this describing the case where polling station 42 was the only polling station in the session?

Lionqueen94:
> > > Noticed that now we have modals informing the user what is about to happen, great! When I follow my steps, it works now.
> > > One suggestion: if you go to the finished state for the corrigendum and then remove 42, we could inform the user that this action will put the committee session status back to in progress?
> > 
> > 
> > @jorisleker What do you think of this suggestion?
> 
> Clarifying question. If you delete polling station 42 (including investigation and data entries) for a session that was completed, than won't the session remain completed? Or is this describing the case where polling station 42 was the only polling station in the session?

well I changed it that the session status is reverted to in progress on deletion of a polling station, because it's possible you downloaded the documents already and that that polling station is listed in the documents. So my thinking was that the documents should be regenerated => status set back to in progress and you have to finish the data entry phase again and download the newly generated documents. Is that assumption incorrect?

jorisleker:
> well I changed it that the session status is reverted to in progress on deletion of a polling station, because it's possible you downloaded the documents already and that that polling station is listed in the documents. So my thinking was that the documents should be regenerated => status set back to in progress and you have to finish the data entry phase again and download the newly generated documents. Is that assumption incorrect?

No that's a valid assumption, and I think alerting the user that previously generated documents are not valid anymore / do not match the status of the election in Abacus is needed. I am not sure we should do this in a modal because there are a lot of warnings in there already., Maybe we could add an alert to the page _after_ the user has deleted the polling station?

Lionqueen94:
> > well I changed it that the session status is reverted to in progress on deletion of a polling station, because it's possible you downloaded the documents already and that that polling station is listed in the documents. So my thinking was that the documents should be regenerated => status set back to in progress and you have to finish the data entry phase again and download the newly generated documents. Is that assumption incorrect?
> 
> No that's a valid assumption, and I think alerting the user that previously generated documents are not valid anymore / do not match the status of the election in Abacus is needed. I am not sure we should do this in a modal because there are a lot of warnings in there already., Maybe we could add an alert to the page _after_ the user has deleted the polling station?

Can you give a suggestion for the copy for this alert? I should be able to add some text to the current success alert for example? Or show a specific alert in this case?

jorisleker:
> > > well I changed it that the session status is reverted to in progress on deletion of a polling station, because it's possible you downloaded the documents already and that that polling station is listed in the documents. So my thinking was that the documents should be regenerated => status set back to in progress and you have to finish the data entry phase again and download the newly generated documents. Is that assumption incorrect?
> > 
> > 
> > No that's a valid assumption, and I think alerting the user that previously generated documents are not valid anymore / do not match the status of the election in Abacus is needed. I am not sure we should do this in a modal because there are a lot of warnings in there already., Maybe we could add an alert to the page _after_ the user has deleted the polling station?
> 
> Can you give a suggestion for the copy for this alert? I should be able to add some text to the current success alert for example? Or show a specific alert in this case?

Yes, [something like this](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=16835-15839&t=yN0YQqgR79g2FK9m-11) (Figma)?

image

jorisleker (approved):
Tested:

- ✅ When i delete polling station 42 I get the modal to confirm deletion. It mentions that the investigation and data entry will be deleted as well;
- ✅ When I delete the polling station _after_ I've completed data entry, the polling station list page shows the alert 'Maak een nieuw proces verbaal voor deze zitting'.
- ✅ If I detete the polling station _before_ completing data entry, the polling station list page shows a confirmation alert. 

Everything as designed. Did not do a code review

stacktraceghost:
Did a code review, looks good. 

Created a test environment with several elections and checked every modal and warning message. The only thing I've noticed is that we now give a warning to recreate the documents when deleting a polling station, but not when updating. @jorisleker maybe we can make this more consistent in the future? 

Since the PR is quite big and this is the only code review, I would propose to wait for a second code review.

stacktraceghost (dismissed):

Lionqueen94:
> Did a code review, looks good.
> 
> Created a test environment with several elections and checked every modal and warning message. The only thing I've noticed is that we now give a warning to recreate the documents when deleting a polling station, but not when updating. @jorisleker maybe we can make this more consistent in the future?
> 
> Since the PR is quite big and this is the only code review, I would propose to wait for a second code review.

According to the flowchart we only change the status from finished to in progress on a data entry/result delete or an investigation delete. Which is now how it's implemented. So not when just deleting or updating a polling station only (also you won't be in a finished state if the polling station doesn't have a linked data entry+result or investigation). If the committee status change should also be done on any update to a polling station (name, for example), then we should update the flowchart regarding this. @jorisleker What do you think?

stacktraceghost:
> > Did a code review, looks good.
> > Created a test environment with several elections and checked every modal and warning message. The only thing I've noticed is that we now give a warning to recreate the documents when deleting a polling station, but not when updating. @jorisleker maybe we can make this more consistent in the future?
> > Since the PR is quite big and this is the only code review, I would propose to wait for a second code review.
> 
> According to the flowchart we only change the status from finished to in progress on a data entry/result delete or an investigation delete. Which is now how it's implemented. So not when just deleting or updating a polling station only (also you won't be in a finished state if the polling station doesn't have a linked data entry+result or investigation). If the committee status change should also be done on any update to a polling station (name, for example), then we should update the flowchart regarding this. @jorisleker What do you think?

Example:
1. Go to abacus-test.nl, create a new environment
2. Go to election "Corrigendum 2026" and finish the committee session
3. Go to a polling station
image

4. Change the name
5. We are back in progress again

Which I think makes sense, since content of the documents change. But I agree that it is missing on the diagram for a polling station (not for investigation). 

This warning already says that new documents will be generated, but after saving it does not mention it anymore on the polling station list page. While when deleting it does. Just a little thing, but less consistent.

Lionqueen94:
> > > Did a code review, looks good.
> > > Created a test environment with several elections and checked every modal and warning message. The only thing I've noticed is that we now give a warning to recreate the documents when deleting a polling station, but not when updating. @jorisleker maybe we can make this more consistent in the future?
> > > Since the PR is quite big and this is the only code review, I would propose to wait for a second code review.
> > 
> > 
> > According to the flowchart we only change the status from finished to in progress on a data entry/result delete or an investigation delete. Which is now how it's implemented. So not when just deleting or updating a polling station only (also you won't be in a finished state if the polling station doesn't have a linked data entry+result or investigation). If the committee status change should also be done on any update to a polling station (name, for example), then we should update the flowchart regarding this. @jorisleker What do you think?
> 
> Example:
> 
>     1. Go to abacus-test.nl, create a new environment
> 
>     2. Go to election "Corrigendum 2026" and finish the committee session
> 
>     3. Go to a polling station
> 
> 
> image
> 
>     4. Change the name
> 
>     5. We are back in progress again
> 
> 
> Which I think makes sense, since content of the documents change. But I agree that it is missing on the diagram for a polling station (not for investigation).
> 
> This warning already says that new documents will be generated, but after saving it does not mention it anymore on the polling station list page. While when deleting it does. Just a little thing, but less consistent.

Updated the flowchart and added warnings for add and update as well in 90f79da3c438716468d6943a8052ec1a295d080e

praseodym:
@Lionqueen94 this branch has a conflict after #2600 was merged

jorisleker (approved):
New approval because of merge conflict, assuming nothing changed in reviewed code

documentatie/flowcharts/committee-session-state.md
```diff
@@ -18,9 +18,10 @@ stateDiagram-v2
   DataEntryPaused --> DataEntryInProgress: click resume <br/> data entry
   DataEntryPaused --> Created: delete last <br/> PS/Inv
   DataEntryPaused --> DataEntryFinished: click finish <br/> data entry
-  DataEntryFinished --> DataEntryInProgress: add new <br/> PS/Inv
-  DataEntryFinished --> DataEntryInProgress: click resume <br/> data entry/update Inv
-  DataEntryFinished --> DataEntryInProgress: delete polling <br/> station result/delete Inv
-  DataEntryFinished --> Created: delete last <br/>PS/Inv
+  DataEntryFinished --> DataEntryInProgress: add <br/> PS/Inv
+  DataEntryFinished --> DataEntryInProgress: update <br/> PS/Inv
```
stacktraceghost:
```suggestion
  DataEntryFinished --> DataEntryInProgress: add/update <br/> PS/Inv
```

To keep the flowchart in shape and readable

backend/src/investigation/api.rs
```diff
@@ -332,61 +363,45 @@ async fn polling_station_investigation_delete(
 
     let committee_session = validate_and_get_committee_session(&mut tx, polling_station_id).await?;
 
-    let investigation = get_polling_station_investigation(&mut tx, polling_station_id).await?;
+    get_polling_station_investigation(&mut tx, polling_station_id).await?;
```
lama-d3v, line 366: Polling station ID could potentially not exist here. Move line 367 to line 363?
And is this covered in a test?

backend/src/investigation/api.rs
```diff
@@ -72,6 +72,36 @@ async fn validate_and_get_committee_session(
...
 }
 
+pub async fn delete_investigation_for_polling_station(
+    conn: &mut SqliteConnection,
+    audit_service: &AuditService,
+    committee_session: &CommitteeSession,
+    polling_station_id: u32,
+) -> Result<(), APIError> {
+    if let Some(investigation) =
+        delete_polling_station_investigation(conn, polling_station_id).await?
```
lama-d3v, line 82: This function deletes all investigations for the polling station. If there are multiple committee sessions for a polling station, than only the investigations should be deleted for that specific committee session.
And maybe also rename this function to plural `delete_polling_station_investigations`

backend/src/investigation/api.rs
```diff
@@ -332,61 +363,45 @@ async fn polling_station_investigation_delete(
...
     // Delete investigation
-    if delete_polling_station_investigation(&mut tx, polling_station_id).await? {
-        audit_service
-            .log(
-                &mut tx,
-                &AuditEvent::PollingStationInvestigationDeleted(investigation.clone()),
-                None,
-            )
-            .await?;
+    delete_investigation_for_polling_station(
```
lama-d3v, line 371: Shouldn't the data entries be deleted first, and then the investigations?

backend/src/investigation/api.rs
```diff
@@ -332,61 +363,45 @@ async fn polling_station_investigation_delete(
...
+    delete_data_entry_and_result_for_polling_station(
+        &mut tx,
+        &audit_service,
+        &committee_session,
+        polling_station.id,
+    )
+    .await?;
+
+    // Change committee session status if last investigation is deleted
+    if list_investigations_for_committee_session(&mut tx, committee_session.id)
```
lama-d3v, line 389: Could it be possible that the committee session status is "in progress" when all investigations are deleted. In that case, should the status be reverted to "created" as well?

backend/src/investigation/repository.rs
```diff
@@ -101,16 +100,22 @@ pub async fn get_polling_station_investigation(
...
     conn: &mut SqliteConnection,
     polling_station_id: u32,
-) -> Result<bool, Error> {
-    let rows_affected = query!(
-        r#"DELETE FROM polling_station_investigations WHERE polling_station_id = ?"#,
+) -> Result<Option<PollingStationInvestigation>, Error> {
+    query_as!(
+        PollingStationInvestigation,
+        r#"
+            DELETE FROM polling_station_investigations
```
lama-d3v, line 107: (as mentioned earlier), there should be filtered on the correct committee session in this query.

backend/src/polling_station/mod.rs
```diff
@@ -296,14 +296,21 @@ async fn polling_station_delete(
...
+    )
+    .await?;
+
+    delete_investigation_for_polling_station(
+        &mut tx,
+        &audit_service,
+        &committee_session,
+        polling_station.id,
+    )
+    .await?;
```
lama-d3v, line 313: Maybe you could remove the duplicate of https://github.com/kiesraad/abacus/blob/de5020dbac28004ceda3a9952e62096e9c89a18c/backend/src/investigation/api.rs#L371

backend/tests/polling_station_integration_test.rs
```diff
@@ -905,40 +903,92 @@ async fn test_delete_for_committee_session_with_in_progress_status_as_administra
 }
 
 #[test(sqlx::test(fixtures(path = "../fixtures", scripts("election_2", "users"))))]
-async fn test_delete_with_data_entry_fails(pool: SqlitePool) {
+async fn test_delete_with_data_entry_works(pool: SqlitePool) {
```
lama-d3v, line 906: Maybe also add tests for first and second committee session and check that the correct data is deleted?

documentatie/flowcharts/committee-session-state.md
```diff
@@ -18,9 +18,9 @@ stateDiagram-v2
...
   DataEntryPaused --> Created: delete last <br/> PS/Inv
   DataEntryPaused --> DataEntryFinished: click finish <br/> data entry
-  DataEntryFinished --> DataEntryInProgress: add new <br/> PS/Inv
-  DataEntryFinished --> DataEntryInProgress: click resume <br/> data entry/update Inv
-  DataEntryFinished --> DataEntryInProgress: delete polling <br/> station result/delete Inv
-  DataEntryFinished --> Created: delete last <br/>PS/Inv
+  DataEntryFinished --> DataEntryInProgress: add/update <br/> PS/Inv
+  DataEntryFinished --> DataEntryInProgress: delete polling <br/> station result/Inv
+  DataEntryFinished --> DataEntryInProgress: click resume <br/> data entry
+  DataEntryFinished --> Created: delete last <br/> PS/Inv
```
lama-d3v, line 24: is `delete last` correct?

frontend/src/testing/api-mocks/ElectionStatusMockData.ts
```diff
@@ -1,51 +1,80 @@
...
+  },
+  {
+    polling_station_id: 6,
+    status: "first_entry_in_progress",
+    first_entry_user_id: 2,
+    first_entry_progress: 60,
+  },
+  {
+    polling_station_id: 7,
+    status: "second_entry_not_started",
```
lama-d3v, line 51: There are 2 items with status `second_entry_not_started `. Is that correct?

frontend/src/i18n/locales/nl/polling_station.json
```diff
@@ -3,21 +3,24 @@
...
   "import_instructions": "Kies het bestand met de stembureaus.",
   "current_form": "het huidige formulier",
   "delete": "Stembureau verwijderen",
   "delete_modal": {
-    "title": "Stembureau verwijderen?",
-    "confirm": "Je kan dit niet ongedaan maken. Weet je zeker dat je dit stembureau wilt verwijderen?"
+    "action_cannot_be_reverted": "Deze actie kan niet worden teruggedraaid.",
+    "existing_data_entry": "<strong>Let op:</strong> Er is in deze zitting invoer voor dit stembureau gedaan. De invoer wordt ook verwijderd. ",
+    "existing_investigation": "<strong>Let op:</strong> Er is in deze zitting een onderzoek voor dit stembureau aangemaakt. Het onderzoek wordt ook verwijderd. ",
+    "existing_investigation_and_data_entry": "<strong>Let op:</strong> Er is in deze zitting een onderzoek voor dit stembureau aangemaakt én invoer gedaan. Het onderzoek en de invoer worden ook verwijderd. ",
```
lama-d3v, line 18: Remove extra space after `.` at line endings

frontend/src/features/polling_stations/components/PollingStationUpdatePage.tsx
```diff
@@ -60,19 +72,30 @@ export function PollingStationUpdatePage() {
...
-    });
-
+    if (currentCommitteeSession.status === "data_entry_finished") {
+      pushMessage({
+        type: "warning",
+        title: t("generate_new_results"),
+        text: `${t("polling_station.message.polling_station_deleted", {
+          number: pollingStation.number,
+          name: pollingStation.name,
+        })}. ${t("documents_are_invalidated")}`,
```
lama-d3v, line 82: Is it correct that both these translations are used here?

frontend/src/features/polling_stations/components/PollingStationDeleteModal.tsx
lama-d3v, line 1: Should there be a specific test file for this modal?

lama-d3v (dismissed):

backend/src/investigation/api.rs
Lionqueen94: The polling station gets duplicated for a new committee session, so there is always a unique constraint between the polling station number and the committee session id. Also the investigation id is the polling station id, so there is maximum 1 investigation per polling station.

documentatie/flowcharts/committee-session-state.md
Lionqueen94: Yes, you go from any state to Created on deleting the last of a polling station/investigation

frontend/src/i18n/locales/nl/polling_station.json
Lionqueen94: This is on purpose, to make sure it renders correctly. The whitespace gets removed when I added it separately in the modal.

frontend/src/features/polling_stations/components/PollingStationUpdatePage.tsx
Lionqueen94: Yes I did this on purpose so that the general invalidated translation can be used in each case, instead of adding it to each individual message

frontend/src/features/polling_stations/components/PollingStationDeleteModal.tsx
Lionqueen94: It's tested in the unit test for the page, so I don't think this is needed.

frontend/src/testing/api-mocks/ElectionStatusMockData.ts
Lionqueen94: Yes, one is changed to another user to make sure there are 2 entries to be tested in the storybook tests. It's nice to check that the user in the data corresponds with the rendering.

backend/src/investigation/api.rs
Lionqueen94: I call the data entry delete before the investigation delete in the delete polling station endpoint

backend/src/investigation/api.rs
Lionqueen94: Yes, that's why it is reverted if there are no investigations left, doesn't matter what state the committee session is in.

backend/src/investigation/repository.rs
Lionqueen94: This is not needed, see https://github.com/kiesraad/abacus/pull/2539#discussion_r2623336502

backend/src/investigation/api.rs
Lionqueen94: This is already checked in:
image
Or do you mean something else?

backend/src/polling_station/mod.rs
Lionqueen94: What do you mean? Both endpoints call the same function that handles the deleting of the investigation defined in investigation/api.rs

backend/tests/polling_station_integration_test.rs
Lionqueen94: The data is linked to a unique polling station id, which is linked to the committee session, there is no relation to the polling station in the earlier committee session. So I don't think this is really needed @jschuurk-kr what do you think?

frontend/src/i18n/locales/nl/polling_station.json
Lionqueen94: To explain it a bit more, I have a function called `renderWarning` that optionally renders the warning with a `tx` and I would need to add another if statement to add the `{" "}` which was a bit of a waste, when I can just add it to the end of these i18n values (since they always need the space anyway).

praseodym:
> **⚠️ Your code did not improve towards your objective of 3.5 stars.**

Are there improvements we can identify to improve the maintainability of the code in this PR? (If this is significant work it's not for this PR, but for a future refactoring.)

Lionqueen94:
> > **⚠️ Your code did not improve towards your objective of 3.5 stars.**
> 
> Are there improvements we can identify to improve the maintainability of the code in this PR? (If this is significant work it's not for this PR, but for a future refactoring.)

@praseodym I'm not sure of any improvements right now, maybe this is something we can have a look at next week and create some issues for?

backend/tests/polling_station_integration_test.rs
jschuurk-kr: Agree that which committee session the polling station is linked to, should not matter.

I do have some concerns about the test coverage, but let's have that conversation separate from this PR.

My concerns are related to (as an example) `polling_station_delete()`. It does quite a few things: deleting the data entry, result and/or investigation, delete the polling station, log the deletion of the polling station, change the committee status if needed.
Our code coverage is happy, since that code gets executed by `polling_station_integration_test.rs` (and maybe also some unit tests, I did not check thoroughly). However, those integration tests do not check everything that's done in `polling_station_delete()`. For example the test `test_delete_with_data_entry_works` asserts on the response of the API call and there being one status fewer returned by `/elections/{election_id}/status`. This also means that our integration tests end up being the only tests we have for `delete_data_entry_and_result_for_polling_station()` and `delete_investigation_for_polling_station()`, which takes some clicking around to determine.

So in conclusion, I think it would be interesting to review the coverage of our backend tests. As part of that conversation I'd like to calculate our code coverage when excluding our integration tests. Not that that's a more correct number for code coverage, but I do think that in unit tests there tends to be a closer relation between the code being executed and what's being asserted.
And I'm wondering if a function like `polling_station_delete()` is doing too many things in the function itself. Maybe the deleting of the polling station and its related data should go in a separate function, which we can then unit test.

praseodym (approved):

jorisleker (approved):
approval based on previous reviews of UX / frontend. Did not review changes last merges

PR #2540, opened 2025-11-27
Fix investigation and election import progress lists rendering

Lionqueen94:
- With the implementation of https://github.com/kiesraad/abacus/pull/2470 2 other places where the ProgressList was used now look ugly. This PR fixes the rendering of the investigation create/update navigation and the election import navigation
- Also the issue is fixed that a page scrollbar appears when there should not be one (in case of > 10 lists or a smaller screen)

Currently on release:
image
image

Fixed:
image
image


frontend/src/components/ui/ProgressList/ProgressList.module.css
```diff
@@ -4,6 +4,7 @@
   flex-direction: column;
   overflow: hidden;
   gap: 1rem;
+  max-height: 70vh;
```
stacktraceghost: This changes the behavior of the list when doing data entry with many political groups. This component is tricky with its behavior and we should discuss how we want it to act.

Example with max-height:
It doesn't stretch anymore.
Image

Example without max-height:
It stretches while scrolling down until it shows all groups (if it fits). 
Image

Besides that, I think `70vh` is based on a specific resolution.

Lionqueen94: Maik and I fixed it in dd0e8aa4790c72b3e6ea8a2889d6af01e4fd25ef please test a lot haha. Only with resizing back the scrollbar does not disappear, but this is pretty unlikely

frontend/src/components/ui/AppLayout/StickyNav.tsx
```diff
@@ -11,24 +11,39 @@ export function StickyNav({ children }: StickyNavProps) {
...
-        nav.style.paddingTop = navTop < 0 ? `${-navTop}px` : "";
+
+        const parentPaddingBottom = parseInt(getComputedStyle(parent).paddingBottom);
+        const parentBottom = parent.getBoundingClientRect().bottom;
+
+        // Calculate visible parent bottom padding, based on element distance to bottom of viewport
+        const visiblePadding = Math.max(parentPaddingBottom + window.innerHeight - parentBottom, 0);
+
+        // If parent bottom padding is not fully into view, set padding of nav to 16, else use visible padding
+        const bottomOffset = visiblePadding === 0 ? 16 : visiblePadding;
```
oliver3: There is now a strange jump between approaching zero and becoming zero and therefore 16px. Better to use the 16 in the `Math.max()` above

oliver3:
This seems to work well, but is there any benefit with this solution as compared to the situation before the regression in https://github.com/kiesraad/abacus/pull/2470 ? This solution seems more complex but maybe it fixes something else that I did not find?

Lionqueen94:
> This seems to work well, but is there any benefit with this solution as compared to the situation before the regression in #2470 ? This solution seems more complex but maybe it fixes something else that I did not find?

If you make the screen very small, the progress items would just be hidden. Try this on the import election screen. I think this was already the case before the previous PR changes. This is not happening anymore in this PR (you will get a scrollbar)

PR #2551, opened 2025-12-01
Change GR26 specific texts

praseodym:
Change/remove GR26 specific texts from the user interface and documentation.


Lionqueen94 (approved):

backend/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "abacus"
-version = "0.1.0"
+version = "1.1.0"
```
Lionqueen94, line 3: should we still update this in our release to 1.0.0?

praseodym: I don't think it's used anywhere, except for publishing to https://crates.io/ (which we don't do for Abacus)

PR #2556, opened 2025-12-01
Enable Clippy code quality lints

### LINKED ISSUE #2552: Clippy code quality metrics lints
Enable Clippy lint for code quality metrics such as function length and complexity.

praseodym:
Enable the Clippy [cognitive_complexity](https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#cognitive_complexity) and [too_many_lines](https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#too_many_lines) lints for code quality.

All current code that the lint found issues with has been explicitly allowed. We can fix that code in separate pull requests.

Resolves #2552

### DIFF
--- a/backend/src/pdf_gen/embedded/mod.rs
+++ b/backend/src/pdf_gen/embedded/mod.rs
@@ -80,6 +80,7 @@ fn convert_datetime<Tz: chrono::TimeZone>(date_time: chrono::DateTime<Tz>) -> Op
     )
 }
 
+#[allow(clippy::cognitive_complexity)]
 fn compile_pdf(world: &mut world::PdfWorld, model: PdfModel) -> Result<PdfGenResult, PdfGenError> {
     debug!("Starting Typst compilation for {}", model.as_model_name());
 


backend/clippy.toml
```diff
@@ -1 +1,3 @@
 allow-unwrap-in-tests = true
+cognitive-complexity-threshold = 20
+too-many-lines-threshold = 60
```
stacktraceghost, line 3: Are these lower than default (respectively 25, 100) based on Sigrids measurements?

praseodym: Yes, although the measurements does not seem to match exactly.

Sigrid lists yellow 'alerts' for units with a McCabe complexity of ≥6 and unit size ≥16 lines of (normalized) code. Units with a complexity of ≥25 or with ≥60 lines have red 'alerts', and then there are orange alerts somewhere in between.

I've chosen these values to include at least all 'reds' and also not to have too many functions to fix right now.

stacktraceghost (approved):

PR #2559, opened 2025-12-02
Add cli version / use exe version for installer

stacktraceghost:
This PR adds `-V` / `--version` to show the version info:

```
> abacus.exe -V
1.0.0-dirty
```

The InnoSetup script will now use this value as the version.

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -134,7 +134,6 @@ The following dependencies (crates) are used:
 - `async_zip`: creating a zip of the EML_NL and PDF PV.
 - `strum`: Converting enums from their string representation and back
 
-
 Additionally, the following development dependencies are used:
 
 - `test-log`: show tracing messages while running tests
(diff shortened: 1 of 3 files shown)


backend/README.md
```diff
@@ -191,6 +190,7 @@ Options:
   -p, --port <PORT>          Server port, optional [env: ABACUS_PORT=] [default: 8080]
   -d, --database <DATABASE>  Location of the database file, will be created if it doesn't exist [env: ABACUS_DATABASE=] [default: db.sqlite]
   -a, --airgap-detection     Enable airgap detection [env: ABACUS_AIRGAP_DETECTION=]
+  -V, --version              Show version
```
Lionqueen94, line 193: Why is this the only letter that is capitalized? 😅

stacktraceghost: I've chosen `-V` because it is often used as version flag, whereas `-v` is used for verbose.

Issue #1008, opened 2025-02-10
Investigate Biome as replacement for ESLint and Prettier


praseodym:
[Biome](https://biomejs.dev) can replace a big part of what we do with ESLint and Prettier (see [Migrate from ESLint and Prettier](https://biomejs.dev/guides/migrate-eslint-prettier/)). The goal of this issue is to investigate how Biome compares to ESLint and Prettier and whether it could be a (partial) replacement.

Part of #2561:
- [x] Apply format fixes
- [x] Fix remaining lint issues (or [suppress](https://biomejs.dev/analyzer/suppressions/) them, or disable the rule)
- [x] Configure GitHub Actions to use Biome instead of Prettier and in addition to ESLint
- [x] Configure Lefthook
- [x] Remove Prettier config

Still TODO:
- #2668
- #2669
- [x] Decide if there are optional Biome rules that we want to enable (see icons on e.g. the [Biome JavaScript Rules docs](https://biomejs.dev/linter/javascript/rules/))

praseodym:
Closing, remaining work is part of #2668 and #2669.

PR #2561, opened 2025-12-02
Configure Biome for linting and formatting

praseodym:
[Biome](https://biomejs.dev/) is a super fast linter and formatter for web projects (written in Rust!). On my machine it can lint and format Abacus in about 130ms, whereas Prettier takes 3.8s and ESLint takes 13.7s.

The Biome configuration consists of formatting and linting configuration. The formatting configuration was created by [automatically migrating from Prettier](https://biomejs.dev/guides/migrate-eslint-prettier/) and some manual tweaks. The current configuration matches our Prettier config quite well, `npx @biomejs/biome format` only reports 8 files that would be reformatted.

Automatic migration of the linting configuration created a very verbose file, so I chose to write it by hand. I enabled the recommended rules and then disabled ones that report many errors. Most of the currently disabled lints should also be re-enabled to fix (potential) issues in our codebase.

A disadvantage is that Biome doesn't (yet) support every lint rule that we have currently configured in ESLint, in particular the library-specific ones from the Storybook and Playwright plugins are not available. We will keep ESLint around for just these rules. Cleaning up the ESLint config to remove duplication with Biome is not in scope for this PR but still part of #1008.

To try out locally:
- Check out this branch
- Run `npm ci` to install Biome
- Run Biome using `npm run lint:biome`
- Install the [first-party extension in VS Code, IntelliJ/RustRover, Zed](https://biomejs.dev/guides/editors/first-party-extensions/#intellij) or one of the [third-party extensions](https://biomejs.dev/guides/editors/third-party-extensions/)
  - For Zed, configure the [LSP `config_path`](https://biomejs.dev/reference/zed/#configuration) as `frontend/biome.jsonc` 
- Change some code to experience how fast Biome can lint
- Remove disabled lints from `biome.json` to check out other findings that we missed with ESLint

Please review per commit, some commits (e.g. import sorting) are very big but don't contain manual changes.

Relates to #1008

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -107,12 +107,14 @@ jobs:
           node-version: 22
       - name: Install dependencies
         run: npm ci
+      - name: Run Biome
+        run: npx biome ci
+      - name: Run TypeScript compiler
+        run: npm run lint:tsc
+      - name: Run ESLint
+        run: npm run lint:eslint
       - name: Install Playwright Chromium Browser
         run: npx playwright install --with-deps chromium
-      - name: Check formatting
-        run: npx @biomejs/biome format
-      - name: Lint
-        run: npm run lint
       - name: Test
         run: npm test
       - name: Build


oliver3:
Good to see that it has support for linting JSON files as well!

```json
  "overrides": [
    {
      "includes": [
        "src/i18n/locales/**/*.json"
      ],
      "assist": {
        "enabled": true,
        "actions": {
          "source": { "useSortedKeys": "on"}
        }
      }
    },
```

oliver3:
I think it is a good idea to switch to Biome for our project because of its speed and combining linting and formatting in one tool and the (recommended) rules seem to uncover some more potential issues.

Having separate dedicated ESLint configs for the remaining use cases (storybook, playwright) would be nice if it does not slow the whole process down again (lefthook, CI).

One extra suggestion: let's use JSONC for the config so that we can also add comments for the choices we make regarding rules etc.

stacktraceghost:
I am impressed by the speed and feedback Biome gives and I fully agree with @oliver3 .

PR #2564, opened 2025-12-02
Update to Vitest 4.0

### LINKED ISSUE #2546: Upgrade to Vitest 4.0
https://vitest.dev/blog/vitest-4.html

Lionqueen94:
Closes #2546


frontend/package-lock.json
```diff
@@ -19,42 +19,43 @@
...
-        "@storybook/react-vite": "^10.0.6",
+        "@storybook/addon-vitest": "^10.1.3",
+        "@storybook/react-vite": "^10.1.3",
         "@testing-library/dom": "^10.4.1",
         "@testing-library/jest-dom": "^6.9.1",
         "@testing-library/react": "^16.3.0",
         "@testing-library/user-event": "^14.6.1",
         "@trivago/prettier-plugin-sort-imports": "^6.0.0",
-        "@types/node": "^22.19.1",
+        "@types/node": "^24.10.1",
```
praseodym: We're not using Node.js 24, so this change should be reverted

frontend/package.json
praseodym, line 1: There are not just Vitest updates in this diff, please undo the other changes

frontend/src/testing/setup.ts
```diff
@@ -47,4 +47,5 @@ afterEach(() => {
   cleanup();
   server.resetHandlers();
   server.events.removeAllListeners();
+  vi.clearAllMocks();
```
praseodym, line 50: What changed that requires this call?

Lionqueen94: The navigate mock does not get cleared anymore, causing 34 tests to fail. Adding this fixed the issue. There are a lot of [changes to mocking](https://vitest.dev/guide/migration.html#changes-to-mocking) so we think it has to do with that

frontend/package-lock.json
Lionqueen94: Done in 0d49f13f812b9c5796c8dc3c2b2f1aea20e2312c

frontend/package.json
Lionqueen94: Done, the storybook updates are still needed to work with the updated vitest

praseodym: Thanks, the `package-lock.json` diff looks a lot cleaner now

praseodym (approved):

Issue #2541, opened 2025-11-28
More strict `YesNo` struct


oliver3:
To prevent [issues such as these](https://github.com/kiesraad/abacus/pull/2538/commits/6996f007e49b62f4495e7016ef8f6cb118b54415), we should investigate if the `YesNo` can be used in a more strict way, i.e. making `yes` and `no` private and only accessible through a method like `as_bool() -> Result<Option<bool>>` that enforces dealing with the both set or neither set situations at the usage location.

PR #2583, opened 2025-12-04
Make YesNo fields private

oliver3:
Make YesNo fields private and accessible through helper functions to enforce unwrapping before using their values.

- Resolves https://github.com/kiesraad/abacus/issues/2541
- Make `yes` and `no` private fields and add helper functions to enforce dealing with an `Option` as seen in `eml_510.rs`
- Add helpers `is_empty()` and `is_both()` to check for validity without unwrapping their value
- Add helpers  `new()`, ~~`empty()`~~ and `both()` for tests and generator (on top of the existing `yes()` and `no()`)
- The `FieldPath` for comparing `YesNo` structs does not include `.yes` or `.no` anymore which is consistent with the validation paths, ~~and `CheckboxesSubsection.tsx` has been updated accordingly~~

PS: Don't forget to Show diff for `validation.rs`, apparently 3370 lines of Rust code is by default collapsed in the Files changed 😅

### DIFF
--- a/backend/src/data_entry/structs.rs
+++ b/backend/src/data_entry/structs.rs
@@ -12,6 +12,7 @@ use crate::{
     election::{CandidateNumber, PGNumber, PoliticalGroup},
     error::ErrorReference,
 };
+pub use yes_no::YesNo;
 
 #[derive(Serialize, Deserialize, Clone, ToSchema, Debug, FromRow, Default)]
 #[serde(deny_unknown_fields)]
@@ -348,31 +349,22 @@ pub struct CSOFirstSessionResults {
 impl CSOFirstSessionResults {
     /// The admitted voters have been recounted in this session
     pub fn admitted_voters_have_been_recounted(&self) -> bool {
-        self.counting_differences_polling_station
-            .unexplained_difference_ballots_voters
-            .yes
-            || self
-                .counting_differences_polling_station
+        matches!(
+            self.counting_differences_polling_station
+                .unexplained_difference_ballots_voters
+                .as_bool(),
+            Some(true)
+        ) || matches!(
+            self.counting_differences_polling_station
                 .difference_ballots_per_list
-                .yes
-            || self
-                .differences_counts
+                .as_bool(),
+            Some(true)
+        ) || matches!(
+            self.differences_counts
                 .difference_completely_accounted_for
-                .no
-    }
-
-    /// The results have been investigated for another reason
-    pub fn investigated_other_reason(&self) -> bool {
-        self.extra_investigation
-            .extra_investigation_other_reason
-            .yes
-    }
-
-    /// There have been recounted ballots in this session
-    pub fn ballots_have_been_recounted(&self) -> bool {
-        self.extra_investigation
-            .ballots_recounted_extra_investigation
-            .yes
+                .as_bool(),
+            Some(false)
+        )
     }
 }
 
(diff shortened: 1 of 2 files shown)


Lionqueen94 (approved):

praseodym (commented):
Looks good but there are still a few frontend test failures that need to be resolved

backend/src/data_entry/api.rs
```diff
@@ -188,21 +188,8 @@ fn initial_current_data_entry(
...
-                .compare_votes_cast_admitted_voters
-                .votes_cast_smaller_than_admitted_voters = false;
-            copy.differences_counts
-                .difference_completely_accounted_for
-                .yes = false;
-            copy.differences_counts
-                .difference_completely_accounted_for
-                .no = false;
+            copy.differences_counts.compare_votes_cast_admitted_voters = Default::default();
+            copy.differences_counts.difference_completely_accounted_for = Default::default();
```
stacktraceghost, line 192:
```suggestion
            copy.differences_counts.difference_completely_accounted_for = YesNo::empty();
```

Should we use `empty()` for this `YesNo` here as well?

backend/src/eml/eml_510.rs
```diff
@@ -258,17 +258,25 @@ fn create_investigations_from_results(
     let mut investigations = Vec::new();
 
     if let Some(first_session_result) = result.as_cso_first_session() {
-        if first_session_result.investigation_other_reason_is_answered() {
+        if let Some(extra_investigation_other_reason) = first_session_result
```
stacktraceghost, line 260: Note there is a behavioural change here. Before `investigation_other_reason_is_answered()` would return true if `yes || no`, so also if they were both true. Then `first_session_result.investigated_other_reason()` below would return `.yes` (so true if yes is checked, false if yes is not checked). So if both were checked, an item would be pushed with value true.

In the new situation, this if-statement only passes if yes or no is checked, **but not both**. And then it puts true if yes was checked and false if no was checked. So the case where both checkboxes were checked and an item would be pushed with value true is going away here.

I just want to check if this is what we expect it to do? Same on the next statement.

backend/src/data_entry/structs.rs
stacktraceghost: Continuing on the previous note, I also want to point out here that this statement also returns true if both checkboxes are checked. I was thinking about changing this to `as_bool()` as well, but then it would only pass if `Y=true/N=false`, while it currently also passes if `Y=true/N=true`. Same in `ballots_have_been_recounted` below.

oliver3:
> Looks good but there are still a few frontend test failures that need to be resolved

Reverted the frontend change because Checkboxes are not just for yes/no so checking for errors/warnings on separate options is still needed (e.g. `compare_votes_cast_admitted_voters`)

backend/src/eml/eml_510.rs
oliver3: When we are generating the EML structure, it should not be possible for yes and no to both be true. 

Ideally we should have a different data representation after the validation, that cannot represent it anymore (good practices like "parse, don't validate" and "make illegal states unrepresentable"). 

When trying `Result<Option<bool>>` to make the distinction, this resulted in less clear code, which was not worth it for a situation that should not occur.

backend/src/data_entry/api.rs
oliver3: Reverted to just having default() for the empty state

backend/src/data_entry/structs.rs
oliver3: Thanks for pointing these out. These methods are all used either in mod `summary` or `eml` where they should have been validated and true/true should not occur anymore, and should use `as_bool()`.

I've moved the `YesNo` struct into its own mod to catch usages in other data entry structs as well in https://github.com/kiesraad/abacus/pull/2583/commits/8b3a9f526c05a3c36747424148bc0f1716d92b52

PR #2590, opened 2025-12-05
Changed Modal close button title from `cancel` to `close`

### LINKED ISSUE #2571: Rename modal close button to something other than Annuleren
The "close" button (top-right &times;) in our `Modal` has as the accessible name "Annuleren", which is also a common word to put in one of the buttons inside of the modal. In these cases our RTL and Playwright tests have to make a distinction between the two buttons by selecting either the first or the second one. 

```ts
this.closeButton = page.getByRole("dialog").getByRole("button", { name: "Annuleren" }).first();
this.cancelButton = page.getByRole("dialog").getByRole("button", { name: "Annuleren" }).nth(1);
```

```ts
await user.click(screen.getAllByRole("button", { name: "Annuleren" })[0]!);
```

We can improve this situation by giving the close button an accessible name like "Sluiten" instead.

Image

Lionqueen94:
Closes #2571

- Modal X button title is now `Venster sluiten`


stacktraceghost (dismissed):

jorisleker (changes requested):
See comment in #2590

lama-d3v (dismissed):
Tested all close buttons that were changed in the deployment. Also reviewed the changed code. Looks good to me 👍

PR #2591, opened 2025-12-05
Table.LinkRow integration into Table.Row and added download icon option to Table.ClickRow

### LINKED ISSUE #2499: Change chevron to download icon for empty models downloads
The chevrons in the empty models rows should be changed to download icons:
Image

### LINKED ISSUE #2570: Simplify using either `Table.LinkRow` or `Table.Row` with same content depending on condition
There are more and more situations where a table row in the code should be a navigation link or not (depending e.g. on the role of the logged in user), but the content of the row is the same. Currently this is implemented as

```tsx
        {condition ? (
          <Table.LinkRow to={url}>{rowContent}</Table.LinkRow>
        ) : (
          <Table.Row>{rowContent}</Table.Row>
        )}
```

We should find another way to prevent this duplication of content, e.g. by adding an optional `to` prop to the Row that can take either a `To` or undefined or false, so that you can do something like
```tsx
  <Table.Row to={condition && url}>{rowContent}</Table.Row>
```

Lionqueen94:
Closes #2570
Closes #2499

### Description
- Removed `Table.LinkRow` and converted `Table.Row` to include an optional link parameter
- Added download icon option to `Table.ClickRow`
- [x] @jorisleker Can you update the Figma?
image

### DIFF
--- a/frontend/src/components/ui/Table/Table.module.css
+++ b/frontend/src/components/ui/Table/Table.module.css
@@ -67,9 +67,7 @@
     &:hover {
       background: var(--color-hover);
     }
-  }
 
-  tr.chevronIcon {
     /* Cell with the chevron icon */
     td:last-of-type {
       padding-right: 4.5rem;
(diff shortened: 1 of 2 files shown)


jorisleker:
Thanks @Lionqueen94.
Icon in Figma [has been updated ](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=9780-54761&t=yN0YQqgR79g2FK9m-11)

Issue #2555, opened 2025-12-01
Add newtypes around ids in the backend


rnijveld:
- Currently our ids are mostly just u32 or some similar integer type
- This can cause confusion for methods requiring a single id, and has in the past caused bugs to occur because of confusion for which id was supposed to be passed.

A better alternative for Rust code:
- Use a newtype wrapper (e.g. `struct PollingStationId(u32)`) for each of the different id types we have
- Make sure the wrapped type can be easily exposed for those cases where we need the original id
- For inspiration we can use the [code I wrote for the nts-pool project](https://github.com/pendulum-project/nts-pool/blob/main/nts-pool-management/src/models/util/mod.rs) that generates all the relevant wrapper utilities for id newtypes.

praseodym:
Would it be a good idea to add a todo list to the issue with all the newtypes that we want to add?

Issue #2557, opened 2025-12-01
Switch to uuids for all internal ids


rnijveld:
Currently we use regular integer values for all ids. This has some disadvantages:
- Ids can be accidentally switched out for each other, where an id can be used in a place where it wasn't supposed to be
- Integer ids are susceptible to things such as bitflips or data corruption accidentally causing unintended behavior
- Partial data from one abacus installation might accidentally map onto another abacus installation

uuids remove most of these issues, a bitflip/data corruption is highly unlikely to result in another valid uuid, partial data is highly unlikely to map onto another instance, and accidentally switching ids for different types is very unlikely with uuids.

Disadvantage: some people find uuids less readable than regular integer values, but generally developers don't deal with literal uuids too often. For tests this is generally pretty easy to solve by using deterministic uuid generation and uuids stored in constants. They could also somewhat clutter urls in browsers, but we typically don't have more than one or two ids in a URL, making them not much longer than the typical URL users will generally see while browsing the web anyway.

jorisleker:
Noticed that we currently show the ids of users in the activity log overview table. Think we should remove this from the table when implementing this issue, because showing uuid's here will break the UI. Have not checked if there are other places in the application were we show id's in the interface, but I think the 'nummer' column on this same page might also be an id that is affected by this change

<img width="1233" height="598" alt="Image" src="https://github.com/user-attachments/assets/2e119bcd-a5c7-47b1-a1ca-b4734f391046" />

praseodym:
Let's implement newtypes first (#2555) and then reconsider picking up this one.

PR #2597, opened 2025-12-09
Allow gaps in PG/Candidate numbers, but enforce increasing order

### LINKED ISSUE #2549: Do not use candidate number and political group number for indexing
We currently use candidate number and political group number for indexing, but we should not do this. The assumption was made that candidate and political group numbers always start at 1 and always are sequential. For country wide elections this can be untrue because not every political group participates in every electoral district (causing holes in the normally sequential group numbers). Additionally, the kieswet allows for political groups and candidates to be removed (in exceptional cases) after assignment of numbers, in which case that assumption also does not hold. Additionally, this should just be a label value anyway, and it is just bad practice to use it for calculations.

Some of this work was already done on [this commit](https://github.com/kiesraad/abacus/commit/dadb2e7724507232f508f970146e43044fb5ef9d) on the `do-not-use-pg-number-for-indexing` branch.

Optionally: store the pg and candidate numbers as strings (or even better on the rust side: a newtype containing a string) and just add some validation on importing that they are numbers, making it unlikely we ever use these for indexing ever again.

michielp1807:
TODO:
- [x] Allow gaps in backend
- [x] Add tests to test gaps are allowed, but increasing order is enforced
- [x] Fix Playwright tests
- [x] Make sure that gaps do not cause issues in the front-end (add some new tests?)

Closes #2549

Also changes the types `PGNumber` and `CandidateNumber` to newtypes to make sure they are not used for indexing (somewhat related to #2555).

### DIFF
--- a/backend/tests/investigation_integration_test.rs
+++ b/backend/tests/investigation_integration_test.rs
@@ -8,7 +8,7 @@ use abacus::{
         VotersCounts, VotesCounts,
         status::{ClientState, DataEntryStatusName},
     },
-    election::{ElectionDetailsResponse, ElectionId, PGNumber},
+    election::{ElectionId, PGNumber},
 };
 use axum::http::StatusCode;
 use reqwest::Response;


backend/src/data_entry/validation.rs
```diff
@@ -2903,13 +2927,29 @@ mod tests {
...
-            assert!(result.is_err());
-            assert!(
-                result
-                    .unwrap_err()
-                    .message
-                    .eq("candidate numbers are not consecutive"),
+            assert!(result.is_ok());
+
+            // Change number of the second candidate on the first list to a non-increasing number
+            political_group_votes[0].candidate_votes[1].number = CandidateNumber::new(1);
```
stacktraceghost:
```suggestion
```

These lines seem left behind/unnecessary

backend/src/election/structs.rs
```diff
@@ -169,7 +169,30 @@ pub enum VoteCountingMethod {
     DSO,
 }
 
-pub type PGNumber = u32;
+#[derive(
+    Serialize, Deserialize, Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash,
+)]
+pub struct PGNumber(u32);
```
stacktraceghost: We can use `crate::util::id!(PGNumber)` / `crate::util::id!(CandidateNumber)` for this.

backend/src/data_entry/comparison.rs
```diff
@@ -329,7 +331,7 @@ mod tests {
             },
             votes_counts: VotesCounts {
                 political_group_total_votes: vec![PoliticalGroupTotalVotes {
-                    number: 1,
+                    number: PGNumber::new(1),
```
stacktraceghost: The implementation for `ElectionId` is using `ElectionId::from(1)`, lets use one solution.

backend/src/election/structs.rs
michielp1807: Thanks! I updated it to use the `id!` macro. When #2557 gets implemented, we will likely need a separate macro for the `PGNumber`s and `CandidateNumber`s and the other internal IDs which can use UUIDs (e.g. a `numeric_id!` and a `uuid!` macro).

michielp1807:
Everything in the front-end seems to work as expected with gaps in the party numbers and candidate numbers:

image

I've also added a Playwright test for an election with gaps in the party numbers and candidate numbers.

praseodym (dismissed):
Code and tests look good, but I haven't tested by using the frontend myself

praseodym (approved):

PR #2605, opened 2025-12-10
Bump wasmi from 0.51.2 to 0.51.5 in /backend (backport of #2596)

praseodym:
Backport of #2596 to the 1.0 release branch. This updates wasmi from 0.51.2 to 0.51.5 to fix GHSA-g4v2-cjqp-rfmq.

I don't think this vulnerability affects Abacus because we don't use any WebAssembly plugins for Typst, but to avoid any confusion (e.g. `cargo audit` reporting the vulnerability) I think it is good to also fix it in 1.0.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -28,6 +28,21 @@ jobs:
       CARGO_TERM_COLOR: always
       SQLX_OFFLINE: "true"
     steps:
+      # Free up runner disk space, so that our job has enough space to run
+      # https://dev.to/mathio/squeezing-disk-space-from-github-actions-runners-an-engineers-guide-3pjg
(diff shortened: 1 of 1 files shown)


Lionqueen94:
@praseodym I don't understand why the pipeline fails, when I locally checkout this branch, reset my database and run the backend tests, there are no issues.

praseodym:
> @praseodym I don't understand why the pipeline fails, when I locally checkout this branch, reset my database and run the backend tests, there are no issues.

The CI runner's disk is full, I'll also have to backport a change to free up disk space at the start of the test run.

praseodym:
> The CI runner's disk is full, I'll also have to backport a change to free up disk space at the start of the test run.

See #2608

Issue #2602, opened 2025-12-10
Password can be changed without entering current password


praseodym:
The users' password can be changed without entering current password. 

We should investigate for which roles this happens (only admin, or also typist?) and come up with a minimal fix that can be backported to the 1.0 release branch.

jschuurk-kr:
Tested on commit with tag v1.0.0, d220eb9f2b857ea4fdaa7ed7f2e1e7a994ac1a1b.

The issue only applies to users with role admin. They can change any user's password to a temporary password, so that includes their own account. So the steps to reproduce are:

1. Log in as admin user
2. Go to user management (`/users`)
3. Go to the details page of the logged in user
4. Click "Wijzig wachtwoord" and change the password
5. The user is logged out automatically
6. Log in with your temporary password
7. Set a new password
8. User is logged in again

Possible solutions:
1. Leave as is. Admin users can reset any user's password. An admin user resetting their own password should not be harder than resetting another admin's password.
2. Do not allow an admin user to reset their own password. They can only reset the password on other people's accounts.
3. Make the admin user confirm their current password before resetting (a) their own password; (b) any admin account's password; (c) any account's password (then also for changes by coordinators).
4. ...?

The reason this issue only applies to admin accounts, is because Abacus does not have any functionality for a user to reset their own password. So password resets can only happen through user management. A coordinator can only manage typist accounts. An admin can manage all accounts. Hence, the only way for a user to change their own account is by an admin using the account management-feature on their own account.

praseodym:
For all roles it is possible to manually navigate to the `/account/setup` URL and change your own password, even after it has initially been changed. The full name can also be changed even when already set.

Related finding: when the administrator sets a temporary password for a user, the user gets redirect to `/account/setup` after logging in. This page then also allows the full name to be changed, even when it was already set. Should this be possible?

Related to #2606, which changes the account setup flow.

praseodym:
> The issue only applies to users with role admin. They can change any user's password to a temporary password, so that includes their own account.

I've created #2619 for that issue.

This issue should cover only the account setup endpoint (https://github.com/kiesraad/abacus/issues/2602#issuecomment-3641228307).

stacktraceghost:
The deliverable for this issue is:
- Only users who are incomplete should be able to access `/account/setup`.

PR #2606, opened 2025-12-10
Enforce password change when needed

### LINKED ISSUE #2601: Password change on first login can be skipped
The mandatory password change on first login can be skipped by going to a different URL. Abacus should block subsequent API requests, but currently does not. For the bugfix in 1.0 the frontend UX does have to be optimal, blocking API requests when the password has not yet been changed is sufficient.

We should investigate why this happens and come up with a minimal fix that can be backported to the 1.0 release branch.

oliver3:
- Fixes #2601 
- Fix test assertion for `UNAUTHORIZED` when password needs change
- Enforce `fullname` and no `needs_password_change` for the `User` `from_request_parts`
- Introduce `IncompleteUser` to be able to not enforce it for updating the user account's fullname and password

### DIFF
--- a/backend/src/authentication/user.rs
+++ b/backend/src/authentication/user.rs
@@ -137,7 +137,7 @@ where
     }
 }
 
-/// Implement the FromRequestParts trait for PartialUser,
+/// Implement the FromRequestParts trait for IncompleteUser,
 /// for endpoints that are needed to fully set up the account
 impl<S> FromRequestParts<S> for IncompleteUser
 where


praseodym:
Looks good!

@marlonbaeten could you take a look at this? Specifically, do you think this is a good way to fix #2602 and do you see an even shorter way to do so (we want to keep backports to 1.0 as small as possible)?

backend/src/authentication/role.rs
```diff
@@ -48,6 +48,10 @@ pub struct Typist(pub User);
 #[allow(unused)]
 pub struct AdminOrCoordinator(pub User);
 
+/// A user with potentially no fullname or needs_password_change=true
+#[allow(unused)]
+pub struct PartialUser(pub User);
```
stacktraceghost: I would like to propose a different name. Maybe something like `UserRequiresAction`, `UserNotReady`.

stacktraceghost (commented):
✅ Tested if I can still go to other pages before finishing the setup of my account. This is not possible anymore. Tested by:
1. Logging in as admin
2. Creating a new user
3. Logging in as the new user
4. Going to `/elections`, which now redirects me to the login page

Other things I did notice, but not neccesary for this PR:
- When creating a new user or setting up your account and entering the same password as your username, you get the wrong error message that your password is not long enough. 
- `/account/setup` is always accessible for users and you can change your password there. However there is no link to it?

backend/src/authentication/user.rs
```diff
@@ -129,10 +129,31 @@ where
             return Err(AuthenticationError::Unauthenticated.into());
         };
 
+        if user.fullname.is_none() || user.needs_password_change() {
```
praseodym, line 132: To check: does the full name come back as `None`, or `Some("")` (empty string)?

backend/src/authentication/role.rs
oliver3: Would `IncompleteUser` work?

backend/src/authentication/user.rs
oliver3: For a newly created anonymous user:

```
 &user = User {
    id: 7,
    username: "gebruiker",
    fullname: None,
    role: Typist,
    needs_password_change: true,
    ..
```

backend/src/authentication/role.rs
oliver3: Renamed in https://github.com/kiesraad/abacus/pull/2606/commits/04f9b85d809f5e55d4a3ea78c2293f520be5d318

backend/src/authentication/user.rs
```diff
@@ -129,10 +129,31 @@ where
...
+            return Err(AuthenticationError::Unauthenticated.into());
+        }
+
         Ok(user.clone())
     }
 }
 
+/// Implement the FromRequestParts trait for PartialUser,
```
Lionqueen94:
```suggestion
/// Implement the FromRequestParts trait for IncompleteUser,
```

Lionqueen94 (commented):
Tested trying to go to another page as a new user and you now get logged out and redirected to the login page 👍  also code looks good, besides the one comment above

backend/src/authentication/user.rs
oliver3: Thanks!

oliver3:
> * `/account/setup` is always accessible for users and you can change your password there. However there is no link to it?

This is also mentioned in https://github.com/kiesraad/abacus/issues/2602#issuecomment-3641228307 so let's continue the discussion there and not fix it in this PR.

oliver3:
>     * When creating a new user or setting up your account and entering the same password as your username, you get the wrong error message that your password is not long enough.

Created issue https://github.com/kiesraad/abacus/issues/2614

PR #2607, opened 2025-12-10
Do not allow multiple sessions per user

### LINKED ISSUE #2603: Do not allow multiple sessions per user
Investigate what needs to be done to allow only a single session per user (i.e. invalidate other sessions when logging in). Check if this breaks tests and how this affects the UX. Preferably with a minimal implementation that can be backported to the 1.0 release branch.

stacktraceghost:
Resolves #2603

With this PR, all other sessions of the user that successfully logs in will be removed and thus invalidated.

Changes:
- Deletes all other sessions of the user that sucessfully logs in
- Fixes integration tests that fail due to this change, since common functions often log themselves in
- Added an integration test to check if the old session gets invalidated and if a login of user B does not affect user A
- E2E test fixtures will now use sessions from storage states instead of using `loginAs` themselves, since this was invalidating the existing sessions of the storage states.
  - As a result of this, there are less login actions and the E2E tests run faster

Question: invalidation of an old session does not add a record to the audit like a logout does, do we want this?


praseodym:
Open question: do we want to create a database constraint to prevent more than one session per user?

stacktraceghost:
> Open question: do we want to create a database constraint to prevent more than one session per user?

Maybe in main with some error handling? Since there is only one way to login, the business logic should ensure a single session.

frontend/e2e-tests/tests/authentication.e2e.ts
```diff
@@ -12,14 +13,14 @@ test.describe("authentication", () => {
     await page.goto("/account/login");
 
     const loginPage = new LoginPgObj(page);
-    await loginPage.username.fill("admin1");
-    await loginPage.password.fill("Admin1Password01");
+    await loginPage.username.fill("admin2");
+    await loginPage.password.fill(getTestPassword("admin2"));
```
jschuurk-kr, line 17: Are we using the admin2 user here to avoid a conflict with other tests if we would use the admin1 user?

If so, it would be good to add a comment to the test. Or create a fresh user through a fixture, similar to the `newTypist` fixture.

frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -55,16 +61,50 @@ const investigations = [
...
+      await expect(userListPgObj.alert).toContainText(`${username} is toegevoegd met de rol Invoerder`);
+    }
+  });
+
   test("create election and a new polling station", async ({ page }) => {
     await page.goto("/account/login");
 
     const loginPage = new LoginPgObj(page);
-    await loginPage.login("admin1", getTestPassword("admin1"));
+    await loginPage.login("admin2", getTestPassword("admin2"));
```
jschuurk-kr, line 107: Same question as in other test: is this to avoid a conflict with other tests? If so, we need a comment or a fixture setting up a new user.

frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -121,7 +161,7 @@ test.describe("full flow", () => {
     await page.goto("/account/login");
 
     const loginPage = new LoginPgObj(page);
-    await loginPage.login("coordinator1", getTestPassword("coordinator1"));
+    await loginPage.login("coordinator2", getTestPassword("coordinator2"));
```
jschuurk-kr, line 164: Same question as in other test: is this to avoid a conflict with other tests? If so, we need a comment or a fixture setting up a new user.

frontend/e2e-tests/tests/full-flow.e2e.ts
jschuurk-kr: If we want to do more with fixtures in the full flow test, it might make sense to use [`test.step`](https://playwright.dev/docs/api/class-test#test-step) instead of `test` for the separate parts of the test.

frontend/e2e-tests/tests/authentication.e2e.ts
stacktraceghost: Yes, this is to avoid conflict, same below. The problem with the `newTypist` fixture is that the user is not complete yet (password needs to be set), which increases the lines of code a lot. We need that for the full flow. 

This PR changes one line of code, but subsequently needs to adjust many tests. I'd like to propose to do this improvement in a separate PR to main.

jschuurk-kr: Makes sense, let's do a separate PR.

frontend/e2e-tests/tests/full-flow.e2e.ts
jschuurk-kr: see https://github.com/kiesraad/abacus/pull/2607#discussion_r2610040549

frontend/e2e-tests/tests/full-flow.e2e.ts
jschuurk-kr: see https://github.com/kiesraad/abacus/pull/2607#discussion_r2610040549

stacktraceghost:
Merge conflicts resolved. When we backport this PR we should cherry pick the first 4 commits.

stacktraceghost:
> Question: invalidation of an old session does not add a record to the audit like a logout does, do we want this?

Answer: good this way. The audit log has enough information.

Issue #2604, opened 2025-12-10
Log uploaded file hashes to audit log


praseodym:
Log uploaded EML_NL file hashes (election definition, candidates list, polling stations list) to audit log. This should be a minimal change that can be backported to the 1.0 release branch.

PR #2610, opened 2025-12-11
Log imported EML_NL hashes to audit log

oliver3:
- Fixes https://github.com/kiesraad/abacus/issues/2604
- Log election and candidates EML hashes to audit log
- Log polling station EML hash to audit log
  - For polling stations import: change `polling_stations/import` request to include original file data instead of previously validated polling stations, to ensure that the hash is not tampered with
  - For election import: keep validation before creating election, but use original data when creating polling stations to be able to log the hash
- Added `allow(clippy::too_many_lines)` to reduce the number of changes in this PR

To test:
- Import an election with and without polling stations, see that this still works, and see that the hashes are shown in the audit log
- Import polling stations, see that this still works, and see that the hash is shown in the audit log


backend/tests/polling_station_integration_test.rs
```diff
@@ -42,14 +42,14 @@ async fn import_polling_stations(
     cookie: &HeaderValue,
     election_id: u32,
     file_name: &str,
-    polling_stations: serde_json::Value,
+    data: serde_json::Value,
```
praseodym: Maybe leave this change to a future PR that we don't need to backport?

frontend/src/i18n/locales/nl/log.json
```diff
@@ -76,6 +76,7 @@
     "fullname": "Volledige naam",
     "import_election_id": "Verkiezings-ID",
     "import_file_name": "Bestandsnaam",
+    "import_file_hash": "Bestandshash",
```
praseodym:
```suggestion
    "import_file_hash": "Bestandshash (SHA256)",
```

backend/src/audit_log/audit_event.rs
```diff
@@ -114,6 +114,7 @@ pub struct PollingStationDetails {
 pub struct PollingStationImportDetails {
     pub import_election_id: u32,
     pub import_file_name: String,
+    pub import_file_hash: String,
```
praseodym: As discussed: undo this change and log the hash in the audit log message instead

frontend/src/i18n/locales/nl/log.json
praseodym: No longer relevant when https://github.com/kiesraad/abacus/pull/2610/files#r2609822722 is changed

backend/tests/polling_station_integration_test.rs
praseodym: As discussed for this PR we keep the field name unchanged, to limit the changes we need to backport

backend/src/audit_log/audit_event.rs
oliver3: Done in https://github.com/kiesraad/abacus/pull/2610/commits/7d34d5c2adcec8d6f8605e19e028b3dd379c7263 to reduce number of changes in this PR

backend/tests/polling_station_integration_test.rs
oliver3: Done in https://github.com/kiesraad/abacus/pull/2610/commits/7d34d5c2adcec8d6f8605e19e028b3dd379c7263 to reduce number of changes in this PR

Lionqueen94 (approved):
Tested by importing a new election with and without polling stations, works well.

backend/src/polling_station/mod.rs
```diff
@@ -406,7 +404,10 @@ pub async fn create_imported_polling_stations(
                 import_number_of_polling_stations: u64::try_from(polling_stations.len())
                     .map_err(|_| EMLImportError::NumberOfPollingStationsNotInRange)?,
             }),
-            None,
+            Some(format!(
+                "Polling stations file hash: {}",
```
lama-d3v, line 409: Maybe add translations for this?

backend/src/election/api.rs
```diff
@@ -391,11 +392,16 @@ pub async fn election_import(
     // Create new election
     let mut tx = pool.begin_immediate().await?;
     let election = crate::election::repository::create(&mut tx, new_election).await?;
+    let message = format!(
+        "Election file hash: {}, candidates file hash: {}",
```
lama-d3v, line 400: Maybe add translations for this in order to make it more readable for administrators/coordinators in the logs?

backend/src/polling_station/mod.rs
oliver3: I suggest not to do that for this PR, to keep the backport to 1.0 as small as possible

backend/src/election/api.rs
oliver3: I suggest not to do that for this PR, to keep the backport to 1.0 as small as possible

Lionqueen94 (commented):
Please fix the merge conflict @oliver3 :)

PR #2615, opened 2025-12-12
Change structs in integration tests to json

### LINKED ISSUE #2191: Change structs in integration tests to JSON
According to the [test-tooling documentation](https://github.com/kiesraad/abacus/blob/main/documentatie/ontwikkelproces/test-tooling.md#integratietests), integration tests should use JSON and never structs. In multiple integration tests, structs are used.

We could:
- Change structs to JSON
- Make the documentation more explicit about usage of structs: only in requests or also for responses?

Lionqueen94:
Closes #2191 

## Fixed
- tests/
  - shared/
    - [x] mod.rs
  - utils/
    - [ ] mod.rs -> to be fixed in https://github.com/kiesraad/abacus/issues/2628
  - [ ] airgap_detection_test.rs -> to be fixed in https://github.com/kiesraad/abacus/issues/2628
  - [ ] authorization_test.rs -> to be fixed in https://github.com/kiesraad/abacus/issues/2627
  - [x] committee_session_integration_test.rs
  - [x] data_entries_integration_test.rs
  - [x] data_entries_resolve_integration_test.rs
  - [x] election_admin_integration_tests.rs
  - [x] election_integration_test.rs
  - [x] investigation_integration_test.rs
  - [x] polling_station_integration_test.rs
  - [x] user_integration_test.rs

### DIFF
--- a/backend/tests/committee_session_integration_test.rs
+++ b/backend/tests/committee_session_integration_test.rs
@@ -103,7 +103,7 @@ async fn test_committee_session_delete_ok_status_created(pool: SqlitePool) {
 
     let committee_session =
         get_election_committee_session(&addr, &coordinator_cookie, election_id).await;
-    assert_eq!(committee_session["id"], 704);
+    assert_eq!(committee_session["id"], committee_session_id);
     assert_eq!(committee_session["status"], "created");
 
     let response = delete_committee_session(


backend/tests/authorization_test.rs
Lionqueen94: @praseodym This one still has an import of an enum and a function, so no struct, is that ok, or do all imports to abacus have to be removed?

praseodym: No we should also remove those imports

backend/tests/shared/mod.rs
```diff
@@ -2,109 +2,98 @@
...
-        PoliticalGroupTotalVotes, PollingStationResults, VotersCounts, VotesCounts, YesNo,
-        status::{ClientState, DataEntryStatusName},
-    },
-    election::{CandidateNumber, ElectionDetailsResponse, ElectionId, PGNumber},
-};
-
 use axum::http::{HeaderValue, StatusCode};
 use hyper::header::CONTENT_TYPE;
-use reqwest::{Body, Response};
-use serde_json::json;
```
praseodym, line 22: I think this import was fine (unqualified `json!` is just as clear as `serde_json::json!`), but `Value` is ambigious so for that one `serde_json::Value` is preferred.

praseodym (approved):

backend/tests/shared/mod.rs
Lionqueen94: It's now fully consistent (always using `serde_json::`) but I'll keep this in mind for the future 👍

backend/tests/committee_session_integration_test.rs
```diff
@@ -89,24 +98,21 @@ async fn test_committee_session_create_current_committee_session_not_finalised(p
...
-    let url = format!(
-        "http://{addr}/api/elections/{election_id}/committee_sessions/{committee_session_id}"
-    );
-    let response = reqwest::Client::new()
-        .delete(&url)
-        .header("cookie", &coordinator_cookie)
-        .send()
-        .await
-        .unwrap();
+    assert_eq!(committee_session["id"], 704);
```
lama-d3v:
```suggestion
    assert_eq!(committee_session["id"], committee_session_id);
```

backend/tests/data_entries_integration_test.rs
```diff
@@ -525,14 +543,9 @@ async fn test_election_details_status_no_other_election_statuses(pool: SqlitePoo
         .unwrap();
 
     assert_eq!(response.status(), StatusCode::OK);
-    let body: ElectionStatusResponse = response.json().await.unwrap();
-
-    assert_eq!(
-        body.statuses.len(),
-        1,
-        "there can be only one {:?}",
```
lama-d3v, line 533: This is removed, is that correct?

backend/tests/data_entries_resolve_integration_test.rs
```diff
@@ -273,76 +256,57 @@ async fn test_polling_station_data_entry_get_differences(pool: SqlitePool) {
...
-        .voters_counts
-        .poll_card_count
-        + 2;
+    second_data_entry["data"]["voters_counts"]["poll_card_count"] = serde_json::Value::from(
+        first_data_entry["data"]["voters_counts"]["poll_card_count"]
+            .as_u64()
+            .unwrap()
+            - 2,
+    );
+    second_data_entry["data"]["voters_counts"]["proxy_certificate_count"] = serde_json::Value::from(
```
lama-d3v, line 292: is it correct that `proxy_certificate_count` and `poll_card_count` are used here?

backend/tests/data_entries_integration_test.rs
Lionqueen94: This check is located on line 548 now

backend/tests/data_entries_resolve_integration_test.rs
Lionqueen94: Yes this is fine, you need them to create a difference between the first and second data entry

backend/tests/committee_session_integration_test.rs
Lionqueen94: Fixed in ea180099d437cfbbd0470f7d666acf3e2b8c9abf

PR #2626, opened 2025-12-15
Do not allow multiple sessions per user (backport of #2607)

stacktraceghost:
Backport of https://github.com/kiesraad/abacus/pull/2607 to the 1.0 release branch.


Lionqueen94 (approved):

Lionqueen94:
I think the full flow e2e test is flaky, But it doesn't make sense to me why that would be

PR #2629, opened 2025-12-16
Fix flaky full-flow test

stacktraceghost:
We've found that the change in the full flow e2e-test in https://github.com/kiesraad/abacus/pull/2626 was flaky:
- [Example 1](https://github.com/kiesraad/abacus/actions/runs/20248405992/job/58135298999)
- [Example 2](https://github.com/kiesraad/abacus/actions/runs/20261832557/job/58176215418) 

~This PR adds some `await expect`'s to make sure we are on the right page before proceeding with the next step.~
This PR changes `getByLabel` by `getByRole` in page objects related to the test.

To reproduce this, add many users by using `const typistUsers = Array.from({ length: 1000 }, (_, i) => `test${i + 1}`);`. If the error doesn't show up locally, it can be reproduced by throttling the CPU. Add the following lines to the changed test and run `npm run test:e2e -- --project chrome --grep "full-flow"`. This will slow down the CPU for Chrome-based browsers by 20x for that test.
```
const context = page.context();
const cdpSession = await context.newCDPSession(page);
await cdpSession.send("Emulation.setCPUThrottlingRate", { rate: 20 });
```

With CPU throttling the test fails at the first loop on my computer. With the fixes from this PR it does not.

### DIFF
--- a/frontend/e2e-tests/page-objects/users/UserCreateDetailsPgObj.ts
+++ b/frontend/e2e-tests/page-objects/users/UserCreateDetailsPgObj.ts
@@ -7,9 +7,9 @@ export class UserCreateDetailsPgObj {
   readonly save: Locator;
 
   constructor(protected readonly page: Page) {
-    this.username = page.getByLabel("Gebruikersnaam");
-    this.fullname = page.getByLabel("Volledige naam");
-    this.password = page.getByLabel("Tijdelijk wachtwoord");
+    this.username = page.getByRole("textbox", { name: "Gebruikersnaam" });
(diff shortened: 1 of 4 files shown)


lama-d3v (dismissed):

Lionqueen94 (dismissed):

praseodym:
Wouldn't it be better to strategically replace some `getByLabel`s by `getByRole`?

stacktraceghost:
> Wouldn't it be better to strategically replace some `getByLabel`s by `getByRole`?

Done, replaced `getByLabel` in the used page objects.

praseodym (approved):

PR #2632, opened 2025-12-16
Add check to IncompleteUser for completeness

### LINKED ISSUE #2602: Password can be changed without entering current password

stacktraceghost:
Resolves #2602 

Adds check to `IncompleteUser` to make sure the user is actually incomplete. By adding this, the api endpoint `PUT /api/account` will not allow a user that has a full name and does not need a password change. 

This breaks several tests that try to call this endpoint with a user that is not incomplete. I've fixed the ones in `authentication/mod.rs` and removed `account_integration_test.rs`, since they duplicate them. A test for updating with the same password was missing in `authentication/mod.rs` (which was originally in `account_integration_test.rs`), so I've added it there. 

To check:
- ~Frontend check for `/account/setup`~ -> https://github.com/kiesraad/abacus/pull/2632/commits/651a75f761b97cbf885c5cd8a9d7c3af0944ea19 // Reverted, will be added to main in a separate PR
- Is the error code / status code okay?
- Test if the change work as expected by creating a user, going through the setup flow and see if you can access it afterwards

### DIFF
--- a/backend/src/authentication/error.rs
+++ b/backend/src/authentication/error.rs
@@ -14,8 +14,8 @@ pub enum AuthenticationError {
     SessionKeyNotFound,
     Unauthenticated,
     Unauthorized,
+    UserAlreadySetup,
     UsernameAlreadyExists,
-    UserNotIncomplete,
     UserNotFound,
 }
 
--- a/backend/src/authentication/user.rs
+++ b/backend/src/authentication/user.rs
@@ -151,7 +151,7 @@ where
         };
 
         if user.fullname.is_some() && !user.needs_password_change() {
-            return Err(AuthenticationError::UserNotIncomplete.into());
+            return Err(AuthenticationError::UserAlreadySetup.into());
         }
 
         Ok(IncompleteUser(user.clone()))
--- a/backend/src/error.rs
+++ b/backend/src/error.rs
@@ -279,7 +279,7 @@ impl IntoResponse for APIError {
                         StatusCode::UNAUTHORIZED,
                         to_error("User not found", ErrorReference::UserNotFound, false),
                     ),
-                    AuthenticationError::UserNotIncomplete => (
+                    AuthenticationError::UserAlreadySetup => (
                         StatusCode::CONFLICT,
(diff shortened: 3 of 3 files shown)


frontend/src/features/account/components/AccountSetupPage.tsx
```diff
@@ -8,15 +9,19 @@ import { LoginResponse } from "@/types/generated/openapi";
...
-    void navigate("/elections#new-account");
+    setSetupComplete(true);
   }
 
-  if (!user) {
+  if (setupComplete) {
+    return <Navigate to="/elections#new-account" />;
+  }
+
+  if (!user || (user.fullname && !user.needs_password_change)) {
```
praseodym: I don't think redirecting to the login page makes sense if the user is logged in, we should have a different kind of error for that.

Maybe we also don't need the frontend checks for the backport PR and can do the frontend part separately?

stacktraceghost: Let's do that, otherwise there will be quite many changes.

stacktraceghost: See https://github.com/kiesraad/abacus/pull/2639

backend/src/authentication/error.rs
```diff
@@ -15,6 +15,7 @@ pub enum AuthenticationError {
     Unauthenticated,
     Unauthorized,
     UsernameAlreadyExists,
+    UserNotIncomplete,
```
praseodym: Maybe `UserAlreadyComplete` or `UserAlreadySetup`?

praseodym (dismissed):

backend/src/authentication/error.rs
Lionqueen94: Renamed in 337bf3727825d6c652081c8e597b5fc47e8afe05

praseodym (approved):

PR #2633, opened 2025-12-16
Added missing committee session status change validations

### LINKED ISSUE #2631: Add missing committee session status change validations
Found in #2214 

Add missing validations (`polling_stations.is_empty()`/`investigations.is_empty()`) for going from any committee session state to `Created`

Lionqueen94:
Closes #2631

### DIFF
--- a/backend/src/committee_session/status.rs
+++ b/backend/src/committee_session/status.rs
@@ -60,7 +60,12 @@ pub async fn change_committee_session_status(
     let committee_session =
         crate::committee_session::repository::get(&mut tx, committee_session_id).await?;
     let new_status = match status {
-        CommitteeSessionStatus::Created => committee_session.status.prepare_data_entry()?,
+        CommitteeSessionStatus::Created => {
+            committee_session
+                .status
+                .prepare_data_entry(&mut tx, &committee_session)
+                .await?
+        }
         CommitteeSessionStatus::DataEntryNotStarted => {
             committee_session
                 .status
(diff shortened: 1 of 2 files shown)


backend/src/committee_session/status.rs
```diff
@@ -135,13 +140,31 @@ pub async fn change_committee_session_status(
...
+            | CommitteeSessionStatus::DataEntryFinished => {
+                let polling_stations =
+                    crate::polling_station::repository::list(conn, committee_session.id).await?;
+                if polling_stations.is_empty() {
+                    return Ok(CommitteeSessionStatus::Created);
+                } else if committee_session.is_next_session() {
+                    let investigations =
+                        list_investigations_for_committee_session(conn, committee_session.id)
+                            .await?;
+                    if investigations.is_empty() {
```
stacktraceghost, line 162:
```suggestion
                    if investigations.is_empty()
                        && polling_stations
                            .iter()
                            .all(|ps| ps.id_prev_session.is_some())
                    {
```

In a scenario where I am in a next committee session and I've added a new polling station (so `id_prev_session=null`) but no investigation yet, I would still be able to go to created. 

According to the diagram, adding (so having) a `PS/Inv` should result into `DataEntryNotStarted`, so my assumption is that `Created` should not be possible then.

Lionqueen94: In a next session the status only changes to `DataEntryNotStarted` once you add an investigation. Or at least it should be this way, in the first session adding a polling station is the trigger for the status change and in a next session adding an investigation is the trigger.

backend/src/committee_session/status.rs
praseodym, line 1: Please consider (in a new PR) shortening the extremely long test names, e.g. by removing redundant information (this is module `committee_session::status` so no need to prefix with `committee_session_status_`) and organizing in a few modules within the test mod.

praseodym (approved):

backend/src/committee_session/status.rs
praseodym, line 51: Please check how this function can be refactored into some smaller parts

backend/src/committee_session/status.rs
stacktraceghost: Oh yes, that is true, thanks!

backend/src/committee_session/status.rs
Lionqueen94: Added https://github.com/kiesraad/abacus/issues/2650 to improve this in the future

Lionqueen94: Added issue https://github.com/kiesraad/abacus/issues/2651 for this

PR #2634, opened 2025-12-16
Workflow updates

Lionqueen94:
## Description
- Fixed 2 other deprecated codecov actions
- Added concurrency rule to `build-lint-test` workflow to cancel running checks when a new commit is pushed to the same pull request

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -17,6 +17,10 @@ permissions:
 env:
   TZ: "Europe/Amsterdam"
 
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   backend:
     name: Backend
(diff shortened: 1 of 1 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -219,3 +219,7 @@ jobs:
         run: cargo install cargo-fuzz
       - name: Run the fuzzer
         run: cargo fuzz run data_entry_status -- -max_total_time=60
+
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
praseodym: Can you move this up in the file, e.g. just before `jobs:`, to make it more clear that this setting is configured

Lionqueen94: Done in 32e6e6b12a025a6a70e7ca1d9329940f838d523c

PR #2643, opened 2025-12-18
Fixed error pages rendering

Lionqueen94:
The error pages were missing the `<AppFrame>` component which made the main not stretch across the whole page.

Example from main:
Screenshot from 2025-12-18 11-28-17

Fixed in this PR:
Screenshot from 2025-12-18 11-28-12


jorisleker (approved):
approval based on screenshot. Good catch and improvement

praseodym (approved):

PR #2657, opened 2025-12-22
Update Figma link to most recent designs

jorisleker:
Update Figma links in documentation to most recent designs. The old public Figma was stale. The new link directly links to the file used by the team during development.


jorisleker:
> Note that it currently still has a password

Yup that needs to be removed before merging this PR

jorisleker:
> > Note that it currently still has a password
> 
> Yup that needs to be removed before merging this PR

done

praseodym:
Discussed that we won't backport this change because it does not have any effect on the application itself.

PR #2659, opened 2025-12-22
Improve Rust `use` paths

Lionqueen94:
- `sqlx::Error` is explicitly written everywhere since there are many possible `Error` options
- Removed unnecessary path prefixes


backend/src/eml/eml_110.rs
praseodym, line 1: I think `std::result::Result` was used here because there is also a `Result` struct in the EML module, so I think it is better to leave this unchanged

backend/src/eml/eml_230.rs
praseodym, line 1: Same as https://github.com/kiesraad/abacus/pull/2659/changes#r2639977913

backend/src/lib.rs
```diff
@@ -1,6 +1,5 @@
 use std::{future::Future, net::SocketAddr, str::FromStr};
 
-use airgap::AirgapDetection;
```
praseodym, line 3: Why remove this and write it as `airgap::AirgapDetection`?

backend/src/lib.rs
```diff
@@ -115,13 +114,13 @@ pub fn openapi_router() -> OpenApiRouter<AppState> {
 
     struct SecurityAddon;
 
-    impl utoipa::Modify for SecurityAddon {
+    impl Modify for SecurityAddon {
```
praseodym: `utoipa::Modify` is more clear, please undo

backend/src/lib.rs
Lionqueen94: I thought to make it more consistent since everything is accessed through the module name, but it's shorter in the code to keep this specific import

backend/src/eml/eml_110.rs
Lionqueen94: Fixed in 6ec8fa97ab96d54cb2d1fba8e8f02c58426f7554

backend/src/eml/eml_230.rs
Lionqueen94: Fixed in 6ec8fa97ab96d54cb2d1fba8e8f02c58426f7554

backend/src/lib.rs
Lionqueen94: Fixed in 6ec8fa97ab96d54cb2d1fba8e8f02c58426f7554

backend/src/lib.rs
Lionqueen94: Fixed in 6ec8fa97ab96d54cb2d1fba8e8f02c58426f7554

stacktraceghost (approved):

PR #2665, opened 2025-12-27
Shorten validation functions

Lionqueen94:
## Description
Functions improved (removed `#[allow(clippy::too_many_lines)]`):
- `src/`
  - `data_entry/validation.rs`: 
    - [x] `validate_differences_counts`
    - [x] `CommonPollingStationResults`: `validate`
    - [x] `VotesCounts`: `validate`
    - [x] `test_f305_more_and_fewer_ballots_count`
    - [x] `test_f307_votes_greater_than_voters`
    - [x] `test_f401`

7 instances of `#[allow(clippy::too_many_lines)]` left in the codebase of which 3 related to generating test elections

### DIFF
--- a/backend/src/data_entry/validation.rs
+++ b/backend/src/data_entry/validation.rs
@@ -245,181 +245,184 @@ pub trait ValidateRoot: Validate {
     }
 }
 
-fn validate_differences_checkboxes(
-    differences_counts: &DifferencesCounts,
-    total_voters_count: u32,
-    total_votes_count: u32,
-    validation_results: &mut ValidationResults,
-    path: &FieldPath,
-) {
-    let equal_checked = differences_counts
-        .compare_votes_cast_admitted_voters
-        .admitted_voters_equal_votes_cast;
-    let greater_than_checked = differences_counts
-        .compare_votes_cast_admitted_voters
-        .votes_cast_greater_than_admitted_voters;
-    let smaller_than_checked = differences_counts
-        .compare_votes_cast_admitted_voters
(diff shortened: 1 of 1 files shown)


backend/src/data_entry/validation.rs
```diff
@@ -525,8 +568,64 @@ impl Validate for PollingStationResults {
     }
 }
 
+fn validate_political_group_votes_errors(
```
praseodym: I would move this into a `impl CommonPollingStationResults { }` block (and then first argument will be `results: &self`)

backend/src/data_entry/validation.rs
```diff
@@ -400,32 +392,83 @@ pub fn validate_differences_counts(
...
+                path.field("difference_completely_accounted_for")
                     .to_string(),
             ],
             code: ValidationResultCode::F310,
             context: None,
         });
     }
+}
+
+pub fn validate_differences_counts(
```
praseodym, line 429: I think this function and all functions called in this function can be part of a `impl DifferencesCounts` block

backend/src/data_entry/validation.rs
Lionqueen94: Done in 837375467ffbd22d56ef39c64c06c4e5190e58b5

backend/src/data_entry/validation.rs
Lionqueen94: Done in 837375467ffbd22d56ef39c64c06c4e5190e58b5

praseodym (approved):

PR #2673, opened 2025-12-29
Enable no prototype builtins rule

Lionqueen94:

### DIFF
--- a/backend/src/bin/gen-test-election.rs
+++ b/backend/src/bin/gen-test-election.rs
@@ -1,5 +1,3 @@
-#![allow(clippy::cognitive_complexity)]
-#![allow(clippy::too_many_lines)]
 use abacus::{
     AppError,
     committee_session::CommitteeSession,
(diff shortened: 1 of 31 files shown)


frontend/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
+    "target": "ES2022",
```
praseodym, line 3: Why this change? This has implications for which browsers we support (see [docs](https://www.typescriptlang.org/tsconfig/#target)) so we shouldn't change this too casually.

Lionqueen94: The ES2020 setting conflicts with the no prototype builtins rule, because hasOwn is relatively newly introduced. See https://caniuse.com/?search=es2022. According to https://www.w3schools.com/js/js_2022.asp it's supported in all major browsers since March 2023, so I think that should be fine since we only support the last 2 versions of the major browsers.

Lionqueen94: I think we can even switch to https://www.w3schools.com/js/js_2023.asp since this is supported by all major browsers since July 2023

praseodym: Thanks for the clarification!

praseodym (approved):

stacktraceghost (dismissed):

praseodym (approved):

stacktraceghost (dismissed):

stacktraceghost (dismissed):

praseodym (dismissed):

stacktraceghost (approved):

PR #2674, opened 2025-12-29
Enable no array index key rule

Lionqueen94:

### DIFF
--- a/frontend/src/features/resolve_differences/components/ResolveDifferencesTables.tsx
+++ b/frontend/src/features/resolve_differences/components/ResolveDifferencesTables.tsx
@@ -51,7 +51,7 @@ function SectionTable({ section, first, second, action }: SectionTableProps) {
   return (
     <div>
       {/** biome-ignore lint/suspicious/useIterableCallbackReturn: switch statement is exhaustive, so it always returns a value */}
-      {section.subsections.map((subsection) => {
+      {section.subsections.map((subsection, subsectionIdx) => {
         switch (subsection.type) {
           case "heading":
             // only used to override title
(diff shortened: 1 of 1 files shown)


frontend/src/features/resolve_differences/components/ResolveDifferencesTables.tsx
```diff
@@ -76,7 +76,7 @@ function SectionTable({ section, first, second, action }: SectionTableProps) {
 
             return (
               <DifferencesTable
-                key={`${section.id}-${subsectionIdx}`}
+                key={`${section.id}-${subsection.type}`}
```
stacktraceghost: Are we sure this is unique enough? There could be multiple subsections with the same type.

Lionqueen94: within a section there is only one `DifferencesTable` per type, so it seemed fine on the resolve differences page, I checked it with a lot of differences on many pages. But please double check

stacktraceghost: The one I was worried about is `countingDifferencesPollingStation` in [dataEntryStructure](https://github.com/kiesraad/abacus/blob/enable_no_array_index_key_rule/frontend/src/utils/dataEntryStructure.ts#L215). There are two subsections there with type "checkboxes". However, I see that these checkboxes get appended in the var `checkboxRows` below and only rendered when they are all in there.

Nonetheless I find it a bit risky, since we let the given subsections dictate if there will be a duplicate key or not. For example, if we would give multiple subsections with type radio, this would cause a duplicate key.

Lionqueen94: Shall I just revert it and ignore the lines like I did in the other file?

stacktraceghost: Sounds good to me, I prefer to avoid those risks and only do this if we know for sure it can't be duplicated.

Lionqueen94: Turns out I made the change where it was not needed in this file, fixed in 25a94b47b8468da212b83d2f32f93f309a687171

PR #2682, opened 2025-12-29
Enabled use valid anchor rule

Lionqueen94:

### DIFF
--- a/frontend/src/features/account/components/FirstLoginForm.tsx
+++ b/frontend/src/features/account/components/FirstLoginForm.tsx
@@ -1,5 +1,5 @@
 import { FormEvent, ReactNode, useState } from "react";
-import { Link, useNavigate } from "react-router";
+import { useNavigate } from "react-router";
 
 import { isError } from "@/api/ApiResult";
 import { useApiState } from "@/api/useApiState";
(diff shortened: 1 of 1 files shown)


frontend/src/features/account/components/FirstLoginForm.tsx
```diff
@@ -37,16 +37,16 @@ export function FirstLoginForm({ prev }: FirstLoginFormProps) {
...
+      to="#previous"
       onClick={(e) => {
         e.preventDefault();
         prev();
       }}
     >
       {content}
-    </a>
+    </Link>
   );
```
praseodym, line 51: If I understand https://biomejs.dev/linter/rules/use-valid-anchor/ correctly this should be a button instead, because `#previous` is not valid:

> If the anchor has logic attached to it with an incorrect href the rules suggests to turn it to a button, because that’s likely what the user wants.

Lionqueen94: According to the design it should be a link, I'll try to think of another solution

praseodym: That the design specifies a link style doesn't mean we can't use a button element, we have the `Button.Link` element with `underlined` variant for that style.

Lionqueen94: Fixed in a02b882ae5f295b907f1d674febcb6940a18867f

stacktraceghost (approved):

PR #2683, opened 2025-12-29
Enabled use semantic elements rule

Lionqueen94:
and enabled no interactive element to non-interactive role rule as well


praseodym (approved):
Code looks good, did not check styling in browser

stacktraceghost (dismissed):

praseodym (dismissed):

praseodym (approved):

PR #2689, opened 2025-12-30
Enabled no label without control rule

Lionqueen94:


frontend/src/components/ui/Button/Button.tsx
```diff
@@ -60,13 +60,21 @@ export interface ButtonLabelProps extends React.LabelHTMLAttributes<HTMLLabelEle
   disabled?: boolean;
   variant?: ButtonVariant;
   size?: Size;
+  htmlFor: string;
```
stacktraceghost, line 61: The changes in this file are not necessary, because labelProps already contains `htmlFor` and it will automatically be spread on `<label>` by `...labelProps`.

Lionqueen94: I want to make it a mandatory prop, that's why I added it here

stacktraceghost: Makes sense!

praseodym (approved):

PR #2275, opened 2025-10-15
Delete MSW preview deployment

praseodym:
Delete MSW preview deployment to Cloudflare Pages, because everyone is using https://abacus-test.nl/ instead.

PR #2691, opened 2025-12-30
Migrate to pnpm

### LINKED ISSUE #2686: Investigate using pnpm instead of npm
npm has various issues such as this one that I noticed in #2685:

> **Pin dependency versions and recreate npm lockfile**
> Recreating the lockfile cleans up unused dependencies:
>     * Before: 608 packages
>     * After:  493 packages

In addition, pnpm has some mitigations for supply chain attacks: https://pnpm.io/supply-chain-security

praseodym:
Migrate to pnpm, resolves #2686 

- Convert npm `package-lock.json` to `pnpm-lock.yaml` using `pnpm import`
- Configure pnpm workspace to [mitigate supply chain attacks](https://pnpm.io/supply-chain-security)
- Explicitly add `@vitest/expect` as a dependency, because [with pnpm only direct dependencies are accessible](https://pnpm.io/motivation#creating-a-non-flat-node_modules-directory)
- Move `vitest.d.ts` to new `@types` directory to make it work with pnpm, and update `tsconfig.json` accordingly
- Set engine requirements in `package.json` to prevent usage of npm and unsupported Node.js versions
- Update GitHub Actions workflows
- Update Lefthook config, also to prevent Biome from throwing errors when no relevant files are changed
- Update documentation
- Update Dockerfile

Note that [pnpx](https://pnpm.io/cli/dlx) uses the latest available version of a command. To run a command from the project packages, use `pnpm <command>` instead (e.g. `pnpm vite`). The latter is preferred: `pnpm vite --version` uses version 6.4.1 as defined in the project, whereas `pnpx vite --version` prints version 7.3.0 which is not compatible with our config files.

### DIFF
--- a/frontend/pnpm-workspace.yaml
+++ b/frontend/pnpm-workspace.yaml
@@ -8,6 +8,7 @@ blockExoticSubdeps: true
 minimumReleaseAge: 4320
 trustPolicy: no-downgrade
 engineStrict: true
+sideEffectsCache: false
 overrides:
   # Storybook currently uses @vitest/expect 3.2.4 which doesn't work with our vitest.d.ts
   "@vitest/expect": "4.0.16"


Lionqueen94 (dismissed):

jschuurk-kr:
Installed `pnpm`, deleted my `node_modules` folder and successfully ran `pnpm install`, `pnpm dev:msw`, `pnpm lint`, `pnpm test`, `pnpm serve` and `pnpm test:e2e`.

.github/workflows/build-lint-test.yml
```diff
@@ -49,20 +49,26 @@ jobs:
           df -h
       - uses: actions/checkout@v6
       # Build frontend so that it can be included in the backend build
+      - name: Install pnpm
+        uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4
```
oliver3: Why not use `npm install -g pnpm@latest-10` instead of depending on a third party action?

.github/workflows/build-lint-test.yml
```diff
@@ -49,20 +49,26 @@ jobs:
...
       - name: Setup Node
         uses: actions/setup-node@v6
         with:
           node-version: 22
-      - name: Install npm dependencies
-        run: npm ci
+          cache: "pnpm"
+          cache-dependency-path: ./frontend/pnpm-lock.yaml
+      - name: Install frontend dependencies
+        run: pnpm install
```
oliver3, line 63: I'm not sure if pnpm is detecting that it is running on the CI environment
```
  pnpm install
  shell: /usr/bin/bash -e {0}
  env:
    TZ: Europe/Amsterdam
    CARGO_TERM_COLOR: always
    SQLX_OFFLINE: true
    PNPM_HOME: /home/runner/setup-pnpm/node_modules/.bin
Lockfile is up to date, resolution step is skipped
```

Maybe better to set `--frozen-lockfile` explicitly?

frontend/package.json
```diff
@@ -29,7 +32,7 @@
     "gen:po": "vite-node scripts/gen_po.js",
     "gen:translation-json": "vite-node scripts/gen_translation-json.js",
     "check-server": "vite-node scripts/check_server.js",
-    "serve": "npm run build:msw && npx serve dist",
+    "serve": "pnpm build:msw && pnpx serve dist",
```
oliver3: If I understand correctly, we should not use `pnpx` ?

Dockerfile
```diff
@@ -1,9 +1,12 @@
...
+FROM node:22-slim AS frontend-builder
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
 WORKDIR /build
 COPY . .
 WORKDIR /build/frontend
-RUN npm ci --omit=dev
-RUN npm run build
+RUN pnpm install --prod
```
oliver3: Also here, maybe add `--frozen-lockfile`?

oliver3 (commented):
Works on my machine!

frontend/package.json
praseodym: No, see the last part of the PR description. But the 'serve' command is not part of our dependency tree (also wasn't with npm), so for that one we do need to use `pnpx`.

.github/workflows/build-lint-test.yml
praseodym: Hm, are we expecting different output here? Our e2e tests and Vite builds already use `process.env.CI` and that's also [what pnpm does](https://pnpm.io/cli/install#--frozen-lockfile).

praseodym: `pnpm install --frozen-lockfile` also prints `Lockfile is up to date, resolution step is skipped` locally and so does `env CI=true pnpm install`.

.github/workflows/build-lint-test.yml
praseodym: Good point, I simply followed https://pnpm.io/continuous-integration. For other CI enviroments they are recommending using Corepack, so that could be an alternative to depending on a third-party action.

frontend/package.json
oliver3: Shouldn't we have `serve` then as a dev dependency, to have a stable version with cooldown and that our tooling checks for malware etc?

.github/workflows/build-lint-test.yml
praseodym: Changed to using Corepack without third-party action in f095554f7ccab0a7290f121d3df621679b8df3a2.

.github/workflows/build-lint-test.yml
oliver3: Apparently `--frozen-lockfile` still checks if the Lockfile is out of date and fails if it should be updated

frontend/package.json
praseodym: I'd prefer removing this script instead, because I don't think anyone is using the MSW build any more. What do you think?

oliver3: After looking more closely at this script, I believe it is a script that is not used anymore. I'm using `dev:msw` to run a local Abacus with MSW, and the deployed MSW preview was removed in #2275 (although it seems not to use the `serve` either..)

praseodym: Removed in e85c08cc063f8a7441d59d3ad3d93bb0a541b3af

Dockerfile
praseodym: Added in aba13cf004b316de9716734073d1cf2c35321550.

Lionqueen94 (dismissed):

frontend/pnpm-workspace.yaml
```diff
@@ -0,0 +1,13 @@
+onlyBuiltDependencies:
+  - '@swc/core'
+  - esbuild
+  - lefthook
+  - msw
```
oliver3, line 5: This does not seem to work. If I remove the `mockServiceWorker.js` file and the `node_modules/`, and then run `pnpm i`, the `mockServiceWorker.js` does not seem to be generated. 

Also, the post-commit hook does not seem to be installed on a fresh repo.

oliver3: Works in commit https://github.com/kiesraad/abacus/pull/2691/commits/d21215e46ad5d82616d610d45a884d673465856c

praseodym:
We enabled [`sideEffectsCache: false`](https://pnpm.io/settings#sideeffectscache) so that Lefthook and MSW will always run their post-install scripts. As a follow up we should consider removing Lefthook as a frontend dependency (letting developers install it themselves) and checking if MSW can be used without the post-install script.

PR #2703, opened 2026-01-05
Enabled no important styles rule

Lionqueen94:
## How to test
- Check that the NavBar links do not change colour on hover/active (same as on main)
- Check that the disabled ProgressList links do not change colour on hover (same as on main) -> for example in Data Entry or in Election Import
- Check that the not disabled ProgressList links still look the same as on main on hover

### DIFF
--- a/frontend/biome.jsonc
+++ b/frontend/biome.jsonc
@@ -62,8 +62,7 @@
         "noSvgWithoutTitle": "off"
       },
       "complexity": {
-        "noUselessFragments": "warn",
-        "noImportantStyles": "off"
+        "noUselessFragments": "warn"
       },
       "correctness": {
         "useExhaustiveDependencies": "off"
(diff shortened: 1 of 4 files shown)


jorisleker (approved):
- [x] Checked that the NavBar links do not change colour on hover/active (same as on main) ✅ 
- [x] Checked that the disabled ProgressList links do not change colour on hover (same as on main) -> for example in Data Entry or in Election Import ✅ 
- [x] Checked that the not disabled ProgressList links still look the same as on main on hover ✅

stacktraceghost:
> * [x]  Checked that the NavBar links do not change colour on hover/active (same as on main) ✅
> * [x]  Checked that the disabled ProgressList links do not change colour on hover (same as on main) -> for example in Data Entry or in Election Import ✅
> * [x]  Checked that the not disabled ProgressList links still look the same as on main on hover ✅

Same conclusions here 👍

stacktraceghost (approved):

PR #2707, opened 2026-01-06
Moved airgap detection test from integration tests to airgap module

### LINKED ISSUE #2628: Move `airgap_detection_test.rs` from `/tests/` to api
Related to https://github.com/kiesraad/abacus/issues/2191

This test uses the `utils::serve_api_with_airgap_detection` which uses the `airgap::AirgapDetection` struct, but the integration tests should not use structs from abacus.

Lionqueen94:
Closes #2628


backend/src/router.rs
```diff
@@ -286,3 +286,7 @@ pub fn create_router(
     let router = router.with_state(state);
     Ok(router)
 }
+
+pub fn create_router_without_airgap_detection(pool: SqlitePool) -> Result<Router, AppError> {
+    create_router(pool, AirgapDetection::nop())
+}
```
Lionqueen94, line 292: @praseodym I want to make this function only available for use in the integration tests, but this might not be possible. Do you have a suggestion?

praseodym: Not easily, but I think it's fine to make the function public like this.

PR #2708, opened 2026-01-06
Generating TypeScript values from Rust enum values in openapi.ts

### LINKED ISSUE #2569: Generate TypeScript values from Rust enum values in openapi.ts
- As discussed in https://github.com/kiesraad/abacus/issues/489

Implement as:
```typescript
export const pollingStationStatusValues = ["first_entry", "first_entry_in_progress", "definitive"] as const;
export type PollingStationStatus = typeof pollingStationStatusValues[number];
```

Lionqueen94:
Closes #2569
Closes https://github.com/kiesraad/abacus/issues/489

### DIFF
--- a/frontend/scripts/openapi/generator.ts
+++ b/frontend/scripts/openapi/generator.ts
@@ -126,15 +126,14 @@ function addDefinition(name: string, v: ReferenceObject | SchemaObject) {
 
   if (v.type === "object") {
     result.push(`export interface ${name} ${tsType(v)}`);
+  } else if (v.enum) {
+    const valuesString = `${name[0]?.toLowerCase()}${name.slice(1)}Values`;
+    result.push(
+      `export const ${valuesString} = [${v.enum.map((e) => `"${e}"`).join(", ")}] as const;
+        export type ${name} = (typeof ${valuesString})[number];`,
+    );
   } else {
-    if (v.enum) {
-      result.push(
-        `export const ${name}Values = [${v.enum.map((e) => `"${e}"`).join(", ")}] as const;
-        export type ${name} = typeof ${name}Values[number];`,
-      );
-    } else {
-      result.push(`export type ${name} = ${tsType(v)};`);
-    }
+    result.push(`export type ${name} = ${tsType(v)};`);
   }
 
   return result.join("\n");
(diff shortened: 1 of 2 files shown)


praseodym:
> @praseodym It seems like the parenthesis are added by default, not sure how to remove them 🤔

Looks like the Biome formatter adds these, so don't bother trying to get rid of them.

praseodym (approved):

frontend/scripts/openapi/generator.ts
```diff
@@ -127,7 +127,14 @@ function addDefinition(name: string, v: ReferenceObject | SchemaObject) {
   if (v.type === "object") {
     result.push(`export interface ${name} ${tsType(v)}`);
   } else {
-    result.push(`export type ${name} = ${tsType(v)};`);
+    if (v.enum) {
```
stacktraceghost: Why is this nested? `if...elseif...else` would be okay here I think.

oliver3:
I'm having some doubts about values variable being in PascalCase (upper camel case). For me PascalCase signals that it is a TypeScript type and not a value. 

What do we prefer @Lionqueen94 @praseodym @stacktraceghost ?
- `export const PollingStationTypeValues = [];`
- `export const pollingStationTypeValues = [];`
- `export const POLLING_STATION_TYPE_VALUES = [];`

stacktraceghost:
> I'm having some doubts about values variable being in PascalCase (upper camel case). For me PascalCase signals that it is a TypeScript type and not a value.
> 
> What do we prefer @Lionqueen94 @praseodym @stacktraceghost ?
> 
> * `export const PollingStationTypeValues = [];`
> * `export const pollingStationTypeValues = [];`
> * `export const POLLING_STATION_TYPE_VALUES = [];`

I'd prefer option 2 (`pollingStationTypeValues`).

frontend/scripts/openapi/generator.ts
Lionqueen94: Changed in ea1c9ccf0730f83c7f37eb87cc21262373319c68

Lionqueen94:
> > I'm having some doubts about values variable being in PascalCase (upper camel case). For me PascalCase signals that it is a TypeScript type and not a value.
> > What do we prefer @Lionqueen94 @praseodym @stacktraceghost ?
> > 
> > * `export const PollingStationTypeValues = [];`
> > * `export const pollingStationTypeValues = [];`
> > * `export const POLLING_STATION_TYPE_VALUES = [];`
> 
> I'd prefer option 2 (`pollingStationTypeValues`).

Changed to option 2 in ea1c9ccf0730f83c7f37eb87cc21262373319c68

praseodym (approved):

PR #2710, opened 2026-01-07
Move authorization_test.rs from integration tests to router.rs

### LINKED ISSUE #2627: Move `authorization_test.rs` from `/tests/` to api
Related to https://github.com/kiesraad/abacus/issues/2191

This test uses the `authentication::Role` enum and `get_scopes_from_operation` function, but the integration tests should not use these from abacus.

stacktraceghost:
Resolves #2627

- Moved test `test_route_authorization` to `router.rs`
- Removed `authorization_test.rs` from integration tests


backend/src/lib.rs
```diff
@@ -229,7 +229,7 @@ mod test {
 
     use super::start_server;
 
-    async fn run_server_test<F, Fut>(pool: SqlitePool, test_fn: F)
+    pub async fn run_server_test<F, Fut>(pool: SqlitePool, test_fn: F)
```
praseodym: Can this be `pub(crate)`?

stacktraceghost: Good point, adjusted in https://github.com/kiesraad/abacus/pull/2710/commits/8f24b8072b953ba8aa26c86ec3936d1819803e3b

PR #2720, opened 2026-01-09
Fixes for 2 code quality findings in `/frontend/src/components/ui/ProgressList/ProgressList.tsx`

Lionqueen94:
- Removed unused reference
- Removed "fixed" class that had no effect (flex setting but not inside a flex element)


Lionqueen94:
Closed in favour of https://github.com/kiesraad/abacus/pull/2727

PR #2721, opened 2026-01-09
Fixes for 3 code quality findings in `/backend/tests/utils/mod.rs`

Lionqueen94:
_This PR applies 3/4 suggestions from code quality [AI findings](https://github.com/kiesraad/abacus/security/quality/ai-findings). 1 suggestion was skipped to avoid creating conflicts._

One finding was a duplicate.

### DIFF
--- a/backend/tests/utils/mod.rs
+++ b/backend/tests/utils/mod.rs
@@ -8,7 +8,9 @@ use tokio::net::TcpListener;
 use abacus::{router, shutdown_signal};
 
 pub async fn serve_api(pool: SqlitePool) -> SocketAddr {
-    let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let listener = TcpListener::bind("127.0.0.1:0")
+        .await
+        .expect("Failed to bind TCP listener for test server");
     let addr = listener.local_addr().unwrap();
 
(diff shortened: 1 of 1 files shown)


backend/tests/utils/mod.rs
```diff
@@ -8,19 +8,22 @@ use tokio::net::TcpListener;
...
-    let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let listener = TcpListener::bind("127.0.0.1:0")
+        .await
+        .expect("Failed to bind TCP listener for test server");
     let addr = listener.local_addr().unwrap();
 
     tokio::spawn(async move {
-        let app = router::create_router_without_airgap_detection(pool).unwrap();
+        let app = router::create_router_without_airgap_detection(pool)
+            .expect("failed to create router for test server");
```
stacktraceghost, line 18:
```suggestion
            .expect("Failed to create router for test server");
```

praseodym: These expect messages don’t follow the [recommended message style](https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style) and we also explicitly allow unwrap in test code, so I propose we leave this code as is and close the PR without merging.

PR #2726, opened 2026-01-12
Remove eslint config ignore

Lionqueen94:


oliver3 (changes requested):
It looks like another PR was included in this PR by accident (use cases)

Lionqueen94:
> It looks like another PR was included in this PR by accident (use cases)

Fixed!

PR #2731, opened 2026-01-12
Committee Session ID Newtypes

### LINKED ISSUE #2761: Committee Session ID newtype

cikzh:
Closes #2761 

Adds newtypes around Committee Session ID's

Related #2555

### DIFF
--- a/backend/src/committee_session/repository.rs
+++ b/backend/src/committee_session/repository.rs
@@ -38,7 +38,7 @@ pub async fn get(
 
 pub async fn get_previous_session(
     conn: &mut SqliteConnection,
-    committee_session_id: super::CommitteeSessionId,
+    committee_session_id: CommitteeSessionId,
 ) -> Result<Option<CommitteeSession>, sqlx::Error> {
     query_as!(
         CommitteeSession,
--- a/backend/src/committee_session/status.rs
+++ b/backend/src/committee_session/status.rs
@@ -289,7 +289,7 @@ mod tests {
             let mut conn = pool.acquire().await?;
             let audit_service = AuditService::new(None, Some(Ipv4Addr::new(203, 0, 113, 0).into()));
 
-            let committee_session_id = crate::committee_session::CommitteeSessionId::from(6);
+            let committee_session_id = CommitteeSessionId::from(6);
 
             // DataEntryInProgress --> DataEntryFinished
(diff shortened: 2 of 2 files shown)


backend/src/committee_session/repository.rs
```diff
@@ -35,13 +38,13 @@ pub async fn get(
 
 pub async fn get_previous_session(
     conn: &mut SqliteConnection,
-    committee_session_id: u32,
+    committee_session_id: super::CommitteeSessionId,
```
Lionqueen94:
```suggestion
    committee_session_id: CommitteeSessionId,
```

backend/src/committee_session/status.rs
```diff
@@ -287,26 +289,28 @@ mod tests {
             let mut conn = pool.acquire().await?;
             let audit_service = AuditService::new(None, Some(Ipv4Addr::new(203, 0, 113, 0).into()));
 
+            let committee_session_id = crate::committee_session::CommitteeSessionId::from(6);
```
Lionqueen94:
```suggestion
            let committee_session_id = CommitteeSessionId::from(6);
```

cikzh: Fixed in [862c403](https://github.com/kiesraad/abacus/pull/2731/commits/862c403271be3553b1e6f2822a7be025d4c99f7e)

backend/src/committee_session/repository.rs
cikzh: [862c403](https://github.com/kiesraad/abacus/pull/2731/commits/862c403271be3553b1e6f2822a7be025d4c99f7e)

praseodym (approved):
lgtm! There are some commits in this branch that shouldn't be here (22fd7ae004c91248e5cad07bf6bd8f3ae9e1a315, d0f490ff5cb7cd5f6789fb85f5909d30e327787b, 17ad41578a52276e4c164ce293a8bbadb36575e4) but they will be squashed anyway.

praseodym:
> I did notice a few files where the committee session ID is unchanged. Do these lines need to be updated as well?

These are all in the integration tests, where we use JSON instead of (new)types from the Abacus crate itself.

praseodym:
> but doesn't close #2555

@cikzh next time please use syntax that GitHub also understands (:

PR #2732, opened 2026-01-12
Add abortcontroller to claim post request

Lionqueen94:

### DIFF
--- a/frontend/src/features/data_entry/hooks/useInitialDataEntryState.ts
+++ b/frontend/src/features/data_entry/hooks/useInitialDataEntryState.ts
@@ -11,7 +11,7 @@ export function useInitialDataEntryState(client: ApiClient, dispatch: DataEntryD
     const abortController = new AbortController();
 
     const claimDataEntry = async () => {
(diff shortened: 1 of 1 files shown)


frontend/src/features/data_entry/hooks/useInitialDataEntryState.ts
```diff
@@ -11,7 +11,7 @@ export function useInitialDataEntryState(client: ApiClient, dispatch: DataEntryD
     const abortController = new AbortController();
 
     const claimDataEntry = async () => {
-      const result = await client.postRequest<ClaimDataEntryResponse>(claimRequestPath);
+      const result = await client.postRequest<ClaimDataEntryResponse>(claimRequestPath, abortController);
```
praseodym: iirc it should be the third, not second, parameter

Lionqueen94: Fixed in 520b722356c811d52f6fa0f288fb4c109f354f68

praseodym (approved):

PR #2735, opened 2026-01-13
Polling Station ID Newtypes

cikzh:
## This PR builds upon #2731, merge that one first

Adds newtypes around Polling Station IDs

Relates to #2555

### DIFF
--- a/backend/src/investigation/api.rs
+++ b/backend/src/investigation/api.rs
@@ -114,7 +114,7 @@ pub async fn delete_investigation_for_polling_station(
         (status = 500, description = "Internal server error", body = ErrorResponse),
     ),
     params(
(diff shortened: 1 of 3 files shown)


backend/src/investigation/api.rs
stacktraceghost: In this file the OpenAPI doc isn't updated for `polling_station_id`

stacktraceghost (changes requested):

backend/src/investigation/api.rs
cikzh: Fixed in [1f2d8e6](https://github.com/kiesraad/abacus/pull/2735/commits/1f2d8e67bdf788e722f3918292e7752113232040)

PR #2736, opened 2026-01-13
Refactor tests in election create test

### LINKED ISSUE #2550: Refactor tests in ElectionCreate.test.tsx similar to NumberOfVoters.test.tsx
The tests in ElectionCreate.test.tsx go through all the steps of the election creation wizard, while that's not always necessary. This refactor has already been done for the `NumberOfVoters` component (NumberOfVoters.test.tsx). We should do something similar for the other components.

Lionqueen94:
Closes #2550

## Description
- Moved many tests from `ElectionCreate.test.tsx` to their respective components
- Replaced the Alert in CheckHash, because it shouldn't be one and this allows `await screen.findByRole("alert")` to work in tests for the actual alert
- Removed e2e tests for too large files, since they are now covered in component tests
- Added full flow election create test in `ElectionCreate.test.tsx`
- Added election creation success alert (was also missing in Figma, has now been added -> https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17177-8851&m=dev)
Screenshot from 2026-01-13 15-35-02


jschuurk-kr:
@Lionqueen94 In the e2e tests there are still quite some tests that don't need to be e2e tests:
- in the `test.describe("modal"` block: the only e2e I would expect is one that covers aborting election creation. For everything else it should be possible to cover it in a component test.
- in the `test.describe("navigation",` block: the two tests "navigating to ... should redirect to create" can be component tests. The tests using the back-button we can keep as they are for now, because I'm not sure if we should convert them to component tests or delete them.
- in the `test.describe("polling stations"` there are two tests for the list of polling stations and the button to show all polling stations. These two tests should be component tests.

(edit) I created a new issue for this, because it's not in scope of the original issue and this PR is large enough as is.

frontend/e2e-tests/fixtures.ts
```diff
@@ -244,25 +237,4 @@ export const test = base.extend<Fixtures>({
 
     await use((await userResponse.json()) as User);
   },
-  // biome-ignore lint/correctness/noEmptyPattern: the argument should use the object destructuring pattern
-  eml110aTooLargeTestFile: async ({}, use) => {
```
stacktraceghost, line 248: I think `frontend/e2e-tests/helpers-utils/file-helpers.ts` is not used anymore and can also be removed.

frontend/src/features/election_create/components/CheckHash.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+        redacted_indexes: [4, 11],
+      }}
+      error={error}
+      onSubmit={onSubmit}
+    />,
+  );
+}
+
+function joinHashChunksAsRendered(chunks: string[]): string {
+  let result = chunks.shift()!;
```
stacktraceghost: Suggestion for shorter code
```ts
function joinHashChunksAsRendered(chunks: string[]): string {
  let obfuscationIndex = 1;
  return chunks.map((chunk) => (chunk === "" ? obfuscationIndex++ : chunk)).join("-");
}
```

frontend/e2e-tests/fixtures.ts
Lionqueen94: Done in 8acc47a4d2ee5ff75cfd682fe36013de4ad32eb9

frontend/src/features/election_create/components/CheckHash.test.tsx
Lionqueen94: Nice! Changed in 8acc47a4d2ee5ff75cfd682fe36013de4ad32eb9

frontend/e2e-tests/page-objects/election/ElectionsOverviewPgObj.ts
```diff
@@ -4,17 +4,20 @@ export class ElectionsOverviewPgObj {
...
   constructor(protected readonly page: Page) {
     this.main = page.getByRole("main");
     this.adminHeader = page.getByRole("heading", { name: "Verkiezingen beheren" });
     this.header = page.getByRole("heading", { name: "Verkiezingen" });
-    this.alert = page.getByRole("alert").filter({ hasText: "Je account is ingesteld" });
     this.create = page.getByRole("link", { name: "Verkiezing toevoegen" });
     this.elections = page.getByTestId("overview").locator("tbody").getByRole("row");
+
+    this.alertAccountSetup = page.getByRole("alert").filter({ hasText: "Je account is ingesteld" });
+    this.alertElectionCreated = page.getByRole("alert").filter({ hasText: /^Verkiezing GSB [\w|\s]+ toegevoegd$/ });
```
oliver3, line 20: Why do you use `[\w|\s]` which if I understand correctly matches one or more alphanumeric or whitespaces, but not e.g. punctuation?

frontend/src/features/election_create/components/CheckHash.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    />,
+  );
+}
+
+function joinHashChunksAsRendered(chunks: string[]): string {
+  let obfuscationIndex = 1;
+  return chunks.map((chunk) => (chunk === "" ? obfuscationIndex++ : chunk)).join("-");
+}
+
+describe("CheckHash component", () => {
```
oliver3, line 49: I would prefer a Storybook story + test instead.

frontend/src/features/election_create/components/ElectionCreate.test.tsx
```diff
@@ -162,627 +162,336 @@ async function uploadCandidateDefinition(file: File) {
...
-    // Make sure the warning is not shown
-    expect(await screen.findByText(/Afwijkende verkiezing/i)).toBeVisible();
-
-    // click next
-    await user.click(screen.getByRole("button", { name: "Volgende" }));
-
-    // Expect to see the next page
-    expect(await screen.findByRole("heading", { level: 2, name: "Type stemopneming in Heemdamseburg" })).toBeVisible();
-  });
+    test("Shows warning when uploading a polling stations file with not matching election id", async () => {
```
oliver3, line 401: I think this test can go to the new `UploadPollingStationDefinition.test.tsx` because it asserts behaviour that is local for that step.

oliver3 (commented):
Nice restructuring into separate tests for the individual steps!

frontend/e2e-tests/page-objects/election/ElectionsOverviewPgObj.ts
Lionqueen94: I supposed it's not very likely for the election name to use punctuation and the ones in the tests don't

frontend/src/features/election_create/components/ElectionCreate.test.tsx
Lionqueen94: I tried but it was very difficult to get the component into the right state for this test to work there. So it was easier to keep it in this test. Maybe we can try again this afternoon to see if it's possible?

jschuurk-kr: My experience matches @Lionqueen94's. The state gets relatively large by the time you get to uploading polling stations.

frontend/src/features/election_create/components/CheckHash.test.tsx
jschuurk-kr: Good point, but let's leave that for a future PR.

frontend/e2e-tests/page-objects/election/ElectionsOverviewPgObj.ts
jschuurk-kr: `\s` matches spaces, tabs and linebreaks. So I'd prefer having only the space character here.

As to the punctuation: we can update the regex when the test data requires it. (If we want to get really strict: `\w` does not match `ö`.)

frontend/src/features/election_create/components/ElectionCreate.test.tsx
oliver3: After our chat I realise now that this step has more coupling to the previous steps than I realised, and therefore not that local as I thought.

PR #2744, opened 2026-01-15
More specific error messages for password fields

### LINKED ISSUE #2614: Wrong error message when password is equal to username
@stacktraceghost mentioned while reviewing https://github.com/kiesraad/abacus/pull/2606

> When creating a new user or setting up your account and entering the same password as your username, you get the wrong error message that your password is not long enough.

Lionqueen94:
Closes #2614

## Description
- Added specific backend validation errors for `password == username`, `password < 13 characters` and `new password == old password`
- Showing specific error messages for the above backend errors

### Frontend validation
Empty password:
image

### Backend validation
Too short password:
image

Password == username:
image

New password == old password:
image


backend/src/authentication/error.rs
```diff
@@ -7,10 +7,12 @@ pub enum AuthenticationError {
     InvalidPassword,
     InvalidSessionDuration,
     InvalidUsernameOrPassword,
+    NewPasswordSameAsOldPassword,
```
stacktraceghost: My preference would be to keep a pattern in these names, so they don't get sprinkled around in this list. Maybe: `PasswordRejectionLength`, `PasswordRejectionSameAsOld`, `PasswordRejectionSameAsUsername`? Or maybe rename `NewPasswordSameAsOldPassword` to` PasswordSameAsOldPassword`, but that misses the rejection part.

frontend/src/features/account/components/AccountSetupForm.test.tsx
```diff
@@ -43,87 +43,212 @@ describe("AccountSetupForm", () => {
...
     const submitButton = screen.getByRole("button", { name: "Opslaan" });
     await user.click(submitButton);
 
     expect(loginSuccess).not.toBeInTheDocument();
   });
 
-  test("Required fields", async () => {
-    const { onSaved } = await renderForm();
-
+  test("Fullname prefilled", async () => {
```
stacktraceghost, line 78: In this file and:
- `frontend/src/features/account/components/CreateFirstAdminForm.test.tsx`
- `frontend/src/features/users/components/create/UserCreateDetailsForm.test.tsx`
- `frontend/src/features/users/components/update/UserUpdateForm.test.tsx`

Essentially the same tests are added and repeated for the three error states. It indicates that we could make a component/abstraction to prevent this repetition. And/or the tests can be shortened by using a `test.each` with params `error, expectedText`

frontend/src/utils/validateUserAccount.ts
```diff
@@ -44,10 +46,12 @@ export function validateCreateUser(
   }
 
   if (accountCreate.temp_password.length === 0) {
-    errors.password = t("initialise.password_rules");
+    errors.password = t("account.password_length_rule");
+  } else if (accountCreate.temp_password === accountCreate.username) {
```
oliver3: Why is this specific rule added for the frontend? 

And now that we have specific error references in the backend, can we remove the frontend checks and rely on the backend for all validation rules?

Lionqueen94: I think it might be nice to get an indication what's wrong before sending to the backend. But we can also switch to only backend checks except the required field checks.

backend/src/authentication/error.rs
Lionqueen94: Good one, renamed in 5a5a3da017076b107d5dbad30e1193d2bc534857

frontend/src/features/account/components/AccountSetupForm.test.tsx
Lionqueen94: Changed to test.each in 4 testfiles in 41a3e40da966ee7db75b813f3b340f94d4e61df3

frontend/src/utils/validateUserAccount.ts
Lionqueen94: Removed in 76b6a1b8ff9f0333fb9cc5034e17e9dc3b8596af

oliver3: Looks more clean now thanks!

PR #2745, opened 2026-01-16
Refactor some of the election creation e2e tests into component tests

### LINKED ISSUE #2740: Refactor some of the election creation e2e tests into component tests
In the e2e tests for election creation there are still quite some tests that don't need to be e2e tests:
- in the `test.describe("modal"` block: the only e2e I would expect is one that covers aborting election creation. For everything else it should be possible to cover it in a component test.
- in the `test.describe("navigation",` block: the two tests "navigating to ... should redirect to create" can be component tests. The tests using the back-button we can keep as they are for now, because I'm not sure if we should convert them to component tests or delete them.
- in the `test.describe("polling stations"` there are two tests for the list of polling stations and the button to show all polling stations. These two tests should be component tests.

_Originally posted by @jschuurk-kr in https://github.com/kiesraad/abacus/issues/2736#issuecomment-3750301380_

Lionqueen94:
Closes #2740

## Description
- Removed election create abort modal e2e tests covered in election create unit tests
- Removed election create no state navigation e2e tests and added them to each component unit tests
- Removed election create e2e tests for show more polling stations button and added extra test to component unit test
- Reduced duplication in component tests that upload files
- Improved e2e test import consistency

### DIFF
--- a/frontend/src/features/election_create/components/CheckAndSave.test.tsx
+++ b/frontend/src/features/election_create/components/CheckAndSave.test.tsx
@@ -5,7 +5,7 @@ import * as useElectionCreateContext from "../hooks/useElectionCreateContext";
 import { CheckAndSave } from "./CheckAndSave";
 
 describe("CheckAndSave component", () => {
-  test("Navigates to election create page when no election, electionDefinitionData, candidateDefinitionData and countingMethod", () => {
+  test("Navigates to election create page when no election, electionDefinitionData, candidateDefinitionData or countingMethod", () => {
     const state = {};
     const dispatch = vi.fn();
     vi.spyOn(useElectionCreateContext, "useElectionCreateContext").mockReturnValue({ state, dispatch });


frontend/src/features/election_create/components/CheckAndSave.test.tsx
```diff
@@ -0,0 +1,20 @@
+import { describe, expect, test, vi } from "vitest";
+import { MessagesProvider } from "@/hooks/messages/MessagesProvider";
+import { renderReturningRouter } from "@/testing/test-utils";
+import * as useElectionCreateContext from "../hooks/useElectionCreateContext";
+import { CheckAndSave } from "./CheckAndSave";
+
+describe("CheckAndSave component", () => {
+  test("Navigates to election create page when no election, electionDefinitionData, candidateDefinitionData and countingMethod", () => {
```
stacktraceghost:
```suggestion
  test("Navigates to election create page when no election, electionDefinitionData, candidateDefinitionData or countingMethod", () => {
```

jschuurk-kr (approved):

PR #2746, opened 2026-01-17
Update sigrid yaml

Lionqueen94:
- Added `router.rs`
- Added `app_error.rs`
- Added `data_entry_detail/`
- Excluded `documentatie/`

### DIFF
--- a/sigrid.yaml
+++ b/sigrid.yaml
@@ -60,7 +60,7 @@ components:
   - name: "(Backend) App error"
     include:
       - ".*/backend/src/app_error.rs"
-  - name: "(Backend) Error"
+  - name: "(Backend) error"
     include:
       - ".*/backend/src/error.rs"
   - name: "(Backend) fixtures"


sigrid.yaml
```diff
@@ -121,6 +127,9 @@ components:
   - name: "(Frontend) Modules/data_entry"
     include:
       - ".*/frontend/src/features/data_entry/.*"
+  - name: "(Frontend) Modules/data_entry_detail"
```
praseodym: We are naming the newer modules `features/...`
```suggestion
  - name: "(Frontend) features/data_entry_detail"
```

sigrid.yaml
```diff
@@ -57,15 +57,21 @@ components:
...
+      - ".*/backend/src/app_error.rs"
+  - name: "(Backend) Error"
     include:
       - ".*/backend/src/error.rs"
-  - name: "(Backend) fixtures"
+  - name: "(Backend) Fixtures"
     include:
       - ".*/backend/src/fixtures.rs"
-  - name: "(Backend) lib"
+  - name: "(Backend) Lib"
```
praseodym: Please don't make any unrelated changes, this will create new modules in Sigrid

sigrid.yaml
Lionqueen94: Changed in 0cd092cb019fd40f6e0923c019e486845a21b38a

sigrid.yaml
Lionqueen94: Reverted in 0cd092cb019fd40f6e0923c019e486845a21b38a

sigrid.yaml
```diff
@@ -57,7 +57,10 @@ components:
   - name: "(Backend) bin/gen-openapi"
     include:
       - ".*/backend/src/bin/gen-test-election.rs"
-  - name: "(Backend) error"
```
stacktraceghost, line 60: I've noticed that here `(Backend) error` changes to `(Backend) Error`, doesn't that cause the same problem in Sigrid as in @praseodym comment?

Lionqueen94: Ah missed one! Fixed in 87a505794e59cb238eb4f8085a411c8ebcd7f753

PR #2775, opened 2026-01-22
Run Cargo CI commands on entire workspace

stacktraceghost:
Add argument `--workspace` to cargo commands in CI to use the whole workspace.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,7 +80,7 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y musl-tools
         if: matrix.target.os == 'ubuntu-latest'
       - name: Run tests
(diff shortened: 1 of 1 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -87,11 +87,11 @@ jobs:
       - name: Check rustfmt
         run: cargo --verbose --locked fmt --all -- --check
```
Lionqueen94, line 88: Does rustfmt also need `--workspace`?

stacktraceghost: No, it does not have the `--workspace` param and seems to already pick up apportionment if you run `cargo fmt --all -v`.

I did find one more cargo test in `release.yml` and added `--workspace` there

PR #2778, opened 2026-01-23
Fix OpenAPI types: required, nullable and value_type

Lionqueen94:
- Added missing `#[serde(skip_serializing_if = "Option::is_none")]`
- Removed `#[serde(value_type = ...)]` when not needed
- Added `#[into_params(parameter_in = Query)]` to `LogFilterQuery` to make sure they are defined as query params and not path params. Then the `Option` field is not made nullable.
- Added test `check_openapi_json_for_null_type`
- Fixed `openapi.ts` params generator to add `?` when param is not required

### DIFF
--- a/frontend/scripts/openapi/generator.ts
+++ b/frontend/scripts/openapi/generator.ts
@@ -82,7 +82,8 @@ function addRequest(requestPath: string, request: OperationObject) {
       } else {
         const paramType = tsType(p.schema);
         requestPath = requestPath.replace(`{${p.name}}`, `\${${paramType}}`);
-        result.push(`${p.name}: ${paramType};`);
+        const optional = p.required ? "" : "?";
+        result.push(`${p.name}${optional}: ${paramType};`);
       }
     });
     result.push("}");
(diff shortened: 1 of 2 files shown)


oliver3:
Now that all types are fixed (thanks!!), should we add a test somewhere, or maybe throw an error in the generator if something is `nullable` and not `required`, so that we do not accidentally do this anymore?

Lionqueen94:
> Now that all types are fixed (thanks!!), should we add a test somewhere, or maybe throw an error in the generator if something is `nullable` and not `required`, so that we do not accidentally do this anymore?

There is one `| null` left in the params for the LogQueryFilter. Since they are params it's a bit different. Search for `since: number | null;` It's considered an optional value, but it's not rendered with a ? in the openapi.

Lionqueen94:
> > Now that all types are fixed (thanks!!), should we add a test somewhere, or maybe throw an error in the generator if something is `nullable` and not `required`, so that we do not accidentally do this anymore?
> 
> There is one `| null` left in the params for the LogQueryFilter. Since they are params it's a bit different. Search for `since: number | null;` It's considered an optional value, but it's not rendered with a ? in the openapi.

Fixed in 90d39e1c6a8336815df4dfb12c518e65c2372e28

PR #2779, opened 2026-01-23
Fix polling station home page getting out of date

### LINKED ISSUE #2749: Stembureau selectie scherm is niet altijd actueel
Het stembureau selectie scherm raakte bij een van de tests in de war. Als je 'm zelf refreshed gaat het weer goed. 

Er is een spike voor websockets, maar een fix voor de 1.0 is ook goed om te maken. 

Image

praseodym:
- **Keep feedback message in sync with status**
  Keep the feedback message in the polling station choice form in sync
  with any polling station status updates made after entering a polling
  station number, e.g. when the data entry list is opened triggering a
  refetch.
  

- **Refetch election status when typing**
  Refetch election status when typing the polling station number on the
  data entry home page. This is how the loading indicator was originally
  designed.
  

- **Enable regular refetch**
  This ensures that stale information will get refreshed without user
  action.

Resolves #2749.

~~We need to consider if we want to backport these changes. I think the
code change itself is small enough, but there might be unexpected
performance implications from these additional status refetches.~~
We have decided not to backport these changes to 1.0.

I will refactor the `PollingStationChoiceForm` component to improve
code quality and maintainability in a separate PR.

### DIFF
--- a/frontend/src/features/data_entry_home/components/PollingStationChoiceForm.test.tsx
+++ b/frontend/src/features/data_entry_home/components/PollingStationChoiceForm.test.tsx
@@ -434,24 +434,31 @@ describe("Test PollingStationChoiceForm", () => {
   describe("Polling station in progress", () => {
     test("Show polling stations as 'in progress'", async () => {
       server.use(ElectionRequestHandler);
-      overrideOnce("get", "api/elections/1/status", 200, {
-        statuses: [
-          { polling_station_id: 1, status: "empty" },
-          {
-            polling_station_id: 2,
-            status: "first_entry_in_progress",
-            first_entry_user_id: 1,
-            first_entry_progress: 42,
-          },
-          {
-            polling_station_id: 3,
-            status: "first_entry_in_progress",
-            first_entry_user_id: 1,
-            first_entry_progress: 42,
-          },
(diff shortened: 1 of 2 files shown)


frontend/src/features/data_entry_home/components/PollingStationChoiceForm.test.tsx
```diff
@@ -434,24 +434,31 @@ describe("Test PollingStationChoiceForm", () => {
...
-          {
-            polling_station_id: 3,
-            status: "first_entry_in_progress",
-            first_entry_user_id: 1,
-            first_entry_progress: 42,
-          },
-          { polling_station_id: 4, status: "definitive", first_entry_user_id: 1, second_entry_user_id: 2 },
-        ],
-      } satisfies ElectionStatusResponse);
+      server.use(
```
stacktraceghost, line 437: Seems like we use `overrideOnce` a bit randomly in the codebase. We could make a decision what should be the standard. 

In case of a backport, these changes are not necessary.

Lionqueen94: `overrideOnce` as the name states, only overrides the call once. We need to use `server.use` when we use the live data (auto refresh) since this permanently overrides the call. Otherwise the tests that use live data fail.

PR #2787, opened 2026-01-27
Apportionment structs/functions/tests

### LINKED ISSUE #2766: Add types with only what's needed for apportionment
Without using dependencies like `utoipa` and `serde`. `tracing` is fine to use.

### LINKED ISSUE #2767: Bring back apportionment functions
using the new types

stacktraceghost:
Resolves #2766
Resolves #2767

Changes:
- Brought back apportionment code
- Adjusted structs
  - As minimal as possible, not linked to Abacus
  - Implemented traits
- Reorganized code
  - Structure: split seat assignment and candidate nomination into folders
  - Renamed `PoliticalGroup` to `List`
  - Extracted `DisplayFraction` to Abacus domain layer
- Updated links to law articles
- Added newtypes
- Added apportionment crate to `sigrid.yaml`

- Moved `backend/src/apportionment/*`, `frontend` and other `backend` changes to separate branch ([bring_back_apportionment_api](https://github.com/kiesraad/abacus/tree/bring_back_apportionment_api)) for future epic.

Not in scope: #2785

### DIFF
--- a/backend/apportionment/src/candidate_nomination/mod.rs
+++ b/backend/apportionment/src/candidate_nomination/mod.rs
@@ -127,7 +127,7 @@ fn candidate_nomination_per_list<'a, T: ListVotesTrait>(
             } else {
                 let updated_ranking = update_candidate_ranking(
                     preference_threshold,
-                    candidate_votes_meeting_preference_threshold,
+                    &candidate_votes_meeting_preference_threshold,
                     candidate_votes,
                 );
 
@@ -193,7 +193,7 @@ fn other_candidate_nomination<'a, T: CandidateVotesTrait>(
 
 /// List the candidates nominated with preferential votes
 fn preferential_candidate_nomination<'a, T: CandidateVotesTrait>(
-    candidates_meeting_preference_threshold: &Vec<&'a T>,
+    candidates_meeting_preference_threshold: &[&'a T],
     list_seats: u32,
 ) -> Result<Vec<&'a T>, ApportionmentError> {
     let mut preferential_candidate_nomination: Vec<&T> = vec![];
@@ -234,7 +234,7 @@ fn preferential_candidate_nomination<'a, T: CandidateVotesTrait>(
 /// to the top of the list and keeping the ranking of the rest of candidates on the list the same
 fn update_candidate_ranking<T: CandidateVotesTrait>(
     preference_threshold: Fraction,
-    candidate_votes_meeting_preference_threshold: Vec<&T>,
+    candidate_votes_meeting_preference_threshold: &[&T],
     candidate_votes: &[T],
 ) -> Vec<CandidateNumber> {
     let mut updated_candidate_ranking: Vec<CandidateNumber> = vec![];


backend/apportionment/src/structs.rs
```diff
@@ -0,0 +1,152 @@
...
+            list_votes: list_votes_from_input(input.0),
+            quota: input.1.quota,
+            total_seats_per_list: get_total_seats_from_apportionment_result(input.1),
+        }
+    }
+}
+
+// Could also be done this way, preferences?
+// impl CandidateNominationInput {
+//     pub fn new(input: impl ApportionmentInput, seat_assignment: &SeatAssignmentResult) -> Self {
```
stacktraceghost: @rnijveld Which one would you prefer?

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -0,0 +1,796 @@
...
+            percentage: preference_threshold_percentage,
+            number_of_votes: preference_threshold,
+        },
+        chosen_candidates,
+        list_candidate_nomination,
+    })
+}
+
+fn all_sorted_chosen_candidates(
+    list_votes: &Vec<ListVotes>,
```
praseodym:
```suggestion
    list_votes: &[ListVotes],
```

backend/apportionment/src/lib.rs
```diff
@@ -1,14 +1,32 @@
...
+    seat_assignment::SeatAssignmentResult,
+    structs::{ApportionmentError, ApportionmentInput, CandidateVotesTrait, ListVotesTrait},
+};
+use self::{
+    candidate_nomination::candidate_nomination,
+    seat_assignment::seat_assignment,
+    structs::{ApportionmentOutput, CandidateNominationInput, SeatAssignmentInput},
+};
+
+pub async fn process(
```
praseodym: Looks like this function is not actually async

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -0,0 +1,796 @@
...
+
+/// Candidate nomination
+pub fn candidate_nomination(
+    input: CandidateNominationInput,
+) -> Result<CandidateNominationResult, ApportionmentError> {
+    info!("Candidate nomination");
+
+    // [Artikel P 15 Kieswet](https://wetten.overheid.nl/BWBR0004627/2026-01-01/#AfdelingII_HoofdstukP_Paragraaf3_ArtikelP15)
+    // Calculate preference threshold as a proper fraction
+    let preference_threshold_percentage = if input.number_of_seats >= 19 { 25 } else { 50 };
```
praseodym: I think we should introduce a constant (`LARGE_COUNCIL_THRESHOLD`?) instead of hardcoding the number 19.

Lionqueen94:
@praseodym Should we already add the apportionment crate to `sigrid.yml` in this PR?

praseodym:
> @praseodym Should we already add the apportionment crate to `sigrid.yml` in this PR?

Yes!

backend/apportionment/src/candidate_nomination/mod.rs
Lionqueen94: Changed in 45c9311bd9122b8e7a2e3e5344457754bc9ce2e2

backend/apportionment/src/lib.rs
Lionqueen94: True, fixed in 45c9311bd9122b8e7a2e3e5344457754bc9ce2e2

backend/apportionment/src/candidate_nomination/mod.rs
Lionqueen94: Added in 45c9311bd9122b8e7a2e3e5344457754bc9ce2e2

backend/apportionment/src/lib.rs
cikzh: There is a clippy lint that checks for unused async blocks, but it's in Pedantic: https://rust-lang.github.io/rust-clippy/master/index.html#unused_async

backend/apportionment/src/id_macro.rs
```diff
@@ -0,0 +1,36 @@
+macro_rules! id {
```
cikzh: The macro name is confusing. It's the same as the one we use for database ID's. Maybe something like `int_newtype` or `index_number`?

backend/apportionment/src/lib.rs
praseodym: Thanks, created #2805 to enable that

backend/apportionment/src/id_macro.rs
Lionqueen94: Renamed the macro in 8207b92e8a1b8dfee05bd99affd98bc504bc388a

backend/apportionment/src/structs.rs
```diff
@@ -0,0 +1,154 @@
...
+//             number_of_seats: input.number_of_seats(),
+//             list_votes: list_votes_from_input(&input),
+
+//             quota: seat_assignment.quota,
+//             total_seats: get_total_seats_from_apportionment_result(&seat_assignment),
+//         }
+//     }
+// }
+
+fn list_votes_from_input<T: ApportionmentInput>(input: &T) -> Vec<ListVotes> {
```
rnijveld: I don't think we should be doing this. Now we're copying data again. Instead I would add a generic on the `CandidateNominationInput` and have the data just be the same traits as we already had on the `ApportionmentInput`, that removes all of this copying code, and at the same time it allows making fixtures way easier, because the same kind of stuff can then be used for both apportionment and candidate nomination. Another disadvantage of this is that the semantics of our own CandidateVotes/ListVotes here could be different from what the trait variants from the ApportionmentInput were (although unlikely), but that could result in subtle bugs or harder to do refactors.

backend/apportionment/src/structs.rs
rnijveld: I would use this variant. Main reason would be that `From` would never allow us to switch to references (ok technically possible, but we should not over-use the existing traits in Rust, especially as there isn't really a need to create a CandidateNominationInput from anything else but the seat assignment result plus apportionment input). 

Currently we're taking ApportionmentInput by value, which means that on the Abacus side we would lose access our value. I don't think we really need to be taking that input by value, and we should use a reference instead, keeping the struct available on the abacus side without having to clone it (with potential for the data to get out of sync on the abacus side). Especially if that input also needs to be transferred along together with the results of this apportionment module.

Lionqueen94: Changed in 4fad7bc609ef2a10881313c5b8bac9cc78d4cf39

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+        input.number_of_seats,
+        input.list_votes,
+        preference_threshold,
+        &input.total_seats_per_list,
+    )?;
+    // TODO: needs debug derive, do we want this?
+    // debug!(
+    //     "List candidate nomination: {:#?}",
+    //     list_candidate_nomination
+    // );
```
Lionqueen94: @praseodym This is mostly nice for when you're developing, we can also remove this debug call, since it means we need to add Debug to the traits.

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+    candidate_votes: &Vec<&T>,
+) -> Vec<CandidateNumber> {
+    candidate_votes
+        .iter()
+        .map(|candidate| candidate.number())
+        .collect()
+}
+
+/// List the other candidates nominated
+// TODO: Check lifetime, is this correct?
```
Lionqueen94: @praseodym We tried to remove lifetimes where possible, but it seems to still be needed here, do you agree?

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -0,0 +1,816 @@
...
+            // Check if we can actually nominate all these candidates, otherwise we would need to draw lots
+            if same_votes_candidates.len() > non_assigned_seats as usize {
+                // TODO: #788 if multiple lists have the same largest remainder and not enough residual seats are available, use drawing of lots
+                // TODO: how to fix? Type mismatch [E0308]
+                //  Expected: &Vec< &_>
+                //  Found: Vec<&&T>
+                // info!(
+                //     "Drawing of lots is required for candidates: {:?}, only {non_assigned_seats} seat(s) available",
+                //     candidate_votes_numbers(same_votes_candidates)
+                // );
```
Lionqueen94: @praseodym Do you have a suggestion on how we can fix this error?

backend/apportionment/src/structs.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn candidate_votes(&self) -> &[Self::Cv];
+}
+
+pub trait CandidateVotesTrait: PartialEq {
+    fn number(&self) -> CandidateNumber;
+    fn votes(&self) -> u32;
+}
+
+// TODO: Note, since these are identical to ApportionmentInput,
+//  we can also drop these and use ApportionmentInput directly
```
Lionqueen94: @praseodym What do you think?

backend/apportionment/src/test_helpers.rs
```diff
@@ -0,0 +1,174 @@
...
+    candidate_votes: Vec<Vec<u32>>,
+    total_seats_per_list: Vec<u32>,
+) -> CandidateNominationInput<'static, ListVotesMock> {
+    let seat_assignment_input =
+        seat_assignment_fixture_with_given_candidate_votes(number_of_seats, candidate_votes);
+
+    CandidateNominationInput {
+        number_of_seats: seat_assignment_input.number_of_seats,
+        // TODO: solve this leak
+        list_votes: Box::leak(seat_assignment_input.list_votes.into_boxed_slice()),
```
Lionqueen94: @praseodym Is there any way we can do this without a leak?

backend/apportionment/src/candidate_nomination/mod.rs
praseodym: Adding a Debug derive is ok.

backend/apportionment/src/candidate_nomination/mod.rs
praseodym: Yes they are needed because the return value uses the references to the `candidate_votes` parameter.

backend/apportionment/src/candidate_nomination/mod.rs
praseodym: Maybe something like this:

```rust
            let same_votes_candidates: Vec<&T> = candidates_meeting_preference_threshold
                .iter()
                .copied()
                .filter(|candidate_votes| {
                    !preferential_candidate_nomination.contains(candidate_votes)
                        && candidate_votes.votes()
                            == candidates_meeting_preference_threshold[index].votes()
                })
                .collect();
            // Check if we can actually nominate all these candidates, otherwise we would need to draw lots
            if same_votes_candidates.len() > non_assigned_seats as usize {
                // TODO: #788 if multiple lists have the same largest remainder and not enough residual seats are available, use drawing of lots
                info!(
                    "Drawing of lots is required for candidates: {:?}, only {non_assigned_seats} seat(s) available",
                    candidate_votes_numbers(&same_votes_candidates)
                );
                return Err(ApportionmentError::DrawingOfLotsNotImplemented);
            } else {
                // Nominate candidate to seat
                preferential_candidate_nomination
                    .push(candidates_meeting_preference_threshold[index]);
            }
```

backend/apportionment/src/structs.rs
praseodym: I think using `ApportionmentInput` directly is fine.

backend/apportionment/src/test_helpers.rs
praseodym: **Option 1:** change the fixture function to accept a `SeatAssignmentInputMock`:

```rust
pub fn candidate_nomination_fixture_with_given_number_of_seats<'a>(
    quota: Fraction,
    seat_assignment_input: &'a SeatAssignmentInputMock,
    total_seats_per_list: Vec<u32>,
) -> CandidateNominationInput<'a, ListVotesMock> {
    CandidateNominationInput {
        number_of_seats: seat_assignment_input.number_of_seats,
        list_votes: &seat_assignment_input.list_votes,
        quota,
        total_seats_per_list,
    }
}
```

And create it inside the caller (test) function:

```rust
let seat_assignment_input = seat_assignment_fixture_with_given_candidate_votes(
    number_of_seats,
    candidate_votes,
);
let input = candidate_nomination_fixture_with_given_number_of_seats(
    quota,
    &seat_assignment_input,
    total_seats_per_list,
);
```

**Option 2:** create a separate struct for tests that owns its data:

```rust
pub struct CandidateNominationInputOwned<L: ListVotesTrait> {
    pub number_of_seats: u32,
    pub list_votes: Vec<L>,
    pub quota: Fraction,
    pub total_seats_per_list: Vec<u32>,
}

impl<L: ListVotesTrait> CandidateNominationInputOwned<L> {
    pub fn as_ref(&self) -> CandidateNominationInput<'_, L> {
        CandidateNominationInput {
            number_of_seats: self.number_of_seats,
            list_votes: &self.list_votes,
            quota: self.quota.clone(),
            total_seats_per_list: self.total_seats_per_list.clone(),
        }
    }
}
```

And then in the caller (test) function:

```rust
let owned = candidate_nomination_fixture_owned(...);
let input = owned.as_ref();
```

backend/apportionment/src/candidate_nomination/mod.rs
Lionqueen94: Done in the last amend of 2bcdcff01e78bc8c8b5d40fcbe1e980f005eeade

backend/apportionment/src/candidate_nomination/mod.rs
Lionqueen94: Done in the last amend of https://github.com/kiesraad/abacus/commit/2bcdcff01e78bc8c8b5d40fcbe1e980f005eeade

backend/apportionment/src/candidate_nomination/mod.rs
Lionqueen94: Nice! Done in the last amend of https://github.com/kiesraad/abacus/commit/2bcdcff01e78bc8c8b5d40fcbe1e980f005eeade

backend/apportionment/src/structs.rs
Lionqueen94: Changed in the last amend of https://github.com/kiesraad/abacus/commit/2bcdcff01e78bc8c8b5d40fcbe1e980f005eeade

backend/apportionment/src/test_helpers.rs
Lionqueen94: We went for option 1, we somehow didn't realize it was so easy to fix haha. Done in the last amend of https://github.com/kiesraad/abacus/commit/2bcdcff01e78bc8c8b5d40fcbe1e980f005eeade

backend/apportionment/src/structs.rs
Lionqueen94: We improved the code a lot in 2bcdcff01e78bc8c8b5d40fcbe1e980f005eeade. More improvements might be possible, this will be done in a separate PR (as discussed).

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -0,0 +1,830 @@
...
+            &lists[0].number,
+            &lists[3].candidate_votes[..1],
+            &lists[3].candidate_votes[2..],
+        );
+        check_chosen_candidates(
+            &result.chosen_candidates,
+            &lists[0].number,
+            &lists[4].candidate_votes[..1],
+            &lists[4].candidate_votes[2..],
+        );
```
praseodym, line 493: Shouldn't these tests use `.number` from each subsequent list, i.e. `&lists[1].number` and higher, instead of `&lists[0].number`?

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -0,0 +1,830 @@
...
+        .iter()
+        .filter(|candidate_votes| Fraction::from(candidate_votes.votes()) >= preference_threshold)
+        .collect();
+    candidates_meeting_preference_threshold.sort_by_key(|b| std::cmp::Reverse(b.votes()));
+    candidates_meeting_preference_threshold
+}
+
+/// Create a vector containing just the candidate numbers from an iterator of candidate votes
+fn candidate_votes_numbers<T: CandidateVotesTrait>(
+    candidate_votes: &Vec<&T>,
```
praseodym: Why not `&[T]`?

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -0,0 +1,830 @@
...
+        }
+    }
+    Ok(preferential_candidate_nomination)
+}
+
+/// Update the candidate list, moving the candidates meeting the preference threshold
+/// to the top of the list and keeping the ranking of the rest of candidates on the list the same
+fn update_candidate_ranking<T: CandidateVotesTrait>(
+    preference_threshold: Fraction,
+    candidate_votes_meeting_preference_threshold: Vec<&T>,
```
praseodym: Why not `&[&T]`?

backend/apportionment/src/candidate_nomination/structs.rs
```diff
@@ -0,0 +1,50 @@
...
+/// Contains information about the chosen candidates and
+/// the candidate list ranking for a specific list.
+#[derive(Debug, PartialEq)]
+pub struct ListCandidateNomination<'a, T: CandidateVotesTrait> {
+    /// List number for which this nomination applies
+    pub list_number: ListNumber,
+    /// The number of seats assigned to this group
+    pub list_seats: u32,
+    /// The list of chosen candidates via preferential votes, can be empty
+    // TODO: check lifetimes
```
praseodym: Remove TODO?

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -0,0 +1,830 @@
...
+            &result.list_candidate_nomination[4],
+            list_4_preferential_nominated_candidate_numbers,
+            list_4_other_nominated_candidate_numbers,
+            &[],
+        );
+
+        let lists = input.list_votes;
+        let (list_0_chosen_candidates, list_0_not_chosen_candidates) =
+            get_chosen_and_not_chosen_candidates_for_a_list(
+                lists[0].candidate_votes.as_slice(),
```
praseodym:
```suggestion
                &lists[0].candidate_votes,
```
And same with `.as_slice()` below

backend/apportionment/Cargo.toml
```diff
@@ -7,3 +7,6 @@ rust-version.workspace = true
 
 [dependencies]
 tracing.workspace = true
+
+[dev-dependencies]
+test-log = "0.2.19"
```
praseodym: `test-log` should be added to the workspace dependencies, just like the tracing dependencies

Lionqueen94: Fixed in e1df11600844f39b7dae765d38988acda2469aff

backend/apportionment/src/candidate_nomination/mod.rs
Lionqueen94: Fixed in e1df11600844f39b7dae765d38988acda2469aff

backend/apportionment/src/candidate_nomination/structs.rs
Lionqueen94: Fixed in e1df11600844f39b7dae765d38988acda2469aff

backend/apportionment/src/candidate_nomination/mod.rs
Lionqueen94: Good catch! Fixed in e1df11600844f39b7dae765d38988acda2469aff

backend/apportionment/src/candidate_nomination/mod.rs
Lionqueen94: Changed to `&[&T]` in e1df11600844f39b7dae765d38988acda2469aff

backend/apportionment/src/candidate_nomination/mod.rs
Lionqueen94: Changed to `&[&T]` in e1df11600844f39b7dae765d38988acda2469aff

PR #2789, opened 2026-01-27
Improve backend modularity

oliver3:
Step two of [PR Improve modularity of backend modules #988](https://github.com/kiesraad/abacus/issues/988)

The following steps are done in separate commits, although I'm not sure if that helps in reviewing:
- Add `infra`, `api`, `repository` and `domain` layer modules
- Reformat imports with rustfmt nightly `--config imports_granularity=Crate,group_imports=StdExternalCrate`
- For each module in abacus backend: 
  - move the files within to the most appropriate layer, without pulling files apart
  - fix and reformat imports
  - make sure everything works
- Update sigrid.yaml
- Update imports in abacus fuzz

image

PR #2790, opened 2026-01-28
Data entry structure mapping simplification

### LINKED ISSUE #2575: Investigate data entry structure mapping in backend
Investigate mapping the data entry structure in the backend instead, so the frontend can send a flat map (dictionary) of dot-separated form fields and their values (e.g. `{"section_a.field_one": 1234}`. Also consider how validation results are built based on the current struct, and mapped back to dot-separated field names.

We could check out which options Serde has, or consider writing a custom mapping macro. The result should be a spike to see what's possible.

If instead we decide to keep the mapping in the frontend instead of moving it to the backend, we should clean up the mapping tests in the frontend because there is a lot of overlap in those tests.

stacktraceghost:
Resolves #2575
See comment https://github.com/kiesraad/abacus/issues/2575#issuecomment-3804267792

Changes:
- Use dots for path array notation in frontend (`political_group_votes.1.total` instead of `political_group_votes[1].total`).
  - The most relevant changes can be found in `frontend/src/utils/dataEntryMapping.ts`
  - Other changes are mainly tests
- Refactor path methods/add docs
  - Removed logic to manipulate the given data structure when a path did not exist.
  - Throw errors if a path doesn't exist or a value can't be set
  - Added extra tests
- Fixed [polling-station-results.ts‎](https://github.com/kiesraad/abacus/pull/2790/changes#diff-4f41375b1fc14e0335d932b2e90f25cbd2acae4cbdd2686c05f1a5b7b8c240d0) to match the candidate counts of mocked data and fixed a test with wrong expectations due to duplicate candidates in `polling-station-results.ts‎`

Note: with these changes, the data structure on which we map the form data needs to contain the given paths. This also includes the amount of political groups and candidates. For example, if we try to map `political_group_votes.0.candidate_votes.30.votes` while there are only 10 candidates in the data structure, this will now throw an error.

### DIFF
--- a/frontend/src/features/resolve_differences/testing/polling-station-results.ts
+++ b/frontend/src/features/resolve_differences/testing/polling-station-results.ts
@@ -7,10 +7,12 @@ import type { Candidate, CandidateVotes, PollingStationResults } from "@/types/g
  * @param votes Mapping of candidate number to votes
  * @returns List of CandidateVotes
  */
-const baseCandidateVotesOn = (candidates: Candidate[] | undefined, votes: Record<number, number>): CandidateVotes[] => {
-  // Get candidate numbers based on candidates or from the votes keys
-  const candidateNumbers = candidates?.map((c) => c.number) ?? Object.keys(votes).map((k) => parseInt(k, 10));
-  const expectedLength = Math.max(...candidateNumbers, 0);
+const baseCandidateVotesOn = (candidates: Candidate[], votes: Record<number, number>): CandidateVotes[] => {
+  const expectedLength = Math.max(
+    ...candidates.map((c) => c.number),
+    ...Object.keys(votes).map((k) => parseInt(k, 10)),
+    0,
+  );
 
   return Array.from({ length: expectedLength }, (_, i) => ({ number: i + 1, votes: votes[i + 1] ?? 0 }));
 };
@@ -59,7 +61,7 @@ export function pollingStationResultsMockData(first: boolean): PollingStationRes
       {
         number: 1,
         total: first ? 1512 : 1481,
-        candidate_votes: baseCandidateVotesOn(electionMockData.political_groups[0]?.candidates, {
+        candidate_votes: baseCandidateVotesOn(electionMockData.political_groups[0]!.candidates, {
           1: first ? 1256 : 1258,
           2: 128,
           3: first ? 65 : 63,
@@ -75,7 +77,7 @@ export function pollingStationResultsMockData(first: boolean): PollingStationRes
       {
         number: 2,
         total: 2,
-        candidate_votes: baseCandidateVotesOn(electionMockData.political_groups[1]?.candidates, { 1: 2 }),
+        candidate_votes: baseCandidateVotesOn(electionMockData.political_groups[1]!.candidates, { 1: 2 }),
       },
     ],
   };
--- a/frontend/biome.jsonc
+++ b/frontend/biome.jsonc
@@ -103,8 +103,8 @@
         "**/*.test.ts",
         "**/*.test.tsx",
         "**/*.stories.tsx",
-        "src/testing/**/*.ts",
-        "src/testing/**/*.tsx"
+        "src/**/testing/**/*.ts",
+        "src/**/testing/**/*.tsx"
       ],
       "linter": {
         "rules": {
--- a/frontend/eslint.config.js
+++ b/frontend/eslint.config.js
@@ -103,7 +103,7 @@ export default defineConfig(
       "**/*.test.ts{,x}",
       ".storybook/**/*.ts{,x}",
       "e2e-tests/**/*",
-      "src/testing/**/*.ts{,x}",
+      "src/**/testing/**/*.ts{,x}",
     ],
     rules: {
       "@typescript-eslint/no-non-null-assertion": "off",


frontend/src/utils/dataEntryMapping.ts
```diff
@@ -62,59 +73,75 @@ export function mapResultsToSectionValues(section: DataEntrySection, results: Da
...
+  const segments = path.split(".");
+  const lastProperty = segments.pop();
+  const refParent = traversePath(data, segments, true);
 
-  const finalSegment = segments[segments.length - 1];
-  if (finalSegment !== undefined && isRecord(current)) {
-    current[finalSegment] = processedValue;
+  if (isRecord(refParent) && lastProperty) {
+    refParent[lastProperty] = processValue(value, valueType);
   }
```
oliver3, line 116: So the assumption here is that path does not end in an index such as `.0` ? Or do we consider arrays to be a record as well? In any case, I would feel more insured if there was an else to this `if` to avoid silently doing nothing in that case.

frontend/src/utils/dataEntryMapping.ts
```diff
@@ -62,59 +73,75 @@ export function mapResultsToSectionValues(section: DataEntrySection, results: Da
...
+    while (arr.length < index + 1) {
+      arr.push({});
+    }
+  };
 
-  const result = segments.reduce<unknown>((current, segment) => {
-    if (current === undefined) return undefined;
+  return segments.reduce<unknown>((prev, curr, index, segments) => {
+    if (isRecord(prev)) {
+      if (createMissing) handleMissingRecord(prev, curr, segments[index + 1]);
```
oliver3: I would be more comfortable with this `if` without a block if we would have ESLint rule [`curly: ["error", "multi-line"],`](https://eslint.org/docs/latest/rules/curly#multi-line).

Unfortunately biome's [use-block-statements](https://biomejs.dev/linter/rules/use-block-statements/) cannot be configured to allow single line if without blocks, but maybe we still want that instead? It will result in 14 errors on this branch.

frontend/src/utils/dataEntryMapping.ts
```diff
@@ -135,59 +162,22 @@ function processValue(
   return value;
 }
 
+/**
+ * Checks if a string represents a valid integer
+ * @param str The string to check
+ * @returns True if the string is an integer, false otherwise
+ */
+function stringContainsInteger(str: string): boolean {
```
oliver3: For me the word contains means "somewhere in the string", maybe rename it to something like `stringIsInteger` ?

stacktraceghost: I actually changed it from `stringIsInteger`, because it felt too much like a type guard name (like below). Maybe `stringRepresentsInteger`?

oliver3:
Overall good improvement! And thanks for adding the jsdocs!

frontend/src/utils/dataEntryMapping.ts
oliver3: Good point, and `stringRepresentsInteger` was going to be my other suggestion :) 👍

frontend/src/utils/dataEntryMapping.ts
oliver3: As discussed: this is actually something unrelated to this PR, I'll resolve the conversation.

praseodym:
@stacktraceghost this needs to be rebased now that #2789 was merged

frontend/src/utils/dataEntryMapping.ts
stacktraceghost: The code has been adjusted to also handle arrays and throw errors in other cases.

frontend/src/utils/dataEntryMapping.ts
stacktraceghost: Adjusted!

stacktraceghost:
Removed logic that manipulated the given data structure when a path did not exist. This was old code that isn't relevant anymore. In the current version, we receive the entire data structure from the backend and we map that to the form fields. When we map back, we can expect that all these paths exist in the data structure. This also counts for arrays, since we do not add/delete array items (like political groups or candidates). 

Updated the description of the PR as well.

Lionqueen94 (dismissed):
Very nice improvement!

frontend/src/features/resolve_differences/testing/polling-station-results.ts
```diff
@@ -44,30 +59,23 @@ export function pollingStationResultsMockData(first: boolean): PollingStationRes
...
-          { number: 5, votes: 4 },
-          { number: 6, votes: 4 },
-          { number: 7, votes: 3 },
-          { number: 8, votes: 2 },
-          { number: 9, votes: 0 },
-          { number: 10, votes: first ? 4 : 0 },
-          { number: 11, votes: first ? 2 : 4 },
-          { number: 12, votes: 0 },
-        ],
+        candidate_votes: baseCandidateVotesOn(electionMockData.political_groups[0]?.candidates, {
```
oliver3:
```suggestion
        candidate_votes: baseCandidateVotesOn(electionMockData.political_groups[0]!.candidates, {
```

Exclamation marks are allowed if you change the paths in `eslint.config.js` and `biome.jsonc` from `src/testing/**` to `src/**/testing/**` as it should be.

frontend/src/utils/dataEntryMapping.ts
oliver3: Clean and simple, thanks!!

frontend/src/features/resolve_differences/testing/polling-station-results.ts
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2790/changes/844ce30cc6d5c78fc606cf16dbe0fed7b58cab5b. I still need to check both lists to see which has the highest number.

oliver3: Thanks for updating the configs though :)

PR #2799, opened 2026-01-29
Fixed e2e tests when running multiple browsers concurrently

### LINKED ISSUE #2793: Fix e2e tests when running multiple browsers concurrently
Fix e2e tests so that `pnpm test:e2e` (with all browsers on one Abacus instance) works again. See #2791 for more details.

lama-d3v:
Fixes #2793

### DIFF
--- a/frontend/e2e-tests/tests/full-flow.e2e.ts
+++ b/frontend/e2e-tests/tests/full-flow.e2e.ts
@@ -83,15 +83,14 @@ test.describe("full flow", () => {
     });
     expect(response.status()).toBe(201);
 
-    const username2 = `admin-${browserName}`;
-    const loginResponse = await loginAs(request, username2, "Temp");
+    const loginResponse = await loginAs(request, username, "Temp");
     expect(loginResponse.status()).toBe(200);
 
     const response2 = await request.put("/api/account", {
       data: {
-        username: username2,
+        username: username,
         fullname: "John Doe",
-        password: getTestPassword(username2),
+        password: getTestPassword(username),
       },
     });
 
@@ -117,14 +116,14 @@ test.describe("full flow", () => {
     await userCreateRolePgObj.coordinator.click();
     await userCreateRolePgObj.continue.click();
 
-    const usernameFull = `coordinator-${browserName}`;
+    const coordinatorUsername = `coordinator-${browserName}`;
     const userCreateDetailsPgObj = new UserCreateDetailsPgObj(page);
-    await userCreateDetailsPgObj.username.fill(usernameFull);
-    await userCreateDetailsPgObj.fullname.fill(`Coordinator ${usernameFull}`);
-    await userCreateDetailsPgObj.password.fill(getTestPassword(usernameFull, "Temp"));
+    await userCreateDetailsPgObj.username.fill(coordinatorUsername);
+    await userCreateDetailsPgObj.fullname.fill(`Coordinator ${coordinatorUsername}`);
+    await userCreateDetailsPgObj.password.fill(getTestPassword(coordinatorUsername, "Temp"));
     await userCreateDetailsPgObj.save.click();
 
-    await expect(userListPgObj.alert).toContainText(`${usernameFull} is toegevoegd met de rol Coördinator`);
+    await expect(userListPgObj.alert).toContainText(`${coordinatorUsername} is toegevoegd met de rol Coördinator`);
   });
 
   test(`complete account for coordinator`, async ({ page, browserName }) => {
@@ -167,14 +166,14 @@ test.describe("full flow", () => {
       const userCreateTypePgObj = new UserCreateTypePgObj(page);
       await userCreateTypePgObj.continue.click();
 
-      const usernameFull = `${username}-${browserName}`;
+      const typistUsername = `${username}-${browserName}`;
       const userCreateDetailsPgObj = new UserCreateDetailsPgObj(page);
-      await userCreateDetailsPgObj.username.fill(usernameFull);
-      await userCreateDetailsPgObj.fullname.fill(`Typist ${usernameFull}`);
-      await userCreateDetailsPgObj.password.fill(usernameFull.repeat(3));
+      await userCreateDetailsPgObj.username.fill(typistUsername);
+      await userCreateDetailsPgObj.fullname.fill(`Typist ${typistUsername}`);
+      await userCreateDetailsPgObj.password.fill(typistUsername.repeat(3));
       await userCreateDetailsPgObj.save.click();
 
-      await expect(userListPgObj.alert).toContainText(`${usernameFull} is toegevoegd met de rol Invoerder`);
+      await expect(userListPgObj.alert).toContainText(`${typistUsername} is toegevoegd met de rol Invoerder`);
     }
   });
 


frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -158,7 +163,9 @@ test.describe("full flow", () => {
   });
 
   test("start data entry", async ({ page }) => {
+    const adminContext = await request.newContext();
     await page.goto("/account/login");
+    await adminContext.storageState({ path: "e2e-tests/state/coordinator2.json" });
 
     const loginPage = new LoginPgObj(page);
     await loginPage.login("coordinator2", getTestPassword("coordinator2"));
```
jschuurk-kr: These lines are mixing up `storageState` and logging in.

What happens here is:
1. we create an `adminContext` with the storageState of `coordinator2`;
2. we don't use that `adminContext` in the rest of the test;
3. we log in as `coordinator2`, invalidating the session of the `adminContext`.

frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -63,12 +63,16 @@ const investigations = [
...
+  storageState: "e2e-tests/state/admin2.json",
+});
+
 test.describe.configure({ mode: "serial" });
 
 test.describe("full flow", () => {
   let electionId: number | null = null;
 
-  test(`create typists for flow`, async ({ page }) => {
+  test(`create typists for flow`, async ({ page, browserName }) => {
```
jschuurk-kr: An advantage of adding a random string to the usernames is that we can do it where they are defined as `const` as on line 65 for the typists. You do then lose the information on which browser the test is running, but the test outut/report also contains that information.

lama-d3v: typist3 and typist4 are also used in these tests: https://github.com/kiesraad/abacus/blob/33bb5e5a68906a988a896f17e555ca4e9387c42a/frontend/e2e-tests/tests/full-flow.e2e.ts#L256
https://github.com/kiesraad/abacus/blob/33bb5e5a68906a988a896f17e555ca4e9387c42a/frontend/e2e-tests/tests/full-flow.e2e.ts#L276
Do I need to use a random string for the usernames instead?

frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -61,23 +62,56 @@ const investigations = [
...
 test.describe.configure({ mode: "serial" });
 
 test.describe("full flow", () => {
   let electionId: number | null = null;
 
-  test(`create typists for flow`, async ({ page }) => {
+  test("create test user accounts for browsers", async ({ admin, browserName }) => {
+    const { request } = admin;
+
+    for (const user of testUsers) {
```
jschuurk-kr: This creates five users of which the test uses only two. And then the next test creates two additional browser-specific users.

I'd prefer it if the `admin` fixture is used to create a new admin. That new admin can then create a coordinator and two typists through the page objects.

frontend/e2e-tests/tests/full-flow.e2e.ts
jschuurk-kr: Discussed last week: browsername is sufficient, no random string needed.

frontend/e2e-tests/tests/full-flow.e2e.ts
lama-d3v: Fixed in https://github.com/kiesraad/abacus/pull/2799/commits/adcb7bd2f632ec7c778c1e0e3097a422efdf9145

frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -61,23 +61,100 @@ const investigations = [
...
+      data: {
+        role: "administrator",
+        fullname: "John Doe",
+        username: username,
+        temp_password: getTestPassword(username, "Temp"),
+      },
+    });
+    expect(response.status()).toBe(201);
+
+    const username2 = `admin-${browserName}`;
```
Lionqueen94: This line can be removed and you can use `username` below.

frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -61,23 +61,100 @@ const investigations = [
...
+    await userCreateRolePgObj.continue.click();
+
+    const usernameFull = `coordinator-${browserName}`;
+    const userCreateDetailsPgObj = new UserCreateDetailsPgObj(page);
+    await userCreateDetailsPgObj.username.fill(usernameFull);
+    await userCreateDetailsPgObj.fullname.fill(`Coordinator ${usernameFull}`);
+    await userCreateDetailsPgObj.password.fill(getTestPassword(usernameFull, "Temp"));
+    await userCreateDetailsPgObj.save.click();
+
+    await expect(userListPgObj.alert).toContainText(`${usernameFull} is toegevoegd met de rol Coördinator`);
```
Lionqueen94:
```suggestion
    await expect(userListPgObj.alert).toContainText(`Coordinator ${usernameFull} is toegevoegd met de rol Coördinator`);
```

frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -61,23 +61,100 @@ const investigations = [
...
+    await page.goto(`/users`);
+
+    const userListPgObj = new UserListPgObj(page);
+    await userListPgObj.create.click();
+
+    const userCreateRolePgObj = new UserCreateRolePgObj(page);
+    await userCreateRolePgObj.coordinator.click();
+    await userCreateRolePgObj.continue.click();
+
+    const usernameFull = `coordinator-${browserName}`;
```
Lionqueen94: This is just the `coordinatorUsername`, the full name is with `Coordinator` added in front of it.

frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -90,21 +167,23 @@ test.describe("full flow", () => {
       const userCreateTypePgObj = new UserCreateTypePgObj(page);
       await userCreateTypePgObj.continue.click();
 
+      const usernameFull = `${username}-${browserName}`;
```
Lionqueen94: Same as the above comment, this is just the typistUsername, not the full name as well, since `Typist` is added for the full name.

frontend/e2e-tests/tests/full-flow.e2e.ts
praseodym: Then change the variable to contain the full name, the proposed change makes it look like the message text itself starts with ‘Coordinator’.

frontend/e2e-tests/tests/full-flow.e2e.ts
Lionqueen94: Changed in 025e50482f1c4df9276fa163c3e2dbad03b4c520

frontend/e2e-tests/tests/full-flow.e2e.ts
Lionqueen94: Changed const name in 025e50482f1c4df9276fa163c3e2dbad03b4c520

frontend/e2e-tests/tests/full-flow.e2e.ts
Lionqueen94: Changed const name in 025e50482f1c4df9276fa163c3e2dbad03b4c520

frontend/e2e-tests/tests/full-flow.e2e.ts
Lionqueen94: Changed const name in https://github.com/kiesraad/abacus/pull/2799

frontend/e2e-tests/tests/full-flow.e2e.ts
praseodym: It looks like the `toContainText` check still checks for only the user name instead of the full name

Lionqueen94: Yes because that is actually what's in the alert, I checked with the design and implementation. There are anonymous users as well and they don't have a full name yet, but I forgot about that originally when I commented

PR #2800, opened 2026-01-29
Sigrid: use component_base_dirs for backend

oliver3:
Currently, for Sigrid the backend is mainly split up into the four layers which is too big to be useful in Sigrid.

image

With this PR I want to introduce more fine-grained components, and because all components are new anyway, this is the most opportune moment to try out Sigrid's `component_base_dirs` feature to auto-detect components.

Also, as a first commit, I've used a different syntax in the YAML for list and dict, which I think makes it easier for humans to maintain the list of components. Please comment if this is an improvement or not.

### DIFF
--- a/sigrid.yaml
+++ b/sigrid.yaml
@@ -105,8 +105,8 @@ exclude:
   - ".*/frontend/.*[.]d[.]ts"
   - ".*/frontend/.*[.]config[.]ts"
   - ".*/frontend/package.json"
-  - ".*/frontend/tsconfig[.]ts"
-  - "docker-compose.yml"
+  - ".*/frontend/tsconfig[.]json"
+  - "docker-compose[.]yml"
 
 dependencychecker:
   blocklist:


sigrid.yaml
```diff
@@ -1,174 +1,58 @@
...
-  - name: "(Backend) Integration tests"
-    include:
-      - ".*/backend/tests/.*"
-  - name: "(Backend) SQL Migrations"
-    include:
-      - ".*/backend/migrations/.*"
-  - name: "(Backend) Typst templates"
-    include:
-      - ".*/backend/templates/.*"
+# All subdirectories within those directories will be used as components
```
praseodym, line 1: I do wonder if this will work because e.g. `backend/src/api` does not have subdirectories, but let's give it a try

oliver3: Hm, let's find out indeed!

PR #2804, opened 2026-01-29
 Sigrid: remove component_base_dirs for backend

oliver3:
Unfortunately `component_base_dirs` introduced in https://github.com/kiesraad/abacus/pull/2800 gave some issues:
- Apparently you cannot combine `component_base_dirs` and `components` which resulted in all other (mostly frontend) components ending up in "Remainder".
- File-based Rust modules under the component_base_dirs were not recognised as components.

This PR is to manually specify the backend modules instead.

Also, fixed one exclude path.

### DIFF
--- a/sigrid.yaml
+++ b/sigrid.yaml
@@ -1,14 +1,51 @@
-# All subdirectories within those directories will be used as components
-# See: https://docs.sigrid-says.com/reference/analysis-scope-configuration.html
-component_base_dirs:
-  - "backend"
-  - "backend/src"
-  - "backend/src/api"
-  - "backend/src/bin"
-  - "backend/src/infra"
-  - "backend/src/repository"
-
 components:
+  # Backend infra layer
+  - { include: [ ".*/backend/src/infra/airgap/.*" ], name: "backend/src/infra/airgap" }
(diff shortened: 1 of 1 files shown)


sigrid.yaml
```diff
@@ -106,7 +143,7 @@ exclude:
   - ".*/frontend/.*[.]config[.]ts"
   - ".*/frontend/package.json"
   - ".*/frontend/tsconfig[.]json"
-  - "docker-compose[.]yml"
+  - ".*/docker-compose[.]yml"
```
praseodym, line 146: This file is in the root so I don’t think this works?

Lionqueen94: it currently doesn't exclude the file, I was thinking it should be `.*docker-compose[.]yml` maybe?

praseodym: Hm since the rest is `.*/frontend/` maybe this would work, let's just try

Lionqueen94, line 146:
```suggestion
  - ".*docker-compose[.]yml"
```

praseodym: Let's try without, see https://github.com/kiesraad/abacus/pull/2804#discussion_r2742442129

PR #2807, opened 2026-01-30
Made sure apportionment works with non-consecutive list/candidate numbers

### LINKED ISSUE #2785: Make sure apportionment works with non-consecutive list numbers

Lionqueen94:
Resolves #2785

- Made sure apportionment works with non-consecutive list and candidate numbers
- Added a separate test for non-consecutive list numbers for seat assignment
- Added a separate test for non-consecutive list and candidate numbers for candidate nomination
- Moved `as_candidate_nomination_input` function to `seat_assignment/mod.rs`
- Extracted `get_number_of_candidates` function to prevent duplication

### DIFF
--- a/backend/apportionment/src/seat_assignment/mod.rs
+++ b/backend/apportionment/src/seat_assignment/mod.rs
@@ -716,13 +716,8 @@ pub(crate) mod tests {
 
             use super::get_total_seats_from_apportionment_result;
             use crate::{
-                ApportionmentError,
-                seat_assignment::{seat_assignment, tests::check_total_seats_per_list},
-                structs::ListNumber,
-                test_helpers::{
-                    seat_assignment_fixture_with_given_candidate_votes,
-                    seat_assignment_fixture_with_given_list_numbers_and_candidate_votes,
-                },
+                ApportionmentError, seat_assignment::seat_assignment, structs::ListNumber,
+                test_helpers::seat_assignment_fixture_with_given_candidate_votes,
             };
 
             /// Apportionment with no residual seats  
@@ -735,14 +730,14 @@ pub(crate) mod tests {
             /// 2 - largest remainder: seat assigned to list 5
             #[test]
             fn test_with_list_exhaustion_during_full_seats_assignment() {
-                let input = seat_assignment_fixture_with_given_list_numbers_and_candidate_votes(
+                let input = seat_assignment_fixture_with_given_candidate_votes(
                     15,
                     vec![
-                        (1, vec![500, 500, 500, 500]),
-                        (2, vec![400, 400, 400, 400]),
-                        (4, vec![400, 400, 400]),
-                        (5, vec![400, 400]),
-                        (7, vec![200, 200]),
+                        vec![500, 500, 500, 500],
+                        vec![400, 400, 400, 400],
+                        vec![400, 400, 400],
+                        vec![400, 400],
+                        vec![200, 200],
                     ],
                 );
                 let result = seat_assignment(&input).unwrap();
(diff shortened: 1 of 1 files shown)


backend/apportionment/src/test_helpers.rs
```diff
@@ -84,6 +84,18 @@ impl ListVotesMock {
     }
 }
 
+#[cfg(test)]
```
stacktraceghost: I think this can be removed, because test_helpers itself is already marked with `#[cfg(test)]`

Lionqueen94:
> I think we could add some tests that are meant to explicitly test non-consecutive numbers

I adapted some existing tests to have non-consecutive numbers instead, do you think it's better to add some more tests?

stacktraceghost:
> > I think we could add some tests that are meant to explicitly test non-consecutive numbers
> 
> I adapted some existing tests to have non-consecutive numbers instead, do you think it's better to add some more tests?

I'd prefer to have an explicit test for it, otherwise there is a chance we lose this coverage if those tests get adjusted in the future.

backend/apportionment/src/test_helpers.rs
Lionqueen94: Removed in d27356b3c2ca674f1a6b236baed69b028d81f000

praseodym (commented):
I'm not so sure about changing all tests to add list numbers in the output check - the list numbers get added to the input implicitly, why not keep the output format the same so that the existing tests don't need to be refactored?

We should add additional tests to check if non-sequential list numbers work, but we can do that without adding bloat to every existing test function.

backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -663,14 +698,18 @@ pub(crate) mod tests {
...
                 structs::ListNumber,
-                test_helpers::seat_assignment_fixture_with_given_candidate_votes,
+                test_helpers::{
+                    seat_assignment_fixture_with_given_candidate_votes,
+                    seat_assignment_fixture_with_given_list_numbers_and_candidate_votes,
+                },
             };
 
-            /// Apportionment with no residual seats  
+            /// Apportionment with no residual seats
```
Lionqueen94:
```suggestion
            /// Apportionment with no residual seats  
```

backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -1545,14 +1582,13 @@ pub(crate) mod tests {
...
             use crate::{
-                ApportionmentError,
-                seat_assignment::{get_total_seats_from_apportionment_result, seat_assignment},
-                structs::ListNumber,
+                ApportionmentError, seat_assignment::seat_assignment, structs::ListNumber,
                 test_helpers::seat_assignment_fixture_with_given_candidate_votes,
             };
 
-            /// Apportionment with no residual seats  
+            /// Apportionment with no residual seats
```
Lionqueen94:
```suggestion
            /// Apportionment with no residual seats  
```

backend/apportionment/src/structs.rs
```diff
@@ -66,6 +67,8 @@ pub(crate) fn as_candidate_nomination_input<'a, T: ApportionmentInput>(
         number_of_seats: input.number_of_seats(),
         list_votes: input.list_votes(),
         quota: seat_assignment.quota,
-        total_seats_per_list: get_total_seats_from_apportionment_result(seat_assignment),
+        total_seats_per_list: get_total_seats_per_list_number_from_apportionment_result(
+            seat_assignment,
+        ),
```
praseodym: Maybe this function should be moved to the seat assignment module?

Lionqueen94: That seems to have worked to improve module coupling, thanks!

praseodym (approved):
A Map could be a more appropriate data structure but the number of lists is always small so lgtm

PR #2809, opened 2026-02-02
Remove MSW build and harden pnpm config

praseodym:
Remove MSW build and harden pnpm config

### DIFF
--- a/frontend/pnpm-workspace.yaml
+++ b/frontend/pnpm-workspace.yaml
@@ -16,7 +16,6 @@ packageManagerStrict: true
 packageManagerStrictVersion: true
 strictDepBuilds: true
 strictPeerDependencies: true
-sideEffectsCache: true
 overrides:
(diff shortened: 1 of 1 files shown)


frontend/pnpm-workspace.yaml
```diff
@@ -12,7 +12,11 @@ trustPolicy: no-downgrade
 trustPolicyExclude:
   - semver@6.3.1
 engineStrict: true
-sideEffectsCache: false
+packageManagerStrict: true
+packageManagerStrictVersion: true
+strictDepBuilds: true
+strictPeerDependencies: true
+sideEffectsCache: true
```
oliver3: This is the default setting for `sideEffectsCache`, not sure if we want to add this/all of these

praseodym: Done in a00d6a9b78d36d826870734655ec97d4c51243c0

Issue #988, opened 2025-02-07
Improve modularity of backend modules


praseodym:
Improve modularity of backend modules, in particular identify and remove any cyclic dependencies.

The scope of the election module is currently not clearly defined. It contains API handlers for e.g. PDF and XML reports that should be moved elsewhere.

Next step: create spike as discussed in team meeting.

oliver3:
Tasks done or moved to separate issues

PR #2822, opened 2026-02-03
Committee session status change providers

stacktraceghost:
Part of issue [PR Improve modularity of backend modules #988](https://github.com/kiesraad/abacus/issues/988)
- Added/implemented providers on `SqliteConnection`
- Removed direct repo usage from `impl CommitteeSessionStatus`
- Refactored tests to be more compact for readability

Note: we could move `change_committee_session_status` to the API layer as a util or to a service layer. Saving that for a separate PR.


praseodym:
The providers module should also be added to `sigrid.yaml`

backend/src/domain/committee_session_status.rs
```diff
@@ -89,33 +105,31 @@ async fn delete_committee_session_files(
 }
 
 pub async fn change_committee_session_status(
-    conn: &mut SqliteConnection,
+    tx: &mut SqliteConnection,
     committee_session_id: CommitteeSessionId,
     status: CommitteeSessionStatus,
     audit_service: AuditService,
 ) -> Result<(), APIError> {
-    let mut tx = conn.begin().await?;
```
praseodym, line 97: We should still start a transaction, otherwise there is no guarantee that we get a consistent view of the database

Edit: in this case we might be able to do without if we ensure that the caller of this function always creates a transaction

backend/src/api/providers.rs
```diff
@@ -0,0 +1,49 @@
...
+    repository::{data_entry_repo, investigation_repo, polling_station_repo},
+};
+
+impl CommitteeSessionHasPollingStationsProvider for SqliteConnection {
+    async fn has_polling_stations(
+        &mut self,
+        committee_session_id: CommitteeSessionId,
+    ) -> Result<bool, CommitteeSessionError> {
+        let polling_stations = polling_station_repo::list(self, committee_session_id).await?;
+        Ok(!polling_stations.is_empty())
```
praseodym: We can use a count query instead of retrieving all polling stations

praseodym:
Requested review from @jschuurk-kr for test changes

backend/src/domain/committee_session_status.rs
stacktraceghost: We could change it to `tx: &mut Transaction<'_, Sqlite>`, which requires you to pass a transaction. But ` delete_data_entry_and_result_for_polling_station`  (`api/data_entry.rs`) for example, also uses this method, and should then also require a transaction (or create one). 

Maybe we can discuss in which scope we would like to do this change.

Alternatively, we can revert to creating the transaction ourselves. In both cases the Providers need to be implemented for `Transaction` instead of `SqliteConnection`, or maybe through a generic for anything that is a`SqliteConnection`

backend/src/domain/committee_session_status.rs
```diff
@@ -124,76 +138,65 @@ pub async fn change_committee_session_status(
...
         .await?;
 
-    tx.commit().await?;
-
     Ok(())
 }
 
 impl CommitteeSessionStatus {
-    pub async fn prepare_data_entry(
+    pub async fn prepare_data_entry<T>(
```
jschuurk-kr, line 162: Am I reading this correctly that `prepare_data_entry()` gets you to `Ok(Created)`, but that `ready_for_data_entry()` gets you to `Ok(InPreparation)`?

That seems confusing. Maybe this function then should be called `create_data_entry()`? Or `set_state_to_created()`?

stacktraceghost: Not necessarily in scope of this PR, but I agree that these names are confusing. We could change them (in a separate PR) for example to `to_created`, `to_in_preparation`? Makes it a lot more clear which `CommitteeSessionStatus` we are aiming for.

backend/src/domain/committee_session_status.rs
```diff
@@ -367,800 +402,173 @@ mod tests {
...
 
-        /// InPreparation --> Paused
-        #[test]
-        fn in_preparation_to_paused() {
-            assert_eq!(
-                CommitteeSessionStatus::InPreparation.pause_data_entry(),
-                Err(CommitteeSessionError::InvalidStatusTransition)
-            );
+        #[test(tokio::test)]
+        async fn prepare_data_entry() {
```
jschuurk-kr, line 432: This test only covers valid states. If I add `(InPreparation, Next, false, true, Ok(Created))`, the test still passes.

Arguably that's ok, since we (should) ensure that we never get into that invalid state.

stacktraceghost: Good observation. That would indeed indicate that on another level something would have gone wrong if this would be happening outside of tests (`has_investigations` but not `has_polling_stations`). 

Outside of the tests, that would point towards a faulty implementation of the trait, an issue in the repo or on the data level.

backend/src/domain/committee_session_status.rs
praseodym: As discussed, providers can work with `Transaction`s by dereferencing them into a `SqliteConnection` (e.g. `&mut *tx`).

backend/src/domain/committee_session_status.rs
```diff
@@ -367,800 +402,173 @@ mod tests {
...
+            use CommitteeSessionStatus::*;
+            use SessionType::*;
+            for (from, session_type, has_polling_stations, has_investigations, expected) in [
+                // Created stays Created
+                (Created, First, false, false, Ok(Created)),
+                (Created, First, true, false, Ok(Created)),
+                (Created, Next, true, true, Ok(Created)),
```
jschuurk-kr, line 439: I think we're missing a `(Created, Next, true, false, Ok(Created)),` here, i.e. the state when a next session has been created, so it's associated with polling stations, but doesn't have any investigations yet.

stacktraceghost:
> The providers module should also be added to `sigrid.yaml`

Done in https://github.com/kiesraad/abacus/pull/2822/commits/b62c417b3b91aa454296cba670401824c2136736

backend/src/domain/committee_session_status.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2822/commits/996215ea14c09359d131995199eb3bffe469980b

backend/src/domain/committee_session_status.rs
jschuurk-kr: added in 49b6e99112eaa068dbacd7f4c2ca8f243e006cf5

jschuurk-kr:
> Requested review from @jschuurk-kr for test changes

Tests reviewed and approved after 49b6e99112eaa068dbacd7f4c2ca8f243e006cf5.

backend/src/api/providers.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2822/commits/06a44dee36cde0d93725a9034ce12a8095b6e82f

Found the naming convention of all functions in `investigation_repo.rs` a bit extensive compared to `polling_station_repo.rs`, but this could be refactored in a different PR.

backend/src/domain/committee_session_status.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2822/commits/49b6e99112eaa068dbacd7f4c2ca8f243e006cf5

Added the case above and three more cases to `ready_for_data_entry` to also cover the next session.

backend/src/repository/polling_station_repo.rs
```diff
@@ -9,6 +9,26 @@ use crate::{
...
 
+/// Returns if a committee session has polling stations
+pub async fn has_any(
+    conn: &mut SqliteConnection,
+    committee_session_id: CommitteeSessionId,
+) -> Result<bool, sqlx::Error> {
+    let result = query!(
+        r#"
+        SELECT EXISTS(
+            SELECT 1 FROM polling_stations AS p
```
oliver3: I don't believe this JOIN is necessary, and I would prefer one or two basic tests for this function, although I realise that the other functions have none either.

stacktraceghost: Thanks, copied the list query a bit too quickly. I've adjusted this in https://github.com/kiesraad/abacus/pull/2822/commits/5aac58b4b2c75438aa3fffe1958eb23fc92ef20f

As discussed, the tests need a few more changes. For example, if we want to test if there are any polling stations, we need a polling station, committee session and election. And for these, we need to have `insert_test_*` methods available, which are not all there. Therefore, it would be good to do this in a new PR.

oliver3: Yes, agreed!

PR #2857, opened 2026-02-05
Move SQL queries out of API authentication middleware

oliver3:
Extract a `session_repo` from the authentication middleware, in six acts:
- Move the whole `session` module to the repository, because it contains mostly database queries
- Remove use of `axum` and `cookie` from the repo, by moving the extractor and the get_cookie() method back to the middleware
- Replace the api `AuthenticationError` with `sqlx::Error` in the repo (and split create and save, actually for the next step)
- Rename `Session::new()` to `Session::create()` and use that in combination with `session_repo::save()` instead of the `session_repo::create()`, to be able to split the responsibility of calculating session lifetime and saving to the database
- Move the `Session::create()` and its helpers to the middleware, remove lifetime calculation from `extend_session()`
- Remove use of `RequestSessionData` (with cookie and ip address) from the repo by replacing it with `SessionIdentifier` containing strings

This results in a `session_repo` that contains all the database queries for the session, but without any knowledge of axum, cookies, or lifetime calculations.


Lionqueen94:
The session repo needs to be added to sigrid

backend/src/api/middleware/authentication/session.rs
```diff
@@ -45,357 +26,84 @@ where
...
 
     /// Get a cookie containing this session key
     pub(crate) fn get_cookie(&self) -> Cookie<'static> {
-        CookieBuilder::new(SESSION_COOKIE_NAME, self.session_key.clone())
+        CookieBuilder::new(SESSION_COOKIE_NAME, self.session_key().to_owned())
             .max_age(cookie::time::Duration::seconds(
                 SESSION_LIFE_TIME.num_seconds(),
             ))
             .build()
+            .clone()
```
praseodym: This `.clone()`  looks unnecessary

backend/src/repository/session_repo.rs
```diff
@@ -0,0 +1,345 @@
...
+    session_key: String,
+    user_id: UserId,
+    user_agent: String,
+    ip_address: String,
+    expires_at: DateTime<Utc>,
+    created_at: DateTime<Utc>,
+}
+
+impl Session {
+    pub fn new(
```
praseodym: I think this can also be `pub(crate)`

oliver3:
> The session repo needs to be added to sigrid

Thanks 😅 added in https://github.com/kiesraad/abacus/pull/2857/commits/d3431351224fc9adceab3f899ba30ec34543fe8b

backend/src/api/middleware/authentication/util.rs
praseodym, line 1: It looks like this file is empty but still exists

backend/src/api/middleware/authentication/session.rs
oliver3: Indeed, forgot to revert it after struggling with the 'static lifetime. Removed in https://github.com/kiesraad/abacus/pull/2857/commits/06bf1bc7288cad8a44d95685a2a0ca0851dd7c3a

backend/src/repository/session_repo.rs
oliver3: I've updated all `pub` to `pub(crate)` in this PR, in https://github.com/kiesraad/abacus/pull/2857/commits/8b59322b2f97075377bd57f08badc8b9566c69ca .

backend/src/api/middleware/authentication/util.rs
oliver3: Yes thanks, removed in https://github.com/kiesraad/abacus/pull/2857/commits/b3f0bf276f1a88e46303c394a648d033416dffc0

PR #2864, opened 2026-02-05
Backend layer domain should not use other layers

oliver3:
Move non-domain usages out of `domain::committee_session_status` and `domain::file`, in separate commits:

- Separate `change_committee_session_status()` into new layer `service` because it is more orchestration than domain logic, including audit logging.
- Move `file::create_file()` to its only usage, and inline `file::delete_file()` into `delete_committee_session_files()`
- Move `CommitteeSessionError` out of `api` into `domain`, so that it can be used there and add a `ProviderError` variant

Questions:
- Are services a new layer or should it somehow live under the `api` layer?

### DIFF
--- a/backend/src/api/committee_session.rs
+++ b/backend/src/api/committee_session.rs
@@ -26,7 +26,7 @@ use crate::{
         election_repo,
         investigation_repo::list_investigations_for_committee_session,
     },
-    service::change_committee_session_status::change_committee_session_status,
+    service::change_committee_session_status,
 };
 
 pub fn router() -> OpenApiRouter<AppState> {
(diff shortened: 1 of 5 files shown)


backend/src/api/committee_session.rs
```diff
@@ -26,6 +26,7 @@ use crate::{
         election_repo,
         investigation_repo::list_investigations_for_committee_session,
     },
+    service::change_committee_session_status::change_committee_session_status,
```
Lionqueen94: Maybe name the file `committee_session_status.rs`/`committee_session.rs` to have a bit shorter filename and not have the duplication when importing the function?

oliver3: I've resolved the duplication by only re-exporting the function from the file in https://github.com/kiesraad/abacus/pull/2864/commits/9a32251473b36b80f28bc88fbb20954287843448, is this better?

Lionqueen94 (approved):

Issue #2883, opened 2026-02-10
Remove backend infra layer


oliver3:
The remaining code in the backend infra layer should find a new place, and the infra module should be removed, as described in the architecture sketch from https://github.com/kiesraad/abacus/issues/988

This can be done after the individual issues to remove parts out of the backend infra layer:
- https://github.com/kiesraad/abacus/issues/2882
- https://github.com/kiesraad/abacus/issues/2832
- https://github.com/kiesraad/abacus/issues/2554
- https://github.com/kiesraad/abacus/issues/2753

cikzh:
To prevent excessive merge conflicts, we should coordinate this with https://github.com/kiesraad/abacus/pull/2788

praseodym:
> To prevent excessive merge conflicts, we should coordinate this with [#2788](https://github.com/kiesraad/abacus/pull/2788)

These issues were meant for sometime in the future, so that should be after #2788 has been merged.

PR #2886, opened 2026-02-11
Move PDF generation to separate crate

### LINKED ISSUE #2832: Split out PDF generation to separate crate

praseodym:
Move PDF generation to a separate, generic PDF generation crate. The Abacus-specific Typst templates and model structs are kept in the main Abacus crate for now.

Resolves #2832

### DIFF
--- a/backend/src/bin/gen-pdf.rs
+++ b/backend/src/bin/gen-pdf.rs
@@ -1,8 +1,6 @@
-use abacus::{
-    domain::models::{PdfFileModel, PdfModel},
-    infra::pdf_gen::generate_pdf,
-};
+use abacus::domain::models::{PdfFileModel, PdfModel};
 use clap::Parser;
+use pdf_gen::generate_pdf;
 use tokio::{fs, process::Command};
 
 /// Generate PDFs for all available models using the input JSON files in
@@ -67,7 +65,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         let file_name = format!("{path}/{}.pdf", model.as_model_name());
         let file_model = PdfFileModel::new(file_name.clone(), model);
 
-        let pdf = match generate_pdf(file_model).await {
+        let pdf = match generate_pdf(&file_model).await {
             Ok(pdf) => pdf,
             Err(e) => {
                 eprintln!("Error generating PDF {file_name}: {e:?}");


backend/pdf_gen/src/external.rs
```diff
@@ -49,20 +47,11 @@ pub async fn generate_pdfs(
...
-    let json_data = file_model.model.get_input()?;
-    tokio::fs::write(json_path, json_data).await?;
+/// Uses environment variable `ABACUS_TYPST_BIN` (`typst` by default) to generate a PDF using an
+/// external binary of typst. Sources, fonts and input data are provided via [`PdfGenInput`].
+pub async fn generate_pdf(input: PdfGenInput) -> Result<PdfGenResult, PdfGenError> {
+    // write sources, fonts and JSON input to a temporary directory
+    let tmp_path = prep_tmp_dir(&input).await?;
 
     // construct the typst command to run
     let typst_binary = std::env::var("ABACUS_TYPST_BIN").unwrap_or("typst".into());
```
stacktraceghost, line 57: Since the code has been moved to a separate crate, maybe we can rename `ABACUS_TYPST_BIN` to something more generic.

backend/src/domain/polling_station.rs
```diff
@@ -135,6 +137,32 @@ impl From<String> for PollingStationType {
     }
 }
 
+#[cfg(test)]
+pub fn polling_stations_fixture(
```
stacktraceghost: Below there is also `polling_station_fixture` in `pub(crate) mod tests`. We could place these together and combine to remove some overlap.

praseodym:
> We could move the code from `infra/pdf_gen` to a service, maybe in a different PR?

Yes indeed, I didn't want to do that in this PR but the service layer would be the best place for that code. Edit: this is in scope for #2883.

I'll implement the other suggestions in this PR.

backend/src/domain/polling_station.rs
praseodym: Done in 40a809b9c45dbeee79f88f00dc7e7549a1624172

backend/pdf_gen/src/external.rs
praseodym: I think it's better to leave it as is, to avoid breaking anyone currently setting this environment variable. If we ever publish `pdf_gen` as a standalone crate we could reconsider.

Aside from that, I hope that separating Typst to a separate crate will improve compile/linking performance so that we don't need the external invocation option and can make `embed-typst` the default.

.github/dependabot.yml
```diff
@@ -10,7 +10,7 @@ updates:
       github-actions:
         patterns: ["*"]
   - package-ecosystem: "cargo"
-    directories: ["/backend", "backend/fuzz"]
+    directories: ["/backend", "backend/apportionment", "backend/fuzz", "backend/pdf_gen"]
```
rnijveld: Pretty sure we don't need to add the crates that are in the workspace

backend/pdf_gen/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+pub struct SourceFile {
+    pub path: &'static str,
+    pub content: &'static str,
+}
+
+/// Font data for Typst
+pub struct FontData(pub &'static [u8]);
+
+/// Configuration for a single PDF generation
+pub struct PdfGenInput {
```
rnijveld: Feels like this should have been a trait instead, that prevents having to do the conversions, also allows the paths to be just `&str`.

backend/src/infra/pdf_gen/sources.rs
```diff
@@ -0,0 +1,74 @@
...
+        include_source!("model-p-2a.typ"),
+        #[cfg(test)]
+        include_source!("test-teletex-charset.typ"),
+        #[cfg(test)]
+        include_source!("test-unsupported-chars.typ"),
+    ]
+}
+
+/// Load all fonts available from the `fonts/` directory
+pub(super) fn load_fonts() -> Vec<FontData> {
```
rnijveld: Maybe the macros in this file should also be part of the pdf gen crate? To reduce the leaky abstraction for `FontData` we could also put that in an opaque newtype wrapper that is defined in the pdf gen crate. That would allow none of the typst deps to leak outside the pdfgen crate. Same with SourceFile.

praseodym: `SourceFile` and `FontData` are already our own types defined in the new pdf_gen crate. The macros contain the relative paths to the templates and fonts, so we can't move those to the pdf_gen crate.

.github/dependabot.yml
praseodym: Reverted in 37b45de4d9274d06adfe0e79e6437061de5a36a8

backend/pdf_gen/src/lib.rs
praseodym: Thanks, changed in febcf0395b6ae149bda8e548fcafac5292038cfb

PR #2893, opened 2026-02-13
Add election creation for CSB role

praseodym:
Add election creation for CSB role, resolves #2824.

Please review per commit:
- Add role field to election
- Add CSB role to election validate and import
- Add enum type for election validation response
- Add helper fns for election validate and import

PR #2896, opened 2026-02-13
Deduplicate fields in election validation request

praseodym:
Like #2893 with an extra commit to deduplicate fields in election validation request by adding multiple named fields to the request enum. This does result in nested JSON on the client side, but I think that could allow for some deduplication in the frontend for code paths shared between GSB/CSB creation.

Draft to discuss if we like this approach. If we do, the `ElectionCreationRequest` can be deduplicated in the same fashion. Target is `main` so I can see what Sigrid thinks (that CI job only runs for PR targeting `main`).

### DIFF
--- a/backend/src/api/election.rs
+++ b/backend/src/api/election.rs
@@ -201,15 +201,20 @@ pub async fn election_number_of_voters_change(
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
-#[serde(tag = "role")]
+
 pub enum ElectionCreationValidateRequest {
-    GSB(GSBElectionCreationValidateRequest),
-    CSB(CSBElectionCreationValidateRequest),
+    GSB {
+        election_and_candidates: ElectionAndCandidateData,
+        gsb: GSBElectionCreationValidateRequest,
+    },
+    CSB {
+        election_and_candidates: ElectionAndCandidateData,
+    },
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
 #[serde(deny_unknown_fields)]
-pub struct GSBElectionCreationValidateRequest {
+pub struct ElectionAndCandidateData {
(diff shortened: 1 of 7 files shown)


backend/src/api/election.rs
```diff
@@ -201,15 +201,20 @@ pub async fn election_number_of_voters_change(
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
-#[serde(tag = "role")]
+
 pub enum ElectionCreationValidateRequest {
-    GSB(GSBElectionCreationValidateRequest),
-    CSB(CSBElectionCreationValidateRequest),
+    GSB {
+        election_and_candidates: ElectionAndCandidateData,
```
oliver3, line 207: In this case it is a bit strange to me that there is an election _and_ candidates struct. Another way is maybe something like the following, but then we introduce some duplication again..


```rust
pub enum ElectionCreationValidateRequest {
    GSB {
        #[schema(nullable = false)]
        election: Option<DataWithHash>,
        #[schema(nullable = false)]
        candidates: Option<DataWithHash>,
        gsb: GSBElectionCreationValidateRequest,
    },
    CSB {
        #[schema(nullable = false)]
        election: Option<DataWithHash>,
        #[schema(nullable = false)]
        candidates: Option<DataWithHash>,
    },
}

#[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
#[serde(deny_unknown_fields)]
pub struct DataWithHash {
    data: String,
    #[schema(value_type = Vec<String>)]
    hash: [String; crate::eml::hash::CHUNK_COUNT],
}

```

praseodym: Agreed that the `ElectionAndCandidateData` struct is not ideal, but your suggestion brings back a lot of duplication again.

As discussed yesterday I'm going to close this PR because it's not worth investing more time into this right now.

Issue #2881, opened 2026-02-10
Epic: introduce sub electoral committee entity


praseodym:
# Context

For data entry (#2811) we will need a new entity to represent an electoral committee (GSB/HSB/CSB/etc.) for building higher-level election results (e.g. at CSB level).

Example country-wide election:
```
CSB - Kiesraad
|_ HSB - Kieskring ...
|_ HSB - Kieskring 10 Haarlem
   |_ GSB Haarlem
   |_ GSB Aalsmeer
   |_ GSB ...
|_ HSB - Kieskring 11 Den Helder
   |_ GSB ...
|_ HSB - Kieskring 12 Den Haag
   |_ GSB - Gemeente Den Haag
```

Example municipal election:
```
CSB - Gemeente Juinen
|_ GSB - Gemeente Juinen
```

An election in Abacus can only do data entry for the level right below its own level, i.e. a country-wide CSB can only do HSB data entry and a municipal CSB can only do GSB data entry. Every level also needs to create reports and manual handover (file export, import, manual second data entry) so trying to integrate multiple levels in a single Abacus election would not align with the abstractions we have in our current architecture.

# Architecture

FigJam proposal for type structure: https://www.figma.com/board/qMn1Yi1UgS8VA8TKiYphqq/Flowchart-Abacus--GSB--HSB-en-CSB?node-id=313-559

Proposed Rust types:
```rust
// Committee session has a fn -> &[(Number, Name, DataEntryStatus)] for election status and polling station home page

enum DataEntry {
    type: ModelType
    // votes counts, voter counts, ...
}

struct PollingStation {
    name: String,
    postcode: String,
    // ..
}

struct PollingStationFirstSession {
    commitee_session_id: CommitteeSessionId,
    polling_station: PollingStation,
    data_entry: DataEntryId,
}

struct PollingStationNextSession {
    commitee_session_id: CommitteeSessionId,
    polling_station: PollingStation,
    // None for new polling station in next session
    prev_data_entry: Option<DataEntryId>,
    state: InvestigationStatus,
}

/// Polling station in next committee session is a state machine, that can transition through the investigation process and possibly lead to a new data entry
enum InvestigationStatus {
    None,
    InProgress { reason: String },
    ConcludedWithoutNewResults { reason: String, findings: String },
    ConcludedWithNewResults { reason: String, findings: String, new_data_entry: DataEntryId },
}

struct SubElectoralCommittee {
    name: String,
[...]

PR #2911, opened 2026-02-17
Merge results into data entries table

### LINKED ISSUE #2903: Merge `polling_station_results` into `polling_station_data_entries` table

praseodym:
Merge results into data entries table, so that we can remove the results table. This simplifies the data entry state machine and resolves #2903.

Several SQL queries in the data entry repository use `json_extract` to get the status and/or definitive results from the JSON data entry state. This is not optimal, but I suspect we'll have to rewrite these statements later in epic #2881 anyway.


backend/fuzz/fuzz_targets/data_entry_status.rs
```diff
@@ -411,9 +411,9 @@ fuzz_target!(|transitions: Vec<Transition>| {
                 &polling_station(),
                 &election(),
             ),
-            Transition::FinaliseSecondEntry(correct_user) => state
-                .finalise_second_entry(&polling_station(), &election(), users.second(correct_user))
-                .map(|r| r.0),
+            Transition::FinaliseSecondEntry(correct_user) => {
```
stacktraceghost, line 414: We can remove the `{ }`

backend/src/repository/data_entry_repo.rs
```diff
@@ -336,27 +231,38 @@ async fn fetch_results_for_committee_session(
...
+    // extracted via json_extract(de.state, '$.state.results').
+    //
     // This function returns the original polling station id and the found results. When no results are found,
     // an error is thrown.
     let results = query!(
         r#"
         WITH RECURSIVE polling_stations_chain(original_id, id, id_prev_session, data, investigation) AS (
-            SELECT ps.id AS original_id, ps.id, ps.id_prev_session, r.data, psi.polling_station_id
+            SELECT ps.id AS original_id, ps.id, ps.id_prev_session,
+                   CASE WHEN json_extract(de.state, '$.status') = 'Definitive'
```
stacktraceghost, line 247: We could pass this here/below as parameter $3 and use `DataEntryStatusName::Definitive`

praseodym: `DataEntryStatusName::Definitive` is not a string and `DataEntryStatusName::Definitive.to_string() = "definitive"`, so it's not that easy

backend/fuzz/fuzz_targets/data_entry_status.rs
praseodym: Then it will end up like this after rustfmt, which is I don't think is better:
```rust
            Transition::FinaliseSecondEntry(correct_user) => state.finalise_second_entry(
                &polling_station(),
                &election(),
                users.second(correct_user),
            ),
```

backend/src/repository/data_entry_repo.rs
stacktraceghost: Ah okay, I missed the capitalization there. It would be nice if there would be a direct connection between the value we are checking for and the code. But if it is not easily manageable, we can still rely on our tests.

backend/fuzz/fuzz_targets/data_entry_status.rs
stacktraceghost: Matter of taste and not important enough, so I will resolve this 👍

stacktraceghost (approved):

oliver3 (approved):
Also tested a full committee session and checked if models were generated, all ok!

PR #2916, opened 2026-02-18
Add data entry id as polling station foreign key

### LINKED ISSUE #2914: Add data entry id as polling station foreign key
Instead of the data entry pointing at the polling station, add a data entry id and point the polling station at the data entry.

praseodym:
Instead of the data entry pointing at the polling station, add a data entry id and point the polling station at the data entry. Resolves #2914.

The data entry API still uses the polling station id and the empty data entry is still created by the data entry module. Changing this will be part of a future PR in #2881.

### DIFF
--- a/backend/fuzz/Cargo.lock
+++ b/backend/fuzz/Cargo.lock
@@ -8,7 +8,6 @@ version = "1.1.0"
 dependencies = [
  "apportionment",
  "argon2",
- "async_zip",
  "axum",
  "axum-extra",
  "chrono",
@@ -16,6 +15,7 @@ dependencies = [
  "cookie",
  "hyper",
  "password-hash",
+ "pdf_gen",
  "quick-xml",
  "rand 0.9.2",
  "serde",
@@ -25,7 +25,6 @@ dependencies = [
  "sqlx",
  "strum",
  "tokio",
- "tokio-util",
  "tower",
  "tower-http",
  "tracing",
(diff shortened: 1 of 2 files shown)


backend/src/repository/polling_station_repo.rs
```diff
@@ -41,6 +42,7 @@ pub async fn list(
             c.election_id AS "election_id: ElectionId",
             p.committee_session_id AS "committee_session_id: CommitteeSessionId",
             p.id_prev_session AS "id_prev_session: _",
+            p.data_entry_id AS "data_entry_id: DataEntryId",
```
oliver3, line 45: We sometimes use [inferred types](https://docs.rs/sqlx/latest/sqlx/macro.query_as.html#column-type-override-infer-from-struct-field) and sometimes specify the types ourself. I would prefer inferred types (`AS "data_entry_id: _"`) to avoid specifying the type twice.

stacktraceghost (approved):

backend/src/repository/data_entry_repo.rs
```diff
@@ -81,52 +76,114 @@ pub async fn get_or_default(
...
 pub async fn delete_data_entry(
     conn: &mut SqliteConnection,
     polling_station_id: PollingStationId,
 ) -> Result<Option<PollingStationDataEntry>, sqlx::Error> {
-    query_as!(
+    let mut tx = conn.begin().await?;
+
+    // Get data_entry_id from polling station
+    let row = query!(
+        r#"SELECT data_entry_id AS "data_entry_id: DataEntryId" FROM polling_stations WHERE id = ?"#,
```
oliver3, line 150: I don't think that you need these casts for id's that you then only use in the following queries, or that they add any safety.

backend/src/repository/polling_station_repo.rs
praseodym: I'll change that in the next PR

backend/src/repository/data_entry_repo.rs
praseodym: True, I'll change it in the next PR

oliver3 (approved):

PR #2919, opened 2026-02-18
Data entry renames

praseodym:
Please review per commit:
- Rename `status` module to `data_entry_status`
- Rename `polling_station_data_entries` to `data_entries`
- Use SQLx inferred types as discussed in https://github.com/kiesraad/abacus/pull/2916#discussion_r2822176824 and https://github.com/kiesraad/abacus/pull/2916#discussion_r2822272318

Relates to #2881

### DIFF
--- a/backend/fuzz/fuzz_targets/data_entry_status.rs
+++ b/backend/fuzz/fuzz_targets/data_entry_status.rs
@@ -13,7 +13,7 @@ use abacus::{
             VoteCountingMethod,
         },
         polling_station::{PollingStation, PollingStationId, PollingStationType},
-        status::{CurrentDataEntry, DataEntryStatus, DataEntryTransitionError},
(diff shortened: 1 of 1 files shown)


stacktraceghost (approved):

backend/src/api/data_entry.rs
```diff
@@ -583,7 +583,7 @@ impl ResolveErrorsAction {
     ),
     security(("cookie_auth" = ["coordinator"])),
 )]
-async fn polling_station_data_entries_and_result_delete(
+async fn data_entries_and_result_delete(
```
Lionqueen94, line 586: Officially this is not deleting a separate result anymore. It's not just the data entries delete for a coordinator. Should we change this function name to make this more clear?

praseodym: I'll rename it in the next PR!

PR #2922, opened 2026-02-19
Add HttpOnly and SameSite attributes to session cookie

underdarknl:
Harden against XSS, CSRF issues. Ip binding (local lan, arp spoofing) and replaying is still wide open.


praseodym:
We already set these attributes in the `set_default_cookie_properties` function:

https://github.com/kiesraad/abacus/blob/8235a51ca19165e5d879b46ea8c8b7043c6d9311/backend/src/api/authentication.rs#L93-L99

Did you find any place where these attributes were not set correctly?

praseodym:
Closing, feel free to reopen if this is still an issue

Issue #2917, opened 2026-02-18
Flaky test: e2e full flow


stacktraceghost:
Step `create browser-specific coordinator for flow ` / `create browser-specific admin user account` within `full-flow.e2e.ts` is flaky:
- [Example 1 (2026-02-18)](https://github.com/kiesraad/abacus/actions/runs/22096027850/job/63853426498#step:9:137)
- [Example 2 (2026-02-17)](https://github.com/kiesraad/abacus/actions/runs/22136880546/job/63990972893#step:9:138)
- [Example 3 (2026-02-18)](https://github.com/kiesraad/abacus/actions/runs/22143325216/job/64013530986?pr=2919)

[Related changes (merged 2025-02-13)](https://github.com/kiesraad/abacus/pull/2799)

jschuurk-kr:
Looks like:
- admin account is created
- logging in as admin fails, so test that creates coordinator
- admin account creation is retried, but fails because account already exists

So next steps:
- randomize user names for each test
- see of the logging in as admin succeeds in branch 2432-review-full-flow-tests, where the admin logs out after first log in
  - alternatively: complete first admin account through the browser instead of API in the hope that helps

jschuurk-kr:
Looked through a bunch of pipeline runs. Did not see any flakiness of the full flow test. Closed.

PR #2925, opened 2026-02-19
review full flow tests and address flaky test

### LINKED ISSUE #2432: Review (and possibly expand) the full flow test from a process perspective
For example, the full flow test contains the test `create investigation for ${station.name}`. That test does not check if the 'corrigendum' of the polling station can be downloaded.

While we have other tests covering that functionality, I think there's value in being able to use the full flow test in in walkthroughs with stakeholders. This means the full flow test should capture the election process sufficiently.

So for the example above, this means at least adding a check that the download link for the 'corrigendum' is shown on the page. And maybe also downloading it.

jschuurk-kr:
### Scope

- closes #2432
- relates to #2917
  - no flakiness in the pipeline runs on this PR: https://github.com/kiesraad/abacus/actions/workflows/build-lint-test.yml?query=branch%3A2432-review-full-flow-tests

I recommend reviewing per commit.

### DIFF
--- a/frontend/e2e-tests/tests/full-flow.e2e.ts
+++ b/frontend/e2e-tests/tests/full-flow.e2e.ts
@@ -326,7 +326,7 @@ test.describe("full flow", () => {
     test(`second data entry ${station.name}`, async ({ page }) => {
       await page.goto("/account/login");
 
-      const secondTypist = typistUsers.filter((typist) => typist.username.startsWith("typist4"))[0]!;
+      const secondTypist = typistUsers[1]!;
       const loginPage = new LoginPgObj(page);
       await loginPage.login(secondTypist.username, getTestPassword(secondTypist.username));
 


frontend/e2e-tests/page-objects/investigations/InvestigationOverviewPgObj.ts
```diff
@@ -2,16 +2,18 @@ import type { Locator, Page } from "@playwright/test";
 
 export class InvestigationOverviewPgObj {
   readonly header: Locator;
+  readonly alert: Locator;
   readonly addInvestigationButton: Locator;
   readonly fillInvestigationLink: Locator;
 
   constructor(protected readonly page: Page) {
-    this.header = page.getByRole("heading", { level: 1 });
+    this.header = page.getByRole("heading", { level: 1 }); // Onderzoeken in tweede/derde/... zitting
```
Lionqueen94:
```suggestion
    this.header = page.getByRole("heading", { level: 1 });
```

jschuurk-kr: addressed in 49560cd8 by adding a `name` with a regex to the locator

Lionqueen94 (approved):

frontend/e2e-tests/tests/full-flow.e2e.ts
```diff
@@ -346,14 +318,17 @@ test.describe("full flow", () => {
...
+
+      await logout(page);
     });
 
-    test(`second data entry ${station.name}`, async ({ page, browserName }) => {
+    test(`second data entry ${station.name}`, async ({ page }) => {
       await page.goto("/account/login");
 
-      const username = `typist4-${browserName}`;
+      const secondTypist = typistUsers.filter((typist) => typist.username.startsWith("typist4"))[0]!;
```
praseodym: We already have `firstTypist = typistUsers[0]!`, so why not `typistUsers[1]!` here?

jschuurk-kr: oops, thanks! fixed in 5acbf534

PR #2926, opened 2026-02-20
Fixed #2915 and added default `type="button"` to `Button` and `IconButton`

### LINKED ISSUE #2915: First log in by first admin fails when using Enter-key
### Steps to reproduce
1. Start Abacus without seeding data.
2. Go to /account/initialise.
3. Create admin account.
4. Log in but do not click "Inloggen", instead hit Enter on your keyboard.
5. Find yourself on the page "Account voor beheerder aanmaken" instead of "Verkiezingen beheren".

Luckily this is not reproducible on our `release-1.0-GR26` branch, only on `main`.

### Analysis
(h/t to @oliver3)

`FirstLoginForm.tsx` has:
```TypeScript
  const link = (content: ReactNode) => (
    <Button
      variant="underlined"
      size="md"
      onClick={(e) => {
        e.preventDefault();
        prev();
      }}
    >
      {content}
    </Button>
  );
```

The `Button` does not have a `type="button"`, so it gets triggered by the submit. This sets the `step` state back to `create-admin-account`.

### Proposed solution
Add `type="button"` to our `Button`-component, so it becomes the default behavior of our Button. This does require a check on all usages. Some tests for this would also be nice.

Lionqueen94:
Fixes #2915

Also checked all uses of Button component and added `type="submit"` to CheckAndSave submit button.

### DIFF
--- a/frontend/src/features/election_overview/components/OverviewPage.tsx
+++ b/frontend/src/features/election_overview/components/OverviewPage.tsx
@@ -32,11 +32,7 @@ function AddFirstElection() {
         <div>{tx("election.add_first_election")}</div>
       </FormLayout.Section>
       <FormLayout.Controls>
-        {isAdministrator && (
-          <Button.Link type="submit" to={"./create"}>
-            {t("election.create")}
-          </Button.Link>
-        )}
+        {isAdministrator && <Button.Link to={"./create"}>{t("election.create")}</Button.Link>}
       </FormLayout.Controls>
     </FormLayout>
   );


jschuurk-kr:
There's a biome rule [useButtonType](https://biomejs.dev/linter/rules/use-button-type/) we could enable for this. Not sure if that's worth it, we have four occurrences of `<button>`, since we mostly user our own `Button` and `IconButton`.

jschuurk-kr:
> There's a biome rule [useButtonType](https://biomejs.dev/linter/rules/use-button-type/) we could enable for this. Not sure if that's worth it, we have four occurrences of `<button>`, since we mostly user our own `Button` and `IconButton`.

"This rule is recommended, meaning it is enabled by default." So it is already enabled?

Lionqueen94:
> > There's a biome rule [useButtonType](https://biomejs.dev/linter/rules/use-button-type/) we could enable for this. Not sure if that's worth it, we have four occurrences of `<button>`, since we mostly user our own `Button` and `IconButton`.
> 
> "This rule is recommended, meaning it is enabled by default." So it is already enabled?

It seems that way, adding it specifically on main did not result in any errors...

jschuurk-kr:
> > > There's a biome rule [useButtonType](https://biomejs.dev/linter/rules/use-button-type/) we could enable for this. Not sure if that's worth it, we have four occurrences of `<button>`, since we mostly user our own `Button` and `IconButton`.
> > 
> > 
> > "This rule is recommended, meaning it is enabled by default." So it is already enabled?
> 
> It seems that way, adding it specifically on main did not result in any errors...

The rule does trigger when I add `<button>bonus button</button>` to a component. It also triggers when I put `type="incorrectbuttontype"` in our `Button` component. It does not trigger if I remove the `type` completely, like we have on main. :man_shrugging:

frontend/src/features/election_create/components/UploadPollingStationDefinition.tsx
```diff
@@ -114,7 +114,7 @@ export function UploadPollingStationDefinition() {
               <PollingStationsPreview pollingStations={state.pollingStations} />
             </FormLayout.Section>
             <FormLayout.Controls>
-              <Button type="button" onClick={() => void next()}>
+              <Button type="submit" onClick={() => void next()}>
```
jschuurk-kr: When I click this button, I get redirected to the first page of election creation, i.e. "Importeer verkiezingsdefinitie". Probably same reason why these tests are failing: https://github.com/kiesraad/abacus/actions/runs/22226687268/job/64295323027?pr=2926

jschuurk-kr: resolved in 485aa559de1af286873a89cdac39b36002429cf8

Lionqueen94: Yes, I'm working on fixing it :+1:

frontend/src/features/election_overview/components/OverviewPage.tsx
```diff
@@ -32,7 +32,11 @@ function AddFirstElection() {
         <div>{tx("election.add_first_election")}</div>
       </FormLayout.Section>
       <FormLayout.Controls>
-        {isAdministrator && <Button.Link to={"./create"}>{t("election.create")}</Button.Link>}
+        {isAdministrator && (
+          <Button.Link type="submit" to={"./create"}>
```
praseodym: `Button.Link` is rendered as an `<a>` tag, does it make sense to add `type="submit"` to that?

Lionqueen94: Good point, removed in bbc5bd32fc6e62380a76181444a258d0310a8998

praseodym (approved):
Code lgtm, did not test behaviour in all forms

oliver3 (approved):
Also checked that forms with an `onSubmit` have a button with `type="submit"`

PR #2927, opened 2026-02-20
Move data entry initialisation to polling station/investigation modules

### LINKED ISSUE #2924: Create empty data entry in polling station module
Create an empty data entry entity in the polling station and investigation modules, whenever a polling station or investigation is created for which a data entry is expected.

praseodym:
- Create empty data entry in polling station, and in investigations module when the investigation has new results.
- Rename `data_entries_and_result_delete` to `data_entry_reset`
- Replace previous polling station id (`id_prev_session`) by data entry id (`prev_data_entry_id`)
  In a next session, set the previous data entry id instead of the previous polling station id.

Might be easiest to review per commit.

Resolves #2924


backend/src/api/data_entry.rs
```diff
@@ -297,7 +297,7 @@ async fn data_entry_claim(
     let validation_results = new_state.start_validate(&polling_station, &election)?;
 
     // Save the new data entry state
-    data_entry_repo::upsert(&mut tx, polling_station_id, &new_state).await?;
+    data_entry_repo::update(&mut tx, polling_station_id, &new_state).await?;
 
     let data_entry = get_data_entry(&mut tx, polling_station_id).await?;
```
oliver3: `update()` already returns the updated data_entry

```suggestion
    let data_entry = data_entry_repo::update(&mut tx, polling_station_id, &new_state).await?;
```

backend/src/api/data_entry.rs
```diff
@@ -411,7 +411,7 @@ async fn data_entry_save(
...
```
oliver3: No need to separately get the data_entry again

backend/src/api/data_entry.rs
```diff
@@ -465,15 +465,9 @@ async fn data_entry_delete(
...
+    data_entry_repo::update(&mut tx, polling_station_id, &new_state).await?;
 
-    if new_state == DataEntryStatus::Empty {
-        // The database entry of the data entry is fully deleted when the first entry is deleted
-        delete_data_entry(&mut tx, polling_station_id).await?;
-    } else {
-        // The status of the data entry is updated when the second entry is deleted
-        data_entry = data_entry_repo::upsert(&mut tx, polling_station_id, &new_state).await?;
-    }
+    let data_entry = get_data_entry(&mut tx, polling_station_id).await?;
```
oliver3: No need to get_data_entry(): update returns the updated data_entry.

That being said, originally it logged the data_entry that was to be deleted in case it was the first entry, now it returns the data entry with new empty state. Is that a problem?

stacktraceghost (dismissed):

backend/src/api/data_entry.rs
```diff
@@ -583,7 +575,7 @@ impl ResolveErrorsAction {
     ),
     security(("cookie_auth" = ["coordinator"])),
 )]
-async fn data_entries_and_result_delete(
+async fn data_entry_reset(
```
oliver3, line 565: I would keep the name delete actually, to match the "Invoer verwijderen". (Also, there are still references to delete, e.g. `ErrorReference::DataEntryCannotBeDeleted`, `AuditEvent::DataEntryDeleted`) 

If this was to disambiguate it from the `data_entry_delete()`, I would prefer to rename that one to `data_entry_abort()` to indicate that it belongs to a typist that wants to "Invoer afbreken"

backend/src/api/data_entry.rs
```diff
@@ -1344,10 +1320,14 @@ mod tests {
     async fn test_claim_data_entry_next_session_ok(pool: SqlitePool) {
         let mut conn = pool.acquire().await.unwrap();
         let polling_station_id = PollingStationId::from(742);
-        // Insert investigation
+        // Insert investigation with corrected_results and create empty data entry
+        // (simulates what the investigation conclude flow does)
         insert_test_investigation(&mut conn, polling_station_id, Some(true))
             .await
             .unwrap();
+        data_entry_repo::create_empty(&mut conn, polling_station_id)
```
oliver3: Doubting if this should not be added to the `insert_test_investigation()` when corrected_results = Some(true), to prevent missing create_empty() calls (e.g. in `test_next_new_polling_station_investigation_with_corrected_results_incomplete()` if I understand it correctly?)

backend/src/api/polling_station.rs
```diff
@@ -127,7 +127,12 @@ async fn polling_station_create(
 
     validate_user_is_allowed_to_perform_action(user, &committee_session)?;
 
-    let polling_station = create(&mut tx, election_id, new_polling_station).await?;
+    let mut polling_station = create(&mut tx, election_id, new_polling_station).await?;
+
+    if !committee_session.is_next_session() {
+        let data_entry = data_entry_repo::create_empty(&mut tx, polling_station.id).await?;
+        polling_station.data_entry_id = Some(data_entry.id);
```
oliver3, line 134: Where is this updated `data_entry_id` saved to the database?

backend/src/api/polling_station.rs
```diff
@@ -373,7 +378,14 @@ pub async fn create_imported_polling_stations(
     let file_hash = EmlHash::from(polling_stations_request.polling_stations.as_bytes()).chunks;
 
     // Create new polling stations
-    let polling_stations = create_many(&mut tx, election_id, polling_stations).await?;
+    let mut polling_stations = create_many(&mut tx, election_id, polling_stations).await?;
+
+    if !committee_session.is_next_session() {
+        for polling_station in &mut polling_stations {
+            let data_entry = data_entry_repo::create_empty(&mut tx, polling_station.id).await?;
+            polling_station.data_entry_id = Some(data_entry.id);
```
oliver3, line 386: Edit: same comment as for `polling_station_create()`

backend/src/api/polling_station.rs
oliver3: Oh, I see now that create_empty() also update its polling_station, which was a surprise to me. 

Can we make this more clear in the name and letting it return `(data_entry, updated_polling_station)` to make it explicit and useful?

backend/src/test_data_gen/generators.rs
```diff
@@ -264,6 +264,9 @@ async fn generate_polling_stations(
         )
         .await
         .expect("Failed to create polling station");
+        data_entry_repo::create_empty(conn, ps.id)
+            .await
+            .expect("Failed to create empty data entry");
         polling_stations.push(ps);
```
oliver3, line 271: Not sure if it is needed, but this `ps` does not have a correct `data_entry_id` which may cause confusion later

backend/src/api/data_entry.rs
praseodym: Fixed in 66744e8effaec01c34a0a6078d8b72c33c64ed60

backend/src/api/data_entry.rs
praseodym: Fixed in 66744e8effaec01c34a0a6078d8b72c33c64ed60.

backend/src/api/data_entry.rs
praseodym: Fixed in 66744e8effaec01c34a0a6078d8b72c33c64ed60.

I don't think the audit log change is a problem, the previous value would have been logged in the preceding `DataEntrySaved` audit log entry.

backend/src/api/data_entry.rs
praseodym: We use 'abort' for the modal, the actions in the modal are 'discard' and 'keep', so I don't think 'abort' is the best name either.

I'm happy to open an issue to get consensus on the naming.

backend/src/api/polling_station.rs
praseodym: That is documented in the `create_empty` rustdoc, and I prefer to keep function names concise. 

Because this is a data entry repository function I don't want to let it query and return a polling station struct. We could create a service function instead, that uses the data entry repository to create the data entry and the polling station repository to link the data entry to the polling station and return the updated struct.

backend/src/api/data_entry.rs
praseodym: Created #2936

backend/src/test_data_gen/generators.rs
praseodym: Changed in 5a3cecef58810c9c5507d0360e4a20d578fa1e96

backend/src/api/data_entry.rs
praseodym: Changed in 48fb15fcd4c72bf984144340cbca6f85c02a075e

backend/src/api/polling_station.rs
oliver3: It now updates a polling station in the data entry repo, which in my opinion is "more wrong" than querying (which would actually be a RETURNING in that update).

I think a service function would be better indeed, maybe even consuming a polling_station and returning a new one to signal that the old polling_station is not valid anymore. Something for a new PR?

praseodym: New PR sounds good, created #2939 to track this.

Issue #2834, opened 2026-02-03
Add committee category ("soort stembureau") to create user flow


oliver3:
[Design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=18005-9385&t=oYcZDDYH2fRrYJAr-11)

Issue #2833, opened 2026-02-03
Extend user role enum with election level for Typist and Coordinator


oliver3:
- With database serialization
-

PR #2930, opened 2026-02-23
Extend user roles with election role

oliver3:
Resolves [Extend user role enum with election level for Typist and Coordinator #2833](https://github.com/kiesraad/abacus/issues/2833)

```rust
#[serde(rename_all = "snake_case")]  
#[strum(serialize_all = "snake_case")]  
#[sqlx(rename_all = "snake_case")]
pub enum Role {  
    Administrator,  
    #[serde(rename = "coordinator_gsb")]  
    CoordinatorGSB,  
    #[serde(rename = "coordinator_csb")]  
    CoordinatorCSB,  
    #[serde(rename = "typist_gsb")]  
    TypistGSB,  
    #[serde(rename = "typist_csb")]  
    TypistCSB,  
}
```

I've tried to keep the changes to a minimum and made the following choices. Please let me know if these are acceptable.
- Authorization checks (backend and frontend) have been changed from `Coordinator` to `CoordinatorGSB` and from `Typist` to `TypistGSB`. This means that if (new) functionality has to be enabled for the CSB roles, that has to be added. (also keeping "Least Privilege principle" in mind)
- Components that do not care if the role is for CSB or GSB, use the newly introduced `RoleWithoutElection` which represents the role without the election part (i.e. the original `Role` variants), which is added to the `useUser()` hook return value.
- User list, create, update now only handle the GSB roles, this will change in https://github.com/kiesraad/abacus/issues/2834
- The translation for both `coordinator` and `coordinator_gsb`/`coordinator_csb` is the same "Coördinator" to avoid changes in the screens and especially the tests (same goes for typist). This will change in https://github.com/kiesraad/abacus/issues/2835

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
   backend:
-    name: Backend
+    name: Backend coverage
     runs-on: ubuntu-latest
     defaults:
       run:
@@ -59,8 +59,8 @@ jobs:
       - name: Setup Rust Nightly
         shell: bash
         run: >
-          rustup update nightly &&
-          rustup default nightly
+          rustup update nightly-2026-02-21 &&
+          rustup default nightly-2026-02-21
       - name: Cargo cache
         uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2
         with:
@@ -95,7 +95,7 @@ jobs:
         run: df -h
 
   frontend:
-    name: Frontend
+    name: Frontend coverage
     runs-on: ubuntu-latest
     defaults:
       run:
--- a/frontend/src/components/ui/Button/Button.tsx
+++ b/frontend/src/components/ui/Button/Button.tsx
@@ -9,6 +9,7 @@ export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
   isLoading?: boolean;
   variant?: ButtonVariant;
   size?: Size;
+  type?: "submit" | "reset" | "button";
   leftIcon?: ReactNode;
   rightIcon?: ReactNode;
   children: ReactNode;
@@ -19,6 +20,7 @@ export function Button({
   isLoading,
   variant = "primary",
   size = "lg",
+  type = "button",
   leftIcon,
   rightIcon,
   children,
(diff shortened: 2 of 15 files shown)


frontend/src/app/RootLayout.test.tsx
```diff
@@ -36,15 +36,15 @@ describe("Route authorisation is handled", () => {
     vi.spyOn(console, "error").mockImplementation(() => {});
 
     if (ownRole) {
-      let user: LoginResponse | undefined;
+      let user: UseUserReturn | null = null;
```
Lionqueen94: Why did you add the `null = null` here?

oliver3: - The `| null` is because `useUser()` returns `null` if there is no logged in user and not `undefined`, but that should be part of the `UseUserReturn` type itself, so I will change that.
- The `= null` is to initialize it with null but it seems that TypeScript is smart enough to detect that all cases are handled in the switch below, so it can never end up being null. So I will remove that.

oliver3: Updated in https://github.com/kiesraad/abacus/pull/2930/commits/b947abb14a040223090e814bc9e218ee80e30cae where I also added `NonNullable<>` to the user-mock-data functions to avoid needless null checks in test code that use these mocks, e.g. `getTypistUser().user_id`

frontend/src/hooks/user/useUserRole.ts
```diff
@@ -1,12 +1,17 @@
+import { isAdministrator, isCoordinator, isTypist } from "@/utils/role";
 import { useUser } from "./useUser";
 
 export function useUserRole() {
   const user = useUser();
 
+  if (user === null) {
```
stacktraceghost: Taste preference, but by using `user?.role` in the return like before, you don't need these lines.

stacktraceghost (dismissed):

frontend/src/hooks/user/useUserRole.ts
oliver3: Ah yes, this was before I made the arguments for the isX() optional, I'll go along with your preference! Updated in https://github.com/kiesraad/abacus/pull/2930/commits/bd4ce009a332dbede0036d01e5079ae5e12f4405

PR #2931, opened 2026-02-23
Use Rust Nightly 2026-02-21 for code coverage

praseodym:
- Pin Rust nightly to 2026-02-21 version, workaround for https://github.com/rust-lang/rust/issues/152961
- Add 'coverage' to Codecov job names to differentiate them from the jobs in the 'Build, lint & test' workflow

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
   backend:
-    name: Backend
+    name: Backend coverage
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


oliver3:
I think we need one before that:

> [This compiles on nightly 2026-02-21, but fails on nightly 2026-02-22.](https://github.com/rust-lang/rust/issues/152961)

praseodym:
> I think we need one before that:
> 
> > [This compiles on nightly 2026-02-21, but fails on nightly 2026-02-22.](https://github.com/rust-lang/rust/issues/152961)

Yeah I figured, changed

PR #2934, opened 2026-02-23
Set report type for Codecov test result uploads

praseodym:
Codecov uploads broken since 7f929bb08a0f9c2606e17e7616202f24f797645c:

Screenshot 2026-02-23 at 15 01 36

Apparently we should have added the `report-type: test_results` option, see https://github.com/codecov/test-results-action.

**Draft until fix is verified.**


praseodym:
This is still broken and I'm not sure why. Other people are also reporting problems (https://github.com/codecov/feedback/issues/843).

Fortunately our code coverage results are being processed, but tests results are not and the view in Codecov is empty. We could consider skipping uploading test results and only upload code coverage reports.

praseodym:
Closing in favour of #2961

Issue #2753, opened 2026-01-19
Reduce audit log service cross-module dependencies


praseodym:
Reduce audit log service cross-module dependencies, e.g. by using JSON. Filter in frontend can be done with database query of existing event types.

First step is to discuss approach with the team, then create a spike.

praseodym:
I think we can keep the `AuditEvent` enum, but move all the details data out of that enum to a JSON value. We can then move the audit event details struct to the module where the audit event is created and serialize it to JSON there. This way we still have a 'registry' of all audit events so no database query is needed to discover them.

oliver3:
Circulair dependencies from/to audit_log have been resolved!

PR #2943, opened 2026-02-25
Separate API audit log event levels

oliver3:
Resolves "Extract AuditEventLevel level from (API) Error" in https://github.com/kiesraad/abacus/issues/2753

Is it okay to have two separate audit log event types for API errors & warnings, instead of one?

---

Before audit log refactor:

image

After audit log refactor:

image

After this PR:

image


praseodym:
> Is it okay to have two separate audit log event types for API errors & warnings, instead of one?

I'd say it's better than making everything an error

PR #2948, opened 2026-02-25
Added fully qualified path, removed imports

stacktraceghost:
Locally ran into an issue where I called macro `as_audit_event` on a struct in `structs.rs` and used that struct in another file. In `structs.rs` I forgot `AuditEvent` and `AsAuditEvent`, resulting into errors on the macro. 

If you forget:
- `AsAuditEvent`, you will get an error that the trait is not implemented for your struct and an error within the macro: cannot find trait `AsAuditEvent`
- `AuditEvent`: you will only get an error within the macro: `cannot find struct AuditEvent`. This is unclear, because my rust-analyzer doen't show where it is missing. 

I've added a fully qualified path to `AuditEvent` in macro `as_audit_event` and removed imports in files where they are not necessary anymore. This prevents the situation with the vague error.

Goal was to also get rid of import `AsAuditEvent` as well, however that one is needed for `as_audit_event`.

### DIFF
--- a/backend/src/api/authentication.rs
+++ b/backend/src/api/authentication.rs
@@ -20,7 +20,7 @@ use crate::{
     },
     domain::role::Role,
     error::ErrorReference,
-    infra::audit_log::{AsAuditEvent, AuditEvent, AuditEventType, AuditService, as_audit_event},
+    infra::audit_log::{AsAuditEvent, AuditEventType, AuditService, as_audit_event},
     repository::{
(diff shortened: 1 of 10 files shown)


oliver3 (approved):

stacktraceghost:
As discussed with @oliver3 and @praseodym, we have decided to change the macro into a trait and pass the event type+level with it. I will create a new PR for that.

PR #2949, opened 2026-02-25
Add data entry and investigation services

### LINKED ISSUE #2939: Change 'create empty data entry' into service function
We could create a service function instead, that uses the data entry repository to create the data entry and the polling station repository to link the data entry to the polling station and return the updated struct.

_Originally posted by @praseodym in https://github.com/kiesraad/abacus/pull/2927#discussion_r2842988167_

praseodym:
- Change 'create empty data entry' into service function.
- Move test helpers for creating definitive data entries and investigations with corrected results from the repository layer to the service layer.

Resolves #2939


backend/src/service/data_entry.rs
```diff
@@ -0,0 +1,66 @@
...
+impl From<sqlx::Error> for DataEntryServiceError {
+    fn from(err: sqlx::Error) -> Self {
+        Self::DatabaseError(err)
+    }
+}
+
+/// Create an empty data entry and link it to the given polling station.
+/// If the polling station already has a data entry, this is a no-op and
+/// returns the polling station as-is.
+pub async fn create_empty(
```
oliver3, line 22:
```suggestion
pub async fn create_empty_data_entry(
```

backend/src/service/data_entry.rs
```diff
@@ -0,0 +1,66 @@
...
+    conn: &mut SqliteConnection,
+    polling_station_id: PollingStationId,
+) -> Result<PollingStation, DataEntryServiceError> {
+    let mut tx = conn.begin().await?;
+
+    // If data entry already exists, return polling station as-is
+    if data_entry_repo::data_entry_exists(&mut tx, polling_station_id).await? {
+        let polling_station = polling_station_repo::get_by_id(&mut tx, polling_station_id).await?;
```
oliver3: I don't see a reason why this function should be able to create a data entry in a previous committee_session, other than one specific unit test that would otherwise fail.

I think having both a `polling_station_repo::get_by_id` and `polling_station_repo::get` with identical signature, between which you have to choose for getting a polling station, is not a good idea.

You can also use `data_entry_repo::create_empty()` and `polling_station_repo::link_data_entry()` directly in `create_definitive_data_entry()` without checking for `data_entry_repo::data_entry_exists()` because ~~that does not occur in the tests~~ the tests still pass without it, currently wondering if this is okay..

Then there is no need for the `polling_station_repo::get_by_id()` anymore.

oliver3:
Good improvement overall!

backend/src/service/data_entry.rs
praseodym: I've re-exported it like this to not have a lot of data_entry function names within a data_entry module

backend/src/service/data_entry.rs
praseodym: If we skip the `data_entry_exists` call, we can possibly create orphan data entries - surely that's not an improvement?

I propose we remove `polling_station_repo::get_by_id` and change the single failing unit test to create a data entry using repo functions instead of the service function.

There's still a broader discussion to be held though: should repo functions incorporate business logic, such as `polling_station_repo::get` incorporating checks to fetch only polling stations for the latest committee session?

oliver3: Not checking for `data_entry_repo::data_entry_exists()` will cause "orphan" data entries in the test, so probably still best to either check for `data_entry_exists()` before creating and linking, or something completely different: not allow the test to use `create_definitive_data_entry()` when it is known that the data entry already exists in the fixture, and updating the data entry to the required state instead.

I still believe that these two options are better than introducing the `polling_station_repo::get_by_id()`

praseodym: Changed in 019102c8bfac05283900bee8039a95568f13126b

backend/src/service/data_entry.rs
oliver3: Not sure if I like that pattern, renaming a lot of the functions in the re-export, but let's re-evaluate when we have some more services in our code.

backend/src/service/data_entry.rs
oliver3: I agree with your doubt about having business logic in the repo layer (other than maybe some data consistency). Let's keep this in mind.

Issue #2941, opened 2026-02-24
Restrict apportionment to CSB only


Lionqueen94:
- [ ] Restrict apportionment API endpoint to CSB election with status Completed
- [ ] Restrict apportionment API endpoint to Coordinator CSB only
- [ ] Restrict frontend apportionment pages as well

PR #2953, opened 2026-02-26
Bring back apportionment API

### LINKED ISSUE #2858: Bring back apportionment API
Implement API according to new apportionment crate

Reference: https://github.com/kiesraad/abacus/pull/2208

stacktraceghost:
Resolves #2858 

This PR implements the apportionment API endpoint. 

Changes/steps:
- Brought back apportionment API code
- Implemented associative types for `ListNumber` and `CandidateNumber` so we don't have to dereference or convert our own newtypes to u32
- Added concrete structs and mapping for the output
- Brought back apportionment integration tests, converted them to a version without structs
- Added a test for the output of the apportionment API (to cover mapping)

Currently the endpoint is restricted to GSB, later we can adjust this to CSB. See https://github.com/kiesraad/abacus/issues/2941

### DIFF
--- a/backend/src/api/apportionment/handlers.rs
+++ b/backend/src/api/apportionment/handlers.rs
@@ -75,7 +75,7 @@ pub async fn election_apportionment(
         let summary = ElectionSummary::from_results(&election, &results)?;
         let input = ApportionmentInputData {
             number_of_seats: election.number_of_seats,
-            list_votes: summary.political_group_votes.clone(),
+            list_votes: &summary.political_group_votes,
         };
         let result = apportionment::process(&input)?;
 
--- a/backend/src/api/apportionment/structs.rs
+++ b/backend/src/api/apportionment/structs.rs
@@ -9,12 +9,12 @@ use crate::domain::{
 };
 
 #[derive(Clone, Debug)]
-pub struct ApportionmentInputData {
+pub struct ApportionmentInputData<'a> {
     pub number_of_seats: u32,
-    pub list_votes: Vec<PoliticalGroupCandidateVotes>,
+    pub list_votes: &'a [PoliticalGroupCandidateVotes],
 }
 
-impl apportionment::ApportionmentInput for ApportionmentInputData {
+impl<'a> apportionment::ApportionmentInput for ApportionmentInputData<'a> {
     type List = PoliticalGroupCandidateVotes;
 
     fn number_of_seats(&self) -> u32 {
@@ -22,7 +22,7 @@ impl apportionment::ApportionmentInput for ApportionmentInputData {
     }
 
     fn list_votes(&self) -> &[Self::List] {
-        &self.list_votes
+        self.list_votes
     }
 }
 


backend/src/api/apportionment/handlers.rs
```diff
@@ -0,0 +1,166 @@
...
+) -> Result<Json<ElectionApportionmentResponse>, APIError> {
+    let mut conn = pool.acquire().await?;
+    let election = election_repo::get(&mut conn, id).await?;
+    let current_committee_session =
+        committee_session_repo::get_election_committee_session(&mut conn, election.id).await?;
+
+    if current_committee_session.status == CommitteeSessionStatus::Completed {
+        let results: Vec<(
+            crate::domain::polling_station::PollingStation,
+            crate::domain::data_entry::PollingStationResults,
```
Lionqueen94: Might be nice to import these so the code is a bit shorter?

backend/src/api/apportionment/mapping.rs
```diff
@@ -0,0 +1,154 @@
...
+#[test]
+fn test_sort_candidates_alphabetically() {
+    // (initials, last name)
+    let names: Vec<(&str, &str)> = vec![
+        ("A.", "Duin"),
+        ("M.", "Appel"),
+        ("M.", "Zee"),
+        ("A.", "Zee"),
+        ("N.", "Zee"),
+        ("D.", "Zee"),
```
Lionqueen94: Maybe one option with `N.A.` and one with `N.B.` as initials and the same last name would be nice, to see the sorting work on the second initial as well.

backend/src/api/apportionment/structs.rs
```diff
@@ -0,0 +1,235 @@
...
+        display_fraction::DisplayFraction,
+        election::{Candidate, CandidateNumber, PGNumber},
+        summary::ElectionSummary,
+    },
+    infra::audit_log::{AsAuditEvent, AuditEvent, AuditEventType, as_audit_event},
+};
+
+#[derive(Serialize)]
+pub struct ApportionmentCreated(pub ElectionDetails);
+as_audit_event!(ApportionmentCreated, AuditEventType::ApportionmentCreated);
```
Lionqueen94: Thinking about it, it might be nice to say `ApportionmentGenerated` here instead of `ApportionmentCreated`, since it's something the user does not do themselves (besides inputting the potential deceased candidates and/or lot drawing results). Can also be something to change later btw, maybe discussed with @jorisleker

backend/src/api/apportionment/structs.rs
```diff
@@ -0,0 +1,235 @@
...
+        self.number
+    }
+
+    fn candidate_votes(&self) -> &[Self::Cv] {
+        &self.candidate_votes
+    }
+}
+
+impl CandidateVotesTrait for CandidateVotes {
+    type CandidateNumber = CandidateNumber;
```
Lionqueen94: Here you might want to use the full path of `CandidateNumber` to make it more explicit?

backend/src/api/apportionment/structs.rs
stacktraceghost: Maybe `ApportionmentProcessed` ?

backend/src/api/apportionment/structs.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2953/commits/9e75d6c99e00852d845aa2abfa60ffba2a576e2e

backend/src/api/apportionment/mapping.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2953/commits/9e75d6c99e00852d845aa2abfa60ffba2a576e2e

backend/src/api/apportionment/handlers.rs
stacktraceghost: Oops! Adjusted in https://github.com/kiesraad/abacus/pull/2953/commits/9e75d6c99e00852d845aa2abfa60ffba2a576e2e

backend/src/domain/display_fraction.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn from(display_fraction: DisplayFraction) -> Self {
+        Self {
+            numerator: display_fraction.numerator
+                + display_fraction.integer * display_fraction.denominator,
+            denominator: display_fraction.denominator,
+        }
+    }
+}
+
+impl PartialEq for DisplayFraction {
```
oliver3: I don't think we need to compare `DisplayFractions` in our code, and this is only for test assertions?

If this is true, I expect the comparison to be made on a field‑by‑field basis rather than as a mathematically correct comparison, and we can derive `PartialEq`.

backend/src/api/apportionment/structs.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn number_of_seats(&self) -> u32 {
+        self.number_of_seats
+    }
+
+    fn list_votes(&self) -> &[Self::List] {
+        &self.list_votes
+    }
+}
+
+impl ListVotesTrait for PoliticalGroupCandidateVotes {
```
oliver3: Personally prefer traits not ending in Trait (same for CandidateVotesTrait), but feel free to resolve this comment.

backend/src/domain/display_fraction.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{DisplayFraction, Fraction};
+
+    #[test]
+    fn test_display_fraction_integer_gt_0() {
+        let fraction = DisplayFraction::from(Fraction::new(11, 5));
+        assert_eq!(
```
oliver3: With the `PartialEq` above, these tests do not really prove that this is working; it also passes when asserting with expected `DisplayFraction { integer: 0, numerator: 11, denominator: 5 }` for instance.

backend/src/api/apportionment/structs.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2953/commits/216dff9890c0d96a5cf996e209c90f3e1f61781f

backend/src/domain/display_fraction.rs
stacktraceghost: We don't compare the `DisplayFraction`, so I have decided to remove `PartialEq` and the test methods. I've moved `DisplayFraction` out of domain and towards the apportionment API structs.

Changed in https://github.com/kiesraad/abacus/pull/2953/commits/36fc9190f300523b63fabe2fb10ae74974ab8efb

backend/src/domain/display_fraction.rs
stacktraceghost: Changed in https://github.com/kiesraad/abacus/pull/2953/commits/36fc9190f300523b63fabe2fb10ae74974ab8efb, see previous comment

backend/src/api/apportionment/structs.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/2953/commits/d35081ba31cfc4a7f56e54192650c57eac8d5630

backend/src/api/apportionment/handlers.rs
```diff
@@ -0,0 +1,175 @@
...
+            data_entry_repo::list_results_for_committee_session(
+                &mut conn,
+                current_committee_session.id,
+            )
+            .await?;
+
+        let summary = ElectionSummary::from_results(&election, &results)?;
+        let input = ApportionmentInputData {
+            number_of_seats: election.number_of_seats,
+            list_votes: summary.political_group_votes.clone(),
```
praseodym: Small detail but you can pass this by reference (`list_votes: &summary.political_group_votes`) by changing the `ApportionmentInputData` struct and trait impl:

```rust
#[derive(Clone, Debug)]
pub struct ApportionmentInputData<'a> {
    pub number_of_seats: u32,
    pub list_votes: &'a [PoliticalGroupCandidateVotes],
}

impl<'a> apportionment::ApportionmentInput for ApportionmentInputData<'a> {
    type List = PoliticalGroupCandidateVotes;

    fn number_of_seats(&self) -> u32 {
        self.number_of_seats
    }

    fn list_votes(&self) -> &[Self::List] {
        self.list_votes
    }
}
```

stacktraceghost: Thanks, adjusted in https://github.com/kiesraad/abacus/pull/2953/commits/15cc6f3864e5314fc31e8c81db9e17d9d6e62de1

PR #2955, opened 2026-02-26
Audit log replaced as_audit_event macro with trait

stacktraceghost:
As a follow-up to https://github.com/kiesraad/abacus/pull/2948, I've made the following changes:
- Replaced the `as_audit_event` macro with a trait. 
  - Removed impl `fn level()` on `AuditEventType` (mapping event type -> level got removed)
- Renamed structs to `*AuditData`
  - Removed derives except for `Serialize` for custom structs
  - Moved out of domain where necessary
- Renamed event `PollingStationInvestigation*` to `Investigation*`

Please review carefully if the event levels are correct [compared to the old mapping](https://github.com/kiesraad/abacus/pull/2955/changes?w=1#diff-6bc0ef6302c766332fb1345601ac130d102648234efde465b9b8958ab8a26f3fL162).


backend/src/service/change_committee_session_status.rs
```diff
@@ -2,13 +2,13 @@ use sqlx::{Connection, SqliteConnection};
 
 use crate::{
     APIError,
+    api::{committee_session::CommitteeSessionUpdatedAuditData, report::FileDeletedAuditData},
```
oliver3: I don't think we want references back to the API layer from services. 

In case of  `FileDeletedAuditData` it can simply go to this service, but `CommitteeSessionUpdatedAuditData` is also used in `committee_session.rs` itself. It causes a circulair dependency between `committee_session.rs` and `change_committee_session_status.rs` so I would prefer also `CommitteeSessionUpdatedAuditData` to go to `change_committee_session_status.rs`.

stacktraceghost: Good catch, adjusted in https://github.com/kiesraad/abacus/pull/2955/commits/26e6aed1a2f89a30dfabcd1007645cc3ffe99d6c

Lionqueen94 (approved):

PR #2956, opened 2026-02-26
Audit log event filter fix

stacktraceghost:
The event filter on the log page does not work due to a mistake in the query. 

Changes:
1. [Fixed query](https://github.com/kiesraad/abacus/commit/98f5ed25879c891f1b6503364f5793d729b18f5b)
2. [Refactored tests and added cases for filtering](https://github.com/kiesraad/abacus/pull/2956/commits/6aae6edbdc02d3697854e3f710570c07b9484144)

### DIFF
--- a/backend/src/api/audit.rs
+++ b/backend/src/api/audit.rs
@@ -134,38 +134,22 @@ async fn audit_log_list_users(
 mod tests {
     use std::net::Ipv4Addr;
 
-    use axum::{
-        Router,
-        body::Body,
-        http::{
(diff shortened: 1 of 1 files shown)


praseodym:
Looking at the commit history I think this broke in #1931, where `eventType` was renamed to `event_type`.

stacktraceghost:
> Looking at the commit history I think this broke in #1931, where `eventType` was renamed to `event_type`.

Confirmed, checked out the commit before and after merge.

Issue #2904, opened 2026-02-16
Transition investigation into enum with state machine


praseodym:
Transition investigation into `InvestigationState` enum with state machine, as described in the types in #2881

PR #2962, opened 2026-02-27
Move investigations to polling stations table

praseodym:
Move investigations to polling stations table as a JSON column. This is part of #2904, next step is to introduce the investigations state machine as described in #2881.


backend/fixtures/election_5_with_results.sql
```diff
@@ -476,7 +476,6 @@ VALUES (8, 5, NULL, 501, 'Testgebouw', 41, NULL, 'FixedLocation', 'Testweg 3', '
        (9, 6, 501, NULL, 'Testgebouw', 41, NULL, 'FixedLocation', 'Testweg 3', '1234 QA', 'Grote Stad'),
        (11, 6, NULL, 502, 'Test ander gebouw', 42, NULL, 'FixedLocation', 'Testweg 4', '1234 QA', 'Grote Stad');
 
-INSERT INTO
-    polling_station_investigations
-VALUES
-    (11, "reason", "findings", 1);
+UPDATE polling_stations
+SET investigation_state = '{"polling_station_id":11,"reason":"reason","findings":"findings","corrected_results":true}'
```
Lionqueen94, line 480: Is it still needed to have the `polling_station_id` in the investigation state?

backend/src/repository/investigation_repo.rs
```diff
@@ -9,118 +9,121 @@ use crate::domain::{
...
+    if result.rows_affected() == 0 {
+        return Err(sqlx::Error::RowNotFound);
+    }
+
+    Ok(())
+}
+
 pub async fn create_polling_station_investigation(
     conn: &mut SqliteConnection,
     polling_station_id: PollingStationId,
```
cikzh, line 35: Slight nitpick, but maybe we could add a little more type safety by passing a `CurrentSessionPollingStationId` here. That way we know the ID exists (since we've looked it up in FromRequestParts) and we can be sure the error on row 55 is caused by `investigation_state` being `null`.  It's not a 100% guarantee though, because in theory the Polling Station could be deleted between `FromRequestParts` and this call.

backend/fixtures/election_5_with_results.sql
praseodym: No, it will be removed in the state machine refactor (next PR)

backend/src/repository/investigation_repo.rs
praseodym: Thanks, will change it in the next PR

PR #2964, opened 2026-03-02
Migrate to eml-nl crate

rnijveld:
This removes all EML_NL parsing code from Abacus and replaces it with a couple of associated functions/methods on a couple of the Abacus structs to do the conversions. Many of the checks that Abacus previously did have moved into the eml-nl crate. My suggestion is to keep reviews here limited to just the changes in Abacus. A separate walkthrough of the eml-nl crate will also take place. Changes outside of the eml module should be limited to updated method calls and some updates to our test data because it contained invalid election ids.

### DIFF
--- a/backend/src/eml/mod.rs
+++ b/backend/src/eml/mod.rs
@@ -79,17 +79,18 @@ impl NewElection {
 
         // typically number of voters is not available in the EML, but it could
         // be so we try and extract it anyway
-        let number_of_voters = election
-            .contest
-            .max_votes
-            .copied_value()
+        let max_votes = election.contest.max_votes.copied_value();
+        let number_of_voters = max_votes
             .map(|v| v.get())
             .unwrap_or(0)
             .try_into()
             .unwrap_or(0);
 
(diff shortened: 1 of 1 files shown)


backend/src/eml/mod.rs
```diff
@@ -1,17 +1,818 @@
...
+            .max_votes
+            .copied_value()
+            .map(|v| v.get())
+            .unwrap_or(0)
+            .try_into()
+            .unwrap_or(0);
+
+        // the number of seats must be between 9 and 45 for municipal elections
+        // clippy suggests way less readable alternative
+        #[allow(clippy::manual_range_contains)]
```
praseodym: nit:
```suggestion
        #[allow(clippy::manual_range_contains, reason = "clippy suggests way less readable alternative")]
```

cikzh (approved):
LGTM, only note I have is that eml/mod.rs is getting quite chonky with al kinds of functions and methods. Maybe they belong in their respective domains?

PR #2968, opened 2026-03-02
Change investigation status into state machine

praseodym:
Change investigation status into state machine, without changes to the API types and frontend to keep the PR size as small as possible.

Open question is if we think this is enough to resolve issue #2904. Possible future improvements we can do at a later point are:
- Also change API types to reflect the new state enum
- Move state updating logic to the service layer

### DIFF
--- a/backend/src/domain/investigation.rs
+++ b/backend/src/domain/investigation.rs
@@ -58,16 +58,15 @@ impl InvestigationStatus {
 
     /// InProgress -> ConcludedWithoutNewResults
     ///
-    /// `has_previous_results` must be true (the polling station must have results
-    /// from a previous session), otherwise corrected results are required.
+    /// `new_polling_station` must be false: for a new polling station, corrected results are required (there are no previous results).
     pub fn conclude_without_new_results(
         self,
         findings: String,
-        has_previous_results: bool,
+        new_polling_station: bool,
     ) -> Result<Self, InvestigationTransitionError> {
         match self {
             Self::InProgress(state) => {
-                if !has_previous_results {
+                if new_polling_station {
                     return Err(InvestigationTransitionError::RequiresCorrectedResults);
                 }
                 Ok(Self::ConcludedWithoutNewResults(
@@ -119,11 +118,11 @@ impl InvestigationStatus {
         self,
         reason: String,
         findings: String,
-        has_previous_results: bool,
+        new_polling_station: bool,
     ) -> Result<Self, InvestigationTransitionError> {
         match self {
             Self::ConcludedWithoutNewResults(_) | Self::ConcludedWithNewResults(_) => {
-                if !has_previous_results {
+                if new_polling_station {
                     return Err(InvestigationTransitionError::RequiresCorrectedResults);
                 }
                 Ok(Self::ConcludedWithoutNewResults(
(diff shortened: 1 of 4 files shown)


backend/src/repository/data_entry_repo.rs
```diff
@@ -409,8 +409,8 @@ pub async fn are_results_complete_for_committee_session(
         LEFT JOIN data_entries AS de ON de.id = ps.data_entry_id
         WHERE ps.committee_session_id = ?
         AND ps.investigation_state IS NOT NULL
-        AND (json_extract(ps.investigation_state, '$.corrected_results') IS NULL
-             OR (json_extract(ps.investigation_state, '$.corrected_results') = 1
+        AND (json_extract(ps.investigation_state, '$.status') = 'InProgress'
```
oliver3, line 412: If I understand correctly, this should be `ConcludedWithoutNewResults` instead of `InProgress` to check if all investigations are finished

backend/src/api/report.rs
```diff
@@ -434,8 +437,12 @@ async fn get_files(
 ) -> Result<(Option<File>, Option<File>, Option<File>, DateTime<Utc>), APIError> {
     let mut conn = pool.acquire().await?;
     let committee_session = committee_session_repo::get(&mut conn, committee_session_id).await?;
-    let investigations =
-        list_investigations_for_committee_session(&mut conn, committee_session.id).await?;
+    let investigations: Vec<PollingStationInvestigation> =
```
rnijveld, line 436: This pattern of iterating `list_for_committee_session` and getting `PollingStationInvestigation` in return repeats itself a couple of times, maybe make this an associated function of `PollingStationInvestigation` instead?

backend/src/domain/investigation.rs
```diff
@@ -5,7 +5,228 @@ use axum::{
...
+    }
+
+    /// InProgress -> ConcludedWithoutNewResults
+    ///
+    /// `has_previous_results` must be true (the polling station must have results
+    /// from a previous session), otherwise corrected results are required.
+    pub fn conclude_without_new_results(
+        self,
+        findings: String,
+        has_previous_results: bool,
```
oliver3: This flag confused me a bit. Do I understand it correctly that for newly added polling stations, there have to be new results, so you cannot conclude_without_new_results for new polling station, i.e. a polling station without corresponding previous data entry results?

Maybe we can invert and rename this boolean to `new_polling_station` ?

backend/src/api/report.rs
praseodym: I didn't bother with it before, but done in 65aa7bf1ab42102e2419b9f30fdbdacd4dacb929

backend/src/domain/investigation.rs
praseodym: `has_previous_results` is false for new polling stations, we could rename it to `new_polling_station` and flip the meaning

backend/src/repository/data_entry_repo.rs
praseodym: corrected_results = NULL (not yet determined whether there are correct results) is the old way of saying "in progress", so the logic is a direct translation

oliver3: Ah yes, it checks for `COUNT(*) = 0` so the inverse of what I was thinking. Thanks 👍

backend/src/domain/investigation.rs
praseodym: Changed in 9554d9b5165181d0526e10aabda5e983b657bf48

oliver3: Thanks!

PR #2980, opened 2026-03-03
Use the RoleWithoutCommitteeCategory only for user create

oliver3:
Do not use `RoleWithoutCommitteeCategory` outside of the user create, which was introduced in https://github.com/kiesraad/abacus/pull/2930 (as `RoleWithoutElection`) to reduce the number of changes.

### DIFF
--- a/frontend/src/features/data_entry/components/DataEntrySection.tsx
+++ b/frontend/src/features/data_entry/components/DataEntrySection.tsx
@@ -94,20 +94,14 @@ export function DataEntrySection({ committeeCategory }: DataEntryProps) {
         )}
         {error instanceof ApiError && <ErrorModal error={error} />}
         {formSection.isSaved && memoizedErrors.length > 0 && (
-          <Feedback
-            id="feedback-error"
-            type="error"
-            data={memoizedErrors}
-            userRole={user.roleWithoutCommitteeCategory}
-            shouldFocus={true}
-          />
+          <Feedback id="feedback-error" type="error" data={memoizedErrors} userRole={user.role} shouldFocus={true} />
         )}
         {formSection.isSaved && memoizedWarnings.length > 0 && (
           <Feedback
             id="feedback-warning"
             type="warning"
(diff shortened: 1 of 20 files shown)


frontend/src/features/data_entry/components/DataEntrySection.tsx
```diff
@@ -94,20 +94,14 @@ export function DataEntrySection({ committeeCategory }: DataEntryProps) {
...
         {error instanceof ApiError && <ErrorModal error={error} />}
         {formSection.isSaved && memoizedErrors.length > 0 && (
-          <Feedback
-            id="feedback-error"
-            type="error"
-            data={memoizedErrors}
-            userRole={user.roleWithoutCommitteeCategory}
-            shouldFocus={true}
-          />
+          <Feedback id="feedback-error" type="error" data={memoizedErrors} userRole={user.role} shouldFocus={true} />
```
lama-d3v, line 97: typo in `memoizedErrors`?

frontend/src/features/users/hooks/UserCreateContext.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+} from "@/features/users/hooks/UserCreateContext";
+
+describe("UserCreateContext", () => {
+  test.each(
+    roleWithoutCommitteeCategoryValues,
+  )("String %j should be a RoleWithoutCommitteeCategory", (role: RoleWithoutCommitteeCategory) => {
+    expect(isRoleWithoutCommitteeCategory(role as string)).toBe(true);
+  });
+
+  test('String "typist_gsb" is not a RoleWithoutCommitteeCategory', () => {
```
lama-d3v, line 15: Does only `typist_gsb` needs to be tested here?

oliver3: I think so, one string which is not a RoleWithoutCommitteeCategory

PR #2981, opened 2026-03-04
Introduce `PollingStation{First,Next}Session` structs

### LINKED ISSUE #2906: Implement `PollingStationFirstSession` + `PollingStationNextSession`

praseodym:
Reorganise polling station structs and introduce `PollingStation{First,Next}Session` structs as described in #2881; resolves #2906. Please review per commit.

- **Update backend fuzz lockfile**

- **Reorganise polling station structs**
  - Introduce `PollingStationRow` struct that matches the SQL schema.
  - Remove committee session and data entry fields from `PollingStation`.
    struct, as these fields are not needed in many places.
  - No API changes.

- **Add `PollingStation{First,Next}Session` structs**

- **Add `PollingStationsForSession` for list calls**

- **Add `PollingStationForSession` for get calls**
  Replace `PollingStationRow` by `PollingStationForSession` outside of the
  polling station repository.


backend/src/domain/polling_station.rs
```diff
@@ -18,10 +18,29 @@ use crate::{
 pub type PollingStationNumber = u32;
 id!(PollingStationId);
 
-/// Polling station of a certain [crate::domain::election::Election]
-#[derive(Serialize, Deserialize, ToSchema, Debug, FromRow, Clone)]
+/// Polling station base entity, linked to an election but
+/// independent of the committee session or data entry.
+#[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct PollingStation {
```
oliver3, line 25: Probably won't need all of these derives anymore?

Edit: Not important enough to not-merge, as there are many more structs with superfluous derives.

backend/src/api/polling_station.rs
```diff
@@ -172,7 +176,7 @@ pub fn validate_user_is_allowed_to_perform_action(
     path = "/api/elections/{election_id}/polling_stations",
     request_body = PollingStationRequest,
     responses(
-        (status = 201, description = "Polling station created successfully", body = PollingStation),
+        (status = 201, description = "Polling station created successfully", body = PollingStationResponse),
```
oliver3, line 179: I don't understand why the `PollingStationResponse` does not show up in the OpenAPI files..

backend/src/domain/polling_station.rs
```diff
@@ -141,6 +141,71 @@ impl From<String> for PollingStationType {
...
+    pub committee_session_id: CommitteeSessionId,
+    pub polling_station: PollingStation,
+    pub prev_data_entry_id: Option<DataEntryId>,
+    pub data_entry_id: Option<DataEntryId>,
+    pub investigation_status: Option<InvestigationStatus>,
+}
+
+impl PollingStationNextSession {
+    /// Returns true if this polling station was newly added in this session.
+    pub fn is_new_polling_station(&self) -> bool {
```
oliver3, line 166: Nice :)

backend/src/api/polling_station.rs
praseodym: Because it has been annotated with `#[schema(as = PollingStation)]` to avoid changing the OpenAPI spec

backend/src/domain/polling_station.rs
praseodym: Still needed, this struct is used in the models where it is also serialised to JSON

praseodym:
> Not sure if the structs are all in the correct layer (e.g. `PollingStationResponse` should go to API?)

I was also not sure about that, we can always move them.

> and if the `PollingStationForSession` and `PollingStationsForSession` enums are helping or actually undo'ing some of the type safety

We need some enum like that to avoid having to match on first/next committee session everywhere, I'm open to suggestions though!

backend/src/api/polling_station.rs
oliver3: Ah I see, thanks!

PR #2995, opened 2026-03-05
Add sub electoral committee entity

### LINKED ISSUE #2907: Implement `SubElectoralCommittee` + `SubElectoralCommitteeWithDataEntry`

praseodym:
Add sub electoral committee entity (`SubElectoralCommittee` and `SubElectoralCommitteeFirstSession` as described in #2881), resolves #2907.


praseodym:
Still need to add a number column to the sub committee, should be municipality number for GSB sub committees.

praseodym:
> Still need to add a number column to the sub committee, should be municipality number for GSB sub committees.

Done in 03303805943378ef186f4bdc7f0992ed6c0f1670

backend/migrations/9_sub_committees.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE sub_committees
+(
+    id                    INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    committee_session_id  INTEGER                           NOT NULL,
+    data_entry_id         INTEGER                           NOT NULL,
+    number                TEXT                              NOT NULL,
```
oliver3: The table polling_stations has a `number INTEGER NOT NULL,`, maybe it is best to keep them the same type?
(Maybe polling_stations.number should become TEXT..)

backend/src/repository/sub_committee_repo.rs
```diff
@@ -0,0 +1,127 @@
...
+struct SubCommitteeRow {
+    id: SubCommitteeId,
+    committee_session_id: CommitteeSessionId,
+    data_entry_id: DataEntryId,
+    number: String,
+    name: String,
+    category: CommitteeCategory,
+}
+
+impl From<SubCommitteeRow> for SubCommittee {
```
oliver3: This seems unused

backend/src/service/sub_committee.rs
```diff
@@ -0,0 +1,90 @@
...
+        number,
+        name,
+        category,
+    )
+    .await?;
+    tx.commit().await?;
+    Ok(sub_committee)
+}
+
+pub async fn list_for_first_session(
```
oliver3, line 45: Why this extra service layer?

backend/migrations/9_sub_committees.sql
praseodym: For municipal elections the number gets copied from the election domain id which is a `String`/`TEXT` field because it has leading zeroes, I just assumed we want to preserve those for the UI.

@jorisleker what do you think?

backend/src/service/sub_committee.rs
praseodym: To align it with `list_for_session` in the polling station service

backend/src/repository/sub_committee_repo.rs
praseodym: Good catch, it was left over from an older version. Removed in 5a1c4964c51d6af229c5f1a8051f84ba00d8be6b

backend/migrations/9_sub_committees.sql
jorisleker: In the UI I would prefer to strip the leading zeros. If we don't need to export this data in the eml or pdf in its original format, I am okay with storing this as integers

praseodym: Changed to numeric in 645fd9b6bf575f3984796748c1bee7d7cf905358.

backend/src/service/sub_committee.rs
oliver3: Okay..

PR #3008, opened 2026-03-05
Rename CommitteeSessionStatus "InPreparation" to "Ready"

oliver3:
I kept finding it confusing that the "second state" of the committee session was called `InPreparation`, especially when the `Created` state before that is translated as "Zitting voorbereiden", and the method to go to `Created` is called `prepare_data_entry()` (whereas the method to go to `InPreparation` is called `ready_for_data_entry()`).

There were already some other team discussions about this before, so I propose to rename `InPreparation` to `Ready` to indicate that the session is ready for data entry ("Klaar voor invoer"), to try to alleviate some confusion.


oliver3:
Won't do, Ready is also not the right term. 

Both `Created` and `InPreparation` indicate that the coordinator is preparing a session, with the difference being that you cannot or can go to `DataEntry`.

Issue #2987, opened 2026-03-04
Update committee session transitions to check data entry


jschuurk-kr:
Check data entry instead of polling station or investigation, because we now have a third one: sub-electoral committee (sub-committee?).

PR #3009, opened 2026-03-05
Update committee session transitions to check data entries

oliver3:
Resolves [Update committee session transitions to check data entry](https://github.com/kiesraad/abacus/issues/2987)

Committee session state machine uses either data entries (first session) or investigations (next session) to determine validity of state change.

Update mocks for state machine tests to use Options to assert that a provider method is not called for a certain test scenario.

### DIFF
--- a/backend/src/api/polling_station.rs
+++ b/backend/src/api/polling_station.rs
@@ -29,7 +29,7 @@ use crate::{
     infra::audit_log::{AsAuditEvent, AuditEventLevel, AuditEventType, AuditService},
     repository::{
         committee_session_repo::get_election_committee_session,
-        election_repo,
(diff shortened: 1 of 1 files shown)


stacktraceghost (approved):

oliver3:
> I think we should also change this transition check to use `data_entry_repo::has_any`:

For a next session we should also check for the existence of investigations, implemented in https://github.com/kiesraad/abacus/pull/3009/commits/08290e457208662f6de6dcb8dd5fe1225fd7da64

> And after that the `polling_station_repo::has_any` function can be removed

It is still used in `polling_station_import()` and I think it is more clear than also checking for either data entry or investigations there.

Issue #3025, opened 2026-03-10
Enrich claim endpoint with needed data entry status info


oliver3:
So that the data entry does not need to retrieve all statuses to determine the status badge (and other things?)

PR #3026, opened 2026-03-10
Data entry source (polling station) and status from claim

oliver3:
Resolves [Enrich claim endpoint with needed data entry status info](https://github.com/kiesraad/abacus/issues/3025)

Add `DataEntryStatus` and `DataEntrySource` (polling station or subcommittee) to the claim endpoint response, and use those in data entry header, in order to decouple data entry from polling station and not use election status endpoint to retrieve the data entry status.

### DIFF
--- a/frontend/src/features/data_entry/components/DataEntryHeader.stories.tsx
+++ b/frontend/src/features/data_entry/components/DataEntryHeader.stories.tsx
@@ -1,0 +1,79 @@
+import type { StoryObj } from "@storybook/react-vite";
+import { expect } from "storybook/test";
+import { AppLayout } from "@/components/ui/AppLayout/AppLayout";
+import { DataEntryContext } from "@/features/data_entry/hooks/DataEntryContext";
+import type { DataEntryStateAndActionsLoaded } from "@/features/data_entry/types/types";
+import { TestUserProvider } from "@/testing/TestUserProvider";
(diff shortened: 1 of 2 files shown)


frontend/src/features/data_entry/components/DataEntryPage.tsx
```diff
@@ -41,17 +35,7 @@ export function DataEntryPage() {
 
   return (
     <DataEntryProvider election={election} pollingStation={pollingStation} entryNumber={entryNumber}>
```
praseodym, line 37: Can we also let `DataEntryProvider` use the new source so that it does not need to depend on the polling station?

oliver3:
> .., adding a test for the header would be appreciated

Added a storybook test in https://github.com/kiesraad/abacus/pull/3026/commits/f6f31861b602264386c20873330ab233cc760bac

frontend/src/features/data_entry/components/DataEntryPage.tsx
oliver3: Seems like the logical next step indeed, let's do that in a next PR.

oliver3: Added issue https://github.com/kiesraad/abacus/issues/3029

Issue #2836, opened 2026-02-03
Check API endpoints according to role committee category


oliver3:
See list https://github.com/kiesraad/abacus/issues/2810#issuecomment-3991686479

PR #3028, opened 2026-03-11
Authorize on route level

stacktraceghost:
Related to https://github.com/kiesraad/abacus/issues/2836

When working on the role/committee category authorization in the related issue, a lot of different role combinations were needed. At the moment, we have been defining them as structs with their own impls. Due to the amount of combinations coming up, this becomes too repetitive.

There is also a duplicate configuration necessary: one for the handler, one for utoipa to define the roles for the OpenAPI documentation. These are then validated in `router.rs` tests, also to make sure they match.

With the changes in this PR, we will be able to define roles directly on routes. This definition will also adjust the OpenAPI documentation, so we don't need check if they match. Public routes need to be explicitly declared as well.  

Example:
```rs
pub fn router() -> OpenApiRouter<AppState> {
    use Role::*;

    const ALLOWED_ROLES: &[Role] = &[Administrator, CoordinatorGSB];

    OpenApiRouter::default()
        .routes(routes!(handler_a).authorize(ALLOWED_ROLES))
        .routes(routes!(handler_b).authorize(ALLOWED_ROLES))
        .routes(routes!(handler_account_update).allow_incomplete_user())
        .routes(routes!(handler_public).public()
}
```

In this case:
- `handler_a/handler_b` will only allow the given roles and the OpenAPI documentation will reflect this. 
- `handler_account_update` will only allow an incomplete user (all roles)
- `handler_public` will get a tag `public`, which groups it in the Swagger UI:

image

There is a test available that will fail if neither or both `authorize(...)/allow_incomplete_user()` and `public()` are used on a route.

By checking openapi.json, you can make sure nothing has changed in terms of authorization roles within this PR. There are two exceptions:
1. Endpoints that allowed every role, will now also include CSB roles
2. Public endpoints have a tag `public` attached to them

Things to check:
1. Location of the middleware, is `role.rs` still the right file or should we move it to `middleware.rs`?
2. [These test changes](https://github.com/kiesraad/abacus/pull/3028/changes#diff-ecc8f378479d6058db045fc32c2ad82faf98005b6f186f2af17934dffea1106cL723). The service also denies access if no user is available. 
3. [Box.pin in service](http://github.com/kiesraad/abacus/blob/11d906646eb9744371194ead44fce4994f3b4585/backend/src/api/middleware/authentication/role.rs#L135), done according to [tower documentation](https://docs.rs/tower/latest/tower/trait.Service.html#be-careful-when-cloning-inner-services)

### DIFF
--- a/backend/tests/authorization_integration_test.rs
+++ b/backend/tests/authorization_integration_test.rs
@@ -23,9 +23,12 @@ async fn test_route_public(pool: SqlitePool) {
 #[test(sqlx::test(fixtures(path = "../fixtures", scripts("users"))))]
 async fn test_route_authorized(pool: SqlitePool) {
     let addr = serve_api(pool).await;
-    let coordinator_cookie = coordinator_login(&addr).await;
-
     let url = format!("http://{addr}/api/users");
+
+    let response = reqwest::Client::new().get(&url).send().await.unwrap();
+    assert_eq!(response.status(), StatusCode::UNAUTHORIZED);
+
+    let coordinator_cookie = coordinator_login(&addr).await;
     let response = reqwest::Client::new()
         .get(&url)
         .header("cookie", coordinator_cookie)


backend/src/infra/router.rs
```diff
@@ -295,156 +247,72 @@ pub fn create_router_without_airgap_detection(pool: SqlitePool) -> Result<Router
...
+                serde_json::to_value(req).ok().is_some_and(|v| {
+                    v.as_object()
+                        .is_some_and(|obj| obj.contains_key(authentication::SECURITY_SCHEME_NAME))
+                })
+            })
+        })
     }
 
-    #[test(sqlx::test(fixtures(path = "../../fixtures", scripts("election_2", "users"))))]
-    async fn test_route_authorization(pool: SqlitePool) {
```
praseodym, line 320: This old test called all endpoints to see if the actual authorization that was applied matches the definition, whereas the new test just checks the definition. I think we should try to test the actual implementation by calling the endpoints from a test.

stacktraceghost: The new test just makes sure if one of both (`authorize(...)` or `public()`) has been called on a route. 

For both the OpenAPI definition as the actual authorization, I've introduced [separate tests](https://github.com/kiesraad/abacus/pull/3028/changes#diff-d2a23f8d913468d1a342f06d34ed1013246ad4e68edf5f2edd3252dfed9403f5R208). The only assumption here is that `authorize` will add the layer and that axum will execute this layer. 

Instead of testing the service only under `mod tower_authorization`, which tests `service -> authorization`. We could use `build_method_router` and call `.authorize()` on it so we test `handler -> layer -> service -> authorization`.

backend/src/api/investigation.rs
```diff
@@ -183,10 +190,9 @@ where
     params(
         ("polling_station_id" = PollingStationId, description = "Polling station database id"),
     ),
-    security(("cookie_auth" = ["coordinator_gsb"])),
 )]
 async fn polling_station_investigation_create(
-    _user: CoordinatorGSB,
+    _user: User,
```
oliver3: Can all `_user: User` now be removed as well?

backend/src/infra/router.rs
stacktraceghost: Adjustments made in https://github.com/kiesraad/abacus/pull/3028/commits/3546fbfdd327bf9039e6c12b8ece364ef79ea2b3, the tests now use `build_method_router`. This way it is also verified that `authorize` had added the layer.

In https://github.com/kiesraad/abacus/pull/3028/commits/4d52472ea08e0fd3b8bb75ef644d0a6fd4965e46 I've also added integration tests to check:
- A public endpoint
- An authorized endpoint with correct role
- An authorized endpoint with incorrect role
- An endpoint using `IncompleteUser`

backend/src/api/investigation.rs
stacktraceghost: Good point, removed `_user: User` in https://github.com/kiesraad/abacus/pull/3028/commits/3bb9262cc9b7ca6932f20c90f2aae65d4d9d6148

Also added integration tests in https://github.com/kiesraad/abacus/pull/3028/commits/4d52472ea08e0fd3b8bb75ef644d0a6fd4965e46

stacktraceghost:
Test `test_cant_do_anything_when_password_needs_change` pointed out that after removing `_user: User` it was possible to access an endpoint for `User` also with `IncompleteUser`, since that check wasn't done anymore. Therefore we (thanks @oliver3 !) have made changes in https://github.com/kiesraad/abacus/commit/d252ea709428641e709aa563e078388f7e4d16f5

We have removed the need for `IncompleteUser` and integrated this logic with an extra method called ~~`allow_setup_user()`~~ `allow_incomplete_user()`, which allows access for an incomplete user (for all roles).

Extra integration tests and router tests have been added.

backend/src/api/authentication.rs
```diff
@@ -278,7 +281,6 @@ pub struct AccountUpdateRequest {
       (status = 401, description = "Invalid user session", body = ErrorResponse),
       (status = 500, description = "Internal server error", body = ErrorResponse),
   ),
-  security(("cookie_auth" = ["administrator", "coordinator_gsb", "typist_gsb"])),
 )]
 async fn account(user: Option<User>) -> Result<impl IntoResponse, APIError> {
```
praseodym: Because this endpoint has `.authorize(ALL_ROLES)` I don't think the user parameter should be an `Option` any longer?

stacktraceghost: Good point, that changes the way this handler acts. Before it returned a HTTP 401 with
```
{
    "error": "User not found",
    "fatal": false,
    "reference": "UserNotFound"
}
```

Now it returns a HTTP 401 without body, which is fine I think. I will remove the `Option<User>`, however the `OptionalFromRequestPars` in `role.rs` is still necessary for the `extend_session` middleware.

Changed in https://github.com/kiesraad/abacus/pull/3028/commits/d5de316f321df5fc66e084864fa8aaa57cbbd98b

stacktraceghost:
Still adjusted some naming in https://github.com/kiesraad/abacus/pull/3028/commits/06de19c9873280d8d285138680b42c6a0d5e882a to align with the concept of incomplete user.

PR #3033, opened 2026-03-11
Add support for sub committees in status API

### LINKED ISSUE #2996: Change election status endpoint
Return all data entry statuses enriched with number, name, typist this/prev entry, (progress?).

Check election status page and data entry homepage

praseodym:
- Create CSB committee sessions in data entry state
- Add support for sub committees in the election status API, resolves #2996

### DIFF
--- a/backend/src/repository/polling_station_repo.rs
+++ b/backend/src/repository/polling_station_repo.rs
@@ -629,7 +629,7 @@ pub async fn list_first_session_with_status(
         FROM polling_stations AS p
         JOIN committee_sessions AS c ON c.id = p.committee_session_id
         LEFT JOIN data_entries AS de ON de.id = p.data_entry_id
-        WHERE c.id = $1
+        WHERE c.id = $1 AND c.number = 1
         "#,
         committee_session_id
     )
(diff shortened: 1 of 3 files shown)


oliver3:
I was expecting also a data_entry_id to be returned next to the source and status. Is see however this was not mentioned in the issue..

praseodym:
Initial session status should be InPreparation, not DataEntry

praseodym:
> Initial session status should be InPreparation, not DataEntry

Changed in 9307b89379934e186dc0c65248f99405f0fe1bba

backend/src/repository/polling_station_repo.rs
```diff
@@ -610,6 +613,73 @@ pub async fn duplicate_for_committee_session(
     Ok(())
 }
 
+/// List all polling stations for a first committee session with their data entry status
+pub async fn list_first_session_with_status(
```
stacktraceghost, line 617: To avoid mistakes, we could add where `c.number = 1` or `c.number > 1` in these queries.

praseodym: Good idea, done in f2f0e5e5. Only for polling stations, because for the sub committees we don't use the committee session at all.

stacktraceghost (approved):

PR #3039, opened 2026-03-12
Extend fixtures for CSB, add FixtureUser

stacktraceghost:
Related to #2836

This PR:
- Extends fixtures
  - `election_8_csb` with political groups, sub committee and data entries
  - `users` with CSB coordinators/typists
- Seed now includes `election_8_csb`
- `UserFixture` for integration tests to easily select a user to login with
- `DevHomePage` shows new users and which committee category elections are:

image

Not implemented yet, but will be done in future PRs:
- Roles are not updated for each endpoint
- Filtering/checking on committee category

### DIFF
--- a/backend/tests/apportionment_integration_test.rs
+++ b/backend/tests/apportionment_integration_test.rs
@@ -10,8 +10,8 @@ use test_log::test;
 
 use crate::{
     shared::{
-        coordinator_login, create_result, create_result_with_non_example_data_entry,
-        differences_counts_zero, political_group_votes_from_test_data_auto,
+        FixtureUser, create_result, create_result_with_non_example_data_entry,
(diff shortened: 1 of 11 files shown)


stacktraceghost:
> Is it possible to only show the elections that the current user can access instead of the whole list on the DevHomePage?

This will be done in the next PR for https://github.com/kiesraad/abacus/issues/2836

stacktraceghost:
> The recurring `FixtureUser::` does add some noise in the test, you could consider importing the FixtureUsers variants with `use FixtureUser::*` instead and then
> 
> ```diff
> - let coordinator_cookie = login(&addr, FixtureUser::CoordinatorGSB).await;
> + let coordinator_cookie = login(&addr, CoordinatorGSB).await;
> ```

I will include this in the next PR.

PR #3042, opened 2026-03-13
Allow custom Chrome executable for Playwright

praseodym:
Use Chromium by default and add the `CHROMIUM_CHANNEL` environment variable to override the [channel](https://playwright.dev/docs/browsers#google-chrome--microsoft-edge) (e.g. `chrome` for locally installed Google Chrome). In addition, the `CHROMIUM_BIN` environment variable can be used if Playwright can't find the Chromium browser on the system by itself.

For other browsers Playwright still requires a patched build, so using a system browser doesn't make sense.

### DIFF
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -21,13 +21,14 @@ function returnWebserverCommand(): string {
   return `cd ../backend && cargo run --features memory-serve,embed-typst -- --database ../backend/target/debug/playwright.sqlite --port 8081`;
 }
 
-function chromeUseOptions() {
-  const executablePath = process.env.CHROME_BIN;
+function chromiumUseOptions() {
+  const channel = process.env.CHROMIUM_CHANNEL ?? "chromium";
+  const executablePath = process.env.CHROMIUM_BIN;
   return {
     ...devices["Desktop Chrome"],
     userAgent: "Abacus-User-Agent/1",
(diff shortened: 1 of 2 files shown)


stacktraceghost (approved):

frontend/playwright.config.ts
```diff
@@ -21,6 +21,16 @@ function returnWebserverCommand(): string {
   return `cd ../backend && cargo run --features memory-serve,embed-typst -- --database ../backend/target/debug/playwright.sqlite --port 8081`;
 }
 
+function chromeUseOptions() {
+  const executablePath = process.env.CHROME_BIN;
+  return {
+    ...devices["Desktop Chrome"],
+    userAgent: "Abacus-User-Agent/1",
+    // Use custom Chrome binary (e.g. Chromium) if set or use Playwright default
+    ...(executablePath ? { launchOptions: { executablePath } } : { channel: "chrome" }),
```
jschuurk-kr:
```suggestion
    // Use custom Chrome binary (e.g. Chromium) if set or use Playwright's default path
    ...(executablePath ? { launchOptions: { executablePath } } : { channel: "chrome" }),
```

This is almost too nitpicky, but with Chrome vs Chromium and with local install with/without custom path vs browsers installed by Playwright, I felt like "Playwright default" might lead to confusion in the future.

praseodym: Committed, even though I don't think this is a lot more clear

PR #3059, opened 2026-03-17
Renumber fixture polling station IDs and data entry IDs

oliver3:
While changing the data entry to use data entry ID instead of polling station ID, I've noticed that I missed something and the tests stayed green. This is because for instance the `data_entry_id` for `polling_station_id = 4` is also `4`.

This PR is to make these IDs unique:
- `polling_station_id` will be prefixed with the number of the fixture file, and the number of the committee session it belongs to, e.g. `2` becomes `212` and `9` becomes `529`
- `data_entry_id` will be prefixed with the number of the fixture file and a zero, e.g. `2` becomes `202`.

This corresponds to how it was constructed in `election_7_four_sessions.sql`

IDs generated in the code and not the fixtures are prefixed with `99` to able to make sure every ID in the test is updated, by searching for all `PollingStationId::from()` and `DataEntryId::from()` and making sure there is no single digit ID used anymore.

### DIFF
--- a/backend/tests/polling_station_integration_test.rs
+++ b/backend/tests/polling_station_integration_test.rs
@@ -852,11 +852,17 @@ async fn test_delete_for_committee_session_with_data_entry_status_as_administrat
     let addr = serve_api(pool).await;
     let admin_cookie = login(&addr, FixtureUser::Admin).await;
     let election_id = 2;
+    let polling_station_id = 211;
 
     let committee_session = get_election_committee_session(&addr, &admin_cookie, election_id).await;
     assert_eq!(committee_session["status"], "data_entry");
 
-    let response = delete_polling_station(&addr, &admin_cookie, election_id, 2).await;
+    // Assert that the polling station does exist
+    let response = get_polling_station(&addr, &admin_cookie, election_id, polling_station_id).await;
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let response =
+        delete_polling_station(&addr, &admin_cookie, election_id, polling_station_id).await;
 
     assert_eq!(
         response.status(),


backend/tests/polling_station_integration_test.rs
stacktraceghost, line 284: Some ids in tests with fixture `election_5_with_results` have not been changed here. See
- `test_get_from_previous_committee_session_fails`
- `test_update_from_previous_committee_session_fails`
- `test_delete_from_previous_committee_session_fails`

stacktraceghost:
```suggestion
    let response = delete_polling_station(&addr, &admin_cookie, election_id, 212).await;
```

stacktraceghost:
```suggestion
    let response = delete_polling_station(&addr, &coordinator_cookie, 5, 518).await;
```

oliver3: Ah I wrongly skipped the `polling_station_integration_test.rs` (also because it remained green). I've updated the IDs and added an assertion that the polling station exists, using the `/investigations` endpoint for a previous session.

Done in https://github.com/kiesraad/abacus/pull/3059/commits/645d1816788911f592c32e8b8d2a49ec97d511a0

oliver3: Also updated ID, names and added `exists()` for the `validate_and_get_committee_session()` tests in https://github.com/kiesraad/abacus/pull/3059/commits/984ef5de7a0e7fb5bf6da3263004abdb9d8b5628

oliver3: Updated in https://github.com/kiesraad/abacus/pull/3059/commits/c2b2c90f14811b5ec0946eae65fbae5091f54b92

PR #3064, opened 2026-03-19
Add `x-session-expires-at` header to login endpoint to prevent /account/login redirect

### LINKED ISSUE #3038: User redirected to /account/login after logging in
Tested on: release 1.0.1 and main [e55911f1](https://github.com/kiesraad/abacus/commit/e55911f1ef74831c0ca5345378701b26d0210be2)

Users are logged out automatically after a couple of minutes of inactivity
When trying to login after being logged out automatically, the user is redirected to /account/login instead of the page the were when they were logged out. This results in an error 'no access to this page'

Reproduction steps:
- Log in using a typist account
- Wait for 10 minutes / until Abacus logs you out automatically
- Try to login again

Observed
- user is on `/acoount/login` with 'no access to this page' message

Image

Expected
- user is returned to page they were before being logged out automatically (or /elections if that is easier to implement)

Note: I tested this with a new typist account, and was logged out automatically from `/elections#new-account`

marlonbaeten:
Fixes #3038

When the user is logged out after session expiration, they are redirected to /account/login after logging in.

This is caused by the fact that the `extend_session` middleware does not inject this header on a login request. The session expiration check in the `AuthorizationDialog` component will then redirect a user if no session expiration time is present.

This PR adds the session expiration header to the login endpoint, to fix this problem.

## How to test

On `main`:

- Login and let the session expire (possibly change the session lifetime in `backend/src/api/middleware/authentication/mod.rs` for faster testing)
- Login again and see that the user is redirected (or remains) on the login page

On `fix-3038-login-redirect`:

- Login and let the session expire
- Login again and see that the user is redirected to `/elections`


praseodym:
On this branch, logging in and then going to `/account/login` will still present a login page. I think we should also consider adding a redirect to `/` if the user somehow ends up on the login page when already logged in. Of course this could be a separate PR.

marlonbaeten:
> On this branch, logging in and then going to `/account/login` will still present a login page. I think we should also consider adding a redirect to `/` if the user somehow ends up on the login page when already logged in. Of course this could be a separate PR.

See #3073

Issue #2977, opened 2026-03-03
Use data_entry_id instead of polling_station_id for data entry urls


oliver3:
If we remove polling_station_id, where does the frontend get the polling station name & number from, and the Badge data? Maybe we should extend the claim response with this data? -> #3025 

**Coordinate data entry**

- /api/polling_stations/{polling_station_id}/data_entries 
- /api/polling_stations/{polling_station_id}/data_entries/get
- /api/polling_stations/{polling_station_id}/data_entries/resolve_differences
- /api/polling_stations/{polling_station_id}/data_entries/resolve_differences
- /api/polling_stations/{polling_station_id}/data_entries/resolve_errors

**Data entry**

- /api/polling_stations/{polling_station_id}/data_entries/{entry_number}
- /api/polling_stations/{polling_station_id}/data_entries/{entry_number}
- /api/polling_stations/{polling_station_id}/data_entries/{entry_number}/claim
- /api/polling_stations/{polling_station_id}/data_entries/{entry_number}/finalise

Also extend the election status view with links to view sub committee data entries.

oliver3:
Blocked by #2836 because I'd like to avoid merge issues for that one

PR #3065, opened 2026-03-19
Remove polling station parameter from validation

praseodym:
When implementing #2977 I noticed that we pass a polling station parameter to the data entry validation logic, which blocks implementation of data entry for sub committees. However, it turns out that the polling station parameter is not actually used during validation, so this PR removes it.

The polling station parameter was introduced in #266 to implement W.206 and W.207. Later on W.207 was removed in #1795 and W.208 was removed in #1838, without removing the polling station parameter.


praseodym:
@oliver3 @Lionqueen94 fixed some merge conflicts with #3059, please re-review

PR #3066, opened 2026-03-19
Move auth logic from RootLayout to AuthorizationGate

marlonbaeten:
The frontend logic to redirect a user to the login page (if a user is unauthorised or their session expired) is currently spread over both the `RootLayout` and the `AuthorizationDialog`, which can be confusing.

This PR moves the authorization logic to from `AuthorizationDialog` and `RootLayout` into `AuthorizationGate`, to improve the readability for a developer or a reviewer. The AuthorizationDialog` only renders the modal/dialog itself.


marlonbaeten:
This needs more work - I will open a new PR once that is done.

PR #3073, opened 2026-03-23
Move frontend authorization logic to `AuthorizationGuard`

marlonbaeten:
While fixing #3064 I noticed the authorization logic in the frontend is divided over the `RootLayout` and the `AuthorizationDialog` which makes it difficult to debug or reason about.

This PR extracts the frontend authorization and session-expiration behavior into a new `AuthorizationGuard` component, then wraps RootLayout with that guard so route access, session expiry redirects, login-page redirects, and the expiration dialog are handled in one place. `AuthorizationDialog` was simplified into a presentational component that only renders the warning modal based on props, and the related tests were rewritten/expanded to cover the new guard-based flow and updated router setup.

I added one functional change: when a user is logged in and visits the `/account/login` page they are redirected to `/elections`


frontend/src/app/AuthorizationGuard.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+          ),
+          handle: { roles: ["administrator"] },
+        },
+      ],
+    });
+
+    expect(screen.getByText("Protected content")).toBeVisible();
+  });
+
+  test("redirects an unauthorized user to the login page", async () => {
```
stacktraceghost: Lets also add a test for an authorized user with the wrong role

Everything else looks good!

stacktraceghost: Added the test case in https://github.com/kiesraad/abacus/pull/3073/commits/42dd8480671b62e6fe52661f080c202fb6bd9dfb

stacktraceghost: In response to [this comment](https://github.com/kiesraad/abacus/commit/42dd8480671b62e6fe52661f080c202fb6bd9dfb#r180397570), adjusted the code in https://github.com/kiesraad/abacus/pull/3073/commits/cbbdad72fb64d2b73a4481783002f7eaa3a2c277:
- If the current route is not public and the current role is not allowed for `handle.roles`:
  - An unauthenticated user is redirected to the login page with `{ unauthorized: true }`
   - An authenticated user is presented with an unauthorized error
   
Also hiding the back button on the forbidden page now if there is no browser history.

jschuurk-kr: This raises a question about authorization and the audit log (not in scope of this PR, though).

If as a coordinator I go to the url of a non-existing election, for example `/elections/19`, I get "Verkiezing niet gevonden" and "API fout: Niet gevonden" in the audit log. If I do the same as typist. I get "Geen toegang" and there's no entry in the audit log.

So the question is: should the audit log keep track of any attempt to access a url that leads to either a not authorized or a not found? I don't think we have an answer right now, because we haven't specified the use cases of our audit log yet.

frontend/src/app/AuthorizationDialog.test.tsx
```diff
@@ -71,36 +101,61 @@ describe("AuthorizationDialog", () => {
...
+      expect(router.state.location.pathname).toBe("/account/login");
+    });
 
     const logoutText = within(await screen.findByRole("alert")).getByRole("strong");
     expect(logoutText).toHaveTextContent("Je bent automatisch uitgelogd");
     expect(logoutText).toBeVisible();
   });
 
-  test("Redirect should happen when not authorized", async () => {
+  test("Redirect should happen when not authenticated", async () => {
```
jschuurk-kr: How is this test different from "redirects an unauthenticated user to the login page" in AuthorizationGuard.test.tsx?

frontend/src/app/AuthorizationGuard.test.tsx
stacktraceghost: The audit log does not log things that happen in the frontend, only in the backend. The backend does not know which frontend path was requested. If we would send that to the backend, someone with bad intentions can also simply block that request.

frontend/src/app/AuthorizationDialog.test.tsx
stacktraceghost: Same goes for "Redirect should happen after expiration" (AuthorizationDialog) vs "redirects to the login page when the session has expired" (AuthorizationGuard). Since the redirect logic is in the guard, I have removed the tests from dialog. 

Adjusted in https://github.com/kiesraad/abacus/pull/3073/commits/abb9e3184c3c8a267bf82f7ed99d5b910f2bf225

frontend/src/app/AuthorizationGuard.test.tsx
jschuurk-kr: I was thinking later about how you would implement that kind of thing and yeah... nope.

frontend/src/app/AuthorizationGuard.tsx
```diff
@@ -0,0 +1,110 @@
...
+ *   warning can be shown again on a future near-expiry.
+ *
+ * Rules, in evaluation order:
+ * - If the current route is not public and the current role is not allowed for `handle.roles`:
+ *    - An unauthenticated user is redirected to the login page with `{ unauthorized: true }`
+ *    - An authenticated user is presented with an unauthorized error
+ * - If the current route is not public and the computed session lifetime has expired, the user is redirected to
+ *   `/account/login` with `{ unauthorized: true }`.
+ * - If the user is already logged in and tries to open `/account/login`, the user is redirected to the same
+ *   post-login destination used by the login form: `/account/setup` for first-login users, otherwise `/elections`.
```
jschuurk-kr: The redirect to `/account/setup` does not seem to be working.

Steps to reproduce:
1. Do first login with a new account.
2. From `/account/setup` go to `/account/login`, `/account/setup` or `/elections`.
3. The user is redirected to `/account/login` with the message "Je hebt geen toegang tot deze pagina". The session cookie and the session entry in the database persist.
4. If the user tries to return to `/account/setup`, they are redirected as mentioned in the previous step.

stacktraceghost: Good find, this is also broken on main. The `/api/account` endpoint does not allow access for incomplete users. Looking into this.

stacktraceghost: This was an interesting one. 

In v1.0.1:
- `/api/whoami` was marked in the OpenAPI docs as `authenticated, all roles allowed`
- The handler used `Option<User>`, which allows everyone, no checks on authentication or complete/incomplete. 
- The handler returns `AuthenticationError::UserNotFound` which is code 401
- The `router.rs` tests succeeded, because it expected an unauthenticated user to get a 401, but in this case it was caused by the `AuthenticationError::UserNotFound`

Conclusion: the endpoint is actually public.

In main until [Authorize on route level #3028](https://github.com/kiesraad/abacus/pull/3028):
- `/api/whoami` is now `/api/account`
- The handler is still behaving the same

After merging [Authorize on route level #3028](https://github.com/kiesraad/abacus/pull/3028):
- The assumption was made that this endpoint was only for authenticated users and configured as `.routes(routes!(account).authorize(ALL_ROLES))` (= authenticated, all roles allowed, only complete users)
- The handler only uses `User` now and returns the user (see [here](https://github.com/kiesraad/abacus/pull/3028/changes#diff-986ac11c3322b36fc84f91df75ad8c7520886d27769c2b0e8fc5da7f762ead68L283))

Result: the `/api/account` endpoint now returns a 401 for incomplete users. When reloading the `/account/setup` page, that endpoint is being called, thus the user gets redirected to the login page.

Solution: 
- Make the `/api/account` endpoint public as it used to be
- Revert changes to handler and make it return  `AuthenticationError::UserNotFound` again

In this proces, also concluded that `allow_incomplete_user` was not clear enough and renamed it to `require_incomplete_user`, since it _requires_ an incomplete user and does not additionally allow it. 

Changed in https://github.com/kiesraad/abacus/pull/3073/commits/0fd3eecef7bd642aa48395eb99c3c1a83384c8dd

praseodym:
@stacktraceghost this branch has a merge conflict

frontend/src/app/AuthorizationGuard.tsx
jschuurk-kr: If I'm at `/account/setup` and change the url to `/account/login` I get redirected to account setup page. However, if I change the url to `/elections` I get redirected to the login page with the alert "Je hebt geen toegang tot deze pagina". I am still logged in, though, and can return to `/account/setup` to continue.

Can we redirect to `/account/setup` or to the "Geen toegang"-page instead?

jschuurk-kr: Am tempted to add the following e2e tests:
1. completed logged-in user goes to `/account/login` => redirected to `/account/elections`
1. incomplete logged-in user goes to `/account/login` => redirected to `/account/setup`
1. completed logged-in user goes to `/account/setup` => redirected to `/account/elections`
1. incomplete user goes to `/account/setup` => at `/account/setup`
1. incomplete logged-in user goes to `/account/elections` => redirected to `/account/setup`

@stacktraceghost Should I do that in a new branch/PR? Where we can also address test no.5, because that would fail currently?

stacktraceghost: I can add this case and test it in `AuthorizationGuard.test.tsx`, however I think it is good to indeed add e2e tests as well, since the behaviour can be different based on the API errors that can come back. Although we can also mock those, they are caught in the ErrorBoundary.

Lets indeed do this in a new branch/PR or discuss further.

jschuurk-kr: I wrote the tests against this branch, so I can create a new one once this one has been merged. (Changes are localized enough that branching off a branch is not needed.)

For test case 3 "completed logged-in user goes to /account/setup => redirected to /account/elections", it redirects to the "Geen toegang"-page, which I think is fine as well.

jschuurk-kr: Unresolving for visibility to second approver.

frontend/src/app/RootLayout.test.tsx
oliver3, line 23: These tests also should go elsewhere, as part of the separate PR?

frontend/src/app/AuthorizationDialog.tsx
```diff
@@ -9,49 +6,15 @@ import { formatTimeToGo } from "@/utils/dateTime";
...
+  setHideDialog: (hide: boolean) => void;
+}
 
-  // navigate to login page if the session has expired
-  if (sessionValidFor !== null && sessionValidFor <= 0 && !routeHandle?.public) {
-    return <Navigate to="/account/login" state={{ unauthorized: true }} />;
-  }
+export function AuthorizationDialog({ sessionValidFor, hideDialog, setHideDialog }: AuthorizationDialogProps) {
+  const { loading, extendSession } = useApiState();
+  const { user } = useApiState();
```
oliver3: Not sure if it can cause issues using the same hook twice, but let's use it once

frontend/src/app/AuthorizationDialog.tsx
oliver3: I believe that checking for user and sessionValidFor should now be handled in the Guard and reflected in the hideDialog

frontend/src/app/AuthorizationGuard.tsx
```diff
@@ -0,0 +1,99 @@
...
+        // logout after session expiration
+        if (validFor <= 0 && user !== null) {
+          setUser(null);
+        }
+
+        // reset hide dialog state if the session is valid for more than the expiration dialog time
+        if (validFor > EXPIRATION_DIALOG_SECONDS && hideDialog) {
+          setHideDialog(false);
+        }
+      };
```
oliver3: This is incorrect

oliver3: maybe we should use showDialog and not hideDialog, to try and prevent logical errors such as these

frontend/src/app/AuthorizationGuard.tsx
```diff
@@ -23,9 +25,10 @@ interface AuthorizationGuardProps {
  *   session later becomes valid again for longer than `EXPIRATION_DIALOG_SECONDS`, the hidden state is reset so the
  *   warning can be shown again on a future near-expiry.
  *
- * Redirect rules, in evaluation order:
- * - If the current route is not public and the current role is not allowed for `handle.roles`, the user is
- *   redirected to `/account/login` with `{ unauthorized: true }` (this shows a message on the login page).
+ * Rules, in evaluation order:
+ * - If the current route is not public and the current role is not allowed for `handle.roles`:
+ *    - An unauthenticated user is redirected to the login page with `{ unauthorized: true }`
+ *    - An authenticated user is presented with an unauthorized error
```
oliver3: For me this description does not add anything and adds effort to keep it in sync with the actual implementation. Shall we remove this?

frontend/src/app/AuthorizationGuard.tsx
```diff
@@ -72,14 +76,19 @@ export function AuthorizationGuard({ children }: AuthorizationGuardProps) {
...
     const route = routeMatch?.pathname || "unknown";
-    console.error(
-      `Forbidden access to route ${route}` + ` for ${user?.role ? `role ${user.role}` : "unauthenticated user"}`,
-    );
 
-    return <Navigate to="/account/login" state={{ unauthorized: true }} />;
+    // redirect to login page if not authenticated
+    if (!isAuthenticated) {
+      console.error(`Forbidden access to route ${route} for unauthenticated user`);
+      return <Navigate to="/account/login" state={{ unauthorized: true }} />;
```
oliver3, line 33: so this should be `unauthenticated: true`?

frontend/src/app/AuthorizationGuard.tsx
marlonbaeten: Why is this incorrect? This should reset the manual override "hideDialog" flag once we have an extended (or new) session. This behavior was not changed in this PR: https://github.com/kiesraad/abacus/blob/main/frontend/src/app/AuthorizationDialog.tsx#L36

frontend/src/app/AuthorizationGuard.tsx
marlonbaeten: This state param only signals that a certain notice should be shown. We do not have a notice that the user is not authenticated - if the exact semantics of authentication vs. authorization matter for the user experience we should add a new notification for this case on the login form, but preferably by a new issue / PR.

frontend/src/app/RootLayout.test.tsx
marlonbaeten: Agree, but I think this PR is large enough.

frontend/src/app/AuthorizationDialog.tsx
marlonbaeten: Fixed in e311fe4986399907973103511d4a5fa1228be26c

frontend/src/app/AuthorizationDialog.tsx
marlonbaeten: Moved to a separate hook ` useSessionExpiration`. hideDialog has to be a separate flag: it holds the state that a user explicitly dismissed the dialog.

frontend/src/app/AuthorizationGuard.tsx
marlonbaeten: I think this can help developers that have to understand this part of the code for the first time. But I guess this is the type of style choice the Abacus team has to make ✌🏻

marlonbaeten:
@oliver3 thank you for your feedback 👍🏻 I've added [e311fe4](https://github.com/kiesraad/abacus/pull/3073/commits/e311fe4986399907973103511d4a5fa1228be26c) to address some of your findings.

frontend/src/app/AuthorizationGuard.tsx
jschuurk-kr: If I have two tabs open as the same user, the dialog appears in both, but extending the session does not seem to work. I need to do some more testing to get a clearer picture of the behavior.

marlonbaeten: If we want synchronisation between tabs, I think we need websockets or polling...

jschuurk-kr: Behavior on main is the same.

I think this then could be added to the known issues: if you have multiple tabs open, extending you session trough the dialog does not work. Luckily, you are logged out in both tabs when your session expires, so it's annoying for users, but it's not that sessions are valid beyond expiration.

stacktraceghost: > If we want synchronisation between tabs, I think we need websockets or polling...

[Broadcast Channel API](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) could be used for that.

praseodym: I think we can create an issue for this that we can fix some time in the future

frontend/src/app/ExpirationDialog.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { t, tx } from "@/i18n/translate";
+import { formatTimeToGo } from "@/utils/dateTime";
+
+interface AuthorizationDialogProps {
+  sessionValidFor: number;
+  onClose: () => void;
+  onStayLoggedIn: () => void;
+}
+
+export function AuthorizationDialog({ sessionValidFor, onClose, onStayLoggedIn }: AuthorizationDialogProps) {
```
Lionqueen94: Please also rename the function and props to ExpirationDialog

frontend/src/app/AuthorizationGuard.tsx
jschuurk-kr: issue created: [#3101 Fix extending session through dialog when having multiple tabs open](https://github.com/kiesraad/abacus/issues/3101)

frontend/src/app/ExpirationDialog.tsx
jschuurk-kr: done in 7b65d2d6

frontend/src/app/AuthorizationGuard.tsx
jschuurk-kr: We set `unauthorized` to `true` in four cases:
1. `!isAuthenticated` in the `AuthorizationGuard`
2. `sessionValidFor !== null && sessionValidFor <= 0 && !isPublic` in the `AuthorizationGuard`
3. `error instanceof ApiError && error.code === 401` in `ErrorBoundary`
4. `!user` in `AccountSetupPage`

So 1, 2, and 4 are cases of not authenticated and only 3 is an example of not authorized. Which makes me wonder if there are more places in the code where we are not making a clear distinction.

In any case, my proposal would be to leave this as-is for the scope of this PR.

frontend/src/app/AuthorizationGuard.tsx
jschuurk-kr: After the changes in e311fe4, the comment is no longer accurate. For example, the session expiration logic now lives in `useSessionExpiration.ts`.

Based on the fact that the code is understandable to me without the description, I'd argue we can remove it.

frontend/src/app/AuthorizationGuard.tsx
jschuurk-kr: https://github.com/kiesraad/abacus/pull/3117

PR #3076, opened 2026-03-23
Use data entry id in data entry API

### LINKED ISSUE #2977: Use data_entry_id instead of polling_station_id for data entry urls

praseodym:
Use data entry id instead of polling station id in data entry API, resolves #2977. This allows data entry of sub committees instead of just polling stations. The data entry API logic does have to look up the source of the data entry (either a polling station or sub committee) to check the committee session state and retrieve a possible previous data entry for validation.

The frontend is mostly unchanged and looks up the data entry id from the polling station id through the election status API.

### DIFF
--- a/backend/src/repository/data_entry_repo.rs
+++ b/backend/src/repository/data_entry_repo.rs
@@ -170,7 +170,8 @@ pub async fn resolve_source(
 ) -> Result<DataEntrySourceContext, ResolveSourceError> {
     let mut tx = conn.begin().await?;
 
-    let row = query!(
+    let row = query_as!(
+        ResolveSourceRow,
         r#"
         SELECT
             p.id AS "polling_station_id?: PollingStationId",
@@ -194,22 +195,7 @@ pub async fn resolve_source(
     .await?;
 
     // Extract the source and committee_session_id, validating that exactly one matched
-    let (source, committee_session_id) = resolve_source_from_row(
-        data_entry_id,
-        ResolveSourceRow {
-            polling_station_id: row.polling_station_id,
-            ps_committee_session_id: row.ps_committee_session_id,
-            ps_prev_data_entry_id: row.ps_prev_data_entry_id,
-            ps_name: row.ps_name,
-            ps_number: row.ps_number,
-            sub_committee_id: row.sub_committee_id,
-            sc_number: row.sc_number,
-            sc_name: row.sc_name,
-            sc_category: row.sc_category,
-            sc_committee_session_id: row.sc_committee_session_id,
-        },
-    )?;
-
+    let (source, committee_session_id) = resolve_source_from_row(data_entry_id, row)?;
     let committee_session = committee_session_repo::get(&mut tx, committee_session_id).await?;
     let election = election_repo::get(&mut tx, committee_session.election_id).await?;
 


frontend/e2e-tests/fixtures.ts
```diff
@@ -92,10 +92,11 @@ export const test = base.extend<Fixtures>({
...
-  pollingStationFirstEntryClaimed: async ({ typistOne, pollingStation }, use) => {
+  pollingStationFirstEntryClaimed: async ({ typistOne, pollingStation, election }, use) => {
     const { request } = typistOne;
+    const dataEntryId = await resolveDataEntryId(request, election.election.id, pollingStation.id);
 
-    const firstDataEntry = new DataEntryApiClient(request, pollingStation.id, 1);
+    const firstDataEntry = new DataEntryApiClient(request, dataEntryId, 1);
     await firstDataEntry.claim();
     await use(pollingStation);
   },
```
oliver3, line 102:
```suggestion
  pollingStationFirstEntryClaimed: async ({ typistOne, pollingStation }, use) => {
    const { request } = typistOne;

    const firstDataEntry = new DataEntryApiClient(request, pollingStation.data_entry_id!, 1);
    await firstDataEntry.claim();
    await use(pollingStation);
  },
```

No need to retrieve the dataEntryId here (and on other locations) with `resolveDataEntryId()` because the polling station contains its data_entry_id.

backend/src/repository/data_entry_repo.rs
```diff
@@ -115,65 +158,77 @@ pub async fn update(
...
+pub async fn resolve_source(
     conn: &mut SqliteConnection,
-    polling_station_id: PollingStationId,
-) -> Result<Option<PollingStationDataEntry>, sqlx::Error> {
+    data_entry_id: DataEntryId,
+) -> Result<DataEntrySourceContext, ResolveSourceError> {
     let mut tx = conn.begin().await?;
 
-    // Get data_entry_id from polling station
     let row = query!(
```
oliver3: Why not 
```rust
    let row = query_as!(
        ResolveSourceRow,
```

instead of manually creating ResolveSourceRow  below?

backend/tests/data_entries_integration_test.rs
```diff
@@ -23,10 +23,18 @@ pub mod utils;
 async fn test_data_entry_valid(pool: SqlitePool) {
     let addr = serve_api(pool).await;
     let typist_cookie = login(&addr, TypistGSB).await;
+    let data_entry_id = get_data_entry_id(&addr, &typist_cookie, 2, 211).await;
```
oliver3, line 26: I would prefer just using the known data_entry_ids from the fixtures.

praseodym: Created #3080 to fix this after this PR

frontend/e2e-tests/fixtures.ts
praseodym: Created #3080 to fix this after this PR

backend/src/repository/data_entry_repo.rs
praseodym: `ResolveSourceRow` was added when I split up `resolve_source` to smaller functions, and I forgot to use `query_as!` instead. Fixed in 7487cd615da100b6cb7703f93cd134788caf34ff.

oliver3: Thanks for the explanation!

PR #3075, opened 2026-03-23
Rename PollingStationResults to Results

stacktraceghost:
Related to #2998 

In preparation of the linked issue, renamed `PollingStationResults` to `Results`. This enum will soon contain `GSB(GSBResults)` as well.

PR #3081, opened 2026-03-25
Use data entry id instead of polling station id for data entry and coordinator frontend

oliver3:
- Revert temporary frontend code added in https://github.com/kiesraad/abacus/pull/3076 to convert polling station id to data entry id
- Add data entry source to API responses for coordinator frontend (data entry detail, resolve errors, resolve differences)
- Use `dataEntryId` instead of `pollingStationId` in routes, components, hooks etc for `data_entry`, `data_entry_detail` and `resolve_differences`
- Use `dataEntryId` in links to these features (data entry home, election status) 

NB: We cannot show the polling station number in "data entry not possible" messages because it cannot be returned in the claim error response.

To test, check dat the data entry id is used in the browser url and that it corresponds with the backend calls for the following:
- typist, data entry
- coordinator, view data entry details
- coordinator, resolve errors
- coordinator, resolve differences


praseodym:
@oliver3 this branch has conflicts after merging #3075

PR #3085, opened 2026-03-25
Create data entry structure 

### LINKED ISSUE #2998: Create data entry structure
Backend (new variant) and frontend (dataEntryStructure).

New section for Z Kiesgerechtigden, with appropriate "required" error.

stacktraceghost:
Resolves #2998 

This PR adds the new data entry structure for CSB. The differences with the `CSO[First/NextSession]Results` versus `GSBResults` are:
1. Field Z "Kiesgerechtigden" 
2. No checkboxes on the differences page

As a consequence, the structs have been adjusted like this:
- Added `GSB(GSBResults)` to `enum Results` 
- `GSBResults`
  - Has field `pub number_of_voters: Count`
  - Has its own struct for `pub differences_counts: GSBDifferencesCounts`

Due to these differences, the `impl`'s have been adjusted and split up. A good place to start is by taking a look at `backend/src/domain/results/mod.rs`.

Validation rules have been copied for the structs. Ones that were using fields that are not available in the GSB model, have been removed. The validation rules still need to be checked and adjusted later.

The role authorization for the data entry handlers have already been adjusted in this PR. This way you can go through the entire data entry flow for CSB on the test environment. 

The frontend `frontend/src/utils/dataEntryStructure.ts` supports the new model as well. I've added some type guards and split up functions due to line amount restrictions. 

I've also added snapshots per model. Every time the frontend tests run, the output of every model in `dataEntryStructure.ts` will be compared to the existing snapshot. If these differ, you will get an error and you can update the snapshots. This will make it easy to spot unintended changes during development and reviewing. In this PR for example, I've added the snapshot for the CSO models in the first commit. Since there are no changes in the second commit, you can assume this model is unchanged in the frontend. 

Extra things to check:
- Docstrings at new structs
- Docstrings of `enum Results` and underlying model structs (including links)

### DIFF
--- a/backend/src/api/data_entry.rs
+++ b/backend/src/api/data_entry.rs
@@ -335,7 +335,7 @@ async fn data_entry_claim(
     // If there is a previous data entry with status definitive
     let previous_results: Option<CommonPollingStationResults> = match prev_data_entry_id {
         Some(prev_id) => match data_entry_repo::get_status(&mut tx, prev_id).await? {
-            // Match polling station results onl and get the common results
+            // Match polling station results only and get the common results
             DataEntryStatus::Definitive(d) => match d.results {
                 Results::CSOFirstSession(r) => Some(r.as_common()),
                 Results::CSONextSession(r) => Some(r.as_common()),


backend/src/api/data_entry.rs
```diff
@@ -928,7 +928,7 @@ pub struct ElectionStatusResponse {
 pub struct ElectionStatusResponseEntry {
     /// Data entry id
     pub data_entry_id: DataEntryId,
-    /// Data entry source (polling station or sub committee)
+    /// Data entry source (polling station or subcommittee)
```
praseodym: I'm sorry but why do we have to change comments like this in a 'Fixed failing test' commit?

Lionqueen94: I amended the previous commit, and removed this change. The IDE suggested fixing the spelling error. I'll not do this in the future in the same commit.

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -0,0 +1,1111 @@
...
+        let validation_results = validate(data, 52, 72)?;
+
+        assert!(validation_results.errors.is_empty());
+
+        Ok(())
+    }
+
+    /// CSO | F.306 (Als H > D) I <> H - D
+    #[test]
+    fn test_f306_votes_equals_voters() -> Result<(), DataError> {
```
Lionqueen94, line 695: The validation error you're testing for is the F.305, but the comment and title say it tests for F.306

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -0,0 +1,1111 @@
...
+                context: None,
+            }]
+        );
+
+        Ok(())
+    }
+
+    /// CSO | F.306 (Als H > D) I <> H - D
+    #[test]
+    fn test_f306_votes_smaller_than_voters() -> Result<(), DataError> {
```
Lionqueen94, line 731: The validation errors you're testing for are the F.308 and F.309, but the comment and title say it tests for F.306

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -0,0 +1,1111 @@
...
+                context: None,
+            }]
+        );
+
+        Ok(())
+    }
+
+    /// CSO | F.307 (Als H > D) J is ingevuld
+    #[test]
+    fn test_f307_votes_greater_than_voters_fewer_ballots_count_zero() -> Result<(), DataError> {
```
Lionqueen94, line 823: The validation error you're testing for is the F.306, but the comment and title say it tests for F.307

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -0,0 +1,1111 @@
...
+                }
+            ]
+        );
+
+        Ok(())
+    }
+
+    /// CSO | F.307 (Als H > D) J is ingevuld
+    #[test]
+    fn test_f307_votes_equals_voters() -> Result<(), DataError> {
```
Lionqueen94, line 897: The validation error you're testing for is the F.305, but the comment and title say it tests for F.307

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -0,0 +1,1111 @@
...
+                context: None,
+            }]
+        );
+
+        Ok(())
+    }
+
+    /// CSO | F.307 (Als H > D) J is ingevuld
+    #[test]
+    fn test_f307_votes_smaller_than_voters() -> Result<(), DataError> {
```
Lionqueen94, line 918: The validation error you're testing for is the F.308, but the comment and title say it tests for F.307

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -0,0 +1,1111 @@
...
+                context: None,
+            }]
+        );
+
+        Ok(())
+    }
+
+    /// CSO | F.308 (Als H < D) J <> D - H
+    #[test]
+    fn test_f308_votes_equals_voters() -> Result<(), DataError> {
```
Lionqueen94, line 983: The validation error you're testing for is the F.305, but the comment and title say it tests for F.308

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -0,0 +1,1111 @@
...
+                }
+            ]
+        );
+
+        Ok(())
+    }
+
+    /// CSO | F.309 (Als H < D) I is ingevuld
+    #[test]
+    fn test_f309_votes_equals_voters() -> Result<(), DataError> {
```
Lionqueen94, line 1049: The validation error you're testing for is the F.305, but the comment and title say it tests for F.309

backend/src/domain/results/gsb_results.rs
```diff
@@ -0,0 +1,563 @@
...
+    votes_counts::VotesCounts,
+};
+use crate::domain::election::ElectionWithPoliticalGroups;
+use crate::domain::results::common_validation::difference_admitted_voters_count_and_votes_cast_count_above_threshold;
+use crate::domain::results::gsb_differences_counts::validate_gsb_differences_counts;
+use crate::domain::validate::{
+    DataError, Validate, ValidationResult, ValidationResultCode, ValidationResultContext,
+    ValidationResults,
+};
+use crate::domain::{compare::Compare, field_path::FieldPath};
```
Lionqueen94:
```suggestion
use super::{
    common_validation::difference_admitted_voters_count_and_votes_cast_count_above_threshold,
    count::Count,
    gsb_differences_counts::{validate_gsb_differences_counts, GSBDifferencesCounts},
    political_group_candidate_votes::PoliticalGroupCandidateVotes,
    voters_counts::VotersCounts,
    votes_counts::VotesCounts,
};
use crate::domain::{
    compare::Compare,
    election::ElectionWithPoliticalGroups,
    field_path::FieldPath,
    validate::{
        DataError, Validate, ValidationResult, ValidationResultCode, ValidationResultContext,
        ValidationResults,
    },
};
```

backend/src/domain/results/gsb_differences_counts.rs
stacktraceghost: These tests have been copied from `differences_counts.rs` so we will need to fix these issues over there too.

Edit: ah actually, I have removed rules which cannot be executed on this struct, because some fields are not there anymore. In that process, I've probably only removed the asserts instead of the tests as well. I will check this out. 

The validation rules are not final yet and still need work.

stacktraceghost: Seems like some tests are mislabeled or actually testing other rules. For example this one, like you said. This is also the case for `differences_count.rs`, so I would like to propose to fix these issues in a separate issue/PR. 

Edit: after looking at it a bit longer. I think most of these tests are simply checking the combinations. For example this test is checking if it doesn't trigger when `H = D` instead of `H > D`. This will of course trigger some other validations along the way. If we remove these tests, gaps might start to exist or they are only implicitly tested through other tests.

I think we should leave it as-is, unless we are missing cases.

backend/src/domain/results/gsb_results.rs
stacktraceghost: Thanks! Accepted this change in https://github.com/kiesraad/abacus/pull/3085/commits/dddc3a28cbebf60a3fc41ae51e64e7836fb0b281

Edit: almost, one of the imports was not sorted. Fixed in https://github.com/kiesraad/abacus/pull/3085/commits/86e4910810e161ce9d8689837c4400f1b687371c

backend/src/domain/results/gsb_differences_counts.rs
stacktraceghost: I've tried to consolidate all these combinations in https://github.com/kiesraad/abacus/pull/3112.

backend/src/api/data_entry.rs
```diff
@@ -325,16 +332,23 @@ async fn data_entry_claim(
         _ => None,
     };
 
-    let previous_results: Option<Results> = match prev_data_entry_id {
+    // If there is a previous data entry with status definitive
+    let previous_results: Option<CommonPollingStationResults> = match prev_data_entry_id {
         Some(prev_id) => match data_entry_repo::get_status(&mut tx, prev_id).await? {
-            DataEntryStatus::Definitive(d) => Some(d.results),
+            // Match polling station results onl and get the common results
```
praseodym:
```suggestion
            // Match polling station results only and get the common results
```

PR #3086, opened 2026-03-25
Use data entry id in data entry API tests

praseodym:
Follow up on #3083 to also use data entry id instead of polling station id in the data entry API tests, as requested by @stacktraceghost. Two tests that created a empty data entry on a polling station without (concluded with new results) investigation have been removed, as this should never occur through normal code paths.

### DIFF
--- a/backend/src/api/data_entry.rs
+++ b/backend/src/api/data_entry.rs
@@ -2481,14 +2481,11 @@ mod tests {
 
         #[test(sqlx::test(fixtures(path = "../../fixtures", scripts("election_2"))))]
         async fn test_status_first_entry_in_progress(pool: SqlitePool) {
-            claim(
-                pool.clone(),
-                DataEntryId::from(201),
-                EntryNumber::FirstEntry,
-            )
-            .await;
+            let data_entry_id = DataEntryId::from(201);
 
-            let result = call_data_entry_get(pool.clone(), DataEntryId::from(201))
+            claim(pool.clone(), data_entry_id, EntryNumber::FirstEntry).await;
+
+            let result = call_data_entry_get(pool.clone(), data_entry_id)
                 .await
                 .unwrap();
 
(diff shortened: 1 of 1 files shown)


backend/src/api/data_entry.rs
```diff
@@ -1573,33 +1558,46 @@ mod tests {
     #[test(sqlx::test(fixtures(path = "../../fixtures", scripts("election_2"))))]
     async fn test_finalise_data_entry(pool: SqlitePool) {
         let request_body = example_data_entry();
-        let polling_station_id = PollingStationId::from(211);
 
-        let response = claim(pool.clone(), polling_station_id, EntryNumber::FirstEntry).await;
+        let response = claim(
+            pool.clone(),
+            DataEntryId::from(201),
```
Lionqueen94: It might be nice to add the data_entry_id as a variable (how it was with the polling_station_id), to make the function calls a little shorter, so some of them will fit on one line again as well. Same for the other tests.

backend/src/api/data_entry.rs
```diff
@@ -1350,43 +1340,6 @@ mod tests {
         assert_eq!(data_entry.state.0, DataEntryStatus::Empty);
     }
 
-    #[test(sqlx::test(fixtures(path = "../../fixtures", scripts("election_7_four_sessions"))))]
-    async fn test_claim_data_entry_next_session_err_no_investigation(pool: SqlitePool) {
```
stacktraceghost, line 1354: It is possible to call the API on data entries in a next session, which have no investigation or no corrected results. So checking if an error is given in that case is a valid test.

Since that validation happens in `validate_and_get_data`, we can also test that function. But that wouldn't verify if that function is being called in the handler.

praseodym: That would be a slightly different test: trying to claim a data entry id from a previous session. I'll add that.

If the polling station from the current session doesn't have an investigation or there are no corrected results, there is also no data entry id to be claimed.

stacktraceghost: Ah of course! Perfect, the design of the code already covers it.

praseodym: Done in 9693357042f6c31095c67c2042b939294db49399

backend/src/api/data_entry.rs
praseodym: Not sure if this was really worth the effort but done in 4f4dcd82523d2dec78497922cc42fd489f387663

Issue #2972, opened 2026-03-03
Add backend code to fill P 22-2 PDF


Lionqueen94:
- [x] Add endpoint for downloading zip with P 22-2 (EML 520 will be added later)
  - [x] Zip file name: `vaststelling-uitslag_gr2026_gemeentenaam_gemeente_gemeente-naam-jjjjmmdd-uummss.zip`
    - [x] PDF file name: `Model_P22-2.pdf`
  - [x] Save PDF with `CsbResultsPdf` file type 
- [x] Enrich `candidate_nomination` in backend and update P 22-2 inputs and template accordingly
  - [x] Add list seat number for each chosen candidate
  - [x] Add candidate details for each chosen candidate
  - [x] Add original candidate list to `updated_candidate_ranking` field if empty, so that `election.political_groups` can be removed later

Apportionment tables will be added in a separate issue #3205

PR #3093, opened 2026-03-26
Complete P 22-2 template in Typst

### LINKED ISSUE #2975: Create P 22-2 PDF Deel 2 in Typst
- [x] Make sure country code is rendered after locality when provided:
Image

- [x] Render blanco lijst correctly (add formatting function)
- [x] Only render section 2.9 for the lists that got any seat
- [x] Add result changes to input for pdf
- [ ] Do not perform dynamic lookup of political group

Lionqueen94:
Closes #2975 

## Description
Please check the commits for the different changes that have been implemented and check the pdf diff for the resulting changes in the other models as well.

Other model changes:
- Removed hardcoded "Gemeenteraad" on title page of Na 31-2
- Added "-" between election name and date on title page
- Adjust width of list and checkbox content to the same as paragraph content
- Fixed list rendering in Na 31-2 Bijlage 1 table, to use more width than the now default 75%
- Removed rendering of section numbers for signature chapters
- Updated rendering of date box for signature chapters to match new models

## Request
I would like a review specifically on the highest averages table slicing of the steps (this might be able to be done shorter/easier?)
And I need some assistance changing the font weight for the selected average, since that somehow does not work at the moment, hence the one TODO left in the document.
EDIT: This will be done in issue #2972 - please review rendering of PDFs now

Screenshot of split up highest averages table (when > 6 residual seats, also showing behaviour with long list name):
image
Currently it is rendered by OSV like [this](https://www.utrecht.nl/fileadmin/uploads/documenten/bestuur-en-organisatie/verkiezingen/processen-verbaal-gemeenteraad/Procesverbaal-P22-2-zonder-handtekening-GR26.pdf) (see page 9)
(to test this yourself, add some extra residual seat assignments to `inputs/extra-model-p-22-2-variations/gte-19-seats-and-p9.json` and generate the pdf with that input)

### DIFF
--- a/backend/templates/common/scripts.typ
+++ b/backend/templates/common/scripts.typ
@@ -746,6 +746,7 @@
         top: if y > 0 { 0.5pt + gray },
       ),
       inset: (x: 4pt, y: 8pt),
+      fill: (_, y) => if y > 1 and calc.even(y) { luma(245) },
       table.header(
         table.cell(rowspan: 2, stroke: none, header_text([Lijst])),
         table.cell(align: center, colspan: slice.len(), header_text([Restzetel])),
--- a/backend/templates/model-p-22-2.typ
+++ b/backend/templates/model-p-22-2.typ
@@ -304,6 +304,7 @@ Na toewijzing van de volle zetels blijft een aantal te verdelen zetels over. Dit
         top: if y > 0 { 0.5pt + gray },
       ),
       inset: (x: 4pt, y: 8pt),
+      fill: (_, y) => if y > 1 and calc.even(y) { luma(245) },
       table.header(
         table.cell(stroke: none, header_text([Lijst])),
         table.cell(header_text([Aantal volle zetels])),


Lionqueen94:
> I think this is a good intermediate step, but I think we should try and get rid of as much of the filter and find calls that we have in here. As for the todo in "marking of selected average does not work", I think this is something we should leave for now, as I believe the logic here is not that trivial, and will have to be re-implemented in the Rust backend code anyway.

Exactly, this will be improved/fixed in https://github.com/kiesraad/abacus/issues/2972

backend/src/domain/models/model_p_22_2.rs
```diff
@@ -5,17 +5,41 @@ use crate::domain::{
...
+pub struct LimitedElectionSummary {
+    /// The total number of voters
+    pub voters_counts: VotersCounts,
+    /// The total number of votes
+    pub votes_counts: VotesCounts,
+    /// The differences between voters and votes
+    pub differences_counts: SummaryDifferencesCounts,
+}
+
+/// Contains the result changes that have occurred in the apportionment
```
lama-d3v:
```suggestion
/// Contains the result changes that have occurred in the apportionment.
```

Lionqueen94: Processed all review comments in d11f7e785a58eef3b59917dbb6118806efd03bd7. Will send an email to the model team with the double negative suggestion and the missing `.`
Edit: e-mail sent!

backend/templates/common/scripts.typ
```diff
@@ -384,6 +417,34 @@
...
+  let prefix = "";
+  if (with_prefix == "with_list_prefix") {
+    prefix += "Lijst " + str(election_political_group.number) + " ";
+  } else if (with_prefix == "only_list_number") {
+    prefix += str(election_political_group.number) + " ";
+  }
+
+  if (election_political_group.name == "") {
+    if (election_political_group.candidates.first() != none) {
+      name = prefix + "Blanco (" + candidate_name(election_political_group.candidates.first(), with_first_name: false, with_gender: false) + ")";
```
lama-d3v, line 437: Does this needs to be changed when https://github.com/kiesraad/abacus/issues/3095 is done?

backend/templates/model-n-10-2.typ
```diff
@@ -257,15 +258,15 @@ Zo komt het handtekeningen-blad altijd op een losse pagina, ook als het verslag
...
+#textbox_only_bottom_stroke[Datum en tijd:][Plaats:]
 
 == Verplicht: voorzitter en twee leden van het stembureau
 
 === Voorzitter van het stembureau:
 
 #textbox[Naam:][Handtekening:]
 
-=== Twee leden van het stembureau:
+=== 2 leden van het stembureau:
```
lama-d3v, line 269: I believe this was already recently changed to `Twee` instead of `2`. What's the reason for changing this again?

backend/templates/common/scripts.typ
Lionqueen94: Yes, this can be removed then :)

backend/templates/model-n-10-2.typ
Lionqueen94: In the models it says the numbers, not the texts, for example in:
[Model+Na+14-2+Corrigendum+bij+het+proces-verbaal+van+een+gemeentelijk+stembureau.pdf](https://github.com/user-attachments/files/26388422/Model%2BNa%2B14-2%2BCorrigendum%2Bbij%2Bhet%2Bproces-verbaal%2Bvan%2Been%2Bgemeentelijk%2Bstembureau.pdf)
It's with letters in the bold heading and with numbers in the box title

backend/templates/common/scripts.typ
Lionqueen94: I added it to the issue description :+1:

oliver3:
- Model does not have paragraph numbers for: 1.1.1 1.2.1 1.3.1 
- 2.7.2 Verdeling van de restzetels is missing zebra stripes in the table (maybe check 2.7.3 too, I had only 1 row..)

Lionqueen94:
> * Model does not have paragraph numbers for: 1.1.1 1.2.1 1.3.1
> * 2.7.2 Verdeling van de restzetels is missing zebra stripes in the table (maybe check 2.7.3 too, I had only 1 row..)

Fixed in b64f277d628e5794adf4bcdd38365e285c61e81b and eca893947b0ac20fe566952c8614da17fcad0f53 :smile:

PR #3099, opened 2026-03-27
Fix data entry id fields in polling station structs

### LINKED ISSUE #3079: Fix data entry ids in polling station structs
In #2881 we specified that `PollingStationFirstSession` should have a required data entry id, which is implemented as optional right now. `PollingStationNextSession` should have a data entry id only in as part of the investigation variant, but it currently also has an optional data entry id field in the struct itself. We should fix this to match the original specification.

praseodym:
Update polling structs to match the specification from #2881. Resolves #3079.

- Require data entry id in first session PS
 `PollingStationFirstSession` had an optional data entry id, even though it is required. Polling station creation functions now return an id of the created polling station in the database, after which the caller can create link the data entry and create the `PollingStationFirstSession` struct.
-  Remove data entry id from next session PS
  `PollingStationNextSession` already has a data entry id in the concluded investigation enum variant. The `create_empty` data entry service function has been replaced by the lower-level `ensure_data_entry` polling station repo function.

### DIFF
--- a/backend/src/repository/polling_station_repo.rs
+++ b/backend/src/repository/polling_station_repo.rs
@@ -139,6 +139,18 @@ impl From<PollingStationRow> for PollingStationForSession {
     }
 }
 
+#[derive(Debug)]
+pub enum CreateDataEntryError {
+    Sqlx(sqlx::Error),
+    DataEntryAlreadyLinked,
+}
+
+impl From<sqlx::Error> for CreateDataEntryError {
+    fn from(err: sqlx::Error) -> Self {
+        Self::Sqlx(err)
+    }
+}
+
 /// Returns the committee category for the related election
 pub async fn get_committee_category(
     conn: &mut SqliteConnection,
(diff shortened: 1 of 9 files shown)


backend/src/domain/polling_station.rs
```diff
@@ -236,8 +238,11 @@ impl PollingStationForSession {
 
     pub fn data_entry_id(&self) -> Option<DataEntryId> {
```
oliver3: This Option<> can now be removed as well?

backend/src/repository/polling_station_repo.rs
```diff
@@ -286,12 +286,13 @@ pub async fn get_for_election(
     .map(PollingStationForSession::from)
 }
 
-/// Create a single polling station for an election
+/// Create a single polling station for an election, returning its id.
+/// The caller is responsible for linking a data entry afterwards.
 pub async fn create(
```
oliver3, line 303: Why only return the id instead of the struct?
I think in your second commit you again always need to get the full polling station, so this create can go back to returning that?

backend/src/repository/polling_station_repo.rs
```diff
@@ -346,17 +333,17 @@ pub async fn create(
...
     tx.commit().await?;
-    Ok(PollingStationForSession::from(res))
+    Ok(id)
 }
 
-/// Create many polling stations for an election
-#[allow(clippy::too_many_lines)]
+/// Create many polling stations for an election, returning their ids.
+/// The caller is responsible for creating and linking data entries afterwards.
 pub async fn create_many(
```
oliver3, line 353: Same comment as for `create()`

backend/src/repository/polling_station_repo.rs
```diff
@@ -511,6 +483,29 @@ pub async fn delete(
     Ok(rows_affected > 0)
 }
 
+/// Returns the existing `data_entry_id` for the polling station,
+/// or creates a new empty data entry, links it, and returns its id.
+pub async fn ensure_data_entry(
```
oliver3: I'm having some doubts about this this function that I believe serves two purposes: 
- getting the data entry that should be there, 
- creating the data entry when either creating a polling station or changing the investigation to conclude with new results.

I think it will be more clear if we split these two things up, because then you don't have to think at the call site about what this function will do.

backend/src/repository/polling_station_repo.rs
```diff
@@ -654,15 +649,15 @@ pub async fn list_first_session_with_status(
     query!(
         r#"
         SELECT
-            de.id AS "data_entry_id: DataEntryId",
+            de.id AS "data_entry_id!: DataEntryId",
```
oliver3: I don't think this `!` is necessary?

backend/src/repository/polling_station_repo.rs
praseodym: Because we can't create a `PollingStationFirstSession` struct without a linked data entry

backend/src/repository/polling_station_repo.rs
praseodym: Indeed, removed in 9e4e67c6b330caa6a9af7735f4d9e7297973b212

backend/src/repository/polling_station_repo.rs
oliver3: Ah, didn't realise that, thanks.

backend/src/repository/polling_station_repo.rs
praseodym: Split up in 51fdb2f9c9afb1efc7d0d54db6ae28d7176aa66b

PR #3100, opened 2026-03-30
Added modal when resuming data entry

### LINKED ISSUE #3041: Add modal when resuming data entry for completed committee session
When resuming data entry for an already completed committee session, the documents for that session will be deleted from Abacus and can not be downloaded anymore. Currently, we do not warn the user that their action is destructive. We should.

The modal has been added to [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=19481-50267&t=ABrJT2eT38DGvil4-11).

lama-d3v:
- Fixes #3041 

[Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=19481-50267&t=ABrJT2eT38DGvil4-11)

For GSB:
- Visible on latest committee session when committee session is completed. After clicking "resume data entry" in the model, the status is updated from "completed" to "data_entry", therefor enabling data entry again.
The button (and model) aren't visible when there're multiple committee sessions and when you're viewing the first committee session.

Since there is only one committee session for CSB, it will be always shown on this election result page.

### DIFF
--- a/frontend/src/features/election_management/components/report/ElectionReportPage.tsx
+++ b/frontend/src/features/election_management/components/report/ElectionReportPage.tsx
@@ -1,7 +1,6 @@
 import { useState } from "react";
 import { Navigate } from "react-router";
 import { ApplicationError, NotFoundError } from "@/api/ApiResult";
-import { useCrud } from "@/api/useCrud";
 import { Footer } from "@/components/footer/Footer";
 import { IconCheckVerified } from "@/components/generated/icons";
 import { PageTitle } from "@/components/page_title/PageTitle";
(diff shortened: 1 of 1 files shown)


stacktraceghost:
I think the resume button should be available for CSB as well. Even though there are no next committee sessions, you can still finish the data entry. I think you should be able to resume it as well.

Lionqueen94:
> I think the resume button should be available for CSB as well. Even though there are no next committee sessions, you can still finish the data entry. I think you should be able to resume it as well.

It does show on the design for CSB: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=19158-8513&m=dev

And the button shows for CSB as well, I think the comment in the description regarding CSB needs to be removed ;)

stacktraceghost:
> > I think the resume button should be available for CSB as well. Even though there are no next committee sessions, you can still finish the data entry. I think you should be able to resume it as well.
> 
> It does show on the design for CSB: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=19158-8513&m=dev
> 
> And the button shows for CSB as well, I think the comment in the description regarding CSB needs to be removed ;)

Resolved, turned out to be a mistake in the PR description. The code does show the button indeed. :)

frontend/src/features/election_management/components/report/ElectionReportPage.tsx
```diff
@@ -9,36 +9,37 @@ import { Button } from "@/components/ui/Button/Button";
...
   const committeeSessionId = useNumericParam("committeeSessionId");
   const committeeSession = committeeSessions.find((session) => session.id === committeeSessionId);
   if (!committeeSession) {
     throw new NotFoundError("error.not_found");
   }
+  const [showModal, setShowModal] = useState(false);
 
   const updatePath: COMMITTEE_SESSION_STATUS_CHANGE_REQUEST_PATH = `/api/elections/${committeeSession.election_id}/committee_sessions/${committeeSession.id}/status`;
-  const { update, isLoading } = useCrud({ updatePath, throwAllErrors: true });
+  const { isLoading } = useCrud({ updatePath, throwAllErrors: true });
```
stacktraceghost: This can be removed. `useCrud` sets the `requestState` to `{ status: "idle" }` and only defines the CRUD-operations which we do not use here, so `isLoading` will always be `false`.

Lionqueen94: Good find! Fixed in 38a763545a4a3aa713e33d645c785297ff33df47

Issue #2997, opened 2026-03-05
Update data entry homepage for sub-committees


oliver3:
Only show list of sub-committees for CSB.

[Figma design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17435-8885&m=dev)

Verify that the election progress status bar works as expected.

<<insert design link here>>

PR #3111, opened 2026-03-30
Show data entry picker only for GSB

oliver3:
- Implements https://github.com/kiesraad/abacus/issues/2997
- In the test, let `renderDataEntryHomePage()` set the correct election response based on GSB or CSB, and wait for the correct title to be show
- Add tests to check for the picker plus collapsed list to be shown only for GSB

### DIFF
--- a/frontend/src/features/data_entry_home/components/DataEntryHomePage.test.tsx
+++ b/frontend/src/features/data_entry_home/components/DataEntryHomePage.test.tsx
@@ -18,11 +18,12 @@ import { ElectionListRequestHandler, ElectionRequestHandler } from "@/testing/ap
 import { server } from "@/testing/server";
 import { renderReturningRouter, screen, within } from "@/testing/test-utils";
 import { getTypistUser } from "@/testing/user-mock-data";
-import type {
-  CommitteeCategory,
-  CommitteeSession,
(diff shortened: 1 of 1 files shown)


frontend/src/features/data_entry_home/components/DataEntryHomePage.test.tsx
```diff
@@ -36,22 +61,10 @@ describe("DataEntryHomePage", () => {
...
-      await screen.findByRole("heading", {
-        level: 1,
-        name: electionDetailsMockResponse.election.name,
-      }),
-    ).toBeVisible();
-  });
-
-  test("Alert not visible when unfinished", async () => {
-    renderDataEntryHomePage();
+  test.each(["GSB", "CSB"] as const)("Alert not visible when unfinished (%s)", async (cat: CommitteeCategory) => {
```
stacktraceghost: You can use `committeeCategoryValues`

oliver3: Right you are! Updated in https://github.com/kiesraad/abacus/pull/3111/commits/de5cd7bbd5b04b60acd9dc1933aca5434c968ee9

PR #3116, opened 2026-04-01
Add political group registered name

### LINKED ISSUE #3095: Add `registered_name` to election political groups and populate `name` with display name
The `display_name` fields should be populated once the election is imported/created using the following function in Rust (based on the frontend functions from https://github.com/kiesraad/abacus/blob/f18f0c403f3f64165862d8d03ac9bf75074c5ab2/frontend/src/utils/politicalGroup.ts#L18 and https://github.com/kiesraad/abacus/blob/f18f0c403f3f64165862d8d03ac9bf75074c5ab2/frontend/src/utils/candidate.ts#L7

- [x] This `display_name` should be used wherever we now use `political_group.name` (excl. EML exports).
- [x] The apportionment `list_name` fields should be populated with the `display_name`
  - [x] Also update the input files for the typst models
    - [x] Make one list in each input nameless
  - [x] The VotesTable `name` used in the Typst models should also contain the `display_name`
  - [x] Remove the display name code from `scripts.typst` (added in https://github.com/kiesraad/abacus/pull/3093)
    -  [x] Use the display name where applicable
- [x] In the frontend the function `getPoliticalGroupName` needs to be adjusted to use the `pg.display_name` and just add the list prefix as specified by the parameter

Lionqueen94:
Closes #3095

## Description
- Added `RegisteredPoliticalGroup` struct with a `number`, `registered_name` and `candidates` used in the creation of elections
- Added `registered_name` to `PoliticalGroup` (which is not serialized to the frontend) and make sure unnamed parties have a correct name, which is saved in the `name` field
- Removed frontend code to format political group names (this is now done when importing the election and filling the `name` field)
- Removed unused political groups from P2a input
- Fixed unnamed list rendering in N10-2 (see changed pdf diff, one of the parties in the input is now unnamed)
- Made sure all typst input files have a unnamed political group
- Added missing typst smoke test: `test_na_31_2_inlegvel` and lowered the minimal file size, since this is a pdf with only one page. (Smoke test for P 22-2 will be added in https://github.com/kiesraad/abacus/issues/2972)

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -6422,7 +6422,7 @@
       },
       "PoliticalGroup": {
         "type": "object",
-        "description": "Political group with the name for display purposes and its candidates",
+        "description": "Political group and its candidates (with name as used for display purposes)",
         "required": [
           "number",
           "name",
@@ -6438,7 +6438,7 @@
           },
           "name": {
             "type": "string",
-            "description": "Political group name for display purposes (with 'Blanco' in case of empty registered name)"
+            "description": "Political group name as used for display purposes (with 'Blanco' in case of empty registered name)"
           },
           "number": {
             "type": "integer",
--- a/backend/src/domain/election.rs
+++ b/backend/src/domain/election.rs
@@ -171,29 +171,29 @@ pub enum VoteCountingMethod {
 
 id!(PGNumber);
 
-/// Political group with the registered name imported from the EML and its candidates
+/// Political group and its candidates (with registered name as imported from the EML)
 #[derive(Serialize, Deserialize, ToSchema, Clone, Debug, PartialEq, Eq, Hash)]
 #[serde(deny_unknown_fields)]
 pub struct RegisteredPoliticalGroup {
     /// Political group number
     #[schema(value_type = u32)]
     pub number: PGNumber,
-    /// Registered political group name as imported from the candidate list EML (230)
+    /// Registered political group name as imported from the candidates list EML (230)
     pub registered_name: String,
     /// List of candidates of the political group
     pub candidates: Vec<Candidate>,
 }
 
-/// Political group with the name for display purposes and its candidates
+/// Political group and its candidates (with name as used for display purposes)
 #[derive(Serialize, Deserialize, ToSchema, Clone, Debug, PartialEq, Eq, Hash)]
 #[serde(deny_unknown_fields)]
 pub struct PoliticalGroup {
     /// Political group number
     #[schema(value_type = u32)]
     pub number: PGNumber,
-    /// Political group name for display purposes (with 'Blanco' in case of empty registered name)
+    /// Political group name as used for display purposes (with 'Blanco' in case of empty registered name)
     pub name: String,
-    /// Registered political group name as imported from the candidate list EML (230)
+    /// Registered political group name as imported from the candidates list EML (230)
     #[serde(skip_serializing)]
     pub registered_name: String,
     /// List of candidates of the political group
--- a/frontend/src/types/generated/openapi.ts
+++ b/frontend/src/types/generated/openapi.ts
@@ -1096,12 +1096,12 @@ export interface NewElection {
 export type PGNumber = number;
 
 /**
- * Political group with the name for display purposes and its candidates
+ * Political group and its candidates (with name as used for display purposes)
  */
 export interface PoliticalGroup {
   /** List of candidates of the political group */
   candidates: Candidate[];
-  /** Political group name for display purposes (with 'Blanco' in case of empty registered name) */
+  /** Political group name as used for display purposes (with 'Blanco' in case of empty registered name) */
   name: string;
   /** Political group number */
   number: number;


backend/src/domain/election.rs
```diff
@@ -170,16 +171,67 @@ pub enum VoteCountingMethod {
 
 id!(PGNumber);
 
+/// Political group as it is registered with its candidates
```
praseodym: This docstring does not really explain the purpose of this struct

backend/src/repository/election_repo.rs
```diff
@@ -65,35 +108,55 @@ pub async fn create(
...
+        election_date: row.election_date,
+        nomination_date: row.nomination_date,
+        political_groups: row
+            .political_groups
+            .0
+            .into_iter()
+            .map(|pg| pg.into())
+            .collect(),
+    })
 }
```
praseodym: Seeing the duplication and the too_many_lines lint ignore, maybe a ElectionRow struct would have been better after all?

backend/src/domain/election.rs
Lionqueen94: Clarified some more in ea9252f49c96e897aec427da912def7094a8ffea, is this better?

backend/src/repository/election_repo.rs
Lionqueen94: Yeah it is, added the struct and an impl in ea9252f49c96e897aec427da912def7094a8ffea

backend/src/domain/election.rs
praseodym: I find the name `RegisteredPoliticalGroup` confusing (registered where? do we also have political groups that are not registered?) but lacking a better name we can keep that name as long as we explain it well.

Currently there is no explanation of what a "registered name" is (it's just the name that is used in the EML_NL definition, isn't it?) and the sentence "as it is registered with its candidates" is confusing (makes it sounds like candidates are registered).

Lionqueen94: Improved in this commit dba31daf8bfa7a8de423f2e5c460e3737a4cdc08

backend/src/domain/election.rs
```diff
@@ -171,25 +171,32 @@ pub enum VoteCountingMethod {
 
 id!(PGNumber);
 
-/// Political group as it is registered with its candidates, as saved in the database
+/// Political group with the registered name imported from the EML and its candidates
```
jorisleker:
```suggestion
/// Political group and its candidates (with registered name as imported from the EML)
```

backend/src/domain/election.rs
```diff
@@ -171,25 +171,32 @@ pub enum VoteCountingMethod {
...
     #[schema(value_type = u32)]
     pub number: PGNumber,
+    /// Registered political group name as imported from the candidate list EML (230)
     pub registered_name: String,
+    /// List of candidates of the political group
     pub candidates: Vec<Candidate>,
 }
 
-/// Political group with its candidates
+/// Political group with the name for display purposes and its candidates
```
jorisleker:
```suggestion
/// Political group and its candidates (with name as used for display purposes)
```

backend/src/domain/election.rs
jorisleker: Still feels like the candidates and the registered name belong together. Suggestion to fix that

backend/openapi.json
```diff
@@ -6422,7 +6422,7 @@
       },
       "PoliticalGroup": {
         "type": "object",
-        "description": "Political group with its candidates",
+        "description": "Political group with the name for display purposes and its candidates",
```
jorisleker: see previous suggestion

backend/openapi.json
```diff
@@ -6433,14 +6433,17 @@
...
             "items": {
               "$ref": "#/components/schemas/Candidate"
-            }
+            },
+            "description": "List of candidates of the political group"
           },
           "name": {
-            "type": "string"
+            "type": "string",
+            "description": "Political group name for display purposes (with 'Blanco' in case of empty registered name)"
```
jorisleker: see earlier suggestion

frontend/src/types/generated/openapi.ts
```diff
@@ -1096,11 +1096,14 @@ export interface NewElection {
 export type PGNumber = number;
 
 /**
- * Political group with its candidates
+ * Political group with the name for display purposes and its candidates
```
jorisleker: see previous comments

backend/src/domain/election.rs
Lionqueen94: Changed in 4fd15a39b0dca2ab65b670a3d821b5d86bb5fa12

backend/src/domain/election.rs
Lionqueen94: Changed in 4fd15a39b0dca2ab65b670a3d821b5d86bb5fa12

backend/openapi.json
Lionqueen94: Changed in 4fd15a39b0dca2ab65b670a3d821b5d86bb5fa12

backend/openapi.json
Lionqueen94: Changed in 4fd15a39b0dca2ab65b670a3d821b5d86bb5fa12

frontend/src/types/generated/openapi.ts
Lionqueen94: Changed in 4fd15a39b0dca2ab65b670a3d821b5d86bb5fa12

PR #3117, opened 2026-04-01
Restrict incomplete users to setup and logout pages

jschuurk-kr:
### Scope

An incomplete user (i.e. user on first login or after password reset) that's at `/account/setup` and changes the url to `/elections` would get redirected to the login page with the alert "Je hebt geen toegang tot deze pagina". The user is still logged in, though, and can return to /account/setup to continue. So that's confusing.

This PR addresses that by redirecting incomplete users to `/account/setup` if they try to navigate to any other page than the account setup page or the logout page.

I also added tests to AuthorizationGuard.test.tsx and e2e tests to increase coverage of the AuthorizationGuard.

I recommend reviewing per commit. Note that in the first commit I added the e2e tests making them all pass. In the second commit I fixed the behavior mentioned above and updated the related e2e test accordingly.

### Questions
1. Does this PR introduce too much duplication in coverage between the RTL tests and the e2e tests?
2. Can the logic of the AuthorizationGuard be simplified?

### DIFF
--- a/frontend/src/app/AuthorizationGuard.test.tsx
+++ b/frontend/src/app/AuthorizationGuard.test.tsx
@@ -376,7 +376,7 @@ describe("AuthorizationGuard", () => {
           path: "/account/logout",
           element: (
             <AuthorizationGuard>
-              <div>Elections overview page</div>
+              <div>Logout page</div>
             </AuthorizationGuard>
           ),
           handle: { public: true },


frontend/src/app/AuthorizationGuard.test.tsx
```diff
@@ -350,4 +360,52 @@ describe("AuthorizationGuard", () => {
...
+        role: "administrator",
+        fullname: "",
+        needs_password_change: true,
+      },
+      routes: [
+        {
+          path: "/account/logout",
+          element: (
+            <AuthorizationGuard>
+              <div>Elections overview page</div>
```
Lionqueen94:
```suggestion
              <div>Logout page</div>
```

PR #3122, opened 2026-04-02
Allow coordinator csb to update committee session

jschuurk-kr:
### Scope

The PUT on a committee session was only allowed for GSB coordinators. CSB coordinators should be allowed to do this as well, so they can fill in the session details and generate the official results.

I noticed we don't have any tests for this. I think that's (probably) ok. A backend test does not add that much value over reviewing either the API code or the API specs. And this issue would also have been found by the full flow e2e test we still need to build for CSB.

### DIFF
--- a/backend/openapi.json
+++ b/backend/openapi.json
@@ -1442,7 +1442,7 @@
     },
     "/api/elections/{election_id}/committee_sessions/{committee_session_id}": {
       "put": {
-        "summary": "Update a [CommitteeSession]. (coordinator_gsb)",
+        "summary": "Update a [CommitteeSession]. (coordinator_csb, coordinator_gsb)",
         "operationId": "committee_session_update",
         "parameters": [
           {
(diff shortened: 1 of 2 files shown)


jschuurk-kr:
> We still need to check whether the role matches the committee category (i.e. coordinator CSB can only update CSB commitee sessions)

I thought it already did that. `committee_session_update()`  has `user.role().is_authorized(&get_committee_category(&mut tx, committee_session_id).await?)?;` And it's also covered by the tests at the end of `committee_session.rs`, where different endpoints for a GSB election are hit by a coordinator CSB and by a coordinator GSB.

praseodym (approved):

praseodym:
> > We still need to check whether the role matches the committee category (i.e. coordinator CSB can only update CSB commitee sessions)
> 
> I thought it already did that. `committee_session_update()` has `user.role().is_authorized(&get_committee_category(&mut tx, committee_session_id).await?)?;` And it's also covered by the tests at the end of `committee_session.rs`, where different endpoints for a GSB election are hit by a coordinator CSB and by a coordinator GSB.

Indeed, missed that at first. Approved!

PR #3143, opened 2026-04-08
CSB validation errors: replace F305..F309 with F312

### LINKED ISSUE #3167: replace F.305..F.309 with F.312
replace F.305..F.309 with F.312: PR https://github.com/kiesraad/abacus/pull/3143

oliver3:
Resolves #3167 

For CSB (= in the validation of GSB Results):
- ~~Add F.205~~
- Replace F.305 .. F.309 with F.312

Please note that errors and warnings that should not be shown for CSB will be suppressed in another issue, e.g. W.203 in the screenshot below.

**Question**
Is the behaviour of F.312 how we want it to be? It will show errors in both "Aantal kiezers en stemmen" and "Verschillen D & H", and will e.g. cause the error to appear and disappear in the navigation for "Aantal kiezers en stemmen" when changing D and/or H during data entry.

#### F.312
image


jorisleker:
> Is the behaviour of F.312 how we want it to be? It will show errors in both "Aantal kiezers en stemmen" and "Verschillen D & H", and will e.g. cause the error to appear and disappear in the navigation for "Aantal kiezers en stemmen" when changing D and/or H during data entry.

For GSB we also have rules that test if I and/or J are in correspondence with the difference between D and H. We only show these on the differences page. Logic is that the chance that the error is in I or J is significantly larger then that the error is in D or H, since D and H both have other validation rules checking if the other fields on the voters and votes page add up to D and H. So I would only show F.312 on the "Verschillen D&H" page.

jschuurk-kr:
@oliver3 F.205 needs to become a warning, because in "waterschapsverkiezingen" there can be a GSB with 0 eligible voters.

jschuurk-kr:
> @oliver3 F.205 needs to become a warning, because in "waterschapsverkiezingen" there can be a GSB with 0 eligible voters.

And a different check.

oliver3:
> So I would only show F.312 on the "Verschillen D&H" page.

Changed in https://github.com/kiesraad/abacus/pull/3143/commits/6b57f8c3c1cd35b35b5014941b52f3245ee05164

oliver3:
> @oliver3 F.205 needs to become a warning, because in "waterschapsverkiezingen" there can be a GSB with 0 eligible voters.
> And a different check.

Removed F.205 from the PR for now

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -46,124 +46,28 @@ impl GSBDifferencesCounts {
...
-        {
-            validation_results.errors.push(ValidationResult {
-                fields: vec![path.field("more_ballots_count").to_string()],
-                code: ValidationResultCode::F306,
-                context: None,
-            });
-        }
-
-        if total_votes_count > total_voters_count && self.fewer_ballots_count != 0 {
+        if voters_count != (votes_count - self.more_ballots_count + self.fewer_ballots_count) {
```
praseodym: `Count` is an unsigned integer (u32) so this subtraction can underflow

frontend/src/testing/api-mocks/ValidationResultMockData.ts
```diff
@@ -79,6 +79,15 @@ export const validationResultMockData: ErrorWarningsMap<ValidationResultCode> =
     fields: ["data.differences_counts.difference_completely_accounted_for"],
     code: "F310",
   },
+  F312: {
+    fields: [
+      "data.voters_counts.total_admitted_voters_count",
+      "data.votes_counts.total_votes_cast_count",
```
praseodym: The backend doesn't return these fields so they should be removed in the mock too

praseodym (changes requested):

backend/src/domain/results/gsb_differences_counts.rs
stacktraceghost: We could add a test case for this, which we can then fix. We can also add test cases to other validations to ensure no under/overflow is possible.

frontend/src/testing/api-mocks/ValidationResultMockData.ts
Lionqueen94: Removed in 2912a1a28a72bf84c0ded1fbd5ca528e9a83c49c

backend/src/domain/results/gsb_differences_counts.rs
Lionqueen94: Failing test added in 00c612bb0bfb470032a7b07b54b4c66f9f50f697 and fix added in d3d0575e2a790f4ae6ba9fbd9dd5ea6f9e0ee37d

frontend/src/i18n/locales/nl/feedback_CSB.json
praseodym, line 1: Do we need a typist feedback for F312 too?

Lionqueen94:
```
#### F.312 `totaal aantal kiezers = totaal aantal uitgebrachte stemmen - meer getelde stemmen + minder getelde stemmen`

> Invoerder: **Controleer je antwoorden** (F.312)

> Coördinator: **D is niet gelijk aan H min I plus J** (F.312)  
> Je kan niet verder met dit proces-verbaal. Controleer of er iets is misgegaan bij het opmaken of overdragen van het proces-verbaal van het onderliggende niveau.  

Velden markeren: I en J
```
The typist gets the default "Check your answers" error, so no need to specify a custom one.

praseodym (dismissed):

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -46,124 +46,31 @@ impl GSBDifferencesCounts {
         }
     }
 
-    fn validate_no_differences(
+    /// F312 totaal aantal kiezers =
+    ///   totaal aantal uitgebrachte stemmen - meer getelde stemmen + minder getelde stemmen
```
stacktraceghost: This docstring could be removed. We don't have this in other places where we implement the rules and it is already mentioned in `validate.rs`

backend/src/domain/results/gsb_differences_counts.rs
```diff
@@ -556,192 +463,58 @@ mod tests {
...
-            let has_f306 = result.errors.iter().any(|e| e == &f306);
-            assert_eq!(has_f306, expect_f306, "Failed: {description}");
-        }
-
-        Ok(())
-    }
-
-    /// CSO | F.307 (Als H > D) J is ingevuld
+    /// F.312 totaal aantal kiezers =
+    ///   totaal aantal uitgebrachte stemmen - meer getelde stemmen + minder getelde stemmen
```
stacktraceghost: Could be changed to:
`/// CSB | F.312: totaal aantal kiezers = totaal aantal uitgebrachte stemmen - meer getelde stemmen + minder getelde stemmen`
However, I think in the validation rules documentation this should be != instead of = ([link to comment in PR](https://github.com/kiesraad/abacus-documentatie/pull/54#pullrequestreview-4107067615))

backend/src/domain/validate.rs
```diff
@@ -85,6 +85,8 @@ pub enum ValidationResultCode {
     F309,
     /// CSO: (Als D <> H en verklaring voor verschil niks aangevinkt of 'ja' en 'nee' aangevinkt)
     F310,
+    /// CSB: totaal aantal kiezers = totaal aantal uitgebrachte stemmen - meer getelde stemmen + minder getelde stemmen
```
stacktraceghost: I think in the validation rules documentation this should be `!=` instead of `=` ([link to comment in PR](https://github.com/kiesraad/abacus-documentatie/pull/54#pullrequestreview-4107067615))

backend/src/domain/results/gsb_differences_counts.rs
Lionqueen94: I think we write it how it's supposed to be in the validation rules, so if it is not equal, the error is shown.

Lionqueen94: Ah yes you're right, changed in 812881e91461ef380ade04d6eb8692cde6973c1e

backend/src/domain/validate.rs
Lionqueen94: Changed in 812881e91461ef380ade04d6eb8692cde6973c1e

stacktraceghost (approved):

praseodym (approved):

PR #3174, opened 2026-04-14
Implement results for CSB

### LINKED ISSUE #3000: Implement results for CSB
- Add number_of_voters to ElectionSummary
- Update `list_results_for_committee_session` and `fetch_results_for_committee_session` to return `[DataEntrySource, Results]` instead of polling stations
- Make use of `DataEntrySource` for `ElectionSummary::from_results`
- Use `DataEntrySource` as entries in `SumCount`, deal with fallout in typst for this

rnijveld:


stacktraceghost (approved):

backend/src/repository/data_entry_repo.rs
```diff
@@ -280,55 +234,102 @@ async fn fetch_results_for_committee_session(
...
-    } else {
-        Ok(results)
+        return Err(sqlx::Error::RowNotFound);
+    }
+
+    Ok(results)
+}
+
+/// List the results for all data entries of a committee session for the CSB.
+async fn list_results_for_csb_committee_session(
```
Lionqueen94, line 272: This function could use a unit test, but this might be fine to add with the apportionment issue?

backend/src/domain/summary.rs
```diff
@@ -111,14 +118,33 @@ impl ElectionSummary {
...
                     .polling_station_investigations
-                    .append_result(polling_station, cso_first_result);
+                    .append_result(polling_station_source, cso_first_result);
+            }
+
+            // GSB results contain number of voters which need to be added
+            if let Results::GSB(gsb_result) = result {
+                // this retrieves the current total number of voters, or initializes it to zero if not yet set
+                // and then adds the number of voters for this result to it
+                *totals.number_of_voters.get_or_insert(0) += gsb_result.number_of_voters;
```
Lionqueen94, line 144: This could use a unit test, but this might be fine to add with the apportionment issue?

rnijveld: I think for both this and the previous comment we need fixture data for CSB with data entry results, which we don't have right now as far as I know? I think it might be a good idea to move that to a later issue to unblock progress on a whole lot of other stuff.

rnijveld: I created #3177 for this!

backend/src/repository/data_entry_repo.rs
rnijveld: I created #3177 for this!

backend/src/domain/summary.rs
Lionqueen94: https://github.com/kiesraad/abacus/issues/2941 also needs CSB fixtures, so it might be good to do the fixtures in your newly created issue indeed.

Lionqueen94 (approved):

PR #3175, opened 2026-04-14
Create P 22-2 Bijlage 1 pdf in Typst

### LINKED ISSUE #3102: Add P 22-2 Bijlage 1 PDF
Add Input code and test in backend

Lionqueen94:
Closes #3102


stacktraceghost (dismissed):
Reviewed code only, not contents of the template.

backend/templates/model-p-22-2-bijlage1.typ
jschuurk-kr, line 1: Looking at some P 22-2s from GR26 and found two differences.

1. The numbering of the pages continues from the P22-2 all through the "Bijlages". I think it's fine we don't do this, since Abacus will only produce Bijlage 1. So continuing the page numbering only from the main document to Bijlage 1, but not to the other bijlages has no advantages.

2. The Bijlage 1 contains the same hash as the main part of the P22-2. That should be the hash of the EML 520 "verkozen kandidaten". Our Bijlage 1 does not contain a hash or a timestamp. This matches with what our user story map says: put the EML 520 hash on the main part and don't put the EML 510d on the Bijlage 1.

The combination of these two does raise an issue: if you have the P22-2 and the Bijlage 1 as separate documents, how do you know they belong together? As in, how do you know they were not generated at separate times with different data?

praseodym: > As in, how do you know they were not generated at separate times with different data?

We also include a timestamp right

jschuurk-kr: We don't.
image

Lionqueen94: Added in c2746995a5fee5db5d8cab356ef6212080446a6e

jschuurk-kr (approved):

stacktraceghost (approved):

PR #3182, opened 2026-04-15
Implement W205

stacktraceghost:
Resolves #3166

- Adds rule W205: `W.205 Aantal kiesgerechtigden = leeg of 0`
- i18n and test added

Issue #3171, opened 2026-04-13
Check validation implementation and texts against specs


jschuurk-kr:
### Scope

- depends on merge of https://github.com/kiesraad/abacus-documentatie/pull/54
- Check if rules are triggered for the right models. For example W201-W203 should not be triggered for CSB
- Update i18n texts
- Also update doc comments (for example in validate.rs) to make them reflect the correct models (e.g. `GSB CSO/GSB DSO, CSB`)
- Make warning texts for typists generic so they collapse, see [this comment](https://github.com/kiesraad/abacus/pull/3182#pullrequestreview-4114206325) and [this commit on the specs](https://github.com/kiesraad/abacus-documentatie/commit/d5a7c3494bd8884361eadc9c42404ea76e255051)
- Don't show [feedback.coordinator_actions](https://github.com/kiesraad/abacus/pull/3182#issuecomment-4258985411) in coordinator view for warnings. ([confirmed](https://github.com/kiesraad/abacus/pull/3182#issuecomment-4266396223) by @jorisleker)

PR #3183, opened 2026-04-15
Implement W206

### LINKED ISSUE #3169: Introduce W.206: Kiesgerechtigden kleiner dan stempassen

stacktraceghost:
Resolves #3169

- Adds rule W206: `(Als W.205 niet getoond wordt) Aantal kiesgerechtigden is kleiner dan aantal stempassen`
- When W205 triggers, W206 is not shown
- i18n and test added


jorisleker (dismissed):

jorisleker (dismissed):
I think stricktly speaking we could/should also show a warning on A, because that field is also part of the validation logic. 

But in cases where the user made an error in A, other validation errors will be triggered that that will highlight A as well, so I can't think of a situation where W206 will be the only validation message shown, and the reason it triggered was not in Z. If we want to cover all edge cases we don't foresee now, feel free to add a warning icon on A as well. Otherwise, ok for now!

praseodym (commented):
After merging #3182 we should add some logic to prevent both W205 and W206 from showing up at the same time (W205 should take precedence).

praseodym (dismissed):

backend/src/domain/validate.rs
```diff
@@ -105,6 +105,8 @@ pub enum ValidationResultCode {
     W203,
     /// CSO/DSO: 'Aantal kiezers en stemmen': Totaal aantal uitgebrachte stemmen leeg of 0
     W204,
+    /// CSO/DSO: 'Aantal kiezers en stemmen': Aantal kiesgerechtigden is kleiner dan aantal stempassen
```
jschuurk-kr:
```suggestion
    /// CSB: 'Aantal kiezers en stemmen': Aantal kiesgerechtigden is kleiner dan aantal stempassen
```

jschuurk-kr:
Same [comment as on the W205 PR](https://github.com/kiesraad/abacus/pull/3182#issuecomment-4258985411) applies here about the text shown to the coordinator. => Let's tackle this as part of https://github.com/kiesraad/abacus/issues/3171

praseodym:
@stacktraceghost this branch has conflicts

praseodym (approved):

Lionqueen94 (approved):

PR #3187, opened 2026-04-16
Fix formatting of hashes in template inputs

jschuurk-kr:
### Scope

Most of the hahses in the template inputs for our pdf files did not have a space between each group of four characters. ~The `from` on `EMLHash` in `backend/src/eml/hash.rs` does add those spaces, so our template inputs should have them as well.~ When you generate pdfs through Abacus, the spaces are present, so our template inputs should have them as well.


Lionqueen94 (approved):

Lionqueen94 (changes requested):
You're still missing a few hashes in other json files

Lionqueen94 (approved):

praseodym (approved):

PR #3189, opened 2026-04-16
Moved CommitteeSessionNotCompleted Apportionment error out of crate

### LINKED ISSUE #3181: `ApportionmentError::CommitteeSessionNotCompleted` should not be in apportionment crate
The apportionment crate includes the `ApportionmentError::CommitteeSessionNotCompleted` type. Checking whether the committee session has been completed is a responsibility handled by the Abacus crate, not the apportionment crate (the apportionment crate does not know about the committee session entity). This error type should not be part of the apportionment crate and instead the main Abacus crate should have its own error type for this.

Lionqueen94:
Closes #3181

### DIFF
--- a/backend/src/error.rs
+++ b/backend/src/error.rs
@@ -9,6 +9,7 @@ use axum::{
 };
 use eml_nl::EMLError;
 use hyper::header::InvalidHeaderValue;
+use pdf_gen::{PdfGenError, zip::ZipResponseError};
 use quick_xml::{DeError, SeError};
 use serde::{Deserialize, Serialize};
 use tracing::error;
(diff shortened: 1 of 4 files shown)


backend/src/error.rs
```diff
@@ -100,6 +103,7 @@ pub enum APIError {
     AddError(String, ErrorReference),
     AirgapViolation(String),
     Apportionment(ApportionmentError),
+    ApportionmentApi(ApportionmentApiError),
```
praseodym: I would merge these two enums and have the API map the `ApportionmentError` from the crate to `ApportionmentApiError`, so that this error handling code doesn't need to know about apportionment internals.

Lionqueen94: Good suggestion, it's much better like this I think, see 9dcbabc636201f239ff76764892a60e9dae86956

praseodym (approved):

stacktraceghost (approved):

PR #3200, opened 2026-04-20
Reverse relationship for files table

### LINKED ISSUE #3197: Reverse relationship to files table
Right now the committee sessions table has several foreign key columns to the files table, each for a separate type of file. This solution does not scale to more types of files (e.g. CSB documents).

We should reverse the relationship and have the files table point to a committee session with a foreign key, and add a type column (Rust enum) to identify the type of document.

praseodym:
Reverse relationship for files table, resolves #3197.


backend/src/domain/file.rs
```diff
@@ -3,10 +3,23 @@ use serde::{Deserialize, Serialize};
...
 
 id!(FileId);
 
+/// File type
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, ToSchema, Type)]
+#[serde(rename_all = "snake_case")]
+#[sqlx(rename_all = "snake_case")]
+pub enum FileType {
+    /// GSB results EML (510b)
+    ResultsEml,
```
Lionqueen94: Might be nice to rename this to CountsEml and then it contains the 510b/510d. The ResultsEml can then contain the 520. What do you think @praseodym

Lionqueen94: Another option is to only have CountsEml for the CSB 510d and have the 520 in ResultsEml, but this is a bit less consistent. Then the change like this is fine though.

praseodym: Do you think we should change that in this PR or when we introduce these new files?

praseodym: Added GSB prefix in 75a71a66c3d0ee9e85d406e5f18fbaecb756d25c

Lionqueen94 (approved):

stacktraceghost (approved):

PR #3194, opened 2026-04-17
fix navbar story

jschuurk-kr:
### Scope
- [x] fix NavBar story by providing location to NavBar, see commit 81623ad30494331b2b8c1e1509d749458bf29595
- [x] refactor NavBar AllRoutes story to allow for testing
- [x] split up and write tests for NavBar AllRoutes story

### Before the fix
image

### After the fix
image

PR #3202, opened 2026-04-21
merge Menu and MenuButton stories

jschuurk-kr:
### Scope

As [suggested](https://github.com/kiesraad/abacus/pull/3194#pullrequestreview-4139656130) by @Lionqueen94 in PR #3194:
- merge Menu and MenuButton stories
- add a test


Lionqueen94 (approved):

jschuurk-kr:
Closing this, since merging Menu and MenuButton only makes sense if I also add the test I added in the PR. But we already have that test as the RTL test `"menu works for $pathname"` in `NavBar.test.tsx`.

So without first reviewing and deciding how we want to test our storybook stories, changes like this PR do not add any value.

PR #3203, opened 2026-04-21
Update validation implementation/texts against specs

### LINKED ISSUE #3171: Check validation implementation and texts against specs

stacktraceghost:
Resolves #3171

Based on the changes made in https://github.com/kiesraad/abacus-documentatie/pull/54

Changes in this PR:
- Updated docstrings https://github.com/kiesraad/abacus/commit/c162d8d8b3dd0d404a32bb366ac3cf06b8f05387
  - Includes model
  - Some title updates
- Update feedback https://github.com/kiesraad/abacus/commit/d5fdc20f60922d7b9c30737386b5351d12d12f53 / https://github.com/kiesraad/abacus/commit/938ff5953be2ea17d342803759e951ddd08260fd
  - CSB feedback copy updated
  - GSB feedback
    - Title removed for W201 - W204, defaults to "Controleer je antwoorden" so warnings get merged in display
    - Minor action change for F305 to align with documentation
    - Updated tests accordingly
  - Feedback component
    - Only show default actions (=handelingsperspectieven) for typists (all feedback types, all categories) and coordinators (error, GSB only).
    - Story: added election selector so you can see feedback based on GSB or CSB
- Trigger rules for the right model https://github.com/kiesraad/abacus/commit/859076ba077a51babcd0e5e45449308f997e1031
  - `election_fixture` now accept `committee_category` so we can test other election types
  - Rules which are implemented in shared structs, now have explicit tests to check that they trigger in the correct committee category, see changelog below:

Changelog validation rules:
- Removed W203 from `GSBResults`

| Validation code | Model/file                          | Change |
|-----------------|-------------------------------------|--------|
| F101/F102       | ExtraInvestigation                  | Added if-statement to trigger for GSB only<br>Tests converted to matrix, included case to check it doesn't trigger for CSB |
| F111/F112       | CountingDifferencesPollingStation   | Added if-statement to trigger for GSB only<br>Tests converted to matrix, included case to check it doesn't trigger for CSB |
| F201            | VotersCounts                        | Should trigger for all models<br >Added case to check it also triggers for CSB |
| F202/F203/F204  | VotesCounts                         | Should trigger for all models<br>Added case to check it also triggers for CSB |
| F301 - F310     | DifferencesCounts                   | Should only trigger for GSB CSO/DSO. This file is only included in CSO (and later DSO) models, so left as-is. |
| F312            | GSBDifferencesCounts                | Should only trigger for CSB. This file is only included in the GSB model, so left as-is. |
| F401 - F403     | CommonPollingStationResults and GSBResults | Should trigger for all models<br>Each model has their own (identical) implementation, since these rules are placed at the model root. Left as-is |
| W001            | domain/data_entry.rs                | Should trigger for all models, left as-is |
| W201/W202       | VotesCounts                         | Added if-statement to trigger for GSB only<br>Added test case for both rules to check it doesn't trigger for CSB |
| W203            | CommonPollingStationResults         | Should only trigger for CSB. This file is only included in the GSB model, so left as-is. |
| W204            | VotesCounts                         | Should trigger for all models<br>Added case to check it also triggers for CSB |
| W205/W206       | GSBResults                          | Should only trigger for CSB. This file is only included in the GSB model, so left as-is. |

### DIFF
--- a/backend/src/domain/election.rs
+++ b/backend/src/domain/election.rs
@@ -282,7 +282,8 @@ pub(crate) mod tests {
     /// Create a test election with some political groups and a given number of seats.
     /// The number of political groups is the length of the `political_groups_candidates` slice.
     /// The number of candidates in each political group is equal to the value in the slice at that index.
-    pub fn election_fixture_with_given_number_of_seats(
+    fn election_fixture_with_given_number_of_seats(
+        commitee_category: CommitteeCategory,
         political_groups_candidates: &[u32],
         number_of_seats: u32,
     ) -> ElectionWithPoliticalGroups {
@@ -311,7 +312,7 @@ pub(crate) mod tests {
         ElectionWithPoliticalGroups {
             id: ElectionId::from(1),
             name: "Test".to_string(),
-            committee_category: CommitteeCategory::GSB,
+            committee_category: commitee_category,
             counting_method: Some(VoteCountingMethod::CSO),
             election_id: "Test_2023".to_string(),
             location: "Test".to_string(),
@@ -328,7 +329,14 @@ pub(crate) mod tests {
     /// Create a test election with some political groups.
     /// The number of political groups is the length of the `political_groups_candidates` slice.
     /// The number of candidates in each political group is equal to the value in the slice at that index.
-    pub fn election_fixture(political_groups_candidates: &[u32]) -> ElectionWithPoliticalGroups {
-        election_fixture_with_given_number_of_seats(political_groups_candidates, 29)
+    pub fn election_fixture(
+        commitee_category: CommitteeCategory,
+        political_groups_candidates: &[u32],
+    ) -> ElectionWithPoliticalGroups {
+        election_fixture_with_given_number_of_seats(
+            commitee_category,
+            political_groups_candidates,
+            29,
+        )
     }
 }
--- a/backend/src/domain/election.rs
+++ b/backend/src/domain/election.rs
@@ -283,7 +283,7 @@ pub(crate) mod tests {
     /// The number of political groups is the length of the `political_groups_candidates` slice.
     /// The number of candidates in each political group is equal to the value in the slice at that index.
     fn election_fixture_with_given_number_of_seats(
-        commitee_category: CommitteeCategory,
+        committee_category: CommitteeCategory,
         political_groups_candidates: &[u32],
         number_of_seats: u32,
     ) -> ElectionWithPoliticalGroups {
@@ -312,7 +312,7 @@ pub(crate) mod tests {
         ElectionWithPoliticalGroups {
             id: ElectionId::from(1),
             name: "Test".to_string(),
-            committee_category: commitee_category,
+            committee_category,
             counting_method: Some(VoteCountingMethod::CSO),
             election_id: "Test_2023".to_string(),
             location: "Test".to_string(),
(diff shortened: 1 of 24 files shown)


backend/src/domain/results/votes_counts.rs
```diff
@@ -492,18 +506,22 @@ mod tests {
             }],
         );
 
+        // Not applicable for CSB
+        let validation_results = validate(CSB, &[40, 20, 9], 69, 31, 0, 100)?;
+        assert!(validation_results.errors.is_empty());
```
praseodym: Shouldn't this be `warnings.is_empty()`?

backend/src/domain/results/votes_counts.rs
```diff
@@ -524,56 +542,62 @@ mod tests {
             }],
         );
 
+        // Not applicable for CSB
+        let validation_results = validate(CSB, &[40, 20, 9], 69, 0, 31, 100)?;
+        assert!(validation_results.errors.is_empty());
```
praseodym: Same here: shouldn't this be `warnings.is_empty()`?

backend/src/domain/election.rs
```diff
@@ -282,7 +282,8 @@ pub(crate) mod tests {
     /// Create a test election with some political groups and a given number of seats.
     /// The number of political groups is the length of the `political_groups_candidates` slice.
     /// The number of candidates in each political group is equal to the value in the slice at that index.
-    pub fn election_fixture_with_given_number_of_seats(
+    fn election_fixture_with_given_number_of_seats(
+        commitee_category: CommitteeCategory,
```
praseodym: typo:
```suggestion
        committee_category: CommitteeCategory,
```

(also below)

stacktraceghost: Well spotted, fixed in https://github.com/kiesraad/abacus/pull/3203/commits/39d0c36cc147b0f3649ccc5fe39ee3c4a7c6eacc

backend/src/domain/results/votes_counts.rs
stacktraceghost: Fixed in https://github.com/kiesraad/abacus/pull/3203/commits/39d0c36cc147b0f3649ccc5fe39ee3c4a7c6eacc

backend/src/domain/results/votes_counts.rs
stacktraceghost: Good find, this was actually also wrong in the original test. I've adjusted it and discovered that the test cases are based on 30% instead of 3%, so the first case even fails when checking if there are no warnings.

Adjusted the cases to verify 3% in https://github.com/kiesraad/abacus/pull/3203/commits/39d0c36cc147b0f3649ccc5fe39ee3c4a7c6eacc

I've also done a sanity check along other tests to see if similar mistakes have been made, only found W204.

Lionqueen94 (approved):

praseodym (approved):

jschuurk-kr (approved):

PR #3206, opened 2026-04-21
Apply Clippy lints to all crates in workspace

praseodym:
I noticed that Sigrid found a long function that did not have an Clippy lint allow annotation, and it turns out Clippy lints were not configured for the apportionment and pdf_gen crates. This PR fixes that.

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -102,16 +102,19 @@ test-log.workspace = true
 [build-dependencies]
 memory-serve = { version = "2.1.0", optional = true }
 
-[lints.rust]
+[workspace.lints.rust]
 unsafe_code = "forbid"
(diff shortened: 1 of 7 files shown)


Lionqueen94 (dismissed):

Lionqueen94:
`test_apportionment_process` is also flagged now, while that's a test function, we should probably still exclude those?

praseodym:
> `test_apportionment_process` is also flagged now, while that's a test function, we should probably still exclude those?

Indeed, added an ignore in 792c9d01db467e717273ffabc3bef2527ddf00a5

stacktraceghost (approved):

Lionqueen94 (approved):

Issue #3054, opened 2026-03-17
Epic: Resolve apportionment differences OSV and Abacus


jschuurk-kr:
# Context

Description of apportionment in OSV2020: https://www.kiesraad.nl/documenten/2016/osv/osv-bestanden/determination-of-the-election-result

Related: https://github.com/kiesraad/abacus-internal/issues/267

(Draft) PR that skips certain inputs for the differential fuzzer: #3213

# Architecture

# Scope

Fuzzer still needs to check candidate nomination and compare which parties/candidates needs drawing lots: moved to #3361.

PR #3213, opened 2026-04-22
Skip cases in differential fuzzer that will need resolving

jschuurk-kr:
### Scope

Skip certain inputs in our osv2020 differential fuzzer, to uncover as many differences as possible.

PR status will be draft until I've taken this approach as far as I can make it go. Step after that would be to merge the PR. Then whenever we resolve a difference between osv2020 and abacus, the related skips can be removed from the fuzzer.

See sub-issues in epic #3054 for the found differences.

### Testing
Did not find any mismatches after running the fuzzer for 10 hours with 4 jobs, covering over 15.000.000 inputs (unless that number in the output means something different):

```
#15045261: cov: 1927 ft: 8946 corp: 754 exec/s: 144 oom/timeout/crash: 0/0/0 time: 36158s job: 627 dft_time: 0
^C
==59190== libFuzzer: a child was interrupted; exiting
INFO: exiting: 72 time: 36192s
```

### DIFF
--- a/backend/apportionment/fuzz/fuzz_targets/differential_osv2020.rs
+++ b/backend/apportionment/fuzz/fuzz_targets/differential_osv2020.rs
@@ -6,8 +6,8 @@ use std::{
     sync::{Mutex, OnceLock},
 };
 
-use apportionment::{ApportionmentError, CandidateVotes, process};
-use apportionment_fuzz::{FuzzedApportionmentInput, get_total_seats, init_tracing, run_with_log};
+use apportionment::{process, ApportionmentError, CandidateVotes};
+use apportionment_fuzz::{get_total_seats, init_tracing, run_with_log, FuzzedApportionmentInput};
 use libfuzzer_sys::fuzz_target;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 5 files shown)


Lionqueen94 (dismissed):

backend/apportionment/fuzz/fuzz_targets/differential_osv2020.rs
```diff
@@ -85,104 +85,216 @@ fn report_mismatch(
...
+    let total_votes = data
+        .list_votes
+        .iter()
+        .map(|list| {
+            list.candidate_votes
+                .iter()
+                .map(|cv| cv.votes())
+                .sum::<u32>()
+        })
+        .sum::<u32>();
```
praseodym, line 140:
```suggestion
    let total_votes = data
        .list_votes
        .iter()
        .flat_map(|list| list.candidate_votes.iter())
        .map(|cv| cv.votes())
        .sum::<u32>();
```

praseodym (dismissed):

praseodym (approved):

stacktraceghost (approved):

PR #3225, opened 2026-04-24
Added largest remainder and unique highest average to list seat assignment in backend P 22-2 input

### LINKED ISSUE #3216: Add largest remainder and unique highest average to list seat assignment
- [x] Add `largest_remainder` in backend and update P 22-2 inputs and template accordingly
- [x] Add `unique_highest_average` in backend and update P 22-2 inputs and template accordingly

Lionqueen94:
Closes #3216 

## Description
- Renamed `InitialFullSeatsTable` to `EnrichedSeatAssignment` and added `initial_total_residual_seats` to it
- Removed `TotalSeatsTable`, since this information is already present in `EnrichedCandidateNomination`
- Added `ChosenCandidates` to `EnrichedCandidateNomination`
- Removed `enriched` from `enriched_candidate_nomination` in `ModelP22_2_Input`
- Added `largest_remainder` to `list_seat_assignment` if list meets threshold
- Added `unique_highest_average` to `list_seat_assignment` if initial unique highest average assignment took place
- P 22-2 pdf now shows only initially assigned residual seats for largest remainder assignment and unique highest average assignment. The footnotes indicate whether more seats were (re)assigned after absolute majority reassignment or list exhaustion. Hence many changes in the pdf diff occur.


backend/src/domain/models/enriched_seat_assignment.rs
```diff
@@ -0,0 +1,228 @@
...
+    },
+};
+
+type SeatChangeSteps<'a> = (
+    Vec<&'a SeatChangeStep>,
+    Vec<&'a SeatChangeStep>,
+    Vec<&'a SeatChangeStep>,
+);
+
+fn get_steps(seat_assignment: &SeatAssignment) -> Result<SeatChangeSteps<'_>, APIError> {
```
stacktraceghost: Do we need to collect `initial_highest_average_assigned_seats`? It seems like it is not being used. 

I would like to propose renaming this method to `get_initial_steps` and add a docstring with something like this:
```
/// Retrieves [SeatChangeStep]s of type LargestRemainderAssignment/UniqueHighestAverageAssignment before absolute majority reassignment or list exhaustion removal takes place
```

Maybe we can use a struct with named properties instead of a tuple, so there is less risk of mixing the returned values up.

backend/src/domain/models/enriched_seat_assignment.rs
```diff
@@ -0,0 +1,228 @@
...
+                .map(|(_, column)| column.clone());
+            let unique_highest_average_column = Self::get_unique_highest_average_column(
+                pg_votes.number,
+                initial_full_seats,
+                largest_remainder_column.clone(),
+                initial_unique_highest_average_steps.clone(),
+            );
+            initial_total_full_seats += initial_full_seats;
+
+            columns.push(EnrichedListSeatAssignment {
```
stacktraceghost: I think these are rows instead of columns

backend/src/domain/models/enriched_seat_assignment.rs
praseodym: This function doesn't need to return a `Result`, it never returns an error

backend/src/domain/models/enriched_seat_assignment.rs
```diff
@@ -0,0 +1,228 @@
...
+            let initial_full_seats = Self::get_initial_full_seats(seat_assignment, pg_votes.number);
+            let largest_remainder_column = largest_remainder_columns
+                .iter()
+                .find(|(pg_number, _)| *pg_number == pg_votes.number)
+                .map(|(_, column)| column.clone());
+            let unique_highest_average_column = Self::get_unique_highest_average_column(
+                pg_votes.number,
+                initial_full_seats,
+                largest_remainder_column.clone(),
+                initial_unique_highest_average_steps.clone(),
```
praseodym: This can be passed by reference instead of cloning

backend/src/domain/models/enriched_seat_assignment.rs
```diff
@@ -0,0 +1,228 @@
...
+        for pg_votes in summary.votes_counts.political_group_total_votes.iter() {
+            let initial_full_seats = Self::get_initial_full_seats(seat_assignment, pg_votes.number);
+            let largest_remainder_column = largest_remainder_columns
+                .iter()
+                .find(|(pg_number, _)| *pg_number == pg_votes.number)
+                .map(|(_, column)| column.clone());
+            let unique_highest_average_column = Self::get_unique_highest_average_column(
+                pg_votes.number,
+                initial_full_seats,
+                largest_remainder_column.clone(),
```
praseodym: This can also be passed by reference instead of cloning

backend/src/domain/models/enriched_seat_assignment.rs
```diff
@@ -0,0 +1,228 @@
...
+            .filter(|list_seat_assignment| list_seat_assignment.meets_remainder_threshold)
+            .collect();
+        let mut columns = Vec::new();
+        for standing in final_standing_pgs_meeting_threshold.iter() {
+            let assigned_seat = initial_largest_remainder_steps
+                .iter()
+                .find(|step| step.change.list_number_assigned() == standing.list_number);
+            let largest_remainder_column = LargestRemainderColumn {
+                remainder_votes: standing.remainder_votes.clone(),
+                residual_seats: if assigned_seat.is_some() { 1 } else { 0 },
```
praseodym, line 99: What about `u32::from(assigned_seat.is_some())`

backend/src/domain/models/enriched_seat_assignment.rs
Lionqueen94: Not yet, it's for the next PR, I commented out the lines related to the highest_average_steps for now. And implemented your suggestions in 44e66164a880f728996c28b71a68c6fc964d73d2

backend/src/domain/models/enriched_seat_assignment.rs
Lionqueen94: True, changed in 44e66164a880f728996c28b71a68c6fc964d73d2

backend/src/domain/models/enriched_seat_assignment.rs
Lionqueen94: True! Changed in 44e66164a880f728996c28b71a68c6fc964d73d2

backend/src/domain/models/enriched_seat_assignment.rs
Lionqueen94: Changed in 44e66164a880f728996c28b71a68c6fc964d73d2

backend/src/domain/models/enriched_seat_assignment.rs
Lionqueen94: Changed in 44e66164a880f728996c28b71a68c6fc964d73d2

backend/src/domain/models/enriched_seat_assignment.rs
Lionqueen94: Good suggestion, changed in 44e66164a880f728996c28b71a68c6fc964d73d2

praseodym (approved):

stacktraceghost (approved):

PR #3229, opened 2026-04-28
Bugfix data entry scroll

### LINKED ISSUE #3224: Data entry page doesn't scroll up to feedback
Found by @jorisleker 

Reproduction steps:
- Enter data entry that results in multiple errors. For example F201 + F204: wrong sum AB>D and leave E.x empty. 
- Fix F201 by correcting AB=D
- Press "Next"

Observed:
- Page doesn't scroll to top, so the remaining error is not visible, which makes it seem that only the checkbox "Ik heb mijn invoer gecontroleerd" appears.

Expected:
- The page scrolls to top so the error message is visible

The behavior of not scrolling back to the top is only expected for F401. For all other errors and warnings, the page should scroll to the top.

We could add a Playwright assertion with [toBeInViewport](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-in-viewport) to cover this behavior.

stacktraceghost:
Resolves #3224

This PR fixes the scrolling behaviour for the data entry form, specifically when having validation errors again on the second submit. See reproduction steps in linked issue.

Changes:
- Added e2e test which checks the behaviour including what is in the viewport
- Moved check for F401 to `useDataEntryFormSection` as `trailingError`, so it can be used in the hook that scrolls to top
- Removed `dataEntryStructure`/`results`/`status` from  `useDataEntryFormSection` return type since they were not used

### DIFF
--- a/frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
+++ b/frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
@@ -872,6 +872,85 @@ test.describe("errors and warnings", () => {
     await expect(votersAndVotesPage.acceptErrorsAndWarnings).toBeHidden();
   });
 
+  test("scroll to top when warning/errors, unless accept checkbox or trailing error is visible", async ({
+    page,
+    pollingStation,
+  }) => {
+    await page.goto(`/elections/${pollingStation.election_id}/data-entry/${pollingStation.id}/1`);
+
+    // Fill sections without errors or warnings
+    const extraInvestigationPage = new ExtraInvestigationPage(page);
+    await extraInvestigationPage.fillAndClickNext(noExtraInvestigation);
+
+    const countingDifferencesPollingStationPage = new CountingDifferencesPollingStationPage(page);
+    await countingDifferencesPollingStationPage.fillAndClickNext(noDifferences);
+
+    // Fill form with data that results in errors and a warning
+    const votersAndVotesPage = new VotersAndVotesPage(page);
+    const voters = {
+      poll_card_count: 90,
+      proxy_certificate_count: 0,
+      total_admitted_voters_count: 100,
+    };
+    const votes = {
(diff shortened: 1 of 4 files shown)


praseodym (dismissed):

jschuurk-kr:
Did some exploratory testing on CSB data entry. Tested combinations of warnings and errors, F.401. Scrolling worked as expected.

frontend/e2e-tests/tests/data-entry/data-entry.e2e.ts
```diff
@@ -872,6 +872,85 @@ test.describe("errors and warnings", () => {
...
+      total_votes_cast_count: 50,
+    };
+    await votersAndVotesPage.fillInPageAndClickNext(voters, votes);
+
+    // Expect errors and a warning, should scroll up
+    await expect(votersAndVotesPage.error).toBeInViewport();
+    await expect(votersAndVotesPage.warning).toBeInViewport();
+    await expect(votersAndVotesPage.acceptErrorsAndWarnings).toBeVisible();
+    await expect(votersAndVotesPage.acceptErrorsAndWarnings).not.toBeInViewport();
+
```
oliver3, line 910: Because at the end you check for the `scrollY` not to be zero, I would expect something like this when it _did_ scroll to the top

```suggestion
    expect(await page.evaluate(() => window.scrollY)).toBe(0);

```

stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/3229/commits/d4dde72eaf43b8178314b0086b9eb3b08b07df7f

oliver3 (approved):

jschuurk-kr (approved):
Approving based on @praseodym's approval on commit a1b452a06811f1054f0a3ada4a486b9dd422feb5 and my review of commit d4dde72eaf43b8178314b0086b9eb3b08b07df7f.

PR #3240, opened 2026-04-30
Implement generating results EML for CSB

### LINKED ISSUE #3110: Add generation of EML file to `election_download_zip_results_csb` endpoint
- Add a zip to the `vaststelling-uitslag_gr2026_gemeentenaam_gemeente_gemeente-naam-jjjjmmdd-uummss.zip` and in it the EML file
  - Zip file name: `Resultaat_GR2026_GemeenteNaam.zip`
    - EML file name: `Resultaat_GR2026_GemeenteNaam.eml.xml`
  - Save EML with `CsbResultsEml` file type
- Expand `test_csb_election_zip_download_results_works` integration test to check the xml
- Add 520 hash to P 22-2 pdf and P 22-2 bijlage 1 pdf

rnijveld:
Fixes #3110 

This PR mostly follows the existing structure, a rewrite of the structure to be better abstracted would be better, but that was put in a separate issue to save some time: #3241

I've opted to rework some of the `Into` implementations to work on references instead, because the apportionment keeps references to the results struct at the same time, making it so that consuming functions won't work at the same time. This means some additional clones, but most of the data was `Copy` already, so there isn't that much difference.

Testing:
- When completing an election and downloading the files you should now also get the results EML

### DIFF
--- a/backend/src/api/report.rs
+++ b/backend/src/api/report.rs
@@ -1,3 +1,4 @@
+use apportionment::ApportionmentOutput;
 use axum::{
     extract::{Path, State},
     response::IntoResponse,
@@ -34,7 +35,7 @@ use crate::{
             votes_table::{VotesTables, VotesTablesWithPreviousVotes},
         },
         polling_station::PollingStation,
-        results::Results,
+        results::{Results, political_group_candidate_votes::PoliticalGroupCandidateVotes},
         role::Role,
         summary::{ElectionSummary, ElectionSummaryCSB},
     },
@@ -128,10 +129,12 @@ impl ResultsInput {
             None
         };
 
+        let summary = ElectionSummary::from_results(&election, &results)?;
+
         Ok(ResultsInput {
             committee_session,
             previous_summary,
-            summary: ElectionSummary::from_results(&election, &results)?,
+            summary,
             created_at,
             election,
             polling_stations,
@@ -158,6 +161,18 @@ impl ResultsInput {
         slugify_filename(&format!(
             "{} {}{} {}.eml.xml",
             xml_count_base_name(&self.election),
+            self.election.category.to_eml_code(),
+            self.election.election_date.year(),
+            location_without_whitespace
+        ))
+    }
+
+    fn xml_results_filename(&self) -> String {
+        use chrono::Datelike;
+        let location_without_whitespace: String =
+            self.election.location.split_whitespace().collect();
+        slugify_filename(&format!(
+            "Resultaat {}{} {}.eml.xml",
             self.election.category.to_eml_code(),
             self.election.election_date.year(),
             location_without_whitespace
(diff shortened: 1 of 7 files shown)


backend/src/api/report.rs
```diff
@@ -1171,23 +1251,27 @@ mod tests {
...
         for _ in 1..=2 {
-            let (pdf, eml_total_counts, _) =
+            let (eml_results, pdf, eml_total_counts, _) =
                 get_files_csb_election(&pool, audit_service.clone(), CommitteeSessionId::from(801))
                     .await
                     .expect("should return files");
             let pdf = pdf.expect("should have generated pdf");
             let eml_total_counts =
                 eml_total_counts.expect("should have generated total counts eml");
+            let eml_results = eml_results.expect("should have generated eml");
```
Lionqueen94:
```suggestion
            let eml_results = eml_results.expect("should have generated results eml");
```

backend/src/api/report.rs
```diff
@@ -487,20 +512,55 @@ async fn generate_and_save_files_gsb_election(
...
     audit_service: &AuditService,
     committee_session: CommitteeSession,
     input: ResultsInput,
 ) -> Result<(Option<File>, Option<File>, Option<File>), APIError> {
-    let eml_file: Option<File> = None;
+    if input.election.committee_category != CommitteeCategory::CSB {
+        return Err(APIError::DataIntegrityError(
+            "Generating CSB files can only be done for CSB elections".to_string(),
+        ));
+    }
```
Lionqueen94, line 532: Should we add this check for GSB in the function `generate_and_save_files_gsb_election` as well?

backend/src/api/report.rs
rnijveld: Applied in 04689a7ab4cfcf4eccbb471447a55738e03ab444

backend/src/api/report.rs
rnijveld: Added in 04689a7ab4cfcf4eccbb471447a55738e03ab444

backend/src/api/report.rs
```diff
@@ -337,16 +349,18 @@ impl ResultsInput {
...
 
-    fn into_pdf_file_models_csb(
-        self,
-        // xml_hash: impl Into<String>,
+    fn as_pdf_file_models_csb(
+        &self,
+        apportionment_result: &ApportionmentOutput<'_, PoliticalGroupCandidateVotes>,
+        xml_hash: impl Into<String>,
     ) -> Result<PdfModelListCSB, APIError> {
         // let hash = xml_hash.into();
```
Lionqueen94:
```suggestion
```

backend/src/eml/mod.rs
```diff
@@ -665,6 +669,118 @@ impl ElectionWithPoliticalGroups {
...
+            }
+
+            builder.build()?
+        })
+        .elected(YesNoType::new(true))
+        .ranking(if preferentially_selected {
+            RankingType::Second
+        } else {
+            RankingType::First
+        })
```
Lionqueen94, line 782: Shouldn't this be the other way around? The epic scope states: `Note: ranking 1=preferential candidate nomination, 2=other candidate nomination`

backend/src/api/report.rs
stacktraceghost, line 662: `created_at` doesn't check `results_eml_file`

rnijveld: Fixed in b3a158e720c3c0f73b9a6fd6c86fc261e7bf32fe

backend/src/api/report.rs
rnijveld: Fixed in b3a158e720c3c0f73b9a6fd6c86fc261e7bf32fe

backend/src/eml/mod.rs
rnijveld: Fixed in b3a158e720c3c0f73b9a6fd6c86fc261e7bf32fe

praseodym (approved):

Lionqueen94 (approved):

PR #3249, opened 2026-05-01
Added initial highest average steps and footnotes to backend P 22-2 input

### LINKED ISSUE #3228: Add highest averages table and footnotes to backend P 22-2 input
- [ ] Add `highest_averages_table` in backend and update P 22-2 inputs and template accordingly -> add marking of average of selected list number as well so typst can make the average bold if true
- [ ] Add `footnotes` in backend and update P 22-2 inputs and template accordingly
- [ ] Remove `seat_assignment`, `result_changes_full_seats` and `result_changes_residual_seats` from P 22-2 inputs
- [ ] Change `election` to be of type `Election`, not `ElectionWithPoliticalGroups` (`election.political_groups` won't be needed anymore once all `input.election.political_groups.find()` calls have been replaced) and update P 22-2 inputs and template accordingly

Lionqueen94:
Closes #3228 

## Description
- Added `initial_highest_average_steps` to `EnrichedSeatAssignment` in backend
- Fixed rendering of chosen average
- Added `footnotes` field in backend
- Removed now obsolete `seat_assignment`, `result_changes_full_seats` and `result_changes_residual_seats` fields
- Changed `election` field to be of type `Election` instead of `ElectionWithPoliticalGroups`
- Renamed `enriched_seat_assignment` field to `seat_assignment`


backend/src/domain/models/footnotes.rs
```diff
@@ -0,0 +1,127 @@
...
+use crate::{
+    APIError,
+    domain::{
+        apportionment::{SeatAssignment, SeatChangeStep},
+        election::{PGNumber, PoliticalGroup},
+    },
+};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Footnotes {
```
praseodym: Maybe name this something like `ApportionmentFootnotes` because it only applies to apportionment?

Lionqueen94: Done in 462c860f5e508e60ec550c8b1019af4746a69cfa

backend/src/domain/apportionment.rs
```diff
@@ -78,9 +78,23 @@ impl SeatChange {
             Self::ListExhaustionRemoval(_) => unimplemented!(),
         }
     }
+
+    pub fn list_number_retracted(&self) -> PGNumber {
+        match self {
+            Self::HighestAverageAssignment(_) => unimplemented!(),
+            Self::UniqueHighestAverageAssignment(_) => unimplemented!(),
+            Self::LargestRemainderAssignment(_) => unimplemented!(),
```
praseodym: Should this be `unreachable!()` instead?

backend/src/domain/models/enriched_seat_assignment.rs
```diff
@@ -45,18 +45,26 @@ fn get_initial_steps(seat_assignment: &SeatAssignment) -> InitialSteps<'_> {
...
+        initial_highest_average_steps,
     }
 }
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct EnrichedSeatAssignment {
     quota: DisplayFraction,
     list_seat_assignment: Vec<EnrichedListSeatAssignment>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    initial_highest_average_steps: Option<Vec<SeatChangeStep>>,
```
praseodym, line 57: Looks like this is never `None`

stacktraceghost (dismissed):

backend/src/domain/models/apportionment_footnotes.rs
```diff
@@ -0,0 +1,127 @@
...
+        } else {
+            None
+        }
+    }
+
+    pub fn get_exhausted_lists(
+        list_exhaustion_steps: Vec<&SeatChangeStep>,
+        political_groups: &[PoliticalGroup],
+    ) -> Option<Vec<ExhaustedList>> {
+        let mut unique_exhausted_list_numbers = Vec::new();
```
praseodym: Use a set (`BTreeSet` or `HashSet`) if you want a set of unique values

backend/src/domain/apportionment.rs
Lionqueen94: Changed in 148f3ed038a5178cc7974e606c00d0d9ea48e278

backend/src/domain/models/enriched_seat_assignment.rs
Lionqueen94: Fixed in 148f3ed038a5178cc7974e606c00d0d9ea48e278

backend/src/domain/models/apportionment_footnotes.rs
Lionqueen94: Changed in 148f3ed038a5178cc7974e606c00d0d9ea48e278

backend/apportionment/src/seat_assignment/structs.rs
```diff
@@ -153,16 +154,16 @@ impl<LN: Copy> SeatChange<LN> {
...
 
     /// Get the list number for the list this step has retracted a seat from
     pub fn list_number_retracted(&self) -> LN {
         match self {
-            Self::HighestAverageAssignment(_) => unimplemented!(),
-            Self::UniqueHighestAverageAssignment(_) => unimplemented!(),
-            Self::LargestRemainderAssignment(_) => unimplemented!(),
+            Self::HighestAverageAssignment(_) => unreachable!(),
+            Self::UniqueHighestAverageAssignment(_) => unreachable!(),
+            Self::LargestRemainderAssignment(_) => unreachable!(),
```
cikzh: Just a matter of taste, but you could opt for this syntax:
```rust
Self::HighestAverageAssignment(_)
  | Self::UniqueHighestAverageAssignment(_)
  | Self::LargestRemainderAssignment(_) => unreachable!(),
            
```
Doesn't impact anything meaningful though, so up to you

cikzh (dismissed):
Only one unmeaningful code style comment, if we can move this PR forward, please don't bother to fix :)

backend/src/domain/models/apportionment_footnotes.rs
stacktraceghost: Note that it now compares on both `number` and `name`. Also, for the name we are using `find`, so it would always return the first political group. 

Theoretically if `political_groups` would contain two groups with same numer: `1 - List A` and  `1 - List B`:
- The name would select the first one due to `find`
- The BTreeSet would still contain `1 - List A`

So it works, but it does prefer the first list found, `List A`. 

If we would avoid find, use filter and expect that there is only 1 result, this wouldn't occur. Currently it works on the assumption that `political_groups` never contains a duplicate list number.

stacktraceghost: Made a little local spike which could solve this elegantly:

```rs
/// Extension trait for [Iterator] that asserts the iterator yields exactly one item.
pub trait IteratorExt: Iterator {
    /// Returns the single item in the iterator, panicking with `msg` if zero
    /// or more than one item is yielded.
    fn expect_one(self, msg: &str) -> Self::Item
    where
        Self: Sized,
    {
        let mut it = self;
        let first = it.next().unwrap_or_else(|| panic!("{msg}"));
        assert!(it.next().is_none(), "{msg}");
        first
    }
}

impl<I: Iterator> IteratorExt for I {}
```

Then you could do this (`find` becomes `filter`, `expect` becomes `expect_one`):
```rs
.filter(|pg| pg.number == pg_number)
.expect_one("exactly one political group should match the number")
```

We could also use it elsewhere in the codebase.

praseodym: > Note that it now compares on both `number` and `name`.

That can be solved by implementing the [`Ord`](https://doc.rust-lang.org/stable/std/cmp/trait.Ord.html) trait on `ExhaustedList`, or by using a [`BTreeMap`](https://doc.rust-lang.org/stable/std/collections/struct.BTreeMap.html) with `pg_number` as a key and `name` as a value (removing the `ExhaustedList` struct). We can then rely on the guarantees provided by the data structure, rather than introducing read-time assertions that can panic.

praseodym (dismissed):

backend/apportionment/src/seat_assignment/structs.rs
Lionqueen94: Changed in e68f4dc966c487e639559be14a4082aac0fdce7e

backend/src/domain/models/apportionment_footnotes.rs
Lionqueen94: I switched from a BTreeSet to a BTreeMap in 65b55ab5d7115b815f082f256423d9afd5fc3059

praseodym (dismissed):

praseodym (approved):

stacktraceghost (approved):

PR #3253, opened 2026-05-01
Refactor API error handling

praseodym:
- Refactor API error handling with `ApiErrorResponse` trait and delegate authentication, apportionment and committee session API errors to this trait.
- Remove unused `APIError::InvalidResultsType` enum variant

This resolves a part of the points mentioned in #2884, but not e.g. `APIError` usage in the domain layer.

PR #3260, opened 2026-05-04
Refactor `error_response` to `ErrorResponse::new`

praseodym:
Resolves https://github.com/kiesraad/abacus/pull/3253#discussion_r3180105799

### DIFF
--- a/backend/src/error.rs
+++ b/backend/src/error.rs
@@ -96,9 +96,9 @@ pub struct ErrorResponse {
 }
 
 impl ErrorResponse {
-    pub fn new(error: String, reference: ErrorReference, fatal: bool) -> Self {
+    pub fn new(error: impl ToString, reference: ErrorReference, fatal: bool) -> Self {
         Self {
-            error,
+            error: error.to_string(),
             reference,
             fatal,
         }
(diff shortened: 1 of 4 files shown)


backend/src/error.rs
```diff
@@ -95,6 +95,16 @@ pub struct ErrorResponse {
     pub reference: ErrorReference,
 }
 
+impl ErrorResponse {
+    pub fn new(error: String, reference: ErrorReference, fatal: bool) -> Self {
+        Self {
+            error,
```
cikzh: Maybe we could do this to get rid of all `String::from("text")` and `.to_string()` calls? 
```suggestion
    pub fn new(error: impl ToString, reference: ErrorReference, fatal: bool) -> Self {
        Self {
            error: error.to_string(),
```

cikzh (dismissed):

backend/src/error.rs
cikzh: @praseodym I've created https://github.com/kiesraad/abacus/pull/3263 to address this. If you agree with the change, you can just merge it into  this one

praseodym: Thanks a lot!

stacktraceghost (approved):

oliver3 (approved):

Issue #3276, opened 2026-05-08
Generating of CSB documents is broken


jschuurk-kr:
When you try to download any of the three zip-files with CSB documents, you get a 500 response:  
`{"error":"Internal server error","fatal":false,"reference":"PdfGenerationError"}`  
(Note that the json response is shown in the browser, not the "Abacus is stuk"-page. Reason for this is that the download is a direct API call to the backend.)

Backend error:
```
2026-05-08T07:28:40.745374Z ERROR request{method=GET uri=/api/elections/8/committee_sessions/801/download_zip_results_csb version=HTTP/1.1}: abacus::error: Pdf generation error: CompilationFailed("Typst command failed: error: unexpected argument: by\n
    ┌─ model-p-22-2.typ:434:125\n    │\n
434 │   ..for list_candidate_nomination in input.candidate_nomination.list_candidate_nomination.sorted(key: lcn => lcn.list_seats, by: (l, r) => l >= r) {\n
    │                                                                                                                              ^^^^^^^^^^^^^^^^^^^^\n
\n")
```

### Steps to reproduce
1. start a clean Abacus instance (`cargo run -- --reset-database --seed-data`)
2. as coordinator open election "Test Election >= 19 seats"
3. complete the committee session and generate the documents
4. click any of the three links to download the documents

Reproducible both on `main` (d76c798b) and on PR #3249. Interestingly enough, the CSB full-flow test in PR #3274 does not fail.

praseodym:
The `by` argument for `sorted` was added in Typst 0.14.0 (from the [0.14.0 changelog](https://typst.app/docs/changelog/0.14.0/#scripting): "Added by parameter to [array.sorted](https://typst.app/docs/reference/foundations/array/#definitions-sorted) for sorting with a comparison function"). This means that when running Abacus locally without the `embed-typst` feature, Typst 0.14.0 or later is required or this error will be thrown.

jschuurk-kr:
Updating from Typst 0.13.1 to 0.14.2 indeed resolved the issue. Closed.

PR #3279, opened 2026-05-08
Always use embedded Typst

praseodym:
Remove code paths to use an external Typst binary and make embedded Typst the default. This prevents mistakes like using the wrong Typst version (e.g. #3276, happened to other developers too) and simplifies the codebase.

The original reason behind the external Typst code path was that it is faster. Since then we introduced the `pdf_gen` crate that contains all Typst dependencies which should (theoretically) improve build times, but the linker will still need to link the Typst objects in the final Abacus binary. We should evaluate (incremental) build times and if we decide that the impact on build time is too big then we should not merge this PR.

The 'Diff PDFs' CI workflow fails because it tries to build the `main` branch without the `embed-typst` feature, I think we can ignore this failure because it will be resolved if this PR is merged.


praseodym:
This slows down incremental builds by about 1.5s which was not deemed acceptable.

PR #3280, opened 2026-05-08
Added backend tests for P 22-2 Input

### LINKED ISSUE #3226: Add backend tests for P 22-2 Input tables
- [ ] `enriched_candidate_nomination.rs`
- [ ] `enriched_seat_assignment.rs`
- [ ] `footnotes.rs`

Lionqueen94:
Closes #3226 

- Added tests
- Made sure largest remainder is only added when there is at least one initial largest remainder step


backend/apportionment/src/lib.rs
```diff
@@ -20,7 +20,8 @@ pub use self::{
         HighestAverageAssignedSeat, SeatAssignmentResult, SeatChange, SeatChangeStep,
     },
     structs::{
-        ApportionmentError, ApportionmentInput, ApportionmentOutput, CandidateVotes, ListVotes,
+        ApportionmentError, ApportionmentInput, ApportionmentOutput, CandidateVotes,
+        LARGE_COUNCIL_THRESHOLD, ListVotes,
```
praseodym: Why was `LARGE_COUNCIL_THRESHOLD` added here?

Lionqueen94: Because I use it now in the enriched seat assignment, and it was not available for import without explicitly adding it here. I can also make a duplicate const in the enriched seat assignment file if that's preferable.

praseodym: I see now, the GitHub collapsed the diff of that file so I didn't see that it was used there.

backend/src/domain/models/enriched_seat_assignment.rs
```diff
@@ -143,30 +144,46 @@ impl EnrichedSeatAssignment {
...
-                .find(|(pg_number, _)| *pg_number == pg_votes.number)
-                .map(|(_, largest_remainder)| largest_remainder.clone());
-            let unique_highest_average = Self::get_unique_highest_average(
-                pg_votes.number,
-                initial_full_seats,
-                &largest_remainder,
-                &initial_steps.initial_unique_highest_average_steps,
-            );
+            let largest_remainder =
+                if number_of_seats >= LARGE_COUNCIL_THRESHOLD || seat_assignment.steps.is_empty() {
```
praseodym: Why do we need the `LARGE_COUNCIL_THRESHOLD` logic here? It feels like we are repeating some apportionment logic to build the reporting structs, instead of only using the result of the apportionment crate to do so.

(And also, these are logic changes in a PR for adding tests)

Lionqueen94: I felt like it was a bit of a waste to fill these fields when in typst we check for the threshold and do not use them. But I can try to find a way to check it without using the const

Lionqueen94: Of course it makes more sense to just check if the largest remainder steps is empty or not, so I did that :sweat_smile:

backend/src/domain/models/mod.rs
```diff
@@ -153,3 +153,36 @@ impl PdfModel {
...
+#[cfg(test)]
+pub mod tests {
+    use crate::domain::{
+        election::PoliticalGroup,
+        results::political_group_candidate_votes::{CandidateVotes, PoliticalGroupCandidateVotes},
+    };
+
+    /// Create a value for `political_group_votes` (type `Vec<PoliticalGroup>`)
+    /// for the given political groups, with given candidate votes per list.
+    pub fn create_political_group_votes(
```
oliver3: I would prefer keeping test create functions near their structs, i.e. in this case in `political_group_candidate_votes.rs`, for more easy discovery. 

Maybe you can also reuse `from_test_data_auto()` which does a similar thing for one political group.

backend/src/domain/models/apportionment_footnotes.rs
```diff
@@ -120,3 +120,138 @@ pub struct FootnotePoliticalGroup {
...
+            vec![347, 100],
+            vec![266, 187],
+        ];
+        let election = election_fixture_with_given_number_of_seats(
+            CommitteeCategory::CSB,
+            candidate_votes
+                .iter()
+                .map(|cv| u32::try_from(cv.len()).expect("Should fit in u32"))
+                .collect::<Vec<u32>>()
+                .as_slice(),
```
oliver3: I believe it is more common to use `&` instead of `.as_slice()`

```suggestion
            &candidate_votes
                .iter()
                .map(|cv| u32::try_from(cv.len()).expect("Should fit in u32"))
                .collect::<Vec<u32>>(),
```

oliver3 (dismissed):

backend/src/domain/models/mod.rs
Lionqueen94: Thanks for the tips! I changed it in 4ea634f10751bccf6d4b07c7ba48acb0e6d07fb8

backend/src/domain/models/apportionment_footnotes.rs
Lionqueen94: Changed in 4ea634f10751bccf6d4b07c7ba48acb0e6d07fb8

oliver3 (dismissed):

backend/src/domain/models/apportionment_footnotes.rs
stacktraceghost: Noticed that this and `get_exhausted_lists` is pub, which can be removed. That way `ApportionmentFootnotes:new` can only be used.

Lionqueen94: Good one, removed in 4ea634f10751bccf6d4b07c7ba48acb0e6d07fb8

stacktraceghost (approved):

oliver3 (approved):

PR #3289, opened 2026-05-13
Use a dylib for linking to typst during development

rnijveld:
- Use a dylib instead of the external typst binary during development
- This has the advantage of not needing the external typst binary to be kept up to date
- It should see a similar speedup in iteration speed on incremental updates on the main binary
- The main disadvantage is that the resulting binary is not portable in any way shape or form, given that the Rust dylib format is not stable. So for production builds we should definitely not use the dylib variant. 
- I've made sure that if both dev an static flags are enabled that the static build gets priority.


praseodym:
This looks like a good approach. @rnijveld what needs to be done before we can consider merging this?

This might also be interesting for @kiesraad/e-ks.

praseodym:
@rnijveld CI fails now because even with the `embed-typst` feature enabled something dylib is being done, which is not supported for the musl target:

```
Run cargo --verbose --locked build --target=x86_64-unknown-linux-musl --features memory-serve,embed-typst,storybook --bin abacus
    Updating crates.io index
 Downloading crates ...
  Downloaded libloading v0.9.0
error: cannot produce dylib for `pdf_gen_dylib v1.1.0 (/home/runner/work/abacus/abacus/backend/pdf_gen/dylib)` as the target `x86_64-unknown-linux-musl` does not support these crate types
Error: Process completed with exit code 101.
```

Any ideas why?

praseodym:
@rnijveld because Typst cannot be compiled with ASAN the backend data entry fuzzer broke, so I had to added a stub feature that subs out Typst entirely. What do you think? 

We could make the stub feature the implicit default for when no other features are enabled.

rnijveld:
@praseodym Looks good to me! I can't approve because I'm the original author, but I would if I could!

cikzh (dismissed):

rnijveld:
@praseodym I was helped by @bjorn3 to show that libloading was not necessary for this to provide the speedup we need (as the dylib itself caused the compilation to already be faster, loading the library dynamically was not necessary).

praseodym (dismissed):

praseodym (approved):
Resolved merge conflict

cikzh (approved):

Issue #3067, opened 2026-03-23
Authorisation


jschuurk-kr:
### Test ideas 
- check issue #2947 for what's been tested, potential gaps in coverage, etc. 
- separation GSB and CSB for typist and coordinator
- admin can only reach GSB-pages for GSB election (e.g. investigations) and same for CSB
- check both frontend as backend
- ...

proposed timebox: 60 minutes

oliver3:
## Test frontend urls

I've looked at all frontend routes (urls) with CBS/GSB roles and a CSB/GSB election  in a semi-automated way:
- extracted routes from the frontend code
- Run playwright tests that for a given role, with a given election (either CSB or GSB), see if it is allowed to see the page
- Determine visually if this is the page or error message that was expected

Code snippets to get all routes (put at the end of `app/routes.tsx` and run application frontend)
```tsx
function routePaths(base: string, route: RouteObject): string[] {
  const path = route.path ? `${base}/${route.path}` : base;
  return route.children === undefined ? [path] : route.children.flatMap((r) => routePaths(path, r));
}

console.log(JSON.stringify(routes.flatMap((r) => routePaths("", r))));
```
Basic setup of the playwright test

```ts
const routes = [ /* routes hier */]; 

test.use({  
  storageState: "e2e-tests/state/typist_csb.json",  
});  
  
test.describe("check access", () => {  
  for (const route of routes) {  
    test(`Route ${route} is not accessible for role`, async ({ page, election, dataEntry, newTypist }) => {
      test.skip(route === "/account/logout", "Do not go to /account/logout");  
  
      await page.goto(  
        route  
          .replace(":electionId", String(election.election.id))
          .replace(":pollingStationId", String(dataEntry.source.id))
          .replace(":dataEntryId", String(dataEntry.data_entry_id))
          .replace(":entryNumber", String(1))
          .replace(":committeeSessionId", String(dataEntry.source.committee_session_id))
          .replace(":userId", String(newTypist.id))
          .replace(":sectionId", "differences_counts"),
      );  
      const loginPage = new LoginPgObj(page);  
      await expect(loginPage.heading).toBeVisible({ timeout: 2000 });  
      await expect(loginPage.alert).toContainText("Je hebt geen toegang tot deze pagina");  
    });  
  }  
});
```

I've also had to add `coordinator_csb` and `typist_csb`, and change fixtures to CSB variants when running the test against a CSB election.

## Observations
These are my observations, which are not necessarily issues.

[...]

PR #3302, opened 2026-05-18
AuthorizationGuard fix, remove /account

cikzh:
Fixes to findings found in exploratory testing in #3067
Does not close the issue, as there is still a point of discussion there.
This PR:
- removes the /account endpoint
- added checks in AuthorizationGuard.tsx to check if the correct handle fields are present

Biome and Sigrid are complaining about complexity. I've added a Biome suppress comment, because I think splitting the function would actually decrease readability. Feel free to add a commit to this PR if you have ideas about reducing the complexity scores.


stacktraceghost:
Could you provide an example path that has been fixed?

jschuurk-kr:
> Could you provide an example path that has been fixed?

https://github.com/kiesraad/abacus/issues/3067#issuecomment-4477914148

oliver3:
An example is the one from the screenshot in the original issue : https://alsmede-bevoegd-toelating.abacus-test.nl/elections/1/investigations/print-corrigendum

which still seems to be broken unfortunately @cikzh 

Edit: fixed in https://github.com/kiesraad/abacus/pull/3302/commits/7b978f6b902b03a1960c6a3612f891fdb871d779

https://zetel-europees-benoeming.abacus-test.nl/elections/1/investigations/print-corrigendum

oliver3 (dismissed):

praseodym:
> Biome and Sigrid are complaining about complexity. I've added a Biome suppress comment, because I think splitting the function would actually decrease readability. Feel free to add a commit to this PR if you have ideas about reducing the complexity scores.

I split out the redirect logic in 71d2e838d954d6ab2dbe5da90ae37d321247dc47, please take a look

oliver3 (approved):

Lionqueen94 (approved):

PR #3303, opened 2026-05-18
Frontend: Add deceased candidates flow

### LINKED ISSUE #3160: Frontend: Add deceased candidates flow
> Added an alternative design that won't require us to build a search component. Both flows can be [evaluated here in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17533-40475&t=QrIZ8mZZ0baTv2cJ-11). 
> 
> I think the 'browse' variant is a bit more dull, but more robust and only uses components that we already have or can reuse later for other features.  

 _Originally posted by @jorisleker in [#797](https://github.com/kiesraad/abacus/issues/797#issuecomment-4225043407)_
 
- Design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17533-40475&t=Vs91HeLWnDBxu9XR-4
- Create reuseable candidate component to select candidate according to design
- Call update apportionment state to save deceased candidates
- Redirect apportionment page to deceased candidates flow while deceased candidates flow not completed
- Add playwright test for deceased candidates flow

Lionqueen94:
Closes #3160 

- [x] Design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17533-40475&t=Vs91HeLWnDBxu9XR-4
  - [x] Include all candidates page
  - [x] Add deceased candidate page
  - [x] Deceased candidates page
- [x] Create reusable candidate component to select candidate according to design (make clickable candidate optional)
- [x] Call backend endpoints
- [x] Redirect apportionment page to deceased candidates flow while deceased candidates flow not completed
- [x] Expand full flow e2e test
- [x] Add unit tests for new pages
- [x] Made "underlined" button variant also underlined not on hover
- [x] Fix flickering of table as mentioned in https://github.com/kiesraad/abacus/pull/3303#issuecomment-4480521352


jorisleker:
Did a quick review. Looks good!

Things I would change:
- open list 1 by default, this makes the interaction clearer then just showing the menu with all parties
- when I select a candidate and return to the overview of deceased candidates, the table is visible during rendering and this results in a jumpy layout (see screen recording). Can we prevent this?

https://github.com/user-attachments/assets/c4a9de3f-ddc3-424f-bafc-c451971f6706

Lionqueen94:
> Did a quick review. Looks good!
> 
> Things I would change:
> 
>     * open list 1 by default, this makes the interaction clearer then just showing the menu with all parties
> 
>     * when I select a candidate and return to the overview of deceased candidates, the table is visible during rendering and this results in a jumpy layout (see screen recording). Can we prevent this?
> 
> 
> Schermopname.2026-05-18.om.20.09.46.mov

Point one is changed in ad33fa1beb9f29fa23f9a4464687edbeb95cde16, the rendering of the table without flickering I haven't been able to fix yet (added to the PR description)

jorisleker (dismissed):
UX is as designed now 👌

frontend/src/components/ui/ProgressList/ProgressList.tsx
```diff
@@ -54,8 +56,17 @@ ProgressList.Item = function ProgressListItem({
...
+        cls.item,
+        { active: active },
+        status,
+        { disabled: disabled },
+        addSpace && cls.addSpace,
+        onClick && cls.pointer,
+      )}
       aria-current={active ? "step" : false}
+      onClick={onClick}
+      onKeyDown={onClick} // TODO: Is this fine? Biome said "Enforce to have the onClick mouse event with the onKeyUp, the onKeyDown, or the onKeyPress keyboard event."
```
oliver3: Ideally we wouldn't make the `<li>` clickable but have a button (with no button styling) inside of the `<li>` with the onClick() which will take care of correct a11y for us.

But both adding this onKeyDown and disabling the biome rule would be ok I think.

Lionqueen94: As discussed, I now added a button in the `<li>` and made sure the styling is adjusted. This should be a11y compliant and work as designed. See b06d1ff5451ccb8bbbfc0ecc4068ed411de0482e

oliver3 (approved):

frontend/src/components/ui/ProgressList/ProgressList.tsx
oliver3: Looks good!

jorisleker (approved):

PR #3319, opened 2026-05-21
Changing usage of number of voters to be more explicit. Resolves #3297

### LINKED ISSUE #3297: Refactor ElectionSummary to split between GSB/CSB
Data entry results have been split up between GSB and CSB (i.e. as an enum) because there are different fields depending on whether your election has the CSB/GSB type. This distinction disappears with the ElectionSummary. For consistency it would be better to continue this distinction here as well. Specifically: for a CSB the ElectionSummary has a required number_of_voters field, whereas for a GSB this field is absent. This is currently encoded as an Option field in ElectionSummary, but it would make more sense to have ElectionSummary be an enum where the CSB variant has a non-optional number_of_voters and the GSB variant has no field for number_of_voters at all.

mohs8421:
Resolves #3297 

As the actual issue was a missing clarification on why in this case it is following a different path than in another case, this change makes it more explicit to understand why it is the way it is. Additionally I add a test for this behaviour.

### DIFF
--- a/backend/src/eml/mod.rs
+++ b/backend/src/eml/mod.rs
@@ -536,10 +536,7 @@ impl ElectionWithPoliticalGroups {
     ) -> Result<ElectionCountContest, EMLError> {
         let builder = ElectionCountContest::builder()
             .identifier(ContestIdentifier::geen())
-            .total_eligible_voter_count(match summary.number_of_voters {
-                Some(n) => n,
-                None => self.number_of_voters,
-            })
+            .total_eligible_voter_count(self.get_eligible_voter_count(summary))
             .total_candidate_votes_count(summary.votes_counts.total_votes_candidates_count)
             .total_rejected_votes(
                 RejectedVotesReason::Blank,
@@ -586,6 +583,13 @@ impl ElectionWithPoliticalGroups {
         };
 
         builder.build()
+    }
+
+    /// Depending on the context of the summary coming from GSB or from CSB, the number of voters source is different.
+    /// In case of a GSB summary, the number of voters there should be the valid source.
+    /// Otherwise, it is the given number of voters.
+    fn get_eligible_voter_count(&self, summary: &ElectionSummary) -> u32 {
+        summary.number_of_voters.unwrap_or(self.number_of_voters)
     }
 
     fn as_eml_count_selections(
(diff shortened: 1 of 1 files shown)


backend/src/eml/mod.rs
```diff
@@ -934,4 +941,99 @@ mod tests {
         dbg!(&res);
         assert!(matches!(res, EMLImportError::OnlyMunicipalSupported));
     }
+
+    #[test]
```
Lionqueen94, line 948:
```suggestion
    #[allow(clippy::too_many_lines)]
    #[test]
```

Change and test look good, the test is too long, but it's fine to add the allow :)

backend/src/eml/mod.rs
```diff
@@ -588,6 +585,13 @@ impl ElectionWithPoliticalGroups {
         builder.build()
     }
 
+    /// Depending on the context of the summary comming from GSB or from CSB, the number of voters source is different.
+    /// In case of a GSB summary, the number of voters there should be the valid source.
+    /// Otherwise, it is the given number of voters.
```
Lionqueen94:
```suggestion
    /// Depending on the context of the summary coming from GSB or from CSB, the number of voters source is different.
    /// In case of a GSB summary, the number of voters there should be the valid source.
    /// Otherwise, it is the given number of voters.
```

backend/src/eml/mod.rs
```diff
@@ -908,6 +912,9 @@ pub fn polling_stations_eml_matches_election(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::domain::election::{ElectionCategory, ElectionWithPoliticalGroups};
+    use crate::domain::summary::{SumCount, SummaryDifferencesCounts};
```
Lionqueen94:
```suggestion
    use crate::domain::{
      election::{ElectionCategory, ElectionWithPoliticalGroups};
      summary::{SumCount, SummaryDifferencesCounts}
    };
```

Lionqueen94 (commented):
Just 2 small improvements

mohs8421:
I signed the changes and squashed the smaller fixes.

Lionqueen94 (dismissed):

backend/src/eml/mod.rs
```diff
@@ -908,6 +912,11 @@ pub fn polling_stations_eml_matches_election(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::domain::{
+      election::{ElectionCategory, ElectionWithPoliticalGroups};
+      summary::{SumCount, SummaryDifferencesCounts}
```
Lionqueen94:
```suggestion
      election::{ElectionCategory, ElectionWithPoliticalGroups},
      summary::{SumCount, SummaryDifferencesCounts},
```
Sorry my suggestion was missing the replacement of `;` for `,`. See the `cargo fmt` output

Lionqueen94 (approved):

rnijveld (approved):

PR #3327, opened 2026-05-21
Make autoComplete an inputProperty of InputField and set it for login related inputs

### LINKED ISSUE #2640: Improve experience when using password manager
Using the browser's password manager with Abacus is not a good experience right now, as [described here](https://github.com/kiesraad/abacus/issues/2471#issuecomment-3647174520). We should improve this, but the experience does not need to be perfect. Reason for that is that we instruct admins to disable password managers.

### What a better experience would look like
- When an admin or coordinator sets a temporary password for a user, the field is not filled and the password is not saved. 
- When a user needs to set a new password, the field "Kies nieuw wachtwoord" is not filled.
- When a user logs in, the username and password are either saved (first time) or filled in.
- When a user sets a new password, the new password can be saved in the password manager.
- The password manager does not confuse the username and full name fields.

### Suggested implementation

To fix the frontend: we should use the autocomplete attribute to properly describe what values are expected in each field:

- autocomplete="new-password" for fields where users enter a new password,
- autocomplete="current-password" for the password field,
- autocomplete="username" for the username field,
- autocomplete="name" for the user's fullname.

(See [this spec](https://www.w3.org/TR/wcag/#input-purposes) for documentation on valid values for the autocomplete attribute.)

This will make sure password managers don't autocomplete a users fullname in a username field and vice versa.

Note that an additional change is needed on the username field when setting a new password. Currently that's a `div`, but [`autocomplete`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete) is only available on `<input>` elements that take a text or numeric value as input, `<textarea>` elements, `<select>` elements, and `<form>` elements. So probably the best option here is to change it to an `<input>` element with [`readonly`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly).

atezet:
Fixes #2640 .

So, I simply made `autoComplete` an input property of `InputField`, with `"off"` as the default and set it for the ones listed in the ticket.

### DIFF
--- a/frontend/src/components/ui/InputField/InputField.tsx
+++ b/frontend/src/components/ui/InputField/InputField.tsx
@@ -40,6 +40,7 @@ export function InputField({
   disabled = false,
   margin = "mb-0",
   autoFocus,
+  autoComplete = "off",
   numberInput,
   hideErrorMessage,
   ...inputFieldProps
@@ -51,7 +52,7 @@ export function InputField({
     value,
     autoFocus,
     defaultValue,
-    autoComplete: "off",
+    autoComplete,
     "aria-invalid": !!error,
     "aria-errormessage": error ? `${name}-hint_or_error` : undefined,
   };
--- a/frontend/src/features/account/components/AccountSetupForm.tsx
+++ b/frontend/src/features/account/components/AccountSetupForm.tsx
@@ -102,6 +102,7 @@ export function AccountSetupForm({ user, onSaved }: AccountSetupFormProps) {
               label={t("account.fullname")}
               subtext={t("account.fullname_subtext")}
               hint={t("account.fullname_hint")}
+              autoComplete="name"
               error={validationErrors?.fullname}
               defaultValue={user.fullname}
             />
@@ -110,6 +111,7 @@ export function AccountSetupForm({ user, onSaved }: AccountSetupFormProps) {
               name="username"
               label={t("account.username")}
               hint={t("account.username_hint")}
+              autoComplete="username"
               value={user.username}
               disabled
             />
@@ -119,6 +121,7 @@ export function AccountSetupForm({ user, onSaved }: AccountSetupFormProps) {
               label={t("account.password_new")}
               hint={t("account.password_hint")}
               type="password"
+              autoComplete="new-password"
               error={validationErrors?.password}
             />
             <InputField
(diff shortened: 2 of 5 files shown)


jorisleker (dismissed):
I think we should add `autocomplete=off` to all input fields in abacus. We never want browsers or password managers to autocomplete fields. 

However: password manager and browsers [are known](https://stackoverflow.com/questions/63792397/prevent-autofill-of-passwords-for-all-browsers) for ignoring these attributes, so chances are it won't completely fix the issue.

atezet:
@jorisleker That is what it was before I made this change. But you're right, browsers ignore that anyway.

Lionqueen94:
> I think we should add `autocomplete=off` to all input fields in abacus. We never want browsers or password managers to autocomplete fields.
> 
> However: password manager and browsers [are known](https://stackoverflow.com/questions/63792397/prevent-autofill-of-passwords-for-all-browsers) for ignoring these attributes, so chances are it won't completely fix the issue.

It was already set to "off" in main, and it was still prompting users when password managers were not disabled. So it seems like they indeed ignore the attribute (if set to "off" at least).

jorisleker:
> @jorisleker That is what it was before I made this change. But you're right, browsers ignore that anyway.

Sorry, feedback was based on the comment of your PR, see the code changes now, and they implement best practice. Will change my review to an approval

jorisleker (approved):

Lionqueen94 (approved):

praseodym:
@atezet this PR still has unsigned commits, could you rebase and sign the commits?

oliver3:
Tested it with 1Password against what is described as a better experience in https://github.com/kiesraad/abacus/issues/2640:
- ✅ When an admin or coordinator sets a temporary password for a user, the field is not filled and the password is not saved.
- ✅ When a user needs to set a new password, the field "Kies nieuw wachtwoord" is not filled. ✅ 
- ✅ When a user logs in, the username and password are either saved (first time) or filled in. 
- 🟧 When a user sets a new password, the new password can be saved in the password manager.
  Still confusing username and fullname, see below
- 🟥 The password manager does not confuse the username and full name fields.
    I have the idea that because the username is not an input, the fullname is still used instead, and 1pw wants to create a new entry with the fullname as username

This PR still improves the experience, thanks you!

atezet:
@praseodym done!

Issue #3293, opened 2026-05-15
Allow creating on the fly backups


rnijveld:
Currently creating backups involves pausing all input and interactions by users and the administrator copying the sqlite database files. This is not ideal, as it is disruptive to the input process. It would be much better if an administrator could be able to create on the fly backups directly from within Abacus.

To be determined:
- Is this an option that should be available in the user interface for an administrator?

Related to #3231

praseodym:
Related to #3231

praseodym:
Closing in favour of #3380

PR #3329, opened 2026-05-21
Obsolete: Add database backup system

### LINKED ISSUE #3231: Periodically backup Abacus database
Admins want to periodically back up the Abacus database, ideally without disturbing data entry. 

Suggested before: use `VACUUM INTO ...`

To be decided:
- how often do we backup?
- how do we prevent consecutive backups to fill up available disk space?
  - The size of the Utrecht GR26 database after the first committee session was 32 MB, of which ~22 MB was the EML file.
- how do we show to the admin / coordinators that successfull backups where made? And how do we alert them if it failed?

Considerations
- Ideally, backup does not lead to locking the database so it can be done 'live' / 'without logging out users'

Hiddentale:
## Description
Adds automatic periodic database backups to the backend. A background task runs every 5 minutes and creates a SQLite backup file using `VACUUM INTO`, stored in a `database_backups` directory next to the executable. A retention policy keeps at most 5 backups and removes the oldest when the limit is exceeded. All backup logic lives in `backend/src/infra/backup.rs` and is wired up in `abacus.rs`.

All these constants can be changed in `backup.rs`.

Errors are logged and discarded. A failed backup only means that backup is skipped, the application keeps running normally. Every functionality is unit tested (except if I potentially overlooked some edge case)

## Related Issues
Partially Implements #3293
Closes #3231

## Questions
This PR does not add any frontend capabilities for enacting backups. The frontend user also does not get a notification if a backup is successful. Should this be added in this PR? If so, should the frontend user's backups go in a separate folder from the automatic periodic backups?

I hope the code is up to your standards, any feedback on improvements are highly appreciated!


Hiddentale:
I messed up rebasing when trying to make my previous commits have a verified checkmark And hence pulled in some commits of others. I have no clue how to fix this.

backend/src/bin/abacus.rs
```diff
@@ -110,5 +114,17 @@
 
     let listener = TcpListener::from_std(socket.into())?;
 
+    let backup_config = BackupConfig::new().expect("Failed to setup backup directory");
+    let backup_pool = pool.clone();
+    tokio::spawn(async move {
+        let mut interval = tokio::time::interval(Duration::from_mins(5));
+        loop {
+            interval.tick().await;
+            if let Err(e) = create_local_backup(&backup_pool, &backup_config).await {
```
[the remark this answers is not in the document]
Hiddentale: False positive? Variable 'e' is clearly being used.

backend/src/infra/backup.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+impl From<sqlx::Error> for BackupError {
+    fn from(err: sqlx::Error) -> Self {
+        BackupError::Database(err)
+    }
+}
+
+pub async fn run_backup_scheduler(backup_pool: SqlitePool, backup_config: BackupConfig) {
+    let mut interval = tokio::time::interval(Duration::from_secs(BACKUP_INTERVAL_IN_MINUTES * 60));
```
Hiddentale: Duration::from_mins was stabilised in Rust 1.91.0 but our MSRV is 1.89.0. 

Hence we will have to use from_secs here and multiply by 60

backend/src/infra/backup.rs
```diff
@@ -0,0 +1,190 @@
...
+    let filename = format!("backup_{}.db", Local::now().format("%Y-%m-%d_%H-%M-%S"));
+    let backup_path = &backup_config.directory.join(filename);
+    backup_database(pool, &backup_path).await?;
+    if count_backups(&backup_config)? > AMOUNT_OF_BACKUP_FILES_ALLOWED {
+        remove_oldest_backup(&backup_config)?;
+    }
+    Ok(())
+}
+
+fn create_backup_directory(backup_config: &BackupConfig) -> Result<(), BackupError> {
```
Hiddentale, line 71: I am not 100% sure if this function is necessary, since it is a thin wrapper. I myself like it since it makes it instantly apparent  what we are doing.

backend/src/infra/backup.rs
```diff
@@ -0,0 +1,190 @@
...
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tokio::time::Duration;
+    use tokio::{sync::Mutex, time::sleep};
+
+    static TEST_LOCK: Mutex<()> = Mutex::const_new(());
```
Hiddentale: I had to lock each unit test so they run sequentially, otherwise I would get weird database/filesystem errors.

backend/src/infra/backup.rs
praseodym: Ah that's not correct, in practice our MSRV is 1.95.0 because of rust-eml-nl: https://github.com/kiesraad/rust-eml-nl/blob/90fec094d3ea04a9ae11aaf34e991c29b73c55ac/Cargo.toml#L6

Hiddentale: I got this from a clippy warning from the workflows:
https://github.com/kiesraad/abacus/actions/runs/26808584242/job/79033665017

So there might be something wrong there then, or due to my inexperience I am misunderstanding

praseodym: #3395 should fix this!

backend/src/infra/backup.rs
Hiddentale: This is now not used anymore, since this was ignored by GitHub actions for some reason and hence the tests still ran in parallel.

PR #3334, opened 2026-05-21
Compilation performance improvements

### LINKED ISSUE #3295: Compilation performance improvements
Abacus is not the fastest piece of software to compile (but not the slowest either). Faster compile times allow for faster feedback loops during development. Any suggestion to improve this is welcome. But it would probably be best if someone could make a more detailed analysis of what our root causes for compilation performance are.

Some ideas:

- Split up more of the codebase into separate crates to allow more parallel compilation to take place
- Replace some generics and `impl T` with `Box<dyn T>` and `&dyn T` to reduce the amount of monomorphization taking place, given that our runtime performance is currently more than good enough.

If you do any larger refactors, make sure to discuss these with the Abacus team before implementing them.

mikebohdan:
I trimmed unused features from tokio, hyper, and tracing-subscriber crates. This change cut build time from `303.3s (5m 3.3s)` to `251.5s (4m 11.5s)`. The most gain achieved on using pre-built libsqlite, but it requires libsqlite-dev to be installed in the system.

Tested using unit tests.

Closes https://github.com/kiesraad/abacus/issues/3295


praseodym:
Thanks for your PR @mikebohdan! Shaving ~50s off the build is a great result, and trimming our dependency surface is something we definitely want. We're going to hold off on merging this for now, though. Setting aside that CI isn't green yet, our main concern is that removing features from crates like `tokio`, `hyper`, and `tracing-subscriber` can introduce subtle changes in behaviour that still compile and pass a quick test run. We have a release coming up soon and can't take on the risk of a regression this late in the cycle.

We'd be happy to revisit this once the release is out and we have time to validate the feature reductions properly. Thanks again for the contribution!

PR #3335, opened 2026-05-21
Refactoring validate function signature

mohs8421:
Instead of providing a result to a validate function, the validate function should provide a result. Additionally I flagged the ValidationResults Type as a must_use, to make sure it never goes unhandled.


Lionqueen94 (approved):
Very nice refactor! Definitely makes more sense this way 👍

stacktraceghost (approved):

mohs8421:
I signed the changes, that is the only difference in the force-push

PR #3343, opened 2026-05-27
Update apportionment state machine diagram

### LINKED ISSUE #1265: Create flowchart

stacktraceghost:
Resolves https://github.com/kiesraad/abacus/issues/1265

Reference drawing of lots screens in [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17533-40913&t=djqfI8rtqTJgeQKI-4)

There are three scenarios for drawing lots in the two parts of apportionment:
```
Seat assignment
1. For residual seat assignment (assign seat, based on largest remainder or highest average)
2. For absolute majority seat reassignment (retract seat)

Candidate nomination
3. For candidates with same amount of preferential votes
```

The `DrawingLotsSeats` state has been split to `DrawingLotsResidualSeat` (1) and `DrawingLotsAbsoluteMajority` (2). This makes it easier to know why we are drawing lots. Also we can add more information, like from which list we are restracting a seat in case of `DrawingLotsAbsoluteMajority` (2).

The scenarions of drawing of lots for seat assignment can happen alternately:
- Before going to apportionment, the deceased candidates are entered first. 
- During the initial seat assignment however, these are not kept into account. `DrawingLotsResidualSeat` could happen multiple times during the assignment. 
- Then `DrawingLotsAbsoluteMajority` could happen
- Then deceased candidates are taken into account for list exhaustion, which could then again result in `DrawingLotsResidualSeat`

Because of this, these scenarios have been grouped.

<table>
  <tr>
    <th>Old version</th>
    <th>New version</th>
  </tr>
  <tr>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
</table>


praseodym (dismissed):

Lionqueen94:
@stacktraceghost Another change that needs to be made is from Finalised/DrawingLots... to RegisteringDeceasedCandidates. See the [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=20409-14768&m=dev). We have a link to `Beheer overleden kandidaten` in the Kengetallen table on the apportionment page. Which sets the status back to RegisteringDeceasedCandidates. See https://github.com/kiesraad/abacus/tree/3161-frontend-alerts-and-table-row for the current implementation. Should I change the flowchart in that issue, or shall we combine it into this one?

And a second thing, we won't have a reset from RegisteringDeceasedCandidates, only from the apportionment page, specifically in Finalised state I believe. See [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17533-40589&m=dev)

stacktraceghost:
> @stacktraceghost Another change that needs to be made is from Finalised/DrawingLots... to RegisteringDeceasedCandidates. See the [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=20409-14768&m=dev). We have a link to `Beheer overleden kandidaten` in the Kengetallen table on the apportionment page. Which sets the status back to RegisteringDeceasedCandidates. See https://github.com/kiesraad/abacus/tree/3161-frontend-alerts-and-table-row for the current implementation. Should I change the flowchart in that issue, or shall we combine it into this one?

Feel free to make adjustments in this PR.

> And a second thing, we won't have a reset from RegisteringDeceasedCandidates, only from the apportionment page, specifically in Finalised state I believe. See [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17533-40589&m=dev)

I think we still need it in the diagram. If you continue data entry at any point, the apportionment state will be reset as well.

jschuurk-kr:
> Then deceased candidates are taken into account for list exhaustion, which could then again result in `DrawingLotsResidualSeat`

This possibility is not explicit in the diagram. Which might suggest that after returning to `DrawingLotsResidualSeat` due to list exhaustion, you could also do a second `DrawingLotsAbsoluteMajority`.

Since this is a flowchart for the code / state machine, perhaps it doesn't need to be made explicit, as the logic is containted in the diamond shape those two states branch off from. And to describe for the full apportionment process we have [these diagrams](https://github.com/kiesraad/abacus-documentatie/blob/a5b802023e000d65111f3a776c32d3912ebdf9a7/verkiezingsproces/zetelverdeling-GR-flowchart.md).

praseodym:
> Another change that needs to be made is from Finalised/DrawingLots... to RegisteringDeceasedCandidates.

When making the state machine diagram we agreed that we want to keep things simple and only allow a full reset. I think we should stick to that and not complicate the design with extra transitions that are unlikely to be used.

praseodym:
> The scenarions of drawing of lots for seat assignment can happen alternately:
> 
>     * Before going to apportionment, the deceased candidates are entered first.
> 
>     * During the initial seat assignment however, these are not kept into account. `DrawingLotsResidualSeat` could happen multiple times during the assignment.
> 
>     * Then `DrawingLotsAbsoluteMajority` could happen
> 
>     * Then deceased candidates are taken into account for list exhaustion, which could then again result in `DrawingLotsResidualSeat`

Can you elaborate why deceased candidates make a difference in these steps? From my understanding, during initial seat assignment list exhaustion is not taken into account. When evaluating list exhaustion in a later step, deceased candidates are taken into account right away. So taking deceased candidates into account does not create extra steps.

Lionqueen94:
> > Another change that needs to be made is from Finalised/DrawingLots... to RegisteringDeceasedCandidates.
> 
> When making the state machine diagram we agreed that we want to keep things simple and only allow a full reset. I think we should stick to that and not complicate the design with extra transitions that are unlikely to be used.

@praseodym So just leaving out this manage deceased candidates arrow but still implementing it then right? Since it's still designed like this (see the design links)

praseodym:
> > > Another change that needs to be made is from Finalised/DrawingLots... to RegisteringDeceasedCandidates.
> > 
> > 
> > When making the state machine diagram we agreed that we want to keep things simple and only allow a full reset. I think we should stick to that and not complicate the design with extra transitions that are unlikely to be used.
> 
> @praseodym So just leaving out this manage deceased candidates arrow but still implementing it then right? Since it's still designed like this (see the design links)

I'm not sure what you mean, I propose we leave out the link and don't implement this transition.

jschuurk-kr:
> > > > Another change that needs to be made is from Finalised/DrawingLots... to RegisteringDeceasedCandidates.
> > > 
> > > 
> > > When making the state machine diagram we agreed that we want to keep things simple and only allow a full reset. I think we should stick to that and not complicate the design with extra transitions that are unlikely to be used.
> > 
> > 
> > @praseodym So just leaving out this manage deceased candidates arrow but still implementing it then right? Since it's still designed like this (see the design links)
> 
> I'm not sure what you mean, I propose we leave out the link and don't implement this transition.

The designs look like they consider the registering of deceased candidates as a separate step __before__ apportionment as opposed to the first step __of__ apportionment. I think it makes more sense as the first step of. So when you reset apportionment, you return to "unitiliased" as in the original diagram.

I do think it makes sense to have a link to see the candidates that have been registered as deceased, but that would be read-only.

Lionqueen94:
> > > > > Another change that needs to be made is from Finalised/DrawingLots... to RegisteringDeceasedCandidates.
> > > > 
> > > > 
> > > > When making the state machine diagram we agreed that we want to keep things simple and only allow a full reset. I think we should stick to that and not complicate the design with extra transitions that are unlikely to be used.
> > > 
> > > 
> > > @praseodym So just leaving out this manage deceased candidates arrow but still implementing it then right? Since it's still designed like this (see the design links)
> > 
> > 
> > I'm not sure what you mean, I propose we leave out the link and don't implement this transition.
> 
> The designs look like they consider the registering of deceased candidates as a separate step **before** apportionment as opposed to the first step **of** apportionment. I think it makes more sense as the first step of. So when you reset apportionment, you return to "uninitialised" as in the original diagram.
> 
> I do think it makes sense to have a link to see the candidates that have been registered as deceased, but that would be read-only.

@jorisleker Would you be fine with a "Bekijk overleden kandidaten" instead of a "Beheer overleden kandidaten" button and that there is no option to add or delete the overleden kandidaten once the state is Finalised (aka once you end up at the Zetelverdeling pages). There is still one way to go back, which is by clicking the "Zetelverdeling opnieuw doen" button and re-entering the deceased candidates fully.

jorisleker:
> @jorisleker Would you be fine with a "Bekijk overleden kandidaten" instead of a "Beheer overleden kandidaten" button and that there is no option to add or delete the overleden kandidaten once the state is Finalised (aka once you are at the Zetelverdeling pages). There is still one way to go back, which is by clicking the "Zetelverdeling opnieuw doen" button and re-entering the deceased candidates fully.

I think we would need something like a hint on the deceased candidates page to explain why editing is no longer possible, and which steps you'd need to take if you need to edit the list (ie: a link or button on the read only page to redo seat apportionment). I don't think it is obvious enough that clicking 'zetelverdeling opnieuw doen' on another page would be your next step. 

Okay to design this?

Lionqueen94:
I removed my commit with the manage deceased candidates change :+1:

stacktraceghost:
> > The scenarions of drawing of lots for seat assignment can happen alternately:
> > 
> >     * Before going to apportionment, the deceased candidates are entered first.
> > 
> >     * During the initial seat assignment however, these are not kept into account. `DrawingLotsResidualSeat` could happen multiple times during the assignment.
> > 
> >     * Then `DrawingLotsAbsoluteMajority` could happen
> > 
> >     * Then deceased candidates are taken into account for list exhaustion, which could then again result in `DrawingLotsResidualSeat`
> 
> Can you elaborate why deceased candidates make a difference in these steps? From my understanding, during initial seat assignment list exhaustion is not taken into account. When evaluating list exhaustion in a later step, deceased candidates are taken into account right away. So taking deceased candidates into account does not create extra steps.

You are right that it does not make a difference in the steps. I was looking for a situation where you first draw lots for residual seat, then absolute majority and then residual seat again (due to list exhaustion). But this can also happen without deceased candidates.

The decision to link the two drawing lots states for seats is based on the fact that the states can alternate, so I was looking for a situation when this could happen.

Lionqueen94 (approved):

praseodym (approved):

PR #3349, opened 2026-05-28
Show currently logged in users

### LINKED ISSUE #3251: Show online status in users table
### Background
Right now, it is not possible to see which users have an active / logged in session. This makes it harder for admins and coordinators to see who is logged in.

### Solution
- [ ] Show logged in status in users table (blue dot next to last activity)
- [ ] Logging session timeouts via #3243 will also help giving admins and coordinators a better understanding of who is logged in and who is not.

See [design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=21374-31478&t=W9M0HK7rz0Ja1mh6-11)

Image

yusufraji:
This shows users who are currently logged in. 

Changes:
- adds a new `is_logged_in` field to the User struct and table on the backend
- sets the field to `false` by default
- sets the field to `true` after logging in
- set the field to `false` after logging out

I still need help working on the frontend for the full implementation.

Resolves #3251


yusufraji:
Perhaps @Lionqueen94 or @stacktraceghost could have a look.

ring-ring-ring:
This is cool! Thnx @yusufraji

backend/migrations/11_alter_users.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users
+ADD COLUMN is_logged_in INTEGER NOT NULL DEFAULT 0;
```
stacktraceghost: This column can be added to `4_users.sql`.

backend/src/repository/user_repo.rs
```diff
@@ -392,6 +412,27 @@ pub async fn update_last_activity_at(
     Ok(())
 }
 
+pub async fn update_is_logged_in(
+    conn: &mut SqliteConnection,
+    user_id: UserId,
+) -> Result<(), sqlx::Error> {
+    query!(
+        r#"
+            UPDATE users
```
stacktraceghost: `is_logged_in` is the result of this query, which needs to be kept up to date by calling `update_is_logged_in` on the right moments. What is the advantage of storing the result of this query passively versus running a query that actively looks up the session of a user (which would not risk a state getting out of sync)?

The session of a user can also expire, at that point the value of `is_logged_in` becomes invalid. When running a query, `expires_at` can be compared with the current time.

stacktraceghost (commented):
Thank you for your PR! I've reviewed it and have some comments.

backend/migrations/11_alter_users.sql
cikzh: Moved in [9a3767e](https://github.com/kiesraad/abacus/pull/3349/commits/9a3767e8c566f30c53133f26f5f80bfbb865bee2)

yusufraji: Thanks @cikzh

backend/src/repository/user_repo.rs
yusufraji: Yes, you are right. Things might get out of sync if `update_is_logged_in` isn't called at the right moments. I was just trying to follow the established pattern with `last_activity_at` & `update_last_activity_at`. Thinking about this, I don't think the advantage of keeping things simple (with naive implementation) outweigh correctness. Since computing on the fly guarantees that things never go out of sync.

>The session of a user can also expire, at that point the value of is_logged_in becomes invalid. When running a query, expires_at can be compared with the current time.

I will check the session expiry for `is_logged_in`.

Would you prefer computing `is_logged_in` on the fly? I can update the sql queries where `User` is retrieved.

praseodym: > Would you prefer computing `is_logged_in` on the fly? I can update the sql queries where `User` is retrieved.

I think that would be preferable. With the amount of users and sessions (max 1 per user) this should be cheap enough to compute.

praseodym:
@yusufraji Looks good! The only thing remaining to resolve #3251 would be to add the blue indicator dot for users that are logged in. Your initial PR body says you need help with the frontend implementation, are there any specific points you're struggling with?

yusufraji:
> @yusufraji Looks good! The only thing remaining to resolve #3251 would be to add the blue indicator dot for users that are logged in. Your initial PR body says you need help with the frontend implementation, are there any specific points you're struggling with?

Not necessarily, just a skill issue, since I'm new to react/Typescript and all that's frontend. I looked into it just a little bit before raising the PR. I'll have a look again.

backend/src/repository/user_repo.rs
```diff
@@ -268,21 +275,30 @@ pub async fn get_by_username(
     conn: &mut SqliteConnection,
     username: &str,
 ) -> Result<Option<User>, AuthenticationError> {
+    let now = Utc::now().to_rfc3339();
```
praseodym: We can directly use `Utc::now()` (a `DateTime<Utc>`), no need to create string first.

backend/src/api/authentication.rs
```diff
@@ -480,11 +480,11 @@ async fn logout(
...
     let mut tx = pool.begin_immediate().await?;
 
-    if let Some(session) = session_repo::get_by_key(&mut tx, session_key)
+    let session = session_repo::get_by_key(&mut tx, session_key)
         .await
         .ok()
-        .flatten()
-    {
+        .flatten();
+    if let Some(session) = &session {
```
praseodym: This change is unnecessary and can be reverted

backend/src/repository/user_repo.rs
```diff
@@ -268,21 +275,30 @@ pub async fn get_by_username(
...
+            u.last_activity_at as "last_activity_at: _",
+            (
+                SELECT COUNT(*)
+                FROM sessions s
+                WHERE s.user_id = u.id
+                    AND s.expires_at > datetime($1)
+            ) > 0 as "is_logged_in: bool",
+            u.updated_at as "updated_at: _",
+            u.created_at as "created_at: _"
+        FROM users u WHERE LOWER(u.username) = LOWER($2)
```
praseodym: We should keep the `WHERE username = ? COLLATE NOCASE` so that this query keeps using the `username COLLATE NOCASE` index

praseodym:
@yusufraji let us know if you want to spend more time on this, otherwise I'd be happy to fix these last points and merge this PR.

backend/src/repository/user_repo.rs
yusufraji: I agree. Thanks for pointing this out.

backend/src/repository/user_repo.rs
yusufraji: I changed this initially as I was unable to get past some trait errors.  

```
the trait bound `bool: From<std::option::Option<bool>>` is not satisfied
`bool` implements trait `From<T>`:
  From<format_description::parse::format_item::HourBase>
  From<format_description::parse::format_item::MonthCaseSensitive>
  From<format_description::parse::format_item::PeriodCase>
  From<format_description::parse::format_item::PeriodCaseSensitive>
  From<format_description::parse::format_item::SignBehavior>
  From<format_description::parse::format_item::WeekdayCaseSensitive>
  From<format_description::parse::format_item::WeekdayOneIndexed>
  From<format_description::parse::format_item::YearBase>
and 3 others
required for `std::option::Option<bool>` to implement `Into<bool>`
```

Perhaps you can assist me with this.

yusufraji:
> @yusufraji let us know if you want to spend more time on this, otherwise I'd be happy to fix these last points and merge this PR.

No worries, I've committed changes to address some of the review comments. But for the last comment, please go ahead and make the changes since I am not sure of how to resolve that comment.
> We should keep the WHERE username = ? COLLATE NOCASE so that this query keeps using the username COLLATE NOCASE index

I look forward to learning how I could have implemented this from your changes.

backend/src/repository/user_repo.rs
praseodym: Adding a `!` non-null override to line 296 seems to work:

```rs
    let user = sqlx::query_as!(
        User,
        r#"
        SELECT
            u.id as "id: UserId",
            u.username,
            u.fullname,
            u.role as "role: _",
            u.password_hash,
            u.needs_password_change as "needs_password_change: bool",
            u.last_activity_at as "last_activity_at: _",
            (
                SELECT COUNT(*)
                FROM sessions s
                WHERE s.user_id = u.id
                    AND s.expires_at > datetime($1)
            ) > 0 as "is_logged_in!: bool",
            u.updated_at as "updated_at: _",
            u.created_at as "created_at: _"
        FROM users u WHERE u.username = $2 COLLATE NOCASE
        "#,
        now,
        username
    )
```

yusufraji: owh, interesting. Does this suppress compile time check for the entire query, or just `is_logged_in` field? Also, can you point me to the documentation or a resource to learn more beyond beginner sqlx usage.

praseodym: The `!` just tells sqlx that the `is_logged_in` field is never null, because sqlx couldn't infer that by itself.

This is documented here: https://docs.rs/sqlx/latest/sqlx/macro.query.html#force-not-null

yusufraji: > Adding a `!` non-null override to line 296 seems to work:

@praseodym. I thought you've made this change based on your comment, and long forgotten about this, until noticing today that it's still open. I have made the change.

praseodym:
@yusufraji this has a merge conflict after merging #3432

praseodym:
I notice that the tests fail because of the session unique constraint. #3458 also adds this constraint and fixes failing tests.

pflanze:
> I notice that the tests fail because of the session unique constraint. #3458 also adds this constraint and fixes failing tests.

Yes; this PR primarily added the index for performance reasons. Given that your PR is merged, this PR can be rebased on top of the new main again and then doesn't need to add the constraint anymore.

@yusufraji, we should add running the tests to our run-left-hooks script; I didn't realize it didn't do that already. (Explanation to the others: we turned the left hooks into a shell script so we could commit without the hooks running all the time; we run the script when we clean up the history.)

praseodym:
> (Explanation to the others: we turned the left hooks into a shell script so we could commit without the hooks running all the time; we run the script when we clean up the history.)

You can always commit using `--no-verify` to skip the pre-commit hooks.

praseodym (approved):

pflanze:
> > (Explanation to the others: we turned the left hooks into a shell script so we could commit without the hooks running all the time; we run the script when we clean up the history.)
> 
> You can always commit using `--no-verify` to skip the pre-commit hooks.

OK, I wasn't aware of that. I would probably end up just running with that option all the time, though, since my workflow commits very often and then I clean up things, and so will end up still preferring an explicit script.

BTW I'm considering parallelizing the actions (in my/our script) to make use of my many cores when I actually do run the actions; I didn't see anything about lefthook that would do that, have I missed it?

praseodym:
> BTW I'm considering parallelizing the actions (in my/our script) to make use of my many cores when I actually do run the actions; I didn't see anything about lefthook that would do that, have I missed it?

Lefthook runs all its commands in parallel in our setup:
https://github.com/kiesraad/abacus/blob/fbf31e2e62e54bf96d757e9d0abc1715aada9aab/lefthook.yml#L4

Also see https://lefthook.dev/configuration/parallel/.

pflanze:
This PR is primarily @yusufraji 's work, so take these comments as ideas, I'm not invested in them:

- The commit message of the first commit still mentions adding the unique index, maybe worth removing; really minor, though.

- Given that `is_logged_in` is a derived value, might it be cleaner to leave struct `User` as is and instead add a wrapper that holds the `User` instance and the `is_logged_in` value? In garbage-collected language there's an incentive against wrapping like that since it costs a memory indirection, but that's not the case with Rust, where doing cleaner abstractions like this has no runtime cost. There's probably no point stopping this PR from being merged, but I could evaluate the feasibility in a separate PR, unless you already have reasons not to want that.

yusufraji:
@pflanze TIL we can run `lefthook run pre-commit --force` to run the lefthook commands, which prevents skipping when no files have changed. For this particular use-case, we no longer need the workaround (shell script), and it allows us to benefit from existing `parallel: true` discussed above.

stacktraceghost (approved):

jschuurk-kr:
Did some brief exploratory testing and found one usability issue, that's not blocking for this PR.

The issue is that the page does not re-fetch the user data. So to get up-to-date data a user needs to either refresh the page (relatively slow) or go to "Verkiezing" or "Logs" and then return to "Gebruikers" (not very intuituve). Note that the page does not refresh if you click "Gebruikers" while on the "Gebruikers" page.

I think we should at least mention this in the user documentation (cc @marjoleintamis). Additionally we could:
- add copy to this effect to the page, and/or
- add a button to refetch the data, and/or
- auto-refetch the data every x seconds as we do on other pages.

@jorisleker What are your thoughts?

pflanze:
> @pflanze TIL we can run `lefthook run pre-commit --force` to run the lefthook commands, which prevents skipping when no files have changed. For this particular use-case, we no longer need the workaround (shell script), and it allows us to benefit from existing `parallel: true` discussed above.

Well yeah, except we still want to run the tests, too, and parallelism seems odd to me in that formatting isn't done first as a dependency (lefthook has no concept of dependencies as far as I can see).

Also it might have made more sense to use make since it is commonly available whereas lefthook is a JavaScript dependency that I didn't know and had to install, thus more exposed to supply chain attacks.

But of course we will manage, there's no question about that.

jorisleker:
> Additionally we could:
> 
> add copy to this effect to the page, and/or
> add a button to refetch the data, and/or
> auto-refetch the data every x seconds as we do on other pages.
> @jorisleker What are your thoughts?

Auto-refetch would be cleanest and what users expect.

jschuurk-kr:
> > Additionally we could:
> > add copy to this effect to the page, and/or
> > add a button to refetch the data, and/or
> > auto-refetch the data every x seconds as we do on other pages.
> > @jorisleker What are your thoughts?
> 
> Auto-refetch would be cleanest and what users expect.

created https://github.com/kiesraad/abacus/issues/3473 for this

Issue #3219, opened 2026-04-23
OSV considers a P 9 seat assignment as part of the "max one seat" condition, Abacus doesn't


jschuurk-kr:
Once seat allocation through largest averages/remainders has been completed, article P9 comes into effect. If a list has an absolute majortiy of votes, but not of seats, the last allocated residual seat is re-assigned to the list with the absolute majortiy of votes. After that, list exhaustion is applied. This can result in additional residual seats.

If there are fewer than 19 total seats, the first two steps of residual seat alignment have the condition that a list can at most gain a single seat. OSV includes a seat assigned through P 9 in that condition. Abacus does not.

### Example 1

OSV2020 and Abacus agree about the first part of allocation:
1. Full seats: list 4 gets 2, list 7 gets 3, list 13 gets 6.
1. That leaves one residual seat, which goes to list 4 based on greatest remainder.
1. Now list 13 has an absolute majority of votes, but not of seats, so the residual seat moves from list 4 to list 13.
1. However, list 4 has 2 seats for 1 candidate and list 7 has 3 seats for 1 candidate.
1. So we have three residual seats left to assign.

This is where OSV2020 and Abacus proceed differently. Abacus assigns the first residual seat based on largest remainder. OSV2020 assigns it based on largest average. The reason is that three parties participate for the largest remainder. Two have been exhausted, one got an extra seat because of P9 (absolute majority).

So OSV considers the step with largest remainders over, because the only non-exhausted participating list already got a seat (albeit through P 9). Abacus does not take that P 9-seat into account, so can assign another seat to that list through largest remainders.

```
Output of `std::fmt::Debug`:

	seats: 12, total_votes: 95179851
	list  1:  8 [ 1,  3,  4]
	list  2: 1536 [1536]
	list  3: 327680 [327680]
	list  4: 17769735 [17769735]
	list  5: 1387597 [ 9, 10, 11, 12, 196608, 327680, 393216, 467968, 2083]
	list  6: 655360 [655360]
	list  7: 24836278 [24836278]
	list  8: 13312 [13312]
	list  9:  5 [ 5]
	list 10: 1536 [1536]
	list 11: 1516544 [1516544]
	list 12: 533248 [533248]
[...]

PR #3351, opened 2026-05-29
differential fuzzer: make condition to skip case related to #3219 more specific

jschuurk-kr:
relates to #3219 OSV considers a P 9 seat assignment as part of the "max one seat" condition, Abacus doesn't

### Scope
For #3219 the differential fuzzer skipped all cases where one list had an absolute majority. That is a significantly broader condition than is required. So this PR updates the skip to: fewer than 19 seats, art. P 9 applied, art. P 10 applied.

### Testing
Ran the differential fuzzer for about 2 hours with 4 jobs:
```
#1620879: cov: 2057 ft: 9619 corp: 724 exec/s: 52 oom/timeout/crash: 0/0/0 time: 7419s job: 240 dft_time: 0
```
I did run the test when the code still contained an additional but duplicate condition, i.e. it included a check if there was a list with an absolute majority of votes in the input data. We already check this in the output of OSV and Abacus. Removing it saved a handful lines of code.

### Questions
I suspect the condition can be made even more specific. If after list exhaustion the residual seats are assigned through non-unique highest averages, Abacus and OSV should return the same result. Is that worth the effort, though?

### DIFF
--- a/backend/apportionment/fuzz/fuzz_targets/differential_osv2020.rs
+++ b/backend/apportionment/fuzz/fuzz_targets/differential_osv2020.rs
@@ -130,25 +130,6 @@ fn init() {
 }
 
 fn fuzz(data: FuzzedApportionmentInput) {
-    // Skip cases where any list has an absolute majority of votes
-    // related issue: #3219
-    let total_votes = data
-        .list_votes
-        .iter()
-        .flat_map(|list| list.candidate_votes.iter())
-        .map(|cv| cv.votes())
-        .sum::<u32>();
-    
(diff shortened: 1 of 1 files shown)


Lionqueen94 (approved):

praseodym:
> I suspect the condition can be made even more specific. If after list exhaustion the residual seats are assigned through non-unique highest averages, Abacus and OSV should return the same result. Is that worth the effort, though?

I don't think it's worth it at this moment, we're already looking at extreme edge cases at this point.

praseodym (approved):

PR #3362, opened 2026-06-02
Fixed number of voters in P 22-2 model

### LINKED ISSUE #3360: Fix number of voters in P 22-2 model (should come from summary)
As mentioned in https://github.com/kiesraad/abacus/issues/3070#issuecomment-4545755854 and https://github.com/kiesraad/abacus/issues/3072 the number of voters (Z Kiesgerechtigden) is set to 1 in the P 22-2 model, because it uses the value from the election instead of the value from the summary of the results. This needs to be fixed.

Lionqueen94:
Closes #3360 

The number of voters in `model-p-22-2.json` input is changed from 6000 to 8000 on purpose, so that the change is visible in the pdf diffs.

### DIFF
--- a/backend/src/domain/summary.rs
+++ b/backend/src/domain/summary.rs
@@ -325,6 +325,8 @@ impl From<ElectionSummary> for ElectionSummaryWithoutVotes {
 #[derive(Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct ElectionSummaryCSB {
+    /// The number of voters (i.e. "Kiesgerechtigden")
+    pub number_of_voters: u32,
     /// The total number of voters
     pub voters_counts: VotersCounts,
     /// The total number of votes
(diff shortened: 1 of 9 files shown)


cikzh:
I'm having a hard time following the context of the PR. It took a while spelunking through links to other issues and comments.
Also, are there any tests or guard to ensure the `expect()` doesn't fire on line 343? Would suck if the expect fires when creating a P 22-2 at the end of the process.
Do we have any issues addressing the optional `number_of_voters` field, now that https://github.com/kiesraad/abacus/issues/3297 is closed?

praseodym (approved):
We should consider completely removing number_of_voters from the CSB election so that we don't make this mistake more often.

Lionqueen94:
> I'm having a hard time following the context of the PR. It took a while spelunking through links to other issues and comments. Also, are there any tests or guard to ensure the `expect()` doesn't fire on line 343? Would suck if the expect fires when creating a P 22-2 at the end of the process. Do we have any issues addressing the optional `number_of_voters` field, now that #3297 is closed?

I added some explanation to the issue. Also the CSBElectionSummary is made from GSBResults, which have a non optional number_of_voters field. That's why the expect will never trigger when creating the P 22-2.

oliver3 (approved):

PR #3384, opened 2026-06-07
Generate CSB OSV4-3 CSV file

### LINKED ISSUE #3137: Generate csv file as part of CSB output
### Scope
- generate file after generating count EML
- add file to zip
- store in database
- update backend tests

rnijveld:
This implements the CSB OSV4-3 CSV file (adding similar code for the GSB should be pretty trivial).

### DIFF
--- a/backend/src/api/report/structs.rs
+++ b/backend/src/api/report/structs.rs
@@ -117,6 +117,7 @@ impl CsbFiles {
             .or(self.results_pdf.as_ref())
             .or(self.attachment_pdf.as_ref())
             .or(self.total_counts_eml.as_ref())
+            .or(self.csv_counts.as_ref())
             .map(|f| f.created_at)
             .expect("At least one file should be present")
     }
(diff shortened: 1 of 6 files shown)


ring-ring-ring:
Works on my machine ;-)

backend/src/api/report/structs.rs
```diff
@@ -210,6 +212,19 @@ impl ResultsInput {
...
+            "osv4-3_telling_{}{}_{}.csv",
+            self.election.category.to_eml_code().to_lowercase(),
+            self.election.election_date.year(),
+            self.election
+                .location
+                .split_whitespace()
+                .collect::<String>()
+                .to_lowercase(),
+        )
+    }
```
praseodym, line 228: We also have `as_osv4_3_csv_filename()` in the crate, can we reuse that?

praseodym (changes requested):
The CSV file also needs to be added to `CsbFiles::needs_generation()`

backend/src/api/report/structs.rs
rnijveld: Not really unfortunately due to the way this was designed, unless we generate another document on the fly here, which feels like way to much work for just generating a filename.

praseodym (approved):

Lionqueen94 (approved):

Issue #788, opened 2025-01-02
Epic: drawing of lots in apportionment


praseodym:

jschuurk-kr:
From the [GR 26 results of Roermond](https://www.roermond.nl/file/verslag-en-zetelverdeling-definitieve-uitslag-model-p22-2):

<img width="1352" height="971" alt="Image" src="https://github.com/user-attachments/assets/f4b1f349-f055-4217-a3a7-edada91811d6" />

PR #3401, opened 2026-06-09
Return extra information when drawing lots is required

oliver3:
This PR is to make it possible to send extra information to the frontend when drawing lots is required, to show to the user, and is needed for the "Frontend: drawing lots screens" issues of [Epic: drawing lots in apportionment](https://github.com/kiesraad/abacus/issues/788)

Changes:
- `ListDrawingLotsVariant` and `CandidateDrawingLotsVariant` do not only convey what the variant of the drawing lots it is, but also contain through an "inner struct":
  - extra information to be displayed to the user
  - the `options` (and `list` in case of Candidate) that was previously part of the `ListDrawingLotsRequired` and `CandidateDrawingLotsRequired` types
 - `ListDrawingLotsError` and `CandidateDrawingLotsError` are returned from apportionment functions as the `Err` to also be able to return an error if the variant of a drawn lot is not as expected

For now, the `LargestRemainderResidualSeat` and `HighestAverageResidualSeat` variants have been extended with the information needed, AbsoluteMajority and CandidateDrawingLots will be done in the issues to update the screens.

I recommend starting at `backend/apportionment/src/structs.rs` to see the struct changes in the apportionment crate.

### DIFF
--- a/backend/apportionment/src/structs.rs
+++ b/backend/apportionment/src/structs.rs
@@ -19,64 +19,92 @@ pub type CandidateNumber<LV> = <<LV as ListVotes>::Cv as CandidateVotes>::Candid
 /// Errors that can occur during apportionment
 #[derive(Debug, PartialEq)]
 pub enum ApportionmentError<LN, CN> {
-    ListDrawingLotsRequired(ListDrawingLotsRequired<LN>),
-    CandidateDrawingLotsRequired(CandidateDrawingLotsRequired<LN, CN>),
+    ListDrawingLotsRequired(ListDrawingLotsVariant<LN>),
+    CandidateDrawingLotsRequired(CandidateDrawingLotsVariant<LN, CN>),
 }
 
 /// Used in [ApportionmentError] to indicate that drawing lots for a list is needed,
 /// containing all the information needed to do the drawing
 #[derive(Debug, PartialEq)]
-pub struct ListDrawingLotsRequired<LN> {
-    pub variant: ListDrawingLotsVariant,
+pub enum ListDrawingLotsError<LN> {
+    DrawingLotsRequired(ListDrawingLotsVariant<LN>),
+}
+
+impl<LN, CN> From<ListDrawingLotsError<LN>> for ApportionmentError<LN, CN> {
+    fn from(value: ListDrawingLotsError<LN>) -> Self {
+        match value {
+            ListDrawingLotsError::DrawingLotsRequired(variant) => {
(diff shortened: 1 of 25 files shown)


backend/apportionment/src/structs.rs
```diff
@@ -19,64 +19,92 @@ pub type CandidateNumber<LV> = <<LV as ListVotes>::Cv as CandidateVotes>::Candid
...
+}
+
+/// The candidate that has been drawn and the variant with all information about the drawing
 pub trait CandidateDrawn<LN, CN> {
-    /// The list the candidate needs to be drawn from
-    fn list(&self) -> &LN;
-    /// The candidates that lots are drawn for
-    fn options(&self) -> &[CN];
+    /// The type and information for drawing lots
+    fn variant(&self) -> CandidateDrawingLotsVariant<LN, CN>;
```
Lionqueen94, line 107: We don't need a variant here I believe, or is this done to keep the structure the same as for ListDrawn?

oliver3: To keep the structure the same indeed, to have a struct with all data to compare when using the candidate drawn and to put other data needed for the frontend.

backend/src/domain/apportionment.rs
```diff
@@ -268,6 +268,17 @@ impl From<apportionment::Fraction> for DisplayFraction {
...
+impl From<DisplayFraction> for apportionment::Fraction {
+    fn from(display_fraction: DisplayFraction) -> Self {
+        let numerator =
+            display_fraction.integer * display_fraction.denominator + display_fraction.numerator;
+        Self {
+            numerator,
+            denominator: display_fraction.denominator,
+        }
+    }
+}
```
praseodym, line 280: It feels like we took a wrong turn somewhere if we need to convert `DisplayFraction`s back to `Fraction`s. It seems like this is not currently used, can we remove it and find a different solution instead?

praseodym (changes requested):

backend/src/domain/apportionment.rs
praseodym: Discussed with @oliver3: we do need this to return the drawn lot to the apportionment after the roundtrip to the frontend.

In a future issue we could remove the `DisplayFraction` struct and replace it with a `Fraction` struct with identical internal representation as the one in the apportionment crate. The functionality to take out the whole number from the fraction can be a function on that struct.

praseodym (approved):

Lionqueen94 (approved):

PR #3403, opened 2026-06-10
Process apportionment anew after adding drawn lots

oliver3:
## What & why

When the result of drawing lots has been added to the apportionment state, the new state has to be determined as well. This PR adds this step for the `add_list_drawn` and `add_candidate_drawn` endpoints.

I've also changed the `audit_log` argument for `update_apportionment_state` and `next_apportionment_state` to be a reference, because there is no need for ownership.

## How to test

You can see in the changed unit test that the state is now `Finalised` instead of the old `DrawingLots`, and that two times `ApportionmentStateUpdated` is logged to the audit log.

### DIFF
--- a/backend/src/api/apportionment/handlers/add_candidate_drawn.rs
+++ b/backend/src/api/apportionment/handlers/add_candidate_drawn.rs
@@ -12,7 +12,7 @@ use crate::{
     },
     infra::audit_log::AuditService,
     repository::{election_repo, user_repo::User},
-    service::update_apportionment_state,
+    service::{next_apportionment_state, update_apportionment_state},
 };
 
 /// Add candidate to drawing lots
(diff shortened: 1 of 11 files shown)


Lionqueen94 (approved):

praseodym (approved):
I'm approving the PR because the functionality is correct.

However, I think the `update_state`/`next_state` functions in the service are not doing the right thing and with this PR we're trying to work around that from the API layer. For example, in case an apportionment needs only one drawing of lots, it shouldn't be possible to add the drawn lot to the state but still have the state be stuck in the `ListDrawingLotsRequired` if we forget to call `next_state`.

I'd prefer to change the `update_state`/`next_state` functions, maybe even replacing them by explicit transition functions like the ones described in the [apportionment state machine diagram](https://github.com/kiesraad/abacus/blob/main/documentatie/state-machines/apportionment.md). These functions could be in an `impl ApportionmentState` block just as we have for `DataEntryStatus`, which I still think is a well-designed part of the application.

oliver3:
> I'd prefer to change the `update_state`/`next_state` functions, maybe even replacing them by explicit transition functions like the ones described in the [apportionment state machine diagram](https://github.com/kiesraad/abacus/blob/main/documentatie/state-machines/apportionment.md). These functions could be in an `impl ApportionmentState` block just as we have for `DataEntryStatus`, which I still think is a well-designed part of the application.

It is IMHO already similar to `DataEntryStatus` with the impl block, with the one big difference that `DataEntryStatus` does not need to retrieve "external information" but `ApportionmentState` needs the result from a new apportionment process to make a decision which would be the next state. This can be done in a similar way to what the `CommitteeSessionHasDataEntriesProvider` solves for the `CommitteeSessionStatus`, i.e. a trait to get this external information.

I'll merge this for now, and let's discuss if we want to find some timebox to try that.

PR #3404, opened 2026-06-10
Process drawing lots for list results

### LINKED ISSUE #1263: Backend: process drawing of lots results

oliver3:
## What & why

Resolves [Backend: process drawing lots for list results #1263](https://github.com/kiesraad/abacus/issues/1263)

Use the `lists_drawn` iterator from the `ApportionmentInput` as the drawing lots in case of `HighestAverageResidualSeat` or `LargestRemainderResidualSeat`:
- Try to get the next value from the iterator, else return `DrawingLotsRequired`
- Make sure it is exactly the right variant of drawing lots, else return `InvalidLotDrawing`
- use the list drawn for the highest average residual seat or largest remainder residual seat

## How to test

Unit tests are extended and added to see the results, frontend will be created in following issues in [the epic](https://github.com/kiesraad/abacus/issues/788).

## Reviewer notes

The actual changes are made in `seat_assignment/residual_seat_assignment.rs` of the apportionment crate, the rest is getting the `lists_drawn` there, test changes and propagating the `InvalidLotDrawing` error.

Please check if the tests and their comments are valid scenarios and contain correct assertions.


backend/apportionment/fuzz/fuzz_targets/apportionment.rs
```diff
@@ -160,5 +160,8 @@ fuzz_target!(
         Err(ApportionmentError::ListDrawingLotsRequired(_) | ApportionmentError::CandidateDrawingLotsRequired(_)) => {
             // Expected error: drawing of lots is not yet implemented.
```
Lionqueen94: This comment needs to be updated

backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -646,29 +654,110 @@ pub(crate) mod tests {
...
+                        list_options: vec![1],
+                        list_assigned: vec![1],
+                        remainder_votes: Fraction::new(900, 15),
+                    })
+                );
+                assert_eq!(
+                    result.steps[1].change,
+                    SeatChange::LargestRemainderAssignment(LargestRemainderAssignedSeat {
+                        selected_list_number: 6,
+                        list_options: vec![6],
```
Lionqueen94, line 684: I'd expect the `list_options` to contain all of them, not just the chosen list

backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -646,29 +654,110 @@ pub(crate) mod tests {
...
             }
 
+            /// Apportionment with residual seats assigned with largest remainders method
+            ///
+            /// Full seats: [6, 2, 2, 1, 1, 1, 0, 0] - Remainder seats: 2
+            /// Remainders: [60, 0/15, 0/15, 0/15, 0/15, 0/15, 55, 45]
+            /// 1 - largest remainder: seat assigned to list 1
+            /// 2 - Drawing of lots is required for lists: [2, 3, 4, 5, 6], only 1 seat available
+            #[test]
+            fn test_with_0_remainders_drawing_of_lots_error_copy() {
```
Lionqueen94: Is the "copy" supposed to be in the name?

Lionqueen94 (changes requested):
A few comments, especially the list_options one is one to maybe discuss

backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -113,6 +114,7 @@ pub(crate) fn seat_assignment<T: ApportionmentInput>(
         input.deceased_candidates(),
         residual_seats,
         steps,
+        &mut input.lists_drawn(),
```
praseodym: This should use the same iterator as on line 68, otherwise the P 10 lot will never match if there was also a lot draw for the P 6 assignment.

backend/apportionment/src/seat_assignment/mod.rs
praseodym: I think more generally `SeatChange` should explicitly record that this was the result of a lot draw, so that we can use those steps to reconstruct the entire apportionment without somehow also using the apportionment state.

backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -1478,24 +1570,109 @@ pub(crate) mod tests {
...
+                                options: vec![2, 3, 5, 6]
+                            }
+                        )
+                    )
+                );
+            }
+
+            /// Apportionment with residual seats assigned with highest averages method
+            ///
+            /// Full seats: [13, 3, 3, 3] - Remainder seats: 4
```
praseodym: This adds up to 26 seats while the test itself says 24 seats

praseodym (changes requested):

backend/apportionment/src/seat_assignment/mod.rs
oliver3: No :) forgot to remove the before-changes test, fixed in https://github.com/kiesraad/abacus/pull/3404/commits/c37c0d9fd17d7ef74f553f295afbef88194478d6

backend/apportionment/src/seat_assignment/mod.rs
oliver3: Corrected remainder seats to be 2 in https://github.com/kiesraad/abacus/pull/3404/commits/c37c0d9fd17d7ef74f553f295afbef88194478d6

backend/apportionment/src/seat_assignment/mod.rs
oliver3: Oh yes, that was the whole clue of using an iterator :D thanks for spotting! Fixed in https://github.com/kiesraad/abacus/pull/3404/commits/8805fcf950bab633ca9e93b06bec3ce25c87486d

backend/apportionment/src/seat_assignment/mod.rs
oliver3: This would need a solution for dealing with drawing lots for the `SeatChange` and how functions return that it used a drawing lots. I suggest dealing with this this in a separate issue / PR.

oliver3: Created https://github.com/kiesraad/abacus/issues/3408

backend/apportionment/fuzz/fuzz_targets/apportionment.rs
oliver3: Tried to improve it in https://github.com/kiesraad/abacus/pull/3404/commits/3e9df2032a0cbb834fb5284c356599cf25ff2972

Lionqueen94 (approved):

praseodym (approved):

Issue #1264, opened 2025-03-27
Frontend: drawing lots screens for list


Lionqueen94:
- [ ] https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=20682-55169&m=dev
  - [ ] Add drawing of lots for list page
  - [x] Update copy of drawing of lots alert on ApportionmentPage
- [ ] https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=20478-15334&m=dev
  - [x] After merging https://github.com/kiesraad/abacus/pull/3452 add extra round to HighestAveragesTable
- [ ] Handle InvalidLotDrawing error as well in the frontend
- [x] Add drawing lots variants in frontend `testing/` folder to be used in tests (to make sure the math adds up)
- [x] Add tests for all alerts, redirects and tables in case of drawing lots

PR #3416, opened 2026-06-11
Return preliminary results in drawing lots required

Lionqueen94:
## What & why

This is the extra backend preparation for #1264 in which we need to show the preliminary seat assignment and the alert and page for list drawing lots.

- Added residual seat numbers to highest average and largest remainder variants
- Changed type parameter of SeatAssignmentResult from ListVotes to ListNumber
- Drawing lots situation is now not returned as error from apportionment endpoint and is enriched with election summary and (preliminary) seat assignment

## How to test

Review the updated tests and the apportionment result sent to the frontend instead of the 422 error

## Reviewer notes

Review per commit

### DIFF
--- a/backend/src/api/apportionment/mod.rs
+++ b/backend/src/api/apportionment/mod.rs
@@ -1,4 +1,4 @@
-use apportionment::ApportionmentError;
+use apportionment;
 use axum::http::StatusCode;
 use tracing::error;
 use utoipa_axum::{router::OpenApiRouter, routes};
@@ -22,10 +22,11 @@ use crate::{
 };
 
 /// Errors that can occur before apportionment
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum ApportionmentApiError {
+    ApportionmentNotCompleted,
     CommitteeSessionNotCompleted,
-    DrawingLotsRequired,
+    InvalidLotDrawing,
 }
 
 impl ApiErrorResponse for ApportionmentApiError {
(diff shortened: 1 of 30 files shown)


backend/src/api/apportionment/mod.rs
```diff
@@ -43,27 +52,22 @@ impl ApiErrorResponse for ApportionmentApiError {
...
+                StatusCode::CONFLICT,
                 ErrorResponse::new(
-                    "Drawing of lots is required",
-                    ErrorReference::ApportionmentDrawingOfLotsRequired,
-                    false,
+                    "Drawn lot is invalid",
+                    ErrorReference::ApportionmentInvalidLotDrawing,
+                    true,
                 ),
             ),
```
praseodym, line 62: This is already mapped to `APIError::Conflict` by the `process` function in `backend/src/service/apportionment.rs:223`

praseodym (approved):
Good improvement!

However, we should do something about all the ignored `result_large_err` lints. It might be better to also stop using error types to report that drawing of lots is required in the apportionment crate itself. It's not a blocker for this PR though.

backend/src/api/apportionment/mod.rs
Lionqueen94: The report generation calls process directly, so not through the service. So we thought we need it here as well for that reason.

praseodym: Let's just keep it then!

oliver3 (approved):

oliver3:
I've approved this PR and we will create a separate PR to change the drawing lots required from an error type to an enum.

PR #3423, opened 2026-06-16
Change DrawingLotsRequired to be variants of the Ok type instead of the Err type

oliver3:
## What & why

Using the "error path" to indicate that drawing lots is required resulted in clippy warnings and bloated error types and mappings. This PR changes that to use new Ok enum types to indicate that an apportionment step was completed or that drawing lots is required.

Typically, this means going from a signature like 

```rust
pub enum CandidateDrawingLotsError<LN, CN> {
    DrawingLotsRequired(CandidateDrawingLotsVariant<LN, CN>),
}

pub(crate) fn candidate_nomination<'a, L: ListVotes>(
    input: &CandidateNominationInput<'a, L>,
) -> Result<CandidateNominationResult<'a, L>, CandidateDrawingLotsError<ListNumber<LV>, CandidateNumber<LV>>> {
```

to 

```rust
pub enum CandidateNomination<'a, LV: ListVotes> {
    Completed(CandidateNominationDetails<'a, LV>),
    DrawingLotsRequired(CandidateDrawingLotsVariant<ListNumber<LV>, CandidateNumber<LV>>),
}

pub(crate) fn candidate_nomination<'a, L: ListVotes>(
    input: &CandidateNominationInput<'a, L>,
) -> Result<CandidateNomination<'a, L>, ApportionmentError> {
```

## How to test

Unit tests are changed accordingly and still succeed, the feature can be fully tested after the frontend has been done.

## Reviewer notes

Unfortunately this was a bit all or nothing. You can leave the fuzz tests for last.

### DIFF
--- a/backend/apportionment/src/seat_assignment/mod.rs
+++ b/backend/apportionment/src/seat_assignment/mod.rs
@@ -1897,6 +1897,7 @@ pub(crate) mod tests {
             };
 
             /// Apportionment with no residual seats
+            ///
             /// This test triggers Kieswet Article P 10
             ///
             /// Full seats: [5, 5, 4, 4, 2] - Remainder seats: 0  
--- a/backend/apportionment/src/seat_assignment/mod.rs
+++ b/backend/apportionment/src/seat_assignment/mod.rs
@@ -3,29 +3,29 @@ use std::cmp::Ordering;
 mod residual_seat_assignment;
 mod structs;
 pub use structs::{
-    AbsoluteMajorityReassignedSeat, ApportionmentWarning, AssignRemainderDrawingLotsError,
-    HighestAverageAssignedSeat, ListExhaustionRemovedSeat, ListStanding,
-    ResidualSeatDrawingLotsError, SeatAssignmentResult, SeatChange, SeatChangeStep,
+    AbsoluteMajorityReassignedSeat, ApportionmentWarning, HighestAverageAssignedSeat,
+    ListExhaustionRemovedSeat, ListStanding, SeatAssignmentDetails, SeatChange, SeatChangeStep,
 };
 use tracing::info;
 
-use self::{
-    residual_seat_assignment::assign_remainder,
-    structs::{AbsoluteMajorityResult, RemainderAssignmentResult},
-};
+use self::{residual_seat_assignment::assign_remainder, structs::RemainderAssignmentResult};
 use crate::{
-    ApportionmentInput, ListDrawn, ListVotes,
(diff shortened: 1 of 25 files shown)


backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -1810,11 +1891,12 @@ pub(crate) mod tests {
...
             use super::get_total_seats_from_apportionment_result;
             use crate::{
-                ApportionmentWarning, seat_assignment::seat_assignment,
+                ApportionmentWarning,
+                seat_assignment::{SeatAssignment, seat_assignment},
                 test_helpers::seat_assignment_fixture_with_given_candidate_votes,
             };
 
-            /// Apportionment with no residual seats  
+            /// Apportionment with no residual seats
```
Lionqueen94, line 1899:
```suggestion
            /// Apportionment with no residual seats  
```

backend/src/domain/report/files.rs
```diff
@@ -112,7 +116,11 @@ async fn generate_and_save_files_csb_election(
         state.get_lists_drawn(),
         state.get_candidates_drawn(),
     );
-    let apportionment_result = apportionment::process(&apportionment_input)?;
+    let ApportionmentOutput::Completed(apportionment_result) =
+        apportionment::process(&apportionment_input)?
+    else {
+        return Err(ApportionmentApiError::ApportionmentNotCompleted.into());
+    };
```
Lionqueen94, line 123: Might be good to add a test for this one

Lionqueen94 (dismissed):

Lionqueen94 (commented):
see previous comments

stacktraceghost (dismissed):
Tests succeeed and fuzz targets works.

backend/src/domain/report/files.rs
oliver3: This is not easily done unfortunately, and the real solution would be to not use apportionment from our domain, but to introduce a Provider (like we did with e.g. `CommitteeSessionHasInvestigationsProvider`) that can then be mocked to test its output.

I suggest not doing this in this PR.

backend/apportionment/src/seat_assignment/mod.rs
oliver3: Added a new line in https://github.com/kiesraad/abacus/pull/3423/commits/760803b79d8e9530d5688543b80771cab9c0c086

stacktraceghost (approved):

Lionqueen94 (approved):

PR #3432, opened 2026-06-17
Set overrides for mapping rust types globally in sqlx.toml

### LINKED ISSUE #3421: Investigate use of `sqlx.toml` for type overrides
As noted in #2645, we might be able to use type overrides in [sqlx.toml](https://docs.rs/sqlx/latest/sqlx/_config/_reference/index.html) to simplify our sqlx queries.

yusufraji:
## What & why

Set per-table and per-column overrides for mapping Rust types globally in sqlx.toml file.

Resolves #3421 
 
## Reviewer notes

The type overrides meant that the function signature of `abacus::infra::audit_log::audit_log_events::list_event_names` changes from `Result<Vec<String>, APIError>` to `Result<Vec<AuditEventType>, APIError>`. Rather than changing the function signature, which will lead to downstream failure and changes of tests, I decode the column as `String` instead (only in the function).

P.s: Callers of said function aren't relying on the type (`AuditEventType`), but rather on `String`. This could be addressed separately if it's important enough to work within the type system.

## Checklist

- [x] Single, focused change: unrelated changes split into separate PRs
- [x] I've self-reviewed the diff
- [x] Formatter and linter pass locally
- [x] Tests added/updated and passing
- [ ] Description explains how to test


backend/src/infra/audit_log/audit_log_events.rs
```diff
@@ -243,7 +243,7 @@ pub async fn assert_last_event(
 
 #[cfg(test)]
 pub async fn list_event_names(conn: &mut SqliteConnection) -> Result<Vec<String>, APIError> {
-    sqlx::query_scalar!(r#"SELECT event_name FROM audit_log ORDER BY ROWID ASC"#)
+    sqlx::query_scalar::<_, String>(r#"SELECT event_name FROM audit_log ORDER BY ROWID ASC"#)
```
praseodym: If possible we should stick to `sqlx::query_scalar!` here, the non-macro `sqlx::query_scalar` (without `!`) does not have compile-time validation.

praseodym (commented):
I think this is a good improvement! Please see the one comment about not using the macro function.

praseodym (approved):
Thanks for your contribution!

stacktraceghost (approved):

PR #3441, opened 2026-06-18
Serve Abacus using HTTPS

### LINKED ISSUE #3411: Serve Abacus using HTTPS
Serve Abacus using HTTPS when TLS feature is enabled. Listen on port 443 in production, port 8443 during development.

praseodym:
## What & why

Serve Abacus using HTTPS. Resolves #3411.

## How to test

Build frontend, then `cargo run --features memory-serve,tls`.

And in another terminal: `curl --cacert tls/ca.pem https://localhost:8443/`

Or trust `tls/ca.pem` / `tls/ca.cer` and go to https://localhost:8443/ in your browser.

## Reviewer notes

None.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -133,7 +133,7 @@ jobs:
       - name: Install Nextest
         run: cargo binstall cargo-nextest
       - name: Run tests
-        run: cargo --verbose --locked nextest run --target=x86_64-unknown-linux-musl --workspace --exclude pdf_gen_dylib --no-default-features --features openapi,dev-database,embed-typst
+        run: cargo --verbose --locked nextest run --target=x86_64-unknown-linux-musl --workspace --exclude pdf_gen_dylib --no-default-features --features openapi,dev-database,embed-typst,tls
       - name: Build documentation
         run: cargo doc --workspace --no-deps --all-features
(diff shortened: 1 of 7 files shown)


stacktraceghost (approved):
Tested on Ubuntu (Windows WSL): 
- Builds
- cURL command gives response, no error

Tested on Windows:
- Builds
- cURL returns error: `curl: (60) schannel: the revocation status is unknown`

This seems to be caused by cURL using Schannel instead of openssl, which does revocation checks. 

If you use `curl --ssl-no-revoke --cacert tls/ca.pem https://localhost:8443/`, it works. 

Importing `ca.cer` to the certificate store in Windows, automatically imports it to `Intermediate Certification Authorities`, which is not enough for browsers to accept it. When imported to the store `Trusted Root Certification Authorities`, browsers accept the certificate.  Tested on Edge, Firefox, Chrome.

oliver3 (approved):
Reviewed the code, and tested on MacOS 15 (Sequoia)

Curl accepted the connection with cacert, and rejected it without it.

Imported and trusted the `ca.pem` in the System Keychain and Firefox (after a restart) and Safari showed that the connection to Abacus on localhost was secure, with Firefox reporting that it is an unrecognised certificate issuer.

PR #3442, opened 2026-06-18
Fix bug incorrect list standings in SeatChangeStep for absolute majority reassignment

oliver3:
## What & why

The SeatChangeStep for absolute majority reassignment contained updated list standings instead of the previous list standings.

Two other improvements were made to prevent other issues:
- Make ListStanding fields read-only, use add/remove seat methods instead, to keep next_votes_per_seat up to date
- Assert that a list standing is found when looking one up by its number

## How to test

Unit tests are still working.

### DIFF
--- a/backend/apportionment/src/seat_assignment/structs.rs
+++ b/backend/apportionment/src/seat_assignment/structs.rs
@@ -89,17 +89,17 @@ impl<LN: Copy + Debug> From<ListStanding<LN>> for ListSeatAssignment<LN> {
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct ListStanding<LN> {
     /// List number for which this standing applies
-    pub list_number: LN,
+    list_number: LN,
     /// The number of votes cast for this list
-    pub votes_cast: u64,
+    votes_cast: u64,
     /// The remainder of votes that was not used to get full seats (does not have to be a whole number of votes)
-    pub remainder_votes: Fraction,
+    remainder_votes: Fraction,
     /// Whether the remainder votes meet the threshold to be applicable for largest remainder seat assignment
-    pub meets_remainder_threshold: bool,
+    meets_remainder_threshold: bool,
     /// The number of votes per seat if a new seat would be added to the current residual seats
-    pub next_votes_per_seat: Fraction,
+    next_votes_per_seat: Fraction,
     /// The number of full seats this list got assigned
-    pub full_seats: u32,
+    full_seats: u32,
     /// The current number of residual seats this list got assigned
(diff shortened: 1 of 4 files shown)


backend/apportionment/src/seat_assignment/structs.rs
```diff
@@ -133,15 +133,31 @@ impl<LN: Debug> ListStanding<LN> {
         }
     }
 
-    /// Add a residual seat to the list and return the updated instance
-    pub fn add_residual_seat(self) -> Self {
+    /// Remove a full seat to the list
```
stacktraceghost:
```suggestion
    /// Remove a full seat from the list
```
and 3x below (2x in `info!` and doc of `remove_residual_seat`)

stacktraceghost (commented):
It would be nice to add some tests/assertions to check if the correct standings are added to the steps. This is another point on our list and could be done in a separate PR.

backend/apportionment/src/seat_assignment/structs.rs
```diff
@@ -89,17 +89,17 @@ impl<LN: Copy + Debug> From<ListStanding<LN>> for ListSeatAssignment<LN> {
...
-    pub meets_remainder_threshold: bool,
+    meets_remainder_threshold: bool,
     /// The number of votes per seat if a new seat would be added to the current residual seats
-    pub next_votes_per_seat: Fraction,
+    next_votes_per_seat: Fraction,
     /// The number of full seats this list got assigned
-    pub full_seats: u32,
+    full_seats: u32,
     /// The current number of residual seats this list got assigned
     pub residual_seats: u32,
```
praseodym: Shouldn't this field also be made private?

praseodym:
> It would be nice to add some tests/assertions to check if the correct standings are added to the steps. This is another point on our list and could be done in a separate PR.

Indeed!

backend/apportionment/src/seat_assignment/structs.rs
oliver3: Sounds better :) changed in https://github.com/kiesraad/abacus/pull/3442/commits/50e21bcd9420d02b47879d35d857c925c664b348

backend/apportionment/src/seat_assignment/structs.rs
oliver3: It should, this is the remainder of a debugging session. Removed the `pub` again in https://github.com/kiesraad/abacus/pull/3442/commits/6670d46d3a32398645e174e64d8329834242415e

stacktraceghost (approved):

praseodym (approved):

Issue #3355, opened 2026-06-01
Mutation testing of apportionment and extending/updating apportionment tests where it makes sense


jschuurk-kr:

Hiddentale:
I ran cargo-mutants on the apportionment folder. It caught 8 mutations that survived. 
Could I pick this up and write the unit tests for it? I see that it has to be done before June 21, that should be doable for me.

Lionqueen94:
> I ran cargo-mutants on the apportionment folder. It caught 8 mutations that survived. Could I pick this up and write the unit tests for it? I see that it has to be done before June 21, that should be doable for me.

Hi @Hiddentale It would be awesome if you can pick this up! We would prefer this to be solved in small PRs (for example per file) so they are easy and quick to review. Thanks a lot for offering to pick this up!

Hiddentale:
I'll keep a list here of each survived mutation and update it when I make progress:

- [x] `apportionment/src/seat_assignment/mod.rs:39:25: replace == with != in seat_assignment`
Does not change return value, aka trivially equivalent mutation, no extra testing necessary
(The `==` check is just an early-exit log message, not a branch that changes the return value.)

- [x] `apportionment/src/seat_assignment/mod.rs:62:60: replace > with >= in seat_assignment`
Does not change return value, aka trivially equivalent mutation, no extra testing necessary
       (If residual_seats >= 0 instead of residual_seats > 0 and we set residual_seats = 0, the while loop in assign_remainder     
won't run and hence the return value is still correct)

- [x] `apportionment/src/seat_assignment/mod.rs:227:57: replace > with >= in reassign_residual_seat_for_absolute_majority`
This one is actually a proper mutation that needs a new unit test. When a list has *exactly* half of the votes, it should not check for absolute majority.

- [x] `apportionment/src/seat_assignment/residual_seat_assignment.rs:160:8: delete ! in exhausted_list_numbers`
Trivially equivalent mutation, mutation only triggers a debug log

- [x] `apportionment/src/seat_assignment/residual_seat_assignment.rs:197:60: replace == with != in list_qualifies_for_extra_seat`
Since `number_of_seats_unique_highest_averages` can ever only be 1 or 0, if anyone accidentally replaces == with !=  the whole second OR clause will be permanently false. This is a proper mutation that needs a new unit test.

- [x] `apportionment/src/seat_assignment/residual_seat_assignment.rs:218:22: replace > with >= in list_standings_qualifying_for_largest_remainder`
[...]

PR #3447, opened 2026-06-18
Resolve mutation in reassign_residual_seat_for_absolute_majority  

### LINKED ISSUE #3445: Mutation survives in reassign_residual_seat_for_absolute_majority
Subissue of #3355 

`apportionment/src/seat_assignment/mod.rs:227:57: replace > with >= in reassign_residual_seat_for_absolute_majority`

When a list has exactly half of the votes, it should not check for absolute majority. This is a proper mutation that needs a new unit test.

Hiddentale:
## What & why

Kills a surviving mutation on the Article P 9 Kieswet boundary check. The mutant changes `>` to `>=` on line 284 of `backend/apportionment/src/seat_assignment/mod.rs` making exactly 50% of votes trigger an absolute majority reassignment. Now if anyone accidentally makes this change in the future, this unit test will flag it by failing.

closes #3445
part of #3355 

## Reviewer notes

The vote split `[500, 300, 200]` (total: 1000) gives list 1 exactly 50%. 
7 seats produces 6 full seats and 1 residual seat, so `reassign_residual_seat_for_absolute_majority` is actually called.


Lionqueen94 (approved):
Nice work, thanks!

stacktraceghost (approved):

Issue #3408, opened 2026-06-11
Reflect drawing lots in SeatChangeStep


oliver3:
See https://github.com/kiesraad/abacus/pull/3404#discussion_r3390806907
Add drawing lots variant to SeatChangeStep

To properly implement https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=20682-53654&m=dev I need to have the averages for each list when drawing lots is required. So it would be nice if we can add a step with the usual information, but with an empty selected_list_number and indicating that drawing lots is needed (solved differently in https://github.com/kiesraad/abacus/pull/3452)

Will also fix issue that last assigned residual seat is not shown in ApportionmentTable (list 9 got a residual seat in the final step before drawing lots): (fixed already with https://github.com/kiesraad/abacus/pull/3448)
<img width="848" height="664" alt="Image" src="https://github.com/user-attachments/assets/109dd63a-2dae-4ce8-8eef-9b6fee1322a1" />

PR #3448, opened 2026-06-18
Return apportionment preliminary standings when drawing lots required

oliver3:
## What & why

To be able to show all information necessary to the user when drawing lots is required, the preliminary standings have to be returned. This solves a part of https://github.com/kiesraad/abacus/issues/3408.

We've renamed the `final_standings` to `standings` (first commit) and also return preliminary standings in this field, in case of drawing lots required.

## How to test

Unit tests have been extended with the residual seats from standings, to be able to verify that they match with the residual seat assignment steps.

## Reviewer notes
Review the first commit for the rename, and the second commit for returning standings also in case of drawing lots required.

Please verify that the residual seats in the unit tests are as expected.


stacktraceghost (approved):

Lionqueen94 (approved):
Nice change, also reviewed the tests, all looks good!

PR #3430, opened 2026-06-16
Update typst 0.15

NullP01nt:
## What & why

This PR updates pdf_gen to use Typst 0.15 under the hood
Resolves #3424 

## How to test

Existing tests can be executed using `cargo test`

## Reviewer notes

Not applicable

## Checklist

- [x] Single, focused change: unrelated changes split into separate PRs
- [x] I've self-reviewed the diff
- [x] Formatter and linter pass locally
- [x] Tests added/updated and passing
- [x] Description explains how to test

PR #3449, opened 2026-06-19
Fix PDF diff workflow for PRs from forks

praseodym:
## What & why

Fix PDF diff workflow for PRs from forks by using the workflow summary instead of a PR comment, because PR comments require `pull-requests: write` permissions, which fork PRs do not have. A warning annotation is also added to the Typst template files to make the changes more visible.

This is useful for e.g. #3430.

## How to test

The workflow should run on this PR without finding any PDF differences. It can be tested more thorougly by creating a PR in your own GitHub fork that includes the commit from this branch and some Typst template changes.

### DIFF
--- a/.github/workflows/pdf-diff.yml
+++ b/.github/workflows/pdf-diff.yml
@@ -6,14 +6,12 @@ on:
 
 permissions:
   contents: read
-  pull-requests: write
 
 env:
   TZ: "Europe/Amsterdam"
 
 jobs:
   diff:
-    if: "!github.event.pull_request.head.repo.fork"
     name: Diff PDFs
     runs-on: ubuntu-latest
     defaults:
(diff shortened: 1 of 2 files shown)


oliver3:
It is not very clear to me in a PR from a fork that there are differences detected. Is there a way to make it more prominent that it found differences?

Please see https://github.com/kiesraad/abacus/pull/3454 for a test run from a fork.

image

image

oliver3 (approved):
Tried to find alternatives (unfortunately `setNeutral()` is not yet implemented in [@actions/core](https://www.npmjs.com/package/@actions/core)) but could not find any.

This works as intended, warnings are generated, the summary shows the table with results, and the ZIP file can still be downloaded.

Image

Lionqueen94 (approved):

PR #3450, opened 2026-06-19
Frontend drawing lots for list

### LINKED ISSUE #1264: Frontend: drawing lots screens for list

Lionqueen94:
Closes #1264 

## What & why

- Add drawing of lots for list page (https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=20682-55169&m=dev)
  - Implemented with suggestions given in this [comment](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=17533-41762&m=dev#1819948776)
- Add all alerts and other additions to ApportionmentPage and ApportionmentResidualSeatsPage (https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=20478-15334&m=dev)
- Add drawing lots variants in frontend `testing/` folder to be used in tests
- Add tests for all alerts, redirects and tables in case of drawing lots
- Alert when drawing lots has been done (https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=20478-17902&m=dev)
  - Implemented with suggestions given in this [comment](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=20478-17813&m=dev#1822559731)
image
image

## How to test

- Generate drawing lots variants test elections
- Go through the apportionment flow for:
  - Drawing lots >= 19 seats, HighestAverageResidualSeat (CSB)
  - Drawing lots < 19 seats, LargestRemainderResidualSeat (CSB)
- Check storybook for the added stories for DrawingLotsForList

## Reviewer notes

Probably best reviewed per commit

Question: Is it fine if the `InvalidLotDrawing` error gives the `Abacus is stuk` page? This can only happen if someone is messing around and is not an expected error in the regular flow (even when changing URLs).


frontend/src/features/apportionment/components/drawing_lots/DrawingLotsForList.tsx
```diff
@@ -0,0 +1,62 @@
...
+        variant: "HighestAverageResidualSeat";
+      })
+    | (LargestRemainderResidualSeatDrawingLots & {
+        variant: "LargestRemainderResidualSeat";
+      });
+  options: PoliticalGroup[];
+}
+
+export function DrawingLotsForList({ drawingLotsRequired, options }: DrawingLotsForListProps) {
+  const total_residual_seats = drawingLotsRequired.residual_seat_numbers.at(-1);
```
praseodym, line 25: With `"residual_seat_numbers":[2,3,4]` this returns 4 seats to be assigned, I think it should be `drawingLotsRequired.residual_seat_numbers.length`

frontend/src/features/apportionment/components/residual_seats/ApportionmentResidualSeatsPage.tsx
```diff
@@ -201,12 +233,14 @@ function SmallCouncilSection({
...
           politicalGroups={politicalGroups}
           footNotes={resultChanges.length > 0 ? footNotes : undefined}
+          state={state}
         />
       )}
     </>
   );
 }
 
+// biome-ignore lint/complexity/noExcessiveLinesPerFunction: TODO: Is there any way to make this shorter?
```
praseodym: Of course there will be a way to make it shorter, but I think it's better to create a tech debt issue for this rather than leave a TODO like this in the codebase

frontend/src/features/apportionment/components/ApportionmentPage.tsx
```diff
@@ -79,21 +89,150 @@ function renderFinalisedAlert(
...
+function renderHighestAverageOrLargestRemainderDrawingLotsAlert(
+  residualSeatNumbers: number[],
+  variant: "HighestAverageResidualSeat" | "LargestRemainderResidualSeat",
+) {
+  return (
+    <FormLayout.Alert>
+      <Alert type="warning">
+        <strong className="heading-md">
+          {t("apportionment.drawing_lots_required_alert.title")}
+          {residualSeatNumbers.length > 1 && "s"} {formatList(residualSeatNumbers, t("and"))}
```
praseodym: Not that important for now, but building i18n strings like this might work for Dutch but not for all languages

praseodym (changes requested):

frontend/src/features/apportionment/components/drawing_lots/DrawingLotsForList.tsx
Lionqueen94: This value is the total number of residual seats, so also assigned ones without drawing lots. See https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17533-41772&m=dev

frontend/src/features/apportionment/components/ApportionmentPage.tsx
Lionqueen94: Good point! Shall I change this now?

stacktraceghost: Would be nice to also make this singular/plural like the other i18n strings

Lionqueen94: Fixed in 54affa5bf9ee9f3207eee8849ba0470c672862d6

frontend/src/features/apportionment/components/residual_seats/ApportionmentResidualSeatsPage.tsx
Lionqueen94: Actually fixed in 54affa5bf9ee9f3207eee8849ba0470c672862d6 by shortening variable names

frontend/src/features/apportionment/components/ApportionmentTable.stories.tsx
```diff
@@ -12,6 +12,7 @@ export const Default: StoryObj = {
         fullSeats={gte19Seats.seat_assignment.full_seats}
         residualSeats={gte19Seats.seat_assignment.residual_seats}
         seats={gte19Seats.seat_assignment.seats}
+        notAssignedSeats={0}
```
stacktraceghost, line 17: We could add a variant with `notAssignedSeats` + expect

frontend/src/features/apportionment/utils/utils.tsx
```diff
@@ -16,6 +17,22 @@ export function renderTitleAndHeader(sectionTitle: string) {
...
+export function renderNotAssignedSeatsAlert(notAssignedSeats: number, linkTo: string, linkText: string) {
+  return (
+    <Alert type="notify" small>
+      <p>
+        {getNotAssignedSeatsText(notAssignedSeats)} <Link to={linkTo}>{linkText}</Link>
+      </p>
+    </Alert>
+  );
+}
+
```
stacktraceghost, line 47: Feels like this could be a separate component including the outer `{notAssignedSeats > 0 && (` where these functions are being called

frontend/src/features/apportionment/components/ApportionmentPage.tsx
```diff
@@ -134,46 +270,30 @@ export function ApportionmentPage() {
...
-                </div>
+                {state.type === "Finalised"
+                  ? renderFinalisedAlert(
+                      warnings,
+                      currentCommitteeSession.id,
+                      () => void handleResetApportionmentState(),
+                    )
+                  : state.drawing_lots_required.type === "ListDrawingLotsRequired" &&
+                    state.drawing_lots_required.variant !== "AbsoluteMajorityHighestAverage" &&
+                    state.drawing_lots_required.variant !== "AbsoluteMajorityLargestRemainder" &&
```
stacktraceghost: Maybe we can make a util like 
```ts
export function isListDrawingLotsVariant<TVariant extends ListDrawingLotsVariant["variant"]>(
  state: ApportionmentState | undefined,
  variants: TVariant[],
): state is Extract<ApportionmentState, { type: "DrawingLots" }> & {
  drawing_lots_required: Extract<DrawingLotsRequired, { variant: TVariant }>;
} {
  if (state?.type !== "DrawingLots" || state.drawing_lots_required.type !== "ListDrawingLotsRequired") {
    return false;
  }
  const { variant } = state.drawing_lots_required;
  return variants.some((v) => v === variant);
}
```

And then call
```ts
isListDrawingLotsVariant(state, ["HighestAverageResidualSeat", "LargestRemainderResidualSeat"])
```

And same for other places where a similar comparison is being done

Edit: adjusted since type guard was necessary

stacktraceghost (changes requested):

frontend/src/features/apportionment/components/drawing_lots/DrawingLotsForList.tsx
praseodym: I see, thanks!

frontend/src/features/apportionment/components/ApportionmentPage.tsx
Lionqueen94: Added in 91d23916b7cabc38f38051a2d57defbc868bc89b! This awesome function makes it so much better!

frontend/src/features/apportionment/utils/utils.tsx
Lionqueen94: I'm not sure if this is a good idea. The `getNotAssignedSeatsText` function is also used in a different version of the alert, hence why it's in utils. Or we would have to move that version of the alert to this component also. That could be an option. What do you think?

frontend/src/features/apportionment/components/ApportionmentTable.stories.tsx
Lionqueen94: Done in eee787d50ce6dd1e0f4024f33a44aca2bc1d010e

frontend/src/features/apportionment/utils/utils.tsx
stacktraceghost: Ah I looked over the difference, lets keep it like this for now.

praseodym (approved):

stacktraceghost (approved):
Functionality works, screens look good.

> Question: Is it fine if the InvalidLotDrawing error gives the Abacus is stuk page? This can only happen if someone is messing around and is not an expected error in the regular flow (even when changing URLs).

I can reproduce this by opening two tabs for the drawing of lots and submit in one of them. If you then submit in the other, you will get the error.

Might be nice to show some kind of message in that case, but we could make a separate issue out of it and pick it up later?

jschuurk-kr:
> Functionality works, screens look good.
> 
> > Question: Is it fine if the InvalidLotDrawing error gives the Abacus is stuk page? This can only happen if someone is messing around and is not an expected error in the regular flow (even when changing URLs).
> 
> I can reproduce this by opening two tabs for the drawing of lots and submit in one of them. If you then submit in the other, you will get the error.
> 
> Might be nice to show some kind of message in that case, but we could make a separate issue out of it and pick it up later?

Am not convinced this needs addressing / an issue. There are a lot of improvements that are more important than showing a better error message when people are messing around.

PR #3451, opened 2026-06-19
Write Sigrid PR feedback to job summary

praseodym:
## What & why

Instead of creating a noisy PR comment, write the feedback to the job summary so it can be seen on the Checks tab.

## How to test

See the Checks tab on this PR.

## Reviewer notes

Do we prefer this approach or the comment? I personally do not like getting three automated comments on every PR, this adds a lot of notification noise. The PDF diff workflow is changed in #3449 and Codecov already reports a job summary, so we can [disable comments](https://docs.codecov.com/docs/pull-request-comments#disable-comment).


Lionqueen94 (approved):

oliver3 (approved):
And one github action dependency less!

PR #3452, opened 2026-06-19
Drawing of lots include list of averages/remainders

stacktraceghost:
## What & why

Related issue https://github.com/kiesraad/abacus/issues/3408

Changes:
- `HighestAverageResidualSeatDrawingLots`: 
  - Field `average` renamed to `max_average`.  
  - Field `list_averages` added with the `next_votes_per_seat` value for every list. 
- `LargestRemainderResidualSeatDrawingLots`: 
  - Field `remainder` renamed to `max_remainder`.
  - Field `list_remainders` added with the `remainder_votes` value for every list.

Since `list_standings_qualifying_for_unique_highest_average` in some cases only received qualified standings, the filter logic to do this has been moved inside the function. This way we full list of standings can be used to compile `list_averages`. 

The logic whether to qualify for unique highest averages or regular highest averages has been extracted to a separate function called `lists_qualifying_for_highest_average`. This method now also has its own testcases.

## How to test

Testcases should cover this change, however you can manually see the added `list_averages`/`list_remainders`:
1. Generating test elections through `Genereer meerdere verkiezingen met verschillende vormen van loting`
2. Login as `Coordinator CSB` and choose a variant, click on `Zetelverdeling`
3. Mark no deceased candidates and continue
4. Check the apportionment state response in your webdev toolbar, `drawing_lots_required` should contain `list_averages`/`list_remainders`.

## Reviewer notes

Review per commit.

### DIFF
--- a/backend/apportionment/src/seat_assignment/mod.rs
+++ b/backend/apportionment/src/seat_assignment/mod.rs
@@ -777,15 +777,16 @@ pub(crate) mod tests {
                     variant,
                     ListDrawingLotsVariant::LargestRemainderResidualSeat(
                         LargestRemainderResidualSeatDrawingLots {
+                            max_remainder: Fraction::new(0, 1),
                             residual_seat_numbers: vec![2],
                             options: vec![2, 3, 4, 5, 6],
                             list_remainders: vec![
                                 (1, Fraction::new(60, 1)),
-                                (2, Fraction::new(0, 15)),
-                                (3, Fraction::new(0, 15)),
-                                (4, Fraction::new(0, 15)),
-                                (5, Fraction::new(0, 15)),
-                                (6, Fraction::new(0, 15)),
+                                (2, Fraction::new(0, 1)),
+                                (3, Fraction::new(0, 1)),
+                                (4, Fraction::new(0, 1)),
+                                (5, Fraction::new(0, 1)),
+                                (6, Fraction::new(0, 1)),
                                 (7, Fraction::new(55, 1)),
                                 (8, Fraction::new(45, 1)),
                             ],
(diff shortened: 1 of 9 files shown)


frontend/src/types/generated/openapi.ts
```diff
@@ -1174,7 +1174,7 @@ export interface HighestAverageAssignedSeat {
 }
 
 export interface HighestAverageResidualSeatDrawingLots {
-  average: DisplayFraction;
+  list_averages: unknown[][];
```
stacktraceghost: We need to change the field type (also for `list_remainders`)

stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/3452/commits/d0424c64bd8401e8c95019646dafc78510f879e1

Lionqueen94 (approved):
Looks good! Nice addition to make sure the drawing of lots for averages can be shown in the table

backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -888,9 +906,17 @@ pub(crate) mod tests {
...
-                            remainder: Fraction::new(900, 15),
+                            max_remainder: Fraction::new(420, 7),
                             residual_seat_numbers: vec![2, 3, 4],
-                            options: vec![2, 3, 5, 6]
+                            options: vec![2, 3, 5, 6],
+                            list_remainders: vec![
+                                (1, Fraction::new(500, 25)),
+                                (2, Fraction::new(420, 7)),
+                                (3, Fraction::new(420, 7)),
+                                (4, Fraction::new(420, 7)),
```
oliver3, line 916: I did not expect the same remainder for list 4 again after drawing, and when it is not an option anymore. Is this correct?

Lionqueen94: These are from before the draw, to show what the draw is based on. This is also what we show in the averages table.

Lionqueen94: NB. This is about list remainders and they don't actually change as just discussed

oliver3: Thanks!

oliver3 (approved):

PR #3460, opened 2026-06-23
Drawing lots for absolute majority

### LINKED ISSUE #3273: Backend: process drawing lots for absolute majority reassignment (P9)

oliver3:
## What & why

- Resolves https://github.com/kiesraad/abacus/issues/3273
- Implements using drawing lots for absolute majority (P 9) in the apportionment crate

Implementation details:
- Update `reassign_residual_seat_for_absolute_majority()` to try and use a list from `lists_drawn` when drawing lots is needed
- Split `ListDrawingLotsVariant::AbsoluteMajority` into two variants, corresponding to a seat that is reassigned from a previous largest remainder or highest average seat change
- Add `assign_to` field to the `AbsoluteMajorityDrawingLots` variant details to be used in the frontend
- Add `ApportionmentError::InvalidState` variant and use it when a previous seat change is not one of the expected variants
- Helper functions and traits to keep the line count of `reassign_residual_seat_for_absolute_majority()` within acceptable limits

## How to test

Unit tests are adjusted accordingly, frontend will be built in later issues from https://github.com/kiesraad/abacus/issues/788

## Reviewer notes

Start in `backend/apportionment/src/seat_assignment/mod.rs` to see the core change in `reassign_residual_seat_for_absolute_majority()` and what the results are in the tests below. Then check what other changes were needed to make this happen.

### DIFF
--- a/backend/apportionment/src/seat_assignment/mod.rs
+++ b/backend/apportionment/src/seat_assignment/mod.rs
@@ -2,6 +2,8 @@ use std::cmp::Ordering;
 
 mod residual_seat_assignment;
 mod structs;
+#[cfg(test)]
+pub use structs::LargestRemainderAssignedSeat;
 pub use structs::{
     AbsoluteMajorityReassignedSeat, ApportionmentWarning, HighestAverageAssignedSeat,
     ListExhaustionRemovedSeat, ListStanding, SeatAssignmentDetails, SeatChange, SeatChangeStep,
--- a/backend/apportionment/src/seat_assignment/mod.rs
+++ b/backend/apportionment/src/seat_assignment/mod.rs
@@ -12,7 +12,9 @@ use self::{residual_seat_assignment::assign_remainder, structs::RemainderAssignm
 use crate::{
     ApportionmentError, ApportionmentInput, ListDrawn, ListVotes,
     fraction::Fraction,
-    seat_assignment::structs::{AbsoluteMajority, AbsoluteMajorityResult, RemainderAssignment},
+    seat_assignment::structs::{
+        AbsoluteMajority, AbsoluteMajorityResult, GetListStandingByNumber, RemainderAssignment,
+    },
     structs::{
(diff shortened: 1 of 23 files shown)


oliver3:
Patch coverage is pretty low, I will add some tests to cover helpers ~~and conversion of apportionment from/to abacus crate structs~~.

stacktraceghost (dismissed):

backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -287,52 +293,64 @@ fn reassign_residual_seat_for_absolute_majority<T: ListVotes>(
...
+                },
+            )?;
+
+            // Get a list from the lists_drawn
+            let Some(list_drawn) = lists_drawn.next() else {
+                return Ok(AbsoluteMajority::DrawingLotsRequired(variant));
+            };
+
+            // Assert the required variant including all data
+            list_drawn.variant().ensure_eq(&variant)?;
```
praseodym: It seems like we don't assert that `list_drawn` is one of the options, which we should definitely check to avoid invalid results.

In other drawing lots code we did include this check, e.g.:
https://github.com/kiesraad/abacus/blob/0d1e46775de0ab8e86442c79c4d37339d6b96c23/backend/apportionment/src/seat_assignment/residual_seat_assignment.rs#L354-L359

praseodym (changes requested):

backend/apportionment/src/seat_assignment/mod.rs
oliver3: Using a new `ListDrawingLotsVariant::validate()` in https://github.com/kiesraad/abacus/pull/3460/commits/be1f3222f7eb56ec59fa7cbd9ac214f99aab56f8 to validate both variant equality and that the list drawn is an option (3x)

praseodym (approved):

stacktraceghost (approved):

PR #3436, opened 2026-06-17
Model P 22-2 thousands separation 

NullP01nt:
## What & why

This fixes thousands-separation in model P 22-2.
Resolves #3425 

## How to test

Using "backend/templates/visual-diff.sh" you can visually review the output difference between model-p-22-2-old.pdf and model-p-22-2-new.pdf

Aside from the typesetting using that script, `cargo test infra::pdf_gen` will expose potential impact on the backend side as well.

## Reviewer notes

This is my first time working on larger Typst templates, if I've missed anything, please let me know.

PR #3461, opened 2026-06-23
Fixed rendering of fractions in highest averages table in Model P 22-2

Lionqueen94:
## What & why

The rendering of fractions in the highest averages table had broken with a change in #3436 
This PR fixes this.

## How to test

Render a PDF with one of the gte-19-seats inputs and/or check the below PDFs.

Broken PDF (see page 9):
[templates-3.pdf](https://github.com/user-attachments/files/29266151/templates-3.pdf)

Fixed PDF (see page 9):
[templates-2.pdf](https://github.com/user-attachments/files/29266115/templates-2.pdf)

## Reviewer notes

Unfortunately we missed this in the initial review, it would help to have the PDF diff check render all the variants, but this will be quite a refactor I think :/


praseodym:
> it would help to have the PDF diff check render all the variants, but this will be quite a refactor I think :/

I don't think this is too difficult, I'll see if I can change it to at least render using the inputs in `backend/templates/inputs/extra-model-p-22-2-variations`.

praseodym:
> > it would help to have the PDF diff check render all the variants, but this will be quite a refactor I think :/
> 
> I don't think this is too difficult, I'll see if I can change it to at least render using the inputs in `backend/templates/inputs/extra-model-p-22-2-variations`.

See #3464

stacktraceghost (approved):

praseodym (approved):
Looks like the new PDF diff works well!

PR #3464, opened 2026-06-24
Expand PDF diff with P 22-2 variants

praseodym:
## What & why

Expand PDF diff with P 22-2 variants as noted in https://github.com/kiesraad/abacus/pull/3461#issuecomment-4786694751. All variants are configured in `backend/src/bin/gen-pdf.rs`, which now also writes a manifest file that `pdf-diff.js` uses to link PDF files to the Typst templates.

## How to test

See CI. Because the manifest doesn't exist for `main`, all PDF files will probably get reported as new.

## Reviewer notes

None.

### DIFF
--- a/.github/workflows/scripts/pdf-diff.js
+++ b/.github/workflows/scripts/pdf-diff.js
@@ -14,23 +14,29 @@ exports.pdfDiff = ({ core }) => {
   // Ensure diff folder exists
   fs.mkdirSync(dirs.diff, { recursive: true });
 
-  // Reads .pdf files in the given directory (if it exists) and returns file names
-  const listPdfFiles = (root) => {
-    if (!fs.existsSync(root)) {
-      return [];
+  // Read the manifest gen-pdf writes next to the PDFs
+  const loadManifest = (root) => {
+    const manifestPath = path.join(root, "manifest.json");
+    if (!fs.existsSync(manifestPath)) {
+      return {};
     }
-
-    return fs
-      .readdirSync(root, { withFileTypes: true })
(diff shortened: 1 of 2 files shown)


Lionqueen94 (changes requested):
Amazing that we can add the variants now! :smile: 
It seems like the artifact upload goes wrong though:
`No files were found with the provided path: backend/tmp-pdf-gen/diffs/. No artifacts will be uploaded.`
The zip download results in a 404 right now.

praseodym:
> See CI. Because the manifest doesn't exist for `main`, all PDF files will probably get reported as new.

All files get reported as 'New'

> It seems like the artifact upload goes wrong though:
> `No files were found with the provided path: backend/tmp-pdf-gen/diffs/. No artifacts will be uploaded.`
> The zip download results in a 404 right now.

Because all files are new, there is nothing to diff in this branch

Lionqueen94 (approved):

Lionqueen94:
> > See CI. Because the manifest doesn't exist for `main`, all PDF files will probably get reported as new.
> 
> All files get reported as 'New'
> 
> > It seems like the artifact upload goes wrong though:
> > `No files were found with the provided path: backend/tmp-pdf-gen/diffs/. No artifacts will be uploaded.`
> > The zip download results in a 404 right now.
> 
> Because all files are new, there is nothing to diff in this branch

Ah I expected to have files with fully new text, but I guess nothing to diff against and then not showing the file at all makes sense.

stacktraceghost (approved):

Issue #3159, opened 2026-04-10
Data entry home and status overview improvements GSB


Lionqueen94:
- All success alerts on status overview and data entry home page should auto close after 5 minutes
- ProgressBar percentage should be with floored instead of rounded input percentage: example `Math.floor((totalFirstEntryFinished / total) * 100) : 0`
- Submitting data entry home number should work on "Enter" only also (not just "Shift" + "Enter")

PR #3465, opened 2026-06-24
Round progress percentages appropriately

jschuurk-kr:
## What & why

Solves one of the three findings in #3159:

> ProgressBar percentage should be with floored instead of rounded input percentage: example Math.floor((totalFirstEntryFinished / total) * 100) : 0

Reason it should be floor is to avoid rounding a percentage >= 99.5% to 100%, suggesting a completed status. However, always using floor also means we round a percentage between 0 and 1 down to 0 in some cases where we shouldn't.

So we introduced a function (hat tip to @stacktraceghost for the conversation and the pair programming) `calculateProgressPercentage()` that calculates the percentage and rounds it down, except when it's within a range provided, in which case it rounds up.

There are three occurrences where we calculate a progress percentage. The function is used in two of them:

1. In ElectionProgress, which is used for the progress bars for typists, any percentage >= 0.5 and < 1 is rounded up. So when you have e.g. 600 polling stations, the 0% of progress dissappears as soon as 3 polling stations have been entered.
2. In useElectionStatus, which is used for the progress bar per status on the election status page, any percentage >=0 and < 1 is rounded up. So when you have e.g. 600 polling stations of which 599 are complete and 1 has errrors, part of the progress bar is red.

*Sidenote on the MultiProgressBar: it accepts values that add up to a total larger than 100. It seems to handle that fine. Because of the data we provide it, the total will at most be slightly larger than 100.*

For the third occurrence, calculateDataEntryProgress, which is used for data entry progress of a single entry when a typist submits data entry, I did not introduce the function, since it serves no purpose. It calculates the percentage completed of the pages in data entry. Since it is extremely unlikely we will ever have more than 100 pages, the issue of rounding down to 0 or up to 100 when we shouldn't, will not occur.  
Since my initial solution was changing the round to a floor for all three occurrences, I have kept that change for this occurrence in this PR. At least it is consistent with floor being the default way to round in `calculateProgressPercentage()`.

## How to test

Tests were updated or added for each of the changes.

## Reviewer notes

No notes.

### DIFF
--- a/frontend/src/features/data_entry_home/components/ElectionProgress.test.tsx
+++ b/frontend/src/features/data_entry_home/components/ElectionProgress.test.tsx
@@ -29,4 +29,31 @@ describe("ElectionProgress", () => {
     expect(definitiveBar).toHaveTextContent("1e en 2e invoer klaar");
     expect(definitiveBar).toHaveTextContent("25%");
   });
+
+  test("rounds down the percentages in the progress bar", () => {
+    vi.spyOn(useElectionStatus, "useElectionStatus").mockReturnValue({
+      ...getElectionStatusMockData([
+        { status: "definitive" },
+        { status: "definitive" },
+        { status: "definitive" },
+        { status: "first_entry_finalised" },
+        { status: "first_entry_finalised" },
+        { status: "first_entry_finalised" },
+        { status: "first_entry_in_progress" },
+      ]),
+      refetch: vi.fn(),
+    });
+
+    render(<ElectionProgress />);
+
+    const firstEntryFinishedBar = screen.getByTestId("progressbar-first-entry-finished");
+    expect(firstEntryFinishedBar).toBeInTheDocument();
+    expect(firstEntryFinishedBar).toHaveTextContent("1e invoer klaar");
+    expect(firstEntryFinishedBar).toHaveTextContent("85%");
+
+    const definitiveBar = screen.getByTestId("progressbar-first-and-second-entry-finished");
+    expect(definitiveBar).toBeInTheDocument();
+    expect(definitiveBar).toHaveTextContent("1e en 2e invoer klaar");
+    expect(definitiveBar).toHaveTextContent("42%");
+  });
 });
--- a/frontend/src/features/data_entry_home/components/ElectionProgress.tsx
+++ b/frontend/src/features/data_entry_home/components/ElectionProgress.tsx
@@ -5,6 +5,7 @@ import { ProgressBar } from "@/components/ui/ProgressBar/ProgressBar";
 import { useElectionStatus } from "@/hooks/election/useElectionStatus";
 import { t } from "@/i18n/translate";
 import type { DataEntryStatusName } from "@/types/generated/openapi";
+import { calculateProgressPercentage } from "@/utils/progressPercentage";
 
 type Stat = {
   title: string;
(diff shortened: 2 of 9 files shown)


frontend/src/utils/progressPercentage.ts
```diff
@@ -0,0 +1,16 @@
...
+) {
+  // Calculates the progress percentage for count / total, rounding up for percentages within the threshold, rounding down for all other cases.
+  // This allows us to determine when to round up a percentage between 0 and 1 (e.g. 0.5%) to 1. And it makes sure we don't round a percantage of 99.5% or higher up to 100%.
+  if (total === 0) {
+    return 0;
+  }
+  const percentage = (count / total) * 100;
+  if (percentage >= ceilingThresholdFrom && percentage < ceilingThresholdTo) {
+    return Math.ceil(percentage);
+  } else return Math.floor(percentage);
```
stacktraceghost: Mixed use of curly braces. 

```suggestion
  const shouldRoundUp = percentage >= ceilingThresholdFrom && percentage < ceilingThresholdTo;
  return shouldRoundUp ? Math.ceil(percentage) : Math.floor(percentage);
```

frontend/src/features/data_entry_home/components/ElectionProgress.tsx
```diff
@@ -31,12 +32,14 @@ export function ElectionProgress() {
       {
         title: t("status.first_entry_finished_short"),
         id: "first-entry-finished",
-        percentage: total > 0 ? Math.round((totalFirstEntryFinished / total) * 100) : 0,
+        // The percentage is rounded up when between 0.5 and 1, so that progress is not set to 0% for too long, when the total is large.
+        percentage: total > 0 ? calculateProgressPercentage(totalFirstEntryFinished, total, 0.5, 1) : 0,
```
stacktraceghost: `total > 0` is not necessary anymore on these lines, `calculateProgressPercentage` already catches that.

frontend/src/features/data_entry_home/components/ElectionProgress.test.tsx
```diff
@@ -29,4 +29,31 @@ describe("ElectionProgress", () => {
     expect(definitiveBar).toHaveTextContent("1e en 2e invoer klaar");
     expect(definitiveBar).toHaveTextContent("25%");
   });
+
+  test("rounds down the percentages in the progress bar", () => {
+    vi.spyOn(useElectionStatus, "useElectionStatus").mockReturnValue({
+      ...getElectionStatusMockData([
+        { status: "definitive" },
```
stacktraceghost, line 36: We could do something like this:
```
const statuses = [
  ...Array.from({ length: 99 }, () => ({ status: "definitive" as const })),
  ...Array.from({ length: 100 }, () => ({ status: "first_entry_finalised" as const })),
  { status: "first_entry_in_progress" as const },
];
```

This would recreate the problem we were trying to solve:
- 1e invoer klaar has `(199 / 200) * 100 = 99,5%` which turns into `99%` instead of `100%`
- 1e en 2e invoer klaar has `(99 / 200) * 100 = 49,5%` which turns into `49%` instead of `50%`

Edit: ah `electionStatusesMock` doesn't have enough statuses to do this. So you can ignore this comment.

Lionqueen94 (dismissed):
Looks good!

frontend/src/utils/progressPercentage.ts
jschuurk-kr: nice!

frontend/src/features/data_entry_home/components/ElectionProgress.tsx
jschuurk-kr: good one, fixed in b88bf9b0c185505185cbe4de8d5bc6baa4813766

Lionqueen94 (approved):

stacktraceghost (approved):

PR #3471, opened 2026-06-25
Drawing lots for candidate nominations (P 15)

oliver3:
## What & why

- Implements https://github.com/kiesraad/abacus/issues/3150

## How to test

Unit test `test_with_drawing_of_lots_error()` has been updated with drawing lots for candidates and asserting the result

Unit test `test_preferential_candidate_nomination_scenarios()` tests multiple scenarios for candidate nomination (P 15)

Frontend will be implemented in a later issue from Epic https://github.com/kiesraad/abacus/issues/788

## Reviewer notes

Main change is in `candidate_nomination_per_list()` in `backend/apportionment/src/candidate_nomination/mod.rs` and I suggest starting there.

Two other changes are:
- Change `CandidateNominationInput::total_seats_per_list` from a `Vec` to a `HashMap`
- Change `CandidateVotesMock` to a tuple struct for one-line construction, to improve test readability

I've tried adding the `candidates_drawn` iterator to the `CandidateNominationInput` struct instead of passing it as a separate argument, but the generics became even more complex, and also the whole input would then have to become mutable. I decided that that was not making things more clear.

### DIFF
--- a/backend/apportionment/src/candidate_nomination/mod.rs
+++ b/backend/apportionment/src/candidate_nomination/mod.rs
@@ -6,7 +6,7 @@ pub use structs::{
 use tracing::{debug, info};
 
 use crate::{
-    ApportionmentError, CandidateVotes, ListVotes,
+    ApportionmentError, CandidateDrawn, CandidateVotes, ListVotes,
     fraction::Fraction,
     structs::{
         CandidateDrawingLotsVariant, CandidateNominationInput, CandidateNumber, DeceasedCandidates,
@@ -24,6 +24,9 @@ pub enum CandidateNomination<'a, LV: ListVotes> {
 #[expect(clippy::cognitive_complexity)]
 pub(crate) fn candidate_nomination<'a, L: ListVotes>(
     input: &CandidateNominationInput<'a, L>,
+    candidates_drawn: impl Iterator<
+        Item = &'a (impl CandidateDrawn<ListNumber<L>, CandidateNumber<L>> + 'a),
+    >,
 ) -> Result<CandidateNomination<'a, L>, ApportionmentError> {
     info!("Candidate nomination");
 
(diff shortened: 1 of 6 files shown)


stacktraceghost (approved):

backend/apportionment/src/structs.rs
```diff
@@ -127,13 +127,38 @@ pub trait ListDrawn<LN> {
...
-pub struct CandidateDrawingLotsVariant<LN, CN> {
+pub struct CandidateDrawingLotsVariant<LN: PartialEq, CN: PartialEq> {
     pub list: LN,
     pub options: Vec<CN>,
 }
 
+impl<LN: PartialEq, CN: PartialEq> CandidateDrawingLotsVariant<LN, CN> {
+    /// Validate the candidate drawn against this variant.
+    ///
+    /// Return [[ApportionmentError::InvalidLotDrawing]] if the variant is different
```
praseodym, line 138: non-blocking, no need to fix: single bracket is enough for rustdoc

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -248,55 +254,73 @@ fn other_candidate_nomination<'a, T: CandidateVotes>(
...
+#[derive(Debug)]
 enum PreferentialCandidateNomination<'a, LV: ListVotes> {
     Completed(Vec<&'a LV::Cv>),
     DrawingLotsRequired(CandidateDrawingLotsVariant<ListNumber<LV>, CandidateNumber<LV>>),
 }
 
 /// List the candidates nominated with preferential votes
+/// * list - List number
+/// * candidates - Candidates that meet the preference_threshold,
+///   sorted by votes descending then by candidate number
```
praseodym, line 266: non-blocking: I don't think we explicitly do a secondary sort by candidate number, so this depends on the candidates being sorted by candidate number in the apportionment input. This will be the case for inputs from Abacus, but we should be a more thorough because apportionment can also be used as a library.

praseodym (approved):

backend/apportionment/src/candidate_nomination/mod.rs
stacktraceghost: Created two issues:
1. https://github.com/kiesraad/abacus/issues/3478 
2. https://github.com/kiesraad/abacus/issues/3479

PR #3474, opened 2026-06-26
Add plain HTTP server for redirect and CA download

### LINKED ISSUE #3412: HTTP redirect to HTTPS, HTTP CA download
Redirect HTTP (port 80) to HTTPS but provide CA download over HTTP for ease of installation.

praseodym:
## What & why

Add a plaintext HTTP server that redirects to HTTPS and makes the CA certificates available on HTTP, so that the CA certificate can be imported into client browser trust stores. Resolves #3412.

We need to document that the fingerprint of the downloaded certificate should be verified against the fingerprint logged by the Abacus server, to prevent man-in-the-middle attacks.

## How to test

From a terminal:
```shell
cargo run --features tls
# and in another terminal
curl -v localhost:8080
curl -v localhost:8080/ca.pem
curl -v localhost:8080/ca.cer
curl --cacert tls/ca.pem -v https://localhost:8443/ca.pem
curl --cacert tls/ca.pem -v https://localhost:8443/ca.cer
```

Then check that the redirect works and that the CA certificates are available.

## Reviewer notes

None.

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -114,11 +114,19 @@ server (leaf) certificate is created in memory on every start, signed by the CA
 and covering `localhost`, `abacus.local`, and all routable LAN addresses.
 
 To trust the server, import the CA into the client trust store: `ca.pem` on
-Linux/macOS/Firefox, `ca.cer` (DER) on Windows.
+Linux/macOS/Firefox, `ca.cer` (DER) on Windows. The CA can be downloaded from the
+running server at `/ca.pem` and `/ca.cer`.
 
 With the `tls` feature enabled, the default port is 8443 in debug builds and 443
 in release builds. Binding to 443 requires elevated privileges (e.g. the
 `CAP_NET_BIND_SERVICE` capability on Linux).
+
+Alongside the HTTPS port, Abacus runs a plain HTTP server that provides the CA
+certificate at `/ca.pem` and `/ca.cer` over plain HTTP (so clients can fetch it
+before trusting the server) and redirects every other request to HTTPS. Its port
+is set by `--http-port` / `ABACUS_HTTP_PORT`, defaulting to 80 in release builds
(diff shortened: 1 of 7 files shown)


backend/README.md
```diff
@@ -114,12 +114,20 @@ server (leaf) certificate is created in memory on every start, signed by the CA
 and covering `localhost`, `abacus.local`, and all routable LAN addresses.
 
 To trust the server, import the CA into the client trust store: `ca.pem` on
-Linux/macOS/Firefox, `ca.cer` (DER) on Windows.
+Linux/macOS/Firefox, `ca.cer` (DER) on Windows. The CA can be downloaded from the
+running server at `/ca.pem` and `/ca.cer`.
```
stacktraceghost, line 118:
```suggestion
running plain HTTP server at `/ca.pem` and `/ca.cer`.
```

stacktraceghost (approved):
Tested on Windows and Linux, both work.

> We need to document that the fingerprint of the downloaded certificate should be verified against the fingerprint logged by the Abacus server, to prevent man-in-the-middle attacks.

Is there an issue for this? We also need some documentation for adding the CA cert in Windows to the correct store.

praseodym:
> Is there an issue for this? We also need some documentation for adding the CA cert in Windows to the correct store.

Just created https://github.com/kiesraad/abacus-documentatie/issues/76, thanks for mentioning this!

backend/README.md
praseodym: It also works on the HTTPS server, so strictly speaking this is not correct :)

Lionqueen94 (approved):
I'm not too familiar with setting up tls, but the steps in the description work and the redirect as well, so seems fine.

rnijveld (approved):

PR #3485, opened 2026-06-30
Replace 520 candidate ids with ranking order numbers

### LINKED ISSUE #3429: Fix EML520 discrepancies between OSV and Abacus
discovered in #3070 / #3072

See diff between Abacus and OSV in [this pdf](https://github.com/user-attachments/files/28272698/EML-diff.pdf).

### Scope
- [x] `<AuthorityIdentifier Id="0340">Rhenen</AuthorityIdentifier>`: OSV has `Id="CSB"` instead of the municipal code. (Also the case in the 510d.) Fixed in #3476 
- [x] Each `<Selection>` with a `<AffiliationIdentifier>` contains a `<Ranking>1</Ranking>` which should not be there. Fixed in #3484 
- [x] For the `Id` of the `CandidateIdentifier ` Abacus uses the `Id` from the candidates lists. OSV numbers the elected candidates incrementally. Abacus should adopt the OSV behavior, because when you have voting districts the combination of `Affiliation` and this `Id` does not uniquely identify a candidate. Fixed in #3485 
- [x] (optionally) https://github.com/kiesraad/abacus/issues/3365 Fixed in #3486

rnijveld:
## What & why

Fixes #3429. This fixes the final part of that issue: candidate ids in 520 documents are replaced by ranking order numbers. I considered replacing the types in the EML crate as well, but decided against this because it would introduce another round trip of review work and releases of the EML crate. Additionally, the 520 format is already not the greatest format anyway, I would probably prefer to re-examine that format some time in the future and then update types according to that.

## How to test

Complete a CSB election making sure that there is a candidate that is preferentially selected and will be re-ordered such that candidate numbers won't match the ranking of the candidate. Notice how in the EML520 file download the CandidateId element now reflects the ranking, and not the original candidate id.


praseodym (dismissed):
@rnijveld this PR has a merge conflict

stacktraceghost:
- Fixed apportionment crate tests in https://github.com/kiesraad/abacus/pull/3485/commits/2ac26c09cabb08d334987bdcb22a06ae2ff275cb
  - `check_list_candidate_nomination` assumes we expect an original list if the given `expected_updated_ranking` is empty
  - Otherwise check if the list is of variant updated
- Renamed `UpdatedCandidateRanking` to `CandidateRanking` within apportionment crate to make it more clear (https://github.com/kiesraad/abacus/pull/3485/commits/bb055f0ac14eb0f2668075b3eb2689439a540bc2)
  - Within Abacus we still use `updated_candidate_ranking`, since we only store updated rankings there

Lionqueen94 (approved):
Looks good to me, good change!

praseodym (approved):

PR #3487, opened 2026-06-30
Require drawing of lots even if equal votes and enough seats

### LINKED ISSUE #3480: Backend: Draw lots for candidates with equal preferential votes even when enough empty seats
Quote from @legally-blonde:
> Ha iedereen,
>
> In het geval dat er twee kandidaten zijn die met dezelfde hoeveelheid stemmen de voorkeursdrempel halen, wordt er geloot, ook als er voldoende zetels zijn om toe te wijzen. Op grond van artikel P 15, eerste lid, van de Kieswet wordt er geloot als aantallen gelijk zijn. Aangezien er verder geen uitzonderingen zijn vermeld, geldt dit dus altijd wanneer er gelijke aantallen zijn.
>
> Voor de andere varianten geldt dat alle antwoorden kloppen.
>

stacktraceghost:
## What & why

Resolves #3480

If we had enough seats to nominate all candidates or enough seats to nominate candidates with equal votes, we would do so. However, we should always draw lots whenever there are candidates with equal votes.

As a result, we will always go through the loop of `preferential_candidate_nomination`. 

Fixture `backend/fixtures/election_8_csb_with_results.sql` has been updated to not have equal amount of votes among the candidates to prevent other tests from failing: 
- Candidate 2 -> 41 instead of 40
- Candidate 10 -> 39 instead of 40
- Candidate 5 -> 101 instead of 100
- Candidate 11 -> 99 instead of 100.

## How to test

Check if the test scenarios are covering every possibility. 

## Reviewer notes

Test cases have been added to `test_preferential_candidate_nomination_scenarios` and existing tests have been adjusted so that they do not have an equal amount of votes between the candidates. 

Note: `nomination.push(candidates[index]);` has been changed to `nomination.push(same_votes_candidates_remaining[0]);` because it would otherwise add the same candidate twice in this scenario:
- Candidate votes: 1000, 1000, 900. Seats: 2
- First loop (index=0): drawing of lots is required, options: 1, 2, 3. Selected = 2
- Second loop (index=1): there is only one candidate with 1000 votes, so we will add it with `nomination.push(candidates[index]);`. This results in candidate 2 being added twice.


Lionqueen94 (approved):
Looks good!

praseodym (approved):

PR #3488, opened 2026-06-30
Add HTTP Strict-Transport-Security (HSTS) header

### LINKED ISSUE #3483: Add HSTS header
Add [Strict-Transport-Security header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security) when TLS is enabled.

praseodym:
## What & why

Add [HTTP `Strict-Transport-Security` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security) responses over HTTPS, when the TLS feature is enabled. This ensures that browsers will use HTTPS to access Abacus from then on. Resolves #3483.

## How to test

```
cargo run --features tls
# and in another terminal, check that the HSTS header is present:
curl --cacert tls/ca.pem -v https://localhost:8443/
# and that it is not present on HTTP:
curl -v localhost:8080
```

## Reviewer notes

None.


stacktraceghost (dismissed):

backend/src/lib.rs
sanderboom, line 92:
```suggestion
/// Start the API server over plain HTTP on the given port, using the given database pool.
```

sanderboom (changes requested):
Works on my computer! :) 

Besides the comment change, we could rename `start_server` to `start_server_plain` (or `start_server_http`) to make the distinction from `start_server_tls` slightly more explicit.

backend/src/lib.rs
praseodym: Good suggestion, but I don't want to do unrelated changes in this PR.

praseodym:
> Besides the comment change, we could rename `start_server` to `start_server_plain` (or `start_server_http`) to make the distinction from `start_server_tls` slightly more explicit.

That's a good idea, but this code is not changed in this PR so I'm going to leave it as-is.

praseodym:
Resolved merge conflict in 990a702f5e86f2f69d5e9c9a8056a54aa41ecebc

stacktraceghost (approved):

sanderboom (approved):

PR #3489, opened 2026-06-30
Use CSB fixtures for apportionment tests

stacktraceghost:
## What & why

When fixing tests for #3487, it became clear that apportionment tests were still using GSB fixtures instead of CSB. 

## How to test

CI should pass

## Reviewer notes

N/A


Lionqueen94 (approved):
Good find!

praseodym (approved):
Surprised that the tests worked at all with the GSB fixtures

Lionqueen94:
> Surprised that the tests worked at all with the GSB fixtures

It works because the endpoints check that the user and election are both the same category (either GSB or CSB). So that's why it works using a GSB election and GSB Coordinator.

Issue #3469, opened 2026-06-25
Frontend for SQLite backups


praseodym:
Frontend page to create backup and report success/failure, see [design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=22522-18151&t=PN8F1xmskKFRhiUu-11)

Initial implementation checklist: 

- [x] Add item to main menu: 'Backups'
- [x] Add page on slug `/backups`
- [x] Ensure access only for 'Coordinators' and 'Beheerders'
- [x] The page displays a title, a text and a button 'Nu backup maken'
- [x] On button click:
  - [x] start the API call (`POST` @ `/api/backup`)
  - [x] display spinner + 'backup aan het maken' next to the button
  - [x] disable the button
- [x] On API call response success:
  - [x] hide spinner, hide 'backup aan het maken'
  - [x] display 'Laatste backup gemaakt om <HH:mm>' (from returned `created_at`) next to the button
  - [x] enable the button
- [x] On API call response error:
  - [x] hide spinner, hide 'backup aan het maken'
  - [x] display error message
  - [x] enable the button
- [x] Implement tests/storybook stories

After consultation with @jorisleker:  
- [x] Add a new icon for use in hamburger menu
- [x] Ensure the loading state with spinner remains visible for at least 0.6 seconds

After consultation with @stacktraceghost:
- [x] Change minimal waiting time to 1sec to prevent 'filename already exist' error
- [x] Smaller error Alert
- [x] Improve layout

PR #3495, opened 2026-07-01
Add SQLite backup REST API

### LINKED ISSUE #3468: API endpoint for SQLite backups
Check design to see what details need to be returned.

praseodym:
## What & why

Add SQLite backup REST API, resolves #3468.

## How to test

Run backend and frontend:
```
cargo run -- --reset-database --seed-data
# other terminal
pnpm run dev
```

Log in as admin or coordinator in the browser, then in devtools console:
```js
fetch('/api/backup', { method: 'POST' }).then(r => r.json()).then(console.log)
```

This should return the database backup filename and the backup should be created in the `backend/backups` directory. When not logged in or logged in as typst an error is returned.

## Reviewer notes

Frontend will be done in #3469.

### DIFF
--- a/backend/src/error.rs
+++ b/backend/src/error.rs
@@ -366,7 +366,7 @@ impl From<BackupError> for APIError {
     fn from(err: BackupError) -> Self {
         match err {
             BackupError::AlreadyExists => APIError::Conflict(
-                "A backup for this timestamp already exists".to_string(),
+                "A backup with this filename already exists".to_string(),
                 ErrorReference::BackupAlreadyExists,
             ),
             BackupError::InvalidPath => {
--- a/backend/src/error.rs
+++ b/backend/src/error.rs
@@ -22,6 +22,7 @@ use crate::{
         role::RoleNotAuthorizedError, validate::DataError,
     },
     eml::EMLImportError,
+    infra::backup::BackupError,
     repository::polling_station_repo,
     service::{DataEntryServiceError, PollingStationServiceError, SubCommitteeServiceError},
 };
@@ -41,6 +42,7 @@ pub trait ApiErrorResponse: Debug + Any {
 pub enum ErrorReference {
     AirgapViolation,
     AlreadyInitialised,
+    BackupAlreadyExists,
     ApportionmentNotCompleted,
     ApportionmentCommitteeSessionNotCompleted,
     ApportionmentInvalidLotDrawing,
@@ -356,6 +358,22 @@ impl From<sqlx::Error> for APIError {
                 _ => APIError::SqlxError(err),
             },
             _ => APIError::SqlxError(err),
+        }
+    }
+}
+
+impl From<BackupError> for APIError {
+    fn from(err: BackupError) -> Self {
+        match err {
+            BackupError::AlreadyExists => APIError::Conflict(
+                "A backup for this timestamp already exists".to_string(),
+                ErrorReference::BackupAlreadyExists,
+            ),
+            BackupError::InvalidPath => {
+                APIError::StdError(Box::new(std::io::Error::other("invalid backup path")))
+            }
+            BackupError::Io(err) => APIError::StdError(Box::new(err)),
+            BackupError::Database(err) => APIError::SqlxError(err),
         }
     }
 }
--- a/backend/src/error.rs
+++ b/backend/src/error.rs
@@ -372,6 +372,9 @@ impl From<BackupError> for APIError {
             BackupError::InvalidPath => {
                 APIError::StdError(Box::new(std::io::Error::other("invalid backup path")))
             }
+            BackupError::IntegrityCheckFailed(output) => APIError::StdError(Box::new(
+                std::io::Error::other(format!("backup integrity check failed: {output}")),
+            )),
             BackupError::Io(err) => APIError::StdError(Box::new(err)),
             BackupError::Database(err) => APIError::SqlxError(err),
         }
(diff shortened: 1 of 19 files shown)


backend/src/error.rs
```diff
@@ -360,6 +362,22 @@ impl From<sqlx::Error> for APIError {
     }
 }
 
+impl From<BackupError> for APIError {
+    fn from(err: BackupError) -> Self {
+        match err {
+            BackupError::AlreadyExists => APIError::Conflict(
+                "A backup for this timestamp already exists".to_string(),
```
stacktraceghost:
```suggestion
                "A backup with this filename already exists".to_string(),
```

stacktraceghost (changes requested):
- Tested on Windows and Linux, both work. 
- Tested error by quickly creating multiple backups at the same second

Noticed that the audit log is missing a translation:

Image


In `packaging/windows/abacus.iss` we create a shortcut `Abacus database map`. We could change this to point to `Abacus database backup map` and point to the backup folder. Although the folder does not exist if no backup has been made. It would make it less likely that a user would copy the live database. This could be a separate issue/PR. 

Also the documentation needs to be adjusted, lets make an issue for that.

sanderboom:
Tested on Linux and MacOS, both work.

Some thoughts/nitpicks (not sure if acting on them is actually worthwhile - depends on context):

- Confirming the backup succeeded by running a `PRAGMA quick_check` or `PRAGMA integrity_check` after data is written would add robustness. 
- Possibly empty the `sessions` table (no upside to including its data and `session_key` is equal to the value in the cookie). `DELETE FROM sessions` could e.g. be run after the check above.
- Theoretically there could be filename collisions because of DST (offset or unix epoch or incremental could be used to mitigate).

> Tested error by quickly creating multiple backups at the same second

Besides 409's this gave me 500's:

```
 Promise.all([
    fetch('/api/backup',{method:'POST'}).then(r=>r.status),
    fetch('/api/backup',{method:'POST'}).then(r=>r.status),
     fetch('/api/backup',{method:'POST'}).then(r=>r.status),
     fetch('/api/backup',{method:'POST'}).then(r=>r.status),
     fetch('/api/backup',{method:'POST'}).then(r=>r.status),
     fetch('/api/backup',{method:'POST'}).then(r=>r.status),
     fetch('/api/backup',{method:'POST'}).then(r=>r.status),
     fetch('/api/backup',{method:'POST'}).then(r=>r.status),
  ]).then(console.log) 
```

> `[201, 500, 409, 409, 500, 409, 409, 409]`

backend/src/infra/backup.rs
```diff
@@ -109,18 +103,12 @@ mod tests {
...
-
     #[sqlx::test]
-    async fn local_backup_is_succesfull(pool: SqlitePool) {
+    async fn local_backup_is_successful(pool: SqlitePool) {
         let backup_config = setup_backup_config();
         create_backup_directory(&backup_config).unwrap();
-        create_local_backup(&pool, &backup_config).await.unwrap();
-        assert!(backup_config.directory.read_dir().unwrap().next().is_some());
+        let result = create_local_backup(&pool, &backup_config).await.unwrap();
+        assert!(backup_config.directory.join(&result.filename).exists());
```
sanderboom, line 149:
```suggestion
        let path = backup_config.directory.join(&result.filename);
        assert!(path.exists());

        // Additional sanity check on database integrity.
        let opts = SqliteConnectOptions::new().filename(path).read_only(true);
        let mut connection = SqliteConnection::connect_with(&opts).await.unwrap();
        let row = sqlx::query("PRAGMA integrity_check")
            .fetch_one(&mut connection)
            .await
            .unwrap();
        let check_result: String = row.get(0);
        assert_eq!(check_result, "ok");

        connection.close().await.unwrap();
```

sanderboom: Note: also needs the `Row` import from `sqlx`.

praseodym:
> Noticed that the audit log is missing a translation: [..]

Thanks, added in 872e7565581d4a7258d0a607db1edf12c4376669.

> In `packaging/windows/abacus.iss` we create a shortcut `Abacus database map`. We could change this to point to `Abacus database backup map` and point to the backup folder. Although the folder does not exist if no backup has been made. It would make it less likely that a user would copy the live database. This could be a separate issue/PR.

Created #3500.

> Also the documentation needs to be adjusted, lets make an issue for that.

Created kiesraad/abacus-documentatie#78.

praseodym:
>     * Confirming the backup succeeded by running a `PRAGMA quick_check` or `PRAGMA integrity_check` after data is written would add robustness.

Good idea, implemented in 44782016b0e92304b34c7e12d93dd58e2cd0ffb9.

>     * Possibly empty the `sessions` table (no upside to including its data and `session_key` is equal to the value in the cookie). `DELETE FROM sessions` could e.g. be run after the check above.

I don't think this is needed: the sessions table is small, and if an old backup gets restored the old sessions get removed from the database. The sessions table is sensitive but the same goes for other tables in the database (they should never be modified outside of Abacus).

>     * Theoretically there could be filename collisions because of DST (offset or unix epoch or incremental could be used to mitigate).

I think that risk is negligible and acceptable. UTC timezones would prevent this but are more confusing for users.

> > Tested error by quickly creating multiple backups at the same second
> 
> Besides 409's this gave me 500's [..]

Resolved by adding a mutex, cd3221d1bf062bd7dc42590294c03663c7d69c8c.

backend/src/infra/backup.rs
praseodym: Added in 44782016b0e92304b34c7e12d93dd58e2cd0ffb9

stacktraceghost (approved):

sanderboom (approved):

PR #3496, opened 2026-07-01
Ensure apportionment only accepts CSB election

praseodym:
## What & why

Ensure apportionment only accepts a CSB election and apportionment state is only changed for CSB elections. Prevents and resolves more instances of using GSB fixtures for apportionment, just as #3489.

## How to test

Covered by unit tests.

## Reviewer notes

None.

### DIFF
--- a/backend/src/service/apportionment.rs
+++ b/backend/src/service/apportionment.rs
@@ -17,7 +17,9 @@ use crate::{
         apportionment_state::{ApportionmentState, ApportionmentStateError, DrawingLotsRequired},
         committee_session::CommitteeSessionId,
         committee_session_status::CommitteeSessionStatus,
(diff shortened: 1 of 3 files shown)


backend/src/service/apportionment.rs
praseodym, line 1: This file has a few more tests using GSB fixtures where CSB fixtures should be used. I'll add more assertions to prevent this and update the tests. (thanks @stacktraceghost!)

praseodym: Done in 13be30d, caught a few more places doing apportionment things with a GSB fixture

stacktraceghost (approved):

Lionqueen94 (approved):

Issue #3414, opened 2026-06-11
Add mDNS announcer


praseodym:
Add mDNS announcer so that Abacus is available on `abacus.local` and the TLS certificate can use that hostname without editing the hosts file.

Removed from the TLS epic to reduce scope.

PR #3505, opened 2026-07-02
Use `abacus.internal` as well-known TLS hostname

### LINKED ISSUE #3456: Set well-known hostname
OSV2020 uses `<something>.osv2020.nl` as its hostname for its TLS certificate. We should configure something similar for Abacus if we want to let users use a hostname. Alternatively, we can rely on IP addresses that are already in the subject alternative names of the certificate.

Using `abacus.local` with mDNS is another issue: #3414. If we don't implement mDNS we should consider removing `abacus.local` from the subject alternative names.

praseodym:
## What & why

Use `abacus.internal` as well-known TLS hostname, instead of `abacus.local`. The [`.local` TLD](https://en.wikipedia.org/wiki/.local) is used for mDNS which we do not yet implement (#3414), whereas the [`.internal` TLD](https://en.wikipedia.org/wiki/.internal) is reserved for private-use applications like Abacus.

Resolves #3456

## How to test

Build frontend, run `cargo run --features memory-serve,tls`, then trust `tls/ca.pem` / `tls/ca.cer` in your browser and go to `https://localhost:8443/`.

Then check that your browser shows `abacus.internal` in the certificate subject alternative names. Or add `127.0.0.1 abacus.internal` to your hosts file and check that `https://abacus.internal:8443` works in your browser without a certificate error.

## Reviewer notes

None.

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -111,7 +111,7 @@ cargo build --release --features tls
 On startup Abacus loads (or, on first run, generates) a local certificate
 authority under the directory given by `--tls-dir` (defaults to `tls`). A fresh
 server (leaf) certificate is created in memory on every start, signed by the CA
-and covering `localhost`, `abacus.local`, and all routable LAN addresses.
+and covering `localhost`, `abacus.internal`, and all routable LAN addresses.
 
 To trust the server, import the CA into the client trust store: `ca.pem` on
 Linux/macOS/Firefox, `ca.cer` (DER) on Windows. The CA can be downloaded from the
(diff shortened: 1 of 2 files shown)


stacktraceghost (approved):
Tested on Windows and Linux, in both cases the certificate is valid for `abacus.internal` and not for example `somethingelse.internal`.

rnijveld (approved):
Note that it's definitely still allowed to do non-mdns stuff on `.local`, but it is specifically reserved for link-local stuff, which feels like the wrong level for Abacus, so `.internal` makes much more sense that way as well. I don't think we really should be doing mdns for setting the hostname, we should leave that to the OS (but that is a separate discussion). We could be doing mdns-sd when/if we ever want to create a client application, but I don't think that requires adding an additional hostname and we can just keep using an IP based certificate for that.

PR #3511, opened 2026-07-03
Apportionment list_qualifies_for_extra_seat tests/refactor

stacktraceghost:
## What & why

Related to #788 

The goal was to implement tests for function `list_qualifies_for_extra_seat`. To do this, I've moved some logic into the function itself in https://github.com/kiesraad/abacus/commit/17619f70d71b0c414e382579fafeeb7e687dbca4.

Test cases have been implemented in https://github.com/kiesraad/abacus/commit/2e3a760d22448602c699d39b917db6b1fc9a839e

After which a small refactor has been applied https://github.com/kiesraad/abacus/commit/2296bb191d17b75656134f0fbdddb6d93c691e6b (tests succeeding before and after). 

## How to test

Run tests / see CI

## Reviewer notes

Review per commit.

- Are the test cases complete, is every path tested?
- Does the function do what you expect, looking at the test cases?

### DIFF
--- a/backend/apportionment/src/seat_assignment/residual_seat_assignment.rs
+++ b/backend/apportionment/src/seat_assignment/residual_seat_assignment.rs
@@ -817,151 +817,126 @@ mod tests {
         }
     }
 
-    #[test]
-    #[expect(clippy::too_many_lines)]
-    fn test_list_qualifies_for_extra_seat() {
+    mod list_qualifies_for_extra_seat {
+        use super::*;
+
         const LIST: u32 = 1;
         const OTHER_LIST: u32 = 2;
 
-        struct Case {
-            description: &'static str,
-            check_for_unique_highest_average_seats: bool,
-            previous_steps: Vec<SeatChangeStep<u32>>,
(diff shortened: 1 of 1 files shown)


Lionqueen94 (dismissed):
Looks good!

rnijveld (changes requested):
I think it makes more sense to split out all the cases as separate test functions, given that it allows Rust to execute them in parallel, allows you to see for multiple test cases at the same time that they fail, removes the clippy allow lint and doesn't really cost any extra lines of code.

stacktraceghost:
> I think it makes more sense to split out all the cases as separate test functions, given that it allows Rust to execute them in parallel, allows you to see for multiple test cases at the same time that they fail, removes the clippy allow lint and doesn't really cost any extra lines of code.

Adjusted in https://github.com/kiesraad/abacus/pull/3511/commits/486b086b844b8054986f4cb3c1b89f42735c49a2

Lionqueen94 (approved):

rnijveld (approved):

Issue #3510, opened 2026-07-03
Update Windows installer for TLS


praseodym:
Update Windows installer for TLS:
- Add firewall rule for port 443 
- Add cert-only option to Abacus binary
- Create CA certificate during installer
- Add CA certificate to system trust store using `certutil -f -addstore root ca.cer`
- Undo during uninstall

PR #3512, opened 2026-07-03
Create backups shortcut in Windows installer

### LINKED ISSUE #3500: Windows installer improvements for SQLite backups
In `packaging/windows/abacus.iss` we create a shortcut `Abacus database map`. We could change this to point to `Abacus database backup map` and point to the backup folder. Although the folder does not exist if no backup has been made. It would make it less likely that a user would copy the live database. This could be a separate issue/PR. 

_Originally posted by @stacktraceghost in https://github.com/kiesraad/abacus/pull/3495#pullrequestreview-4608574862_

praseodym:
## What & why

Create backups folder shortcut on the desktop in the Windows installer, to make it easier to find the backups. The `[Dirs]` section in the installer lets the installer create the backups directory so that the shortcut works without running Abacus first.

The backups shortcut replaces the database folder shortcut to reduce confusion between the two.

Resolves #3500.

## How to test

[Create and run the Windows installer](https://github.com/kiesraad/abacus/blob/main/packaging/windows/README.md), or just review the diff and check out this amazing screenshot:

Screenshot 2026-07-03 161322

## Reviewer notes

The Windows installer also needs to be updated for TLS: #3510.

### DIFF
--- a/packaging/windows/abacus.iss
+++ b/packaging/windows/abacus.iss
@@ -17,6 +17,7 @@
 #define MyAppExeName "abacus.exe"
 #define MyAppIcon "abacus.ico"
 #define MyDatabaseFile = "db.sqlite"
+#define MyBackupDirName = "backups"
 
 [Setup]
 ; NOTE: The value of AppId uniquely identifies this application. Do not use the same AppId value in installers for other applications.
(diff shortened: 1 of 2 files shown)


Lionqueen94 (dismissed):
Looks good based on the diff and screenshot 👍

packaging/windows/abacus.iss
```diff
@@ -156,7 +160,7 @@ begin
       if FileExists(ExpandConstant('{app}\{#MyDatabaseFile}')) then
```
stacktraceghost: In case the database file doesn't exist, the question to remove it isn't asked. If there is a weird case where someone has removed the database, but still has backups, the backup folder will stay behind. 

It would be better to check if there are any backups as well. I've tested the following code:
```pascal
function BackupFolderHasFiles(): Boolean;
var
  FindRec: TFindRec;
begin
  Result := False;

  if FindFirst(ExpandConstant('{app}\{#MyBackupDirName}\*'), FindRec) then
  begin
    try
      repeat
        if (FindRec.Name <> '.') and
           (FindRec.Name <> '..') then
        begin
          Result := True;
          Break;
        end;
      until not FindNext(FindRec);
    finally
      FindClose(FindRec);
    end;
  end;
end;
```

And then
```pascal
if FileExists(ExpandConstant('{app}\{#MyDatabaseFile}')) or BackupFolderHasFiles() then
```

Tested:
- Install, not starting Abacus, uninstall -> no question
- Install, start Abacus, uninstall -> question
  - Delete = entire folder gone
  - Keep = only database files left
- Install, not starting Abacus, add file in backup folder, uninstall -> question
  - Delete = entire folder gone
  - Keep = backup folder with files left 
- Install, start abacus, add file in backup folder, uninstall -> question
  - Delete = entire folder gone
  - Keep = database files and backup folder with files left

praseodym: Changed to delete entire data directory in 412f12a0e03cb2fed4d87b2c3698d424b4130a9a, instead of only the database

stacktraceghost (approved):

Lionqueen94 (approved):

PR #3516, opened 2026-07-06
Apportionment add extra information to CandidateDrawingLotsVariant

### LINKED ISSUE #3513: Add extra information to `CandidateDrawingLotsRequired`
- [ ] `CandidateDrawingLotsRequired` variant is missing the number of seats that need to be assigned by drawing lots. Preferably like `seat_numbers: [2, 3, 4]` similar to `residual_seat_numbers` in drawing lots for list
- [ ] Since we need to know the total number of seats assigned to the list as well, adding `total_seats` would also be helpful
- [ ] The number of `votes` the drawing of lots is needed for also needs to be added to the variant.
- [ ] If possible, adding the `list_name` to the variant would also be very helpful.

stacktraceghost:
## What & why

Resolves #3513

Adds three fields to `CandidateDrawingLotsVariant`:
- `total_seats`: The total number of seats for the list
- `number_of_votes`: The number of votes, equal for these candidates
- `seat_numbers`: The seat numbers that the candidates can be drawn for

## How to test

- Check if tests succeed
- Validate if the `seat_numbers` are calculated correctly

## Reviewer notes

`seat_numbers` is calculated by taking:
- The `current_seat_number`, which is the current loop `index + 1`
- Until `current_seat_number + MIN(amount_of_options, number_of_seats_remaining)`

If we have more or equal number of seats than the options, it will only take as many seats as there are options.
If we have less number of seats than the options, take the amount of seats left.

### DIFF
--- a/backend/apportionment/src/candidate_nomination/mod.rs
+++ b/backend/apportionment/src/candidate_nomination/mod.rs
@@ -302,7 +302,17 @@ fn preferential_candidate_nomination<'a, LV: ListVotes>(
                 "Drawing of lots is required for candidates: {options:?}, {seats_remaining} seat(s) available",
             );
 
-            let variant = CandidateDrawingLotsVariant { list, options };
+            let current_seat_number = index + 1;
+            let nr_of_seats = options.len().min(seats_remaining as usize);
+            let seat_numbers = (current_seat_number..current_seat_number + nr_of_seats)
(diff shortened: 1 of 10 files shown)


Lionqueen94 (approved):
Looks good!

praseodym (approved):

PR #3518, opened 2026-07-06
Create and trust CA in Windows installer

### LINKED ISSUE #3510: Update Windows installer for TLS

praseodym:
## What & why

Update the Windows installer to create the TLS CA certificate and add it to the system root trust store. To let the installer create the certificate without starting the full Abacus server, the `--init-tls` CLI option was added. The firewall rule now also opens TCP port 443 (HTTPS) in addition to TCP port 80 (HTTP). Resolves #3510.

## How to test

- The new `--init-tls` option should create the TLS CA certificate in the backend directory: `cargo run --features tls -- --init-tls` (and not recreate it when the CA certificate already exists)
- Run the Windows installer to verify that the TLS CA certificate gets created and added to the trust store.

## Reviewer notes

I can provide you with a built Windows installer if needed.

### DIFF
--- a/backend/src/bin/abacus.rs
+++ b/backend/src/bin/abacus.rs
@@ -99,6 +99,11 @@ struct Args {
     #[arg(long, default_value_t = default_http_port(), env = "ABACUS_HTTP_PORT")]
     http_port: u16,
 
+    /// Create the TLS directory and CA certificate (if missing) and exit without starting the server
+    #[cfg(feature = "tls")]
+    #[arg(long)]
+    init_tls: bool,
+
     /// Seed the database with initial data using the fixtures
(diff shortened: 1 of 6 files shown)


stacktraceghost (approved):

sanderboom (approved):
Checked before/after install/ after uninstall:
- `certutil -store root "Abacus Local CA"`
- `netsh advfirewall firewall show rule name="Abacus server"`

Looks good!

PR #3528, opened 2026-07-07
Apportionment fix standings, added assertions

stacktraceghost:
## What & why

Related to #788 

Noticed that the absolute majority step did not contain the standings **before** the change. This has been fixed and assertions have been added in the code where steps get added. At the end of the `seat_assignment` function we also assert that the `final_standings` are different than the last step.

## How to test

Tests should succeed. See if all places are covered.

## Reviewer notes

N/A

### DIFF
--- a/backend/apportionment/src/seat_assignment/mod.rs
+++ b/backend/apportionment/src/seat_assignment/mod.rs
@@ -133,11 +133,18 @@ pub(crate) fn seat_assignment<T: ApportionmentInput>(input: &T) -> SeatAssignmen
     };
     if let Some(assigned_seat) = assigned_seat {
         // add the absolute majority change to the remainder assignment steps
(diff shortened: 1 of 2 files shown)


praseodym (approved):
Good enough for now but as a more permanent solution I'd prefer the type-based solution you proposed, that can throw proper errors instead of panicking.

Lionqueen94 (approved):

stacktraceghost:
> Good enough for now but as a more permanent solution I'd prefer the type-based solution you proposed, that can throw proper errors instead of panicking.

As clarification: I was considering creating a newtype out of the `Vec<SeatChangeStep>` and overriding the `push` method to do this assertion locally. We can of course also throw an error. Most likely this issue will be caused by a programming mistake.

PR #3545, opened 2026-07-07
Add checks to ensure input is sorted

### LINKED ISSUE #3478: Apportionment crate sort lists/candidates by number
The apportionment crate expects that the given lists and candidates are sorted by number, but doesn't sort those itself. The crate should sort these inputs, since it is a separate crate. 

Related comment (candidates): https://github.com/kiesraad/abacus/pull/3471#discussion_r3480217843

cikzh:
## What & why

Closes #3478
The apportionment crate expects sorted input. Abacus currently does this correctly, but because it is a separate crate and can be used as a library, we should ensure that it is sorted.

## How to test
Check out the unit tests added in lib.rs. You can remove the `reverse()` calls and see that the unit tests fail.

## Reviewer notes

I've opted to simply add `is_sorted_*` checks and return an `ApportionmentError` when the checks fail.

Currently, the `process` function passes in an immutable reference to the input. Sorting the input in the input crate itself would require to either pass in a mutable reference, or cloning the data, creating multiple sources of the input.
Both solutions decrease type safety and maintainability, while we gain nothing, since Abacus already passes in sorted input correctly.

### DIFF
--- a/backend/apportionment/src/lib.rs
+++ b/backend/apportionment/src/lib.rs
@@ -50,6 +50,20 @@ pub enum ApportionmentOutput<'a, T: ApportionmentInput> {
 pub fn process<T: ApportionmentInput>(
     input: &T,
 ) -> Result<ApportionmentOutput<'_, T>, ApportionmentError> {
+    // Check if all list votes are sorted by number
+    if !input.list_votes().is_sorted_by_key(|lv| lv.number()) {
+        return Err(ApportionmentError::UnsortedInput);
+    }
+
+    // Check if all candidate votes are sorted by number
+    if input
+        .list_votes()
+        .iter()
+        .any(|lv| !lv.candidate_votes().is_sorted_by_key(|cv| cv.number()))
+    {
+        return Err(ApportionmentError::UnsortedInput);
+    }
+
     let seat_assignment = match seat_assignment(input)? {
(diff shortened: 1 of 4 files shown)


backend/apportionment/src/lib.rs
```diff
@@ -409,4 +426,30 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_apportionment_process_unsorted_list_candidate_number_fails() {
+        let mut input = seat_assignment_fixture_with_default_50_candidates(
+            15,
+            vec![540, 160, 160, 80, 80, 80, 60, 40],
+        );
+
+        input.list_votes.iter_mut().for_each(|lv| {
```
stacktraceghost: By only reversing one list, the test would cover a mistake like using `.all` instead of `.any`

stacktraceghost (dismissed):

praseodym (dismissed):

backend/apportionment/src/lib.rs
stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/3545/commits/a836407b04ea27cb78086a4ac5261041d85070f3

stacktraceghost (approved):

Lionqueen94 (approved):

PR #3552, opened 2026-07-09
Apportionment individual tests

stacktraceghost:
## What & why

Related to #788 

Added many tests for the smaller functions within the apportionment crate, added per commit.

Overview per function with notes in case anything else was changed:
- Seat assignment `mod.rs`
  - `get_total_seats_per_list_number_from_apportionment_result` https://github.com/kiesraad/abacus/commit/39f5f0c31798bdebc6526f7220ca9ccffec27aa7
    - Limited the scope of the input (not passing entire result)
    - Renamed to  `get_total_seats_per_list_number_from_seat_assignments`
    - Removed test function which was used once
  - `get_number_of_candidates` https://github.com/kiesraad/abacus/commit/810958a13d78ac8d11e58a2c90d74e59640a367b
  - `list_numbers_with_exhausted_seats` https://github.com/kiesraad/abacus/commit/252cef68d05f90c9ee16494d3e17b2dbccb0b2b1
- Seat assignment `‎residual_seat_assignment.rs‎` 
  - `list_assigned_from_previous_step` https://github.com/kiesraad/abacus/commit/bbcc50decc55e0a6d457aef3ab6dcec2a403de6a
    - Pass only list number instead of list
    - Renamed to `extend_list_assigned_from_previous_step` and adjusted docstring
  - `list_largest_remainder_assigned_seats` / `list_unique_highest_average_assigned_seats` https://github.com/kiesraad/abacus/commit/c9cba9e441df5e1df0341a939726b4f07deb3578
    - Merged into one reusable function to select a certain kind of steps
    - Moved the count outside of the function
  - `list_numbers_without_empty_seats` / `exhausted_list_numbers`
    - Merged `list_numbers_without_empty_seats` into `exhausted_list_numbers`, since it is the only place where it is used https://github.com/kiesraad/abacus/commit/9790bc8232bdb5988c344fd988965abe94ff0fa3
    - & Split the parameters
    - & Moved the responsibility to call it or not back to `assign_remainder`, since it is using `exclude_exhausted_lists` as a flag to decide if it should map (when `Some`) or simply return a `vec[]` (when `None`)
    - Simplified the fold to a filter (after the validated test in the previous commit) https://github.com/kiesraad/abacus/commit/8302670dee2612049777c496fcc5155b208ff525
    - & ⚠️ Adjusted the filter to also return a list as exhausted if there are more seats than candidates (instead of only if they are the same). This matches the function name. In the apportionment logic, at the point of calling this function, lists couldn't have more seats than candidates anymore, so this does not affect the apportionment logic.
  - `list_standings_qualifying_for_largest_remainder` + `list_standings_qualifying_for_unique_highest_average` https://github.com/kiesraad/abacus/commit/0c13bf05d82a3dc6d008239dd3d2dc131c812481
- Candidate nomination `mod.rs`
  - `all_chosen_candidates` https://github.com/kiesraad/abacus/commit/bb716fa36d8c1c5361d7725864e490e167b35f16
  - `candidate_votes_meeting_preference_threshold` https://github.com/kiesraad/abacus/commit/801c800a8217eb2dbf69837b305c9320213f96b4
  - `candidate_votes_numbers` / `other_candidate_nomination` https://github.com/kiesraad/abacus/commit/e49e8036902b29ba91f5541ebe36d82abb75cecb

## How to test

Check if all tests are passing

## Reviewer notes

- Check per commit
  - Do you understand what the function does?
  - Does it match what you would expect?
  - Do the tests cover all scenarios of that function?
- Please pay extra attention to the change with the exclamation mark above


backend/apportionment/src/seat_assignment/mod.rs
```diff
@@ -554,6 +554,100 @@ pub(crate) mod tests {
...
+            let standings: Vec<_> = list_votes.iter().map(standing).collect();
+
+            assert_eq!(
+                list_numbers_with_exhausted_seats(&standings, &list_votes, &HashMap::new()),
+                vec![(1, 2), (3, 2)]
+            );
+        }
+
+        #[test]
+        fn test_residual_seats_count_towards_exhaustion() {
```
jschuurk-kr, line 623: Does the P 9-seat count as a residual seat? Reading through `reassign_residual_seat_for_absolute_majority()` it seems like it doesn't.

stacktraceghost: `reassign_residual_seat_for_absolute_majority` calls `remove_residual_seat` on one standing and `add_residual_seat` on another. This manipulates the number of residual seats and updates `next_votes_per_seat` (which is used for the highest averages method).

The method `list_numbers_with_exhausted_seats` (which this test is part of) calls `total_seats()` on the standing, which returns the sum of `standing.full_seats` and `standing.residual_seats`. So for this method, it does not matter how a seat came in place.

backend/apportionment/src/candidate_nomination/mod.rs
```diff
@@ -576,6 +576,82 @@ mod tests {
...
+            CandidateVotesMock(1, 80),
+            CandidateVotesMock(7, 60),
+        ]
+        .iter()
+        .collect();
+
+        assert_eq!(candidate_votes_numbers(&candidate_votes), vec![3, 1, 7]);
+    }
+
+    mod other_candidate_nomination {
```
jschuurk-kr, line 592: I'm missing a test where all candidates have been nominated through preferential votes, so the function returns an empty vector.

stacktraceghost: Added in https://github.com/kiesraad/abacus/pull/3552/commits/a1e237f7d6328967353b2ebd9dc719d0e1331d0a

jschuurk-kr (approved):

Lionqueen94 (approved):

PR #3557, opened 2026-07-10
Remove the pdf-gen dylib

### LINKED ISSUE #3440: pdf_gen_dylib breakage
The pdf_gen dylib refactor from #3289 has caused some breakage:

- `cargo test --workspace` is broken with:
```
error: cannot satisfy dependencies so `std` only shows up once
  |
  = help: having upstream crates all available in one format will likely make this go away
  = note: `pdf_gen_dylib` was unavailable as a static crate, preventing fully static linking
```
- Native Windows compiles with dylib are broken (WSL works fine):
  - msvc linker `= note: LINK : fatal error LNK1189: library limit of 65535 objects exceeded`
  - rust-lld: `= note: rust-lld: error: too many exported symbols (got 261765, max 65535)`

We should find solutions for these issues or reconsider the dylib implementation. A simple solution would be to always embed Typst, but one complication is that our backend fuzzer compiles with ASAN and the Typst crates somehow cannot be compiled with ASAN. We currently have the pdf_gen stub feature for this.

rnijveld:
## What & why

Fixes #3440 

The pdf-gen dylib doesn't really work reliably on Windows. I've attempted a couple of workarounds for that, but none of them work decently well. I've decided that it isn't worth our time to put too much effort into this, so we probably just want to bite the bullet and deal with typst always being statically compiled. I'll take a look at seeing if there are other quick wins to improve incremental compilation, but this will not be a target for that for now.

## How to test

Only removed stuff, so nothing much really changed. The fuzzer should still work even with the removed features, the stub implementation is now the default when the static feature is not provided.


praseodym (approved):
What about the CLI flag that you talked about earlier this week?

stacktraceghost (approved):

rnijveld:
> What about the CLI flag that you talked about earlier this week?

Tried it, but it was a very suboptimal experience, and hard to do with rust-analyzer as well, meaning its unlikely to be maintained properly, and I'd prefer not to run the risk of it not working properly.

PR #3558, opened 2026-07-10
Pin Rust versions

praseodym:
## What & why

Pin Rust versions used in Abacus to prevent breakage like #3554, especially during the release cycle.

## How to test

CI should pass and local builds should still work.

## Reviewer notes

The code coverage workflow was pinned to Rust Nightly 2026-02-21 in #2931 and never updated, this PR does update it.


Lionqueen94 (approved):
Nice improvement

stacktraceghost (approved):

PR #3564, opened 2026-07-10
Footnote for drawing lots for P 9 and bugfix for result changes reflected in tables

### LINKED ISSUE #3507: Footnote for drawing lots for P9 on ApportionmentResidualSeatsPage
Design: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=20697-57011&m=dev
- [x] Add footnote variant of P9 footnote for drawing lots
  - [x] story
- [x] Make sure the P9 before drawing lots footnote is shown on ApportionmentResidualSeatsPage
  - [x] test
- [x] Add footnote superscript to total seats of applicable lists
  - [x] test
  - [x] story HighestAveragesTable
  - [x] story LargestRemaindersTable (also for P9/P10 without drawing lots)
  - [x] story UniqueHighestAveragesTable
- [x] Add result changes to UniqueHighestAveragesTable as well
  - [x] test
  - [x] story (P10)
- [x] Add what list options were when P9 happened by drawing lot (add `(er is geloot tussen lijst {nr} en {nr})`

Lionqueen94:
Closes #3507 

## What & why
- Added drawing lots options to P 9 footnote if assigned by drawing lots
- Moved all footnote translation keys under "footnotes" key
- Added footnote for absolute majority reassignment before drawing lots
- Made sure `UniqueHighestAveragesTable` can also reflect result changes
- Added more stories for `LargestRemainderTable` and `UniqueHighestAveragesTable`

### Copy updates
- [x] Change "na het toewijzen" into "bij het toewijzen" in `Er zijn meerdere partijen die na het toewijzen van de volgende restzetel precies hetzelfde hoogste gemiddelde hebben.` and `De partij met het hoogste gemiddeld aantal stemmen bij het toewijzen van de restzetel krijgt de restzetel` -> also change in design @jorisleker 
- [x] Copy update: https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=17533-41735&m=dev#1838713324

## How to test

- Check the new stories
- Review the added tests (to replicate the issue I found with the result changes, use Election < 19 seats from the P 22-2 variants from the gen-test-election, then make 40 candidates from list 1 deceased, causing 2 seats to be removed due to list exhaustion. Originally this caused the `LargestRemainderTable` to display -1 residual seats for list 1, this is now fixed -> see the added test variant and added test and stories).

## Reviewer notes

- Review per commit


frontend/src/features/apportionment/components/residual_seats/ApportionmentResidualSeatsPage.tsx
```diff
@@ -194,14 +194,15 @@ function SmallCouncilSection({
   footNotes,
   state,
 }: SmallCouncilSectionProps) {
+  const showFootnotes = isP9DrawingLots(state) || resultChanges.length > 0;
```
stacktraceghost: Both `footNotes` and `resultChanges` are passed, after which `showFootnotes` decides that they should not be shown. 

This check is also done at line 279 (for the other component). 

Since the state is being passed, I would generate the footnotes as close as possible to the place where they are needed, so inside this function. That also saves another parameter.

frontend/src/features/apportionment/components/residual_seats/ApportionmentResidualSeatsPage.tsx
stacktraceghost: `largestRemainderSteps`, `uniqueHighestAverageSteps`, `highestAverageSteps` are derived from the `seatAssignment`. I think it is better to only pass the data that is necessary and get the deriven data where its needed. That way we don't have to pass around so much data. 

This could be applied to more components where this is done, like `Footnotes`

frontend/src/features/apportionment/utils/steps.tsx
```diff
@@ -37,15 +37,15 @@ export interface AssignmentSteps {
   largestRemainderSteps: LargestRemainderAssignmentStep[];
   uniqueHighestAverageSteps: UniqueHighestAverageAssignmentStep[];
   highestAverageSteps: HighestAverageAssignmentStep[];
-  absoluteMajorityStep?: AbsoluteMajorityReassignmentStep;
```
stacktraceghost, line 40: I feel like `absoluteMajorityStep` would be more clear, but this might be a preference thing.

frontend/src/features/apportionment/utils/seat-change.tsx
```diff
@@ -79,6 +83,32 @@ export function getResultChanges(
   return resultChanges;
 }
 
+export function splitResultChanges(
```
stacktraceghost, line 86: I am unsure what this function does, could you add some comments? It feels like it is passively trying to figure out what a change was, not sure if this is the best way to do it.

stacktraceghost (changes requested):

frontend/src/features/apportionment/utils/steps.tsx
Lionqueen94: I agree, but the place where it's used needed a shorter name cause I got the excessiveLines error. If you can shorten it in another way please do ;)

frontend/src/features/apportionment/utils/seat-change.tsx
Lionqueen94: A better way would be to figure out in what order the list exhaustion/p9 takes place and if any uniquehighestaverages steps have come before. This worked as well and I tested it well. The missing comments I'll add

frontend/src/features/apportionment/components/residual_seats/ApportionmentResidualSeatsPage.tsx
Lionqueen94: The renderFootnotes function needs to be located in the main component, so I don't think it's possible to render them from one of the functions outside of this

jorisleker (dismissed):
Reviewed the new stories in storybook and tested them on abacus-test for the AbsoluteMajority elections with >=19 and <19 seats. Everythink looks good to me!

frontend/src/features/apportionment/utils/seat-change.tsx
stacktraceghost: I've adjusted `splitResultChanges` and added some comments in https://github.com/kiesraad/abacus/pull/3564/commits/e71a0233d1d0337507cc1fed6f9bf746396fdd42

jorisleker (approved):
Reviewed again, still looks good

Lionqueen94 (commented):
I ✔️ @stacktraceghost's changes

stacktraceghost (approved):

PR #3572, opened 2026-07-13
Always remove empty app dir in Windows installer

### LINKED ISSUE #3567: Uninstall does not remove directory in certain scenario
Something unexpected when installing/reinstalling. Reproduce with:

- Install
- Uninstall - select 'keep database and backups'
- Reinstall
- Uninstall - select 'remove database and backups'

See that the directory `%APPDATA%` contains an empty directory `Abacus`. 

Reason:
> There are several reasons why a directory may not be removed:
> It already existed prior to installation. By default, the uninstaller plays it safe and doesn't remove directories that the installer didn't create.

See: [Inno FAQ](https://jrsoftware.org/isfaq.php#emptydirs)

The `DelTree` does not remove the directory because the uninstaller is still present in the directory. 

Possible solution: 
> `uninsalwaysuninstall`
> Instructs the uninstaller to always attempt to delete the directory if it's empty. Normally the uninstaller will only try to delete the directory if it didn't already exist prior to installation. 

From [Inno docs](https://jrsoftware.org/ishelp/index.php?topic=dirssection)

praseodym:
## What & why

Always remove empty app dir in Windows installer. Resolves #3567.

## How to test

Run installer/uninstaller:
- Install
- Uninstall - select 'keep database and backups'
- Reinstall
- Uninstall - select 'remove database and backups'
- Abacus directory should be removed

## Reviewer notes

None.

### DIFF
--- a/packaging/windows/abacus.iss
+++ b/packaging/windows/abacus.iss
@@ -73,6 +73,7 @@ Source: ".\VC_redist.x64.exe"; DestDir: "{tmp}"; Flags: deleteafterinstall
 Source: ".\{#MyAppIcon}"; DestDir: "{app}"
 
 [Dirs]
+Name: "{app}"; Flags: uninsalwaysuninstall
 Name: "{app}\{#MyBackupDirName}"
 
 [Icons]


sanderboom (approved):

stacktraceghost (approved):
Built the installer in Windows and followed the described steps. The installation folder was removed after the last step.

PR #3571, opened 2026-07-13
Add playwright tests for drawing lots flow and 2 small fixes

Lionqueen94:
Closes #1267 

## What & why

- Adds playwright test for drawing lots for list and candidate
- Adds playwright test for drawing lots for p9

### Fixes
- Incorrect copy for drawing lots for P9 (last list item)
- Font color of paragraphs in Alert (was dark grey, is now black)

## How to test

- Run the new tests (for example by running (`pnpm run test:e2e-ui tests/apportionment`, select only `initialisation-test' and 'setup-test-users', run all, when done, select only one or more browsers and then run all)
- Verify the fixes by checking the stories and/or checking on an AbsoluteMajority drawing lots election from gen-test-election

PR #3573, opened 2026-07-13
Frontend: Undo drawing lots for candidates when enough seats

### LINKED ISSUE #3565: Undo drawing lots for candidates when enough seats
## TODO
- [x] Undo drawing lots for candidates when enough seats (as implemented in https://github.com/kiesraad/abacus/pull/3487)
- [x] Remove reference to article P 15 (see https://github.com/kiesraad/abacus/pull/3543#issuecomment-4952015709 and [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17533-41745&m=dev))
- [x] Adjust candidate drawing lots variant to make sure there are 3 candidates for 2 seats for 1 list, so we can still test the multiple candidate drawing lots notify alert variant -> will be done by @Lionqueen94 
  - [x] gen-test-election variant
  - [x] frontend (`lt-19-seats-and-p15-drawing-lots.ts`)
  - [x] templates (`lt-19-seats-and-p15-drawing-lots.json` -> nice to update to keep the same as its ts equivalent)

Lionqueen94:
Closes #3565 

## What & why

- Undo frontend copy change
- Added one more singular/plural text for drawing lots for candidates
- Same ApportionmentListDetailsPage copy update as in https://github.com/kiesraad/abacus/pull/3564 (https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp?node-id=17533-41735&m=dev#1838713324)
- Updated gen test election for drawing lots and frontend variant and tests
- Updated P 22-2 template variant to match the frontend test variant

## How to test

- Reset your database
- Generate drawing lots variants from gen-test-election
- Use the candidate drawing lots variant from gen-test-election and check that the copy is updated for both (see first 2 commits)

## Reviewer notes

- Check per commit


jorisleker (dismissed):

jorisleker (dismissed):
Tested:
- no drawing of lots when enough seats ✅ 
- drawing of lots when not enough seats ✅ 
- copy changes ok ✅

frontend/src/features/apportionment/components/drawing_lots/DrawingLotsForCandidate.tsx
```diff
@@ -14,20 +14,23 @@ export function DrawingLotsForCandidate({ drawingLotsRequired, options, list }:
   const seatNumber = drawingLotsRequired.seat_numbers[0];
   if (seatNumber) {
     return (
-      <ul className={cn(cls.drawingLotsList, "w-32")}>
+      <ul className={cn(cls.drawingLotsList, "w-32")} aria-label="drawing-lots-information">
```
praseodym: Shouldn't the `aria-label`s be translated?

Lionqueen94: Good one, will add that. It was more to be able to distinguish the lists in the test, I guess I can add an id instead

Lionqueen94: Changed to id in 16519a3c0b37987112b596932a9cc76bf91366c5

jorisleker (approved):
better to use id then aria-label here indeed

stacktraceghost (approved):

praseodym:
@Lionqueen94 @jorisleker I just noticed that the alert has two different font colors:

image

jorisleker:
> @Lionqueen94 @jorisleker I just noticed that the alert has two different font colors:
> 
> image

Funny, thought I saw that too, but forgot too check. Tx for checking. Texts in alerts should all be black (0,0,0)

Lionqueen94:
> > @Lionqueen94 @jorisleker I just noticed that the alert has two different font colors:
> > image
> 
> Funny, thought I saw that too, but forgot too check. Tx for checking. Texts in alerts should all be black (0,0,0)

Fixed in https://github.com/kiesraad/abacus/pull/3571

PR #3587, opened 2026-07-15
Apportionment sort candidates alphabetically with collation

### LINKED ISSUE #3583: Onjuiste sortering namen gekozen kandidaten als deze beginnen met een diakriet
In OSV komt een kandidaat met een achternaam met Ç na de C en voor de D, in Abacus na de Z. 

We willen de sorteringsvolgorde zoals OSV die gebruikt aanhouden, bv met locale-aware unicode collation. 
Komt voor in de P 22 en in de Abacus Zetelverdeling UI

stacktraceghost:
## What & why

Resolves #3583 

Use collation to sort the names of the candidates. Added test cases to verify if the sorting is correct. 

## How to test

Run tests / check CI

## Reviewer notes

- Check the solution
- Verify if the sorting behaves as you expect
- Verify if the test cases cover enough

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -19,6 +19,7 @@ dependencies = [
  "http-body-util",
  "hyper",
  "icu_collator",
+ "icu_locale_core",
  "if-addrs",
  "memory-serve",
  "pdf_gen",
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -18,6 +18,7 @@ dependencies = [
  "eml-nl",
  "http-body-util",
  "hyper",
+ "icu_collator",
  "if-addrs",
  "memory-serve",
  "pdf_gen",
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -79,6 +79,7 @@ chrono.workspace = true
 clap = { version = "4.6", default-features = false, features = ["derive", "std", "help", "env"] }
 eml-nl = { version = "0.6.0" }
 hyper = { version = "1", features = ["full"] }
+icu_collator = "2.2.0"
 tokio.workspace = true
 memory-serve = { version = "2.1.0", optional = true, features = ["force-embed"] }
 utoipa = { version = "5.4.0", features = ["axum_extras", "chrono"] }
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -80,6 +80,7 @@ clap = { version = "4.6", default-features = false, features = ["derive", "std",
 eml-nl = { version = "0.6.0" }
 hyper = { version = "1", features = ["full"] }
 icu_collator = "2.2.0"
+icu_locale_core = "2.2.0"
 tokio.workspace = true
 memory-serve = { version = "2.1.0", optional = true, features = ["force-embed"] }
 utoipa = { version = "5.4.0", features = ["axum_extras", "chrono"] }
--- a/backend/README.md
+++ b/backend/README.md
@@ -175,6 +175,7 @@ The following dependencies (crates) are used:
 - `clap`: library for command-line argument parsing.
 - `cookie`: dependency of axum_extra, for encoding and parsing cookies.
 - `hyper`: fast and correct HTTP implementation.
+- `icu_collator`: locale-aware string comparison
 - `memory-serve`: serves frontend assets from memory, but ad-hoc from disk during development.
 - `password_hash`: password hashing interfaces.
 - `rand`: create a random session key.
--- a/backend/README.md
+++ b/backend/README.md
@@ -176,6 +176,7 @@ The following dependencies (crates) are used:
 - `cookie`: dependency of axum_extra, for encoding and parsing cookies.
 - `hyper`: fast and correct HTTP implementation.
 - `icu_collator`: locale-aware string comparison
+- `icu_locale_core`: locale definitions for `icu_collator`
 - `memory-serve`: serves frontend assets from memory, but ad-hoc from disk during development.
 - `password_hash`: password hashing interfaces.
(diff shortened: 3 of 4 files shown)


praseodym:
Which locale does this use? Do we need to force it to use a specific locale? Also see https://sts10.github.io/2023/01/29/sorting-words-alphabetically-rust.html#:~:text=Resigning%20myself%20to%20the%20fact%20that%20we%20need%20to%20add%20a%20locale%20option%20to%20Tidy

stacktraceghost:
> Which locale does this use? Do we need to force it to use a specific locale? Also see https://sts10.github.io/2023/01/29/sorting-words-alphabetically-rust.html#:~:text=Resigning%20myself%20to%20the%20fact%20that%20we%20need%20to%20add%20a%20locale%20option%20to%20Tidy

```rs
   let collator = Collator::try_new(CollatorPreferences::default(), CollatorOptions::default())
        .expect("collator data should be available");
 ```

`CollatorPreferences::default()` defaults to the locale `und`. It then uses the default [CLDR](https://docs.rs/icu/latest/icu/#compiled-data) set.

Locale `nl` has no locale-specific differences, see [nl.xml](https://github.com/unicode-org/cldr/blob/main/common/collation/nl.xml).
However, `sv` for example does, see [sv.xml](https://github.com/unicode-org/cldr/blob/main/common/collation/sv.xml).

So if you pass the locale `sv` like this:
```rs
    let collator = Collator::try_new(locale!("sv").into(), CollatorOptions::default())
        .expect("collator data should be available");
```

`test_diacritics` will fail, because in Swedish the letters Å and Ö get sorted after Z:
```
assertion `left == right` failed
  left: [("F.", "Aalst"), ("I.", "Boenen"), ("I.", "boer"), ("B.", "Ćelik"), ("E.", "Ekker"), ("C.", "Éland"), ("P.", "Groen"), ("Ö.", "Groen"), ("j.","groot"), ("L.", "Mulder"), ("N.", "Müller"), ("H.", "Ñuñez"), ("O.", "Šimić"), ("G.", "Ünal"), ("A.", "Zee"), ("R.", "Ådal"), ("D.", "Öztürk")]
 right: [("F.", "Aalst"), ("R.", "Ådal"), ("I.", "Boenen"), ("I.", "boer"), ("B.", "Ćelik"), ("E.", "Ekker"), ("C.", "Éland"), ("Ö.", "Groen"), ("P.", "Groen"), ("j.", "groot"), ("L.", "Mulder"), ("N.", "Müller"), ("H.", "Ñuñez"), ("D.", "Öztürk"), ("O.", "Šimić"), ("G.", "Ünal"), ("A.", "Zee")]
```

Using the locale `nl` gives the same result when running the tests. We can of course be specific about using the `nl` locale anyway, in case things change?

jschuurk-kr:
> Using the locale `nl` gives the same result when running the tests. We can of course be specific about using the `nl` locale anyway, in case things change?

Adding the locale `nl` would also answer the questions "why no explicit locale?" and/or "which locale?" for anyone reading the code in the future.

stacktraceghost:
> > Using the locale `nl` gives the same result when running the tests. We can of course be specific about using the `nl` locale anyway, in case things change?
> 
> Adding the locale `nl` would also answer the questions "why no explicit locale?" and/or "which locale?" for anyone reading the code in the future.

Adjusted in https://github.com/kiesraad/abacus/pull/3587/commits/3efa28a315ccffc2b4c76d42ca4e6b17ada5ce70

praseodym (approved):

jschuurk-kr:
Found during testing that our sorting does the following:
- De Vries, A.
- Vries, B.
- De Vries, C.

Will check what OSV does in this case.

jorisleker:
> Found during testing that our sorting does the following:
> 
> * De Vries, A.
> * Vries, B.
> * De Vries, C.
> 
> Will check what OSV does in this case.

Idealiter doen we dat op de manier zoals [hier](https://taaladvies.net/alfabetische-rangschikking-van-namen/) beschreven. 

Vries, A. de
Vries, B.
Vries, C. de

De voorvoegsel voor de achternaam houden is wat ze in België doen. Maar daar is het voorvoegsel ook onderdeel van de achternaam, en dus geen apart veld.

jschuurk-kr:
> Found during testing that our sorting does the following:
> 
>     * De Vries, A.
> 
>     * Vries, B.
> 
>     * De Vries, C.
> 
> 
> Will check what OSV does in this case.

OSV's P 22-2 has the same sorting as Abacus in this case. (Sidenote: you can sort the page "Gekozen kandidaten" in OSV on "Achternaam, Voornaam". This results in a different sorting than in OSV's P 22-2. So I ignored that page for the comparison.)

(edit): OSV also displays the names in the same way. So it does not put the "De" at the end.

There is a difference in handling diacritics, however. The clearest example from my test data is that OSV sorts "Ńaas Naas Ñaas", while Abacus does "Naas Ńaas Ñaas". So OSV puts some diacritics before the regular letter and some after. Not sure if that's caused by them using a German locale or if it has a completely different cause. In any case, I think this difference between OSV and Abacus is acceptable.

We also still have the question if we want to display the names differently, [as Joris suggested above](https://github.com/kiesraad/abacus/pull/3587#issuecomment-4989834776). I'll approve this PR, so we can address this (if we want to) in a separate issue/PR.

jschuurk-kr (approved):

stacktraceghost:
> We also still have the question if we want to display the names differently, [as Joris suggested above](https://github.com/kiesraad/abacus/pull/3587#issuecomment-4989834776). I'll approve this PR, so we can address this (if we want to) in a separate issue/PR.

Created issue https://github.com/kiesraad/abacus/issues/3592

PR #3596, opened 2026-07-16
Apportionment refactor `splitResultChanges` due to residual seat logic bug

### LINKED ISSUE #3588: `splitResultChanges` residual seat logic bug
https://github.com/kiesraad/abacus/blob/7a33053bdc63e4e0fd13dee1ed9981ea95ff46ae/frontend/src/features/apportionment/utils/seat-change.tsx#L99-L102 assumes that a list cannot receive a residual seat through highest averages if it did not receive a seat through largest remainders, which is an incorrect assumption and can cause incorrect frontend rendering.

## TODO
- [x] Make sure steps in dataset are correctly typed and do renames @stacktraceghost 
- [x] Add testcase @stacktraceghost found to test the bug with @Lionqueen94 
- [x] Add state to each `testing/` variant @Lionqueen94 
- [x] Add separate steps back into seat_assignment variable for each `testing/` variant @Lionqueen94

stacktraceghost:
## What & why

Resolves #3588

`splitResultChanges` assumed that a list couldn't receive a residual seat through unique highest averages if it did not receive a seat through largest remainders, this could cause incorrect frontend rendering.

Since the functions `getResultChanges`/`splitResultChanges` became quite complex, these have been refactored into one function called `buildAssignmentTableData`. It divides the steps and seat changes for each table and then applies footnote numbers afterwards. 

## How to test

- Validate if tests succeed
- Go through different apportionment scenarios and check if the footnotes on the residual seats page are as expected

## Reviewer notes

It might be easier to review `seat-change.tsx` by simply looking at the file instead of the changes. 

The following sample can be used to reproduce the bug on main by adding it to `test_data_gen/api.rs` and generating the election, then mark the only candidate of list 3 as deceased.
```rs
args.custom_name = Some("Test".to_string());
let _ = create_test_election_with_votes(
    &args,
    pool,
    10,
    vec![
        vec![515, 0, 0, 0, 0, 0, 0, 0, 0, 0],
        vec![74, 0],
        vec![73],
        vec![72, 0],
        vec![71, 0],
        vec![70, 0],
        vec![60, 0],
        vec![60, 0],
    ],
)
.await;
```


frontend/src/features/apportionment/utils/seat-change.tsx
```diff
@@ -1,118 +1,185 @@
...
-      });
-    });
+  // The full seat removal footnote is only rendered once per list, not for every removal
+  if (change.full_seat) {
+    const existingFullSeatRemoval = tableData.LargestRemainderAssignment.resultChanges.find(
+      (c) => c.type === "FullSeatRemoval" && c.listNumber === change.list_retracted_seat,
+    );
+
+    // If < 19 this change is shown in the largest remainders table
+    // If >= 19 this change is not shown
```
praseodym: It looks like the `Footnotes` component doesn't filter `fullSeatRemovalSteps` to only <19 seats, won't this cause an unreferenced footnote to be rendered?

praseodym (commented):
Haven't completed my review but posting this question already so that you can look at it

frontend/src/features/apportionment/utils/seat-change.tsx
stacktraceghost: You are right, I am able to reproduce this. The `Footnotes` component does not use the output of this function, so I will remove these comment lines. 

Since we only render `LargestRemaindersTable` (which shows full seats) when it is a small council (< 19 seats), I will only show these footnotes under that condition.

I'll add a test case as well.

stacktraceghost: Fixed in https://github.com/kiesraad/abacus/pull/3596/commits/735d0a4494b7afe3e0917b3be7aba831941bae12

Lionqueen94 (approved):

oliver3 (approved):

PR #3627, opened 2026-07-21
Expanded ElectionCategory and added ElectionSubCategory

### LINKED ISSUE #3614: Election category and sub category
- extend election category enum with PS and WS
- add sub category enum
- add sub category to Election

Lionqueen94:
Closes #3614 

## What & why

- Expanded `ElectionCategory` with Provincial and WaterAuthority
- Added `ElectionSubCategory`
- Added `sub_category` to Election
- Added `ElectionCategoryAndSubCategoryMismatch` error and check
- Removed check for number_of_seats between 9 and 45 after discussion and agreement with @jschuurk-kr , since this is stated in individual laws such as the `Gemeentewet` and not in the Kieswet and we can assume this data is correct from the Kandidaatstelling software
- Added `CommitteeCategoryForElectionCategoryNotSupported` error and check for PS/WS only for GSB and not CSB

### Misc
- Alphabetically ordered `EMLImportError`
- Removed `impl From<String> for PollingStationType` since this is obsolete

## How to test

Verify that election creation still works for Municipal elections and that the sub category is added correctly from the eml


backend/src/eml/mod.rs
```diff
@@ -339,6 +345,21 @@ impl ElectionWithPoliticalGroups {
                     eml_nl::utils::ElectionSubcategory::GR2
                 }
             }
+            crate::domain::election::ElectionCategory::Provincial => {
+                // TODO: Needs to be changed to if single district else multiple districts
```
Lionqueen94: @praseodym I don't think this information about districts is available yet. Is this fine for now or do you prefer defaulting to PS1 for now?

backend/src/domain/election.rs
```diff
@@ -175,12 +175,16 @@ pub struct ElectionNumberOfVotersChangeRequest {
 #[strum(serialize_all = "lowercase")]
 pub enum ElectionCategory {
     Municipal,
+    Provincial,
+    WaterAuthority,
```
praseodym, line 186: We should add doc comments to these

backend/src/eml/mod.rs
praseodym: It's probably better to get this information from the election definition and save it to the database instead of trying to calculate it here.

If we don't fix that in this PR I'd much rather default to PS1 or even `panic!` than write something that is subtly incorrect.

Lionqueen94: Changed to default to PS1 for now and expanded the TODO some more in dfcc935b5ad39db131827c34d671b0c5d0047eea

backend/src/domain/election.rs
Lionqueen94: Added in dfcc935b5ad39db131827c34d671b0c5d0047eea

backend/src/eml/mod.rs
praseodym: In which issue are we going to resolve this? I think this is something we can already fix, we don't have to wait for the election tree.

Lionqueen94: Where can I currently find this information?

praseodym: The subcategory is in the election definition, e.g. `<kr:ElectionSubcategory>PS2</kr:ElectionSubcategory>`. Or do you mean something else?

praseodym (dismissed):
lgtm as long as we don't forget about fixing the subcategory

backend/src/eml/mod.rs
Lionqueen94: Sub category added to election in d50e7b03b81db3243428856b471b29298407573b

backend/src/domain/election.rs
```diff
@@ -174,13 +178,67 @@ pub struct ElectionNumberOfVotersChangeRequest {
...
+    }
+}
+
+impl ElectionSubCategory {
+    pub fn to_eml_code(&self) -> &'static str {
+        match self {
+            ElectionSubCategory::AB1 => "AB1",
+            ElectionSubCategory::AB2 => "AB2",
+            ElectionSubCategory::GR1 => "GR1",
+            ElectionSubCategory::GR2 => "AB2",
```
praseodym: typo:
```suggestion
            ElectionSubCategory::GR2 => "GR2",
```
I think we can also use the [strum crate](https://docs.rs/strum/0.28.0/strum/) to generate this code but not sure if that's worth it.
edit: just noticed we also have `#[strum(serialize_all = "lowercase")]` on the `ElectionSubCategory` enum

Lionqueen94: Fixed typo in 2c7e41bcf4d51f71e2387199d08ca14776b2f63d

Shall I remove the `serialize_all = "lowercase"` in this case?

praseodym: Yes remove it if it's not used

Lionqueen94: Removed in 262ee4ac6da8096291c7cc50e5a48a1529972ae7

backend/src/api/election.rs
```diff
@@ -164,17 +164,17 @@ pub async fn election_list(
...
         (status = 500, description = "Internal server error", body = ErrorResponse),
-    ),
-    params(
+  ),
+  params(
         ("election_id" = ElectionId, description = "Election database id"),
-    ),
+  ),
 )]
 pub async fn election_details(
```
praseodym, line 179: Why was the whitespace changed here to something inconsistent? (Was using 4 spaces to indent before)

backend/src/domain/election.rs
```diff
@@ -174,13 +178,53 @@ pub struct ElectionNumberOfVotersChangeRequest {
...
+            "AB1" => Self::AB1,
+            "AB2" => Self::AB2,
+            "GR1" => Self::GR1,
+            "GR2" => Self::GR2,
+            "PS1" => Self::PS1,
+            "PS2" => Self::PS2,
+            _ => panic!("invalid ElectionSubCategory `{value}`"),
         }
     }
 }
```
praseodym: Is this code used or can we remove it?

backend/src/eml/mod.rs
```diff
@@ -58,15 +101,11 @@ impl NewElection {
         // extract common information
         let election = &definition.election_event.election;
         let identifier = &election.identifier;
-        let category = identifier.category.copied_value()?;
+        let category = Self::get_category_or_err(identifier.category.copied_value()?)?;
+        let sub_category = Self::get_sub_category_or_err(identifier.subcategory.copied_value()?)?;
```
praseodym, line 112: I think we should also validate whether the `sub_category` is valid for the `category`, e.g. AB+AB1 is allowed but AB+GR1 is not allowed.

backend/src/eml/mod.rs
```diff
@@ -58,15 +101,11 @@ impl NewElection {
...
         let election_date = identifier.election_date.copied_value()?.date;
         let nomination_date = identifier.nomination_date.copied_value()?.date;
 
-        // we currently only support GR elections
-        if category != eml_nl::utils::ElectionCategory::GR {
-            return Err(EMLImportError::OnlyMunicipalSupported);
-        }
```
praseodym, line 68: We do need to limit PS/WS to only GSB in this epic

backend/src/api/election.rs
Lionqueen94: I didn't change this, my `cargo fmt` did, no clue why :woman_shrugging:  I can try to change it back

backend/src/domain/election.rs
Lionqueen94: It is needed, otherwise all query_as calls in election_repo were erroring (if you know another way to satisfy these, please enlighten me, clippy said it needed a `From<String>`)

backend/src/api/election.rs
Lionqueen94: Fixed via GitHub UI since locally it kept being removed in 8ef3b354429ca89680f1694a06ebc11b6020d518

backend/src/eml/mod.rs
Lionqueen94: Is it an idea to pass the CommitteeCategory up to this function, so this function can verify if we are creating a GSB or CSB election and if this is allowed. Or shall I add a check in the api functions? What do you think is best?

backend/src/api/election.rs
praseodym: There are also many whitespace changes in `backend/tests/election_admin_integration_tests.rs`. Maybe it is some IDE setting?

backend/src/domain/election.rs
praseodym: Let's keep it then

backend/src/api/election.rs
Lionqueen94: My RustRover update probably messed with my indent settings, fixed now

backend/src/eml/mod.rs
praseodym: I would check it in the API, e.g. in `import_csb_election`:

```rust
let mut new_election =
    parse_election_candidates_eml(&edu.election_data, Some(&edu.candidate_data))?;

// PS/WS CSB support will be implemented later
if matches!(
    new_election.category,
    ElectionCategory::Provincial | ElectionCategory::WaterAuthority
) {
    return Err(EMLImportError::CommitteeCategoryNotSupportedForCategory.into());
}
new_election.committee_category = CommitteeCategory::CSB;
```

backend/src/domain/election.rs
praseodym: You can fix it by adding `'sub_category' = "crate::domain::election::ElectionSubCategory"` under `[macros.table-overrides.'elections']` in `sqlx.toml`, then the `From<String>` can be removed

backend/src/api/election.rs
```diff
@@ -709,6 +709,10 @@ async fn create_sub_committees(
             )
             .await?;
         }
+        (
+            CommitteeCategory::CSB,
+            ElectionCategory::Provincial | ElectionCategory::WaterAuthority,
+        ) => {}
```
stacktraceghost: We could throw `unimplemented!("...")` or `todo!("...")`

> The difference between [unimplemented!](https://doc.rust-lang.org/std/macro.unimplemented.html) and [todo!](https://doc.rust-lang.org/std/macro.todo.html) is that while todo! conveys an intent of implementing the functionality later and the message is “not yet implemented”, unimplemented! makes no such claims. Its message is “not implemented”.

Found this in the rust docs, so I would recommend `todo!`

backend/src/domain/election.rs
```diff
@@ -174,13 +178,53 @@ pub struct ElectionNumberOfVotersChangeRequest {
 )]
 #[strum(serialize_all = "lowercase")]
 pub enum ElectionCategory {
+    /// Gemeenteraadsverkiezing
     Municipal,
+    /// Provinciale statenverkiezing
```
stacktraceghost:
```suggestion
    /// Provinciale Statenverkiezing
```
Nitpick: noticed that it is written like this on Kiesraad.nl, could be adjusted below as well.

backend/src/eml/mod.rs
```diff
@@ -47,6 +47,49 @@ use crate::domain::{
...
+                Ok(crate::domain::election::ElectionCategory::Provincial)
+            }
+            _ => Err(EMLImportError::LimitedElectionsSupported),
+        }
+    }
+
+    fn get_sub_category_or_err(
+        sub_category: eml_nl::utils::ElectionSubcategory,
+    ) -> Result<crate::domain::election::ElectionSubCategory, EMLImportError> {
+        match sub_category {
```
stacktraceghost: Nitpick, but this could be made more short like this. Same in other places.

```rs
    use crate::domain::election::ElectionSubCategory as Domain;
    use eml_nl::utils::ElectionSubcategory as Eml;

    let sub_category = match sub_category {
        Eml::AB1 => Domain::AB1,
        Eml::AB2 => Domain::AB2,
        Eml::GR1 => Domain::GR1,
        Eml::GR2 => Domain::GR2,
        Eml::PS1 => Domain::PS1,
        Eml::PS2 => Domain::PS2,
        _ => return Err(EMLImportError::LimitedElectionsSupported),
    };

    Ok(sub_category)
```

backend/src/api/election.rs
Lionqueen94: Other whitespace changes are also fixed in 0c8c4f71668dde3962592116221b98d02bd4fba8

backend/src/domain/election.rs
Lionqueen94: Done in 0c8c4f71668dde3962592116221b98d02bd4fba8

backend/src/eml/mod.rs
Lionqueen94: Added a check and error in 0c8c4f71668dde3962592116221b98d02bd4fba8

backend/src/eml/mod.rs
Lionqueen94: Added in 0c8c4f71668dde3962592116221b98d02bd4fba8

backend/src/api/election.rs
Lionqueen94: Done in 84850e63878e4365b87b310209d8cb822f25a08f

backend/src/domain/election.rs
Lionqueen94: Changed in 84850e63878e4365b87b310209d8cb822f25a08f

backend/src/eml/mod.rs
Lionqueen94: Nice improvement, implemented wherever relevant in 84850e63878e4365b87b310209d8cb822f25a08f

stacktraceghost (approved):

praseodym (approved):

Issue #3640, opened 2026-07-22
Adjust validation logic for voter cards


stacktraceghost:
- Add validation logic
  - That validates that the field should be `None` when `election.category == "Municipal"` (and return `DataError` otherwise)
  - That includes the field in existing validation rules
- Add diff logic

Issue #3619, opened 2026-07-21
Full flow end-to-end tests for PS/WS GSB


jschuurk-kr:
- [x] full flow test WS GSB CSO
- [x] full flow test WS GSB DSO
- [x] full flow test PS1 GSB CSO
- [x] full flow test PS2 GSB DSO
- [x] update full flow test GR GSB to check for no field kiezerspassen

Check if we miss end-to-end tests for things that are different from GR (election creation, kiezerspassen, ...).

---

Data sets from 2023
- https://data.overheid.nl/dataset/verkiezingsuitslag-provinciale-staten-2023#panel-resources
- https://data.overheid.nl/dataset/verkiezingsuitslagen-waterschappen-2023#panel-resources

PR #3643, opened 2026-07-22
Add voter_card_count to data structure

### LINKED ISSUE #3632: Add voter cards to data structure
As decided in https://github.com/kiesraad/abacus/issues/3615
 
- Add field voter cards (C) to `frontend/src/utils/dataEntryStructure.ts` for all types
  - Disabled if `election.category == "Municipal"`
  - Add snapshots (in tests) for every input combination
- Add an optional field `voter_card_count: Option<Count>` to struct `VotersCounts`.
- Adjust tests to handle the new field
- ~~Expand the `data_entry_status` fuzzer with the new election categories~~
  - No need to extend the fuzzer with new combinations. Checked if the fuzzer ran.

stacktraceghost:
## What & why

Resolves #3632 

- Backend
  - Added `voter_card_count` to struct `VotersCounts`
  - Updated tests
  - Updated `data_entry_fuzzer` and lockfile https://github.com/kiesraad/abacus/pull/3643/commits/fb2bafe6b76c876c44896704f7b51342864d5039
- Frontend
  - Added field `C Kiezerspassen` to `dataEntryStructure`. Disabled for `election.category === "Municipal"`
  - Added option to disable field in `dataEntryStructure`
  - Updated `useFormKeyboardNavigation` to skip disabled fields
  - Updated `dataEntryMapping.ts‎` to skip disabled fields, otherwise it throws an error that the field path doesn't exist on the data structure
  - Updated `‎InputGridSubsection.tsx‎` to set `previousValue` to empty string if `row.isDisabled`
  - Extended snapshots with `election.category`, so we can see the differences for every `model x election category` combination https://github.com/kiesraad/abacus/pull/3643/commits/3b257b9fe18befc65720910e3037b9ab91a72671

Validation logic will be adjusted in https://github.com/kiesraad/abacus/issues/3640

## How to test

- Login as typist and do a data entry for GSB Municipal. Check if the field C is disabled
  - When focused on field B, validate that the focus goes to field D when pressing enter
- Stop your data entry, adjust the `election.category` of your election in the database to `Provincial` and claim a data entry again
- Validate that field C is now enabled and remembering data
- When doing a corrigendum, check if the field behaves as expected. You can also try this out in Storybook.

## Reviewer notes

- Review per commit
  - Pay attention to the snapshot changes in https://github.com/kiesraad/abacus/pull/3643/commits/4bdfe8612700e132e47d6147c5aebd3db9aa947c

- Do we need to add `UncountedVotesReason::ValidVoterCards` to `as_eml_count_contest` in `backend/src/eml/mod.rs`?
- Maybe other places? Import/EML/PDFs?


backend/src/domain/results/voters_counts.rs
```diff
@@ -21,6 +21,10 @@ pub struct VotersCounts {
     /// Number of valid proxy certificates ("Aantal geldige volmachtbewijzen")
     #[schema(value_type = u32)]
     pub proxy_certificate_count: Count,
+    /// Number of valid voter cards ("Aantal geldige kiezerspassen")
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[schema(value_type = u32)]
+    pub voter_cards_count: Option<Count>,
```
Lionqueen94: For consistency, it should be `voter_card_count` instead of `voter_cards_count`

Lionqueen94 (changes requested):

backend/src/domain/results/voters_counts.rs
stacktraceghost: Done in https://github.com/kiesraad/abacus/pull/3643/commits/26677c264ae0a067d109ff9dc9c7afb45c8cb1ca

praseodym:
> * Do we need to add `UncountedVotesReason::ValidVoterCards` to `as_eml_count_contest` in `backend/src/eml/mod.rs`?

Yes.

> * Maybe other places? Import/EML/PDFs?

Yes, in the election summary used to generate EML and PDF files.

I'm ok with doing both in the next issue, as long as we do it within this epic.

praseodym (dismissed):

stacktraceghost:
@Lionqueen94 noticed that there was a frontend error when doing data entry for non-municipal elections. The frontend expects the propery on the data structure to be there, however after switching from my previous solution with `null` to skipping serialization if the value is `None`, the field was not there during the claim. This shows the importance of e2e tests for non-municipal elections, which we will do in https://github.com/kiesraad/abacus/issues/3619. 

I've fixed this in https://github.com/kiesraad/abacus/pull/3643/commits/8ab3543f45830c756215f84c7960e06e29d9e721 . I am now setting the `voter_card_count` in the `empty` methods. I prefer this over mutating results in `initial_current_data_entry`, since that could result in missing other places like `investigations.rs`. Tests have been added as well.

stacktraceghost:
> > * Do we need to add `UncountedVotesReason::ValidVoterCards` to `as_eml_count_contest` in `backend/src/eml/mod.rs`?
> 
> Yes.
> 
> > * Maybe other places? Import/EML/PDFs?
> 
> Yes, in the election summary used to generate EML and PDF files.
> 
> I'm ok with doing both in the next issue, as long as we do it within this epic.

Added issue https://github.com/kiesraad/abacus/issues/3672

Lionqueen94 (approved):
The field behaves properly now and the flow works for a Provincial election data entry as well

praseodym (approved):

PR #3649, opened 2026-07-23
Fix `per_page` overflow in audit log pagination

### LINKED ISSUE #3326: Pagination per_page value can overflow
https://github.com/kiesraad/abacus/blob/463dbafd782b44188a971d70dda54a920208eaea/backend/src/infra/audit_log/audit_log_events.rs#L115

This can overflow `u32::MAX` with high values, e.g. `page=100000&per_page=100000` query params. Can be fixed with `saturating_mul`/`checked_mul`. Similar to #3315.

praseodym:
## What & why

Fix `per_page` overflow in audit log pagination. Resolves #3326.

## How to test

Regression test is committed before the fix, so it can be tested by checking out b911b071bc4119e6ee4ad40dade5a1625625a459 and then verifying that 95cf98e2eab9ef69151ed56aabf4f256b5c6871c fixes the bug. Another way is checking that the audit log with `page=100000&per_page=100000` does not crash Abacus.

### DIFF
--- a/backend/src/api/audit.rs
+++ b/backend/src/api/audit.rs
@@ -235,6 +235,27 @@ mod tests {
     }
 
     #[test(sqlx::test(fixtures("../../fixtures/users.sql")))]
+    async fn test_list_paging_no_overflow(pool: SqlitePool) {
+        create_log_entries(pool.clone()).await;
+
+        // (page - 1) * per_page = 9_999_900_000 overflows u32::MAX; must not panic (#3326)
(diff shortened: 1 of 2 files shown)


stacktraceghost (approved):
Tested by checking main on abacus-test with `page=100000&per_page=100000` (500 upstream error) vs this branch (no error).

Lionqueen94 (approved):
Checked on main vs this branch, fixed!

PR #3665, opened 2026-07-23
Add correction states to data-entry-state.md

### LINKED ISSUE #3656: Update flow chart
Update `data-entry-state.md` with new states `FirstEntryCorrection`/`SecondEntryCorrection` and all transitions
 - `discard` goes to `FirstEntryFinalised`, swapping first and second entry if it was `FirstEntryCorrection`
 - `reset` goes to `Empty` similar to in-progress states

oliver3:
## What & why

Resolves https://github.com/kiesraad/abacus/issues/3656
- Update the diagram with the new states for streamlining resolving data entry differences and their transitions
- Remove `reset` transitions to try to prevent one big ball of transitions

## How to test

Check that the necessary transitions for the new states have been added and go to the correct states.

### DIFF
--- a/documentatie/state-machines/data-entry-state.md
+++ b/documentatie/state-machines/data-entry-state.md
@@ -2,12 +2,17 @@
 
 This document describes the states a data entry can have.
 The transition labels describe the endpoint that is used for performing the transition.
+
 The `save` endpoint which is used for [First/Second]EntryInProgress states is kept out, because Mermaid doesn't render self-loops too well.
+
 All states except for `FirstEntryHasErrors` and `EntriesDifferent` also have a `reset` endpoint which transitions to the `Empty` state. 
 
 Note the difference between `discard` and `reset`:
 - `discard` is a typist removing their own _in-progress_ entry (the `data_entry_discard` endpoint). Discarding an in-progress second entry keeps the finalised first entry. 
 - `reset` is a coordinator clearing the whole data entry back to `Empty` (the `data_entry_reset` endpoint). It always removes _both_ entries.
+
+When resolving differences between the first and second entry (`EntriesDifferent` state), the coordinator can choose to
+discard one of the two data entries. The remaining entry will from then on be the first entry, and the data entry is open for a new second entry.
 
 ```mermaid
 stateDiagram-v2
@@ -48,9 +53,4 @@ stateDiagram-v2
   Definitive --> [*]
 ```
 
-When resolving differences between the first and second entry (`EntriesDifferent` state), the coordinator can choose to
-keep one of the entries or discard both. If one of the entries is kept, the other entry is discarded. The remaining entry
-will from then on be the first entry, and the data entry is open for a new second entry.
 
-When the coordinator resets the data entry while there's a second entry (`SecondEntryInProgress` &
-`Definitive` state), both entries will be deleted.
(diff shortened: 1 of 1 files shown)


documentatie/state-machines/data-entry-state.md
```diff
@@ -2,27 +2,30 @@
...
+The `save` endpoint which is used for [First/Second]EntryInProgress states is kept out, because Mermaid doesn't render self-loops too well.
+
+All states except for `FirstEntryHasErrors` and `EntriesDifferent` also have a `reset` endpoint which transitions to the `Empty` state. 
 
 Note the difference between `discard` and `reset`:
 - `discard` is a typist removing their own _in-progress_ entry (the `data_entry_discard` endpoint). Discarding an in-progress second entry keeps the finalised first entry. 
 - `reset` is a coordinator clearing the whole data entry back to `Empty` (the `data_entry_reset` endpoint). It always removes _both_ entries.
 
+When resolving differences between the first and second entry (`EntriesDifferent` state), the coordinator can choose to
+discard one of the two data entries. The remaining entry will from then on be the first entry, and the data entry is open for a new second entry.
```
praseodym: This doesn't yet describe the new correction states

documentatie/state-machines/data-entry-state.md
```diff
@@ -2,27 +2,30 @@
 
 This document describes the states a data entry can have.
 The transition labels describe the endpoint that is used for performing the transition.
-The "save" endpoint which is used for [First/Second]EntryInProgress states is kept out, because Mermaid doesn't render self-loops too well.
+
+The `save` endpoint which is used for [First/Second]EntryInProgress states is kept out, because Mermaid doesn't render self-loops too well.
+
+All states except for `FirstEntryHasErrors` and `EntriesDifferent` also have a `reset` endpoint which transitions to the `Empty` state.
```
praseodym: Essentially the 'discard' and 'discard both entries' actions are the reset transitions for those states

documentatie/state-machines/data-entry-state.md
oliver3: My idea was that this sentence describes the switching of first and second entry, which is not applicable for the new states, only when the coordinator chooses to discard an entry. I'll try to make it more clear.

documentatie/state-machines/data-entry-state.md
oliver3: Good point, do you thing we should rename them to reset to be more consistent?

stacktraceghost: It depends what we want to show in the diagram. The discard actions are the actions the user actually takes. Reset is the effect.

I think we might have had a similar discussion in the past as well and we settled with this. 

Sidenote: we also have a discard and a reset endpoint. Although 'discard both entries' does not call the reset endpoint, since the state machine itself just sets the state to `Empty`.

oliver3: Thanks, I'll keep them separate and try to make this clear in the text.

documentatie/state-machines/data-entry-state.md
oliver3: Updated in https://github.com/kiesraad/abacus/pull/3665/commits/d62fb62c8e216a6aa9429acc4acc941179b63881

documentatie/state-machines/data-entry-state.md
oliver3: Updated in https://github.com/kiesraad/abacus/pull/3665/commits/d62fb62c8e216a6aa9429acc4acc941179b63881 with minor edit in https://github.com/kiesraad/abacus/pull/3665/commits/65ce5842a760d2a02f8889ccde92ffe0ad97e6a1

praseodym (approved):

stacktraceghost (approved):

PR #3666, opened 2026-07-23
Adjust validation logic for voter cards

### LINKED ISSUE #3640: Adjust validation logic for voter cards

stacktraceghost:
## What & why

Resolves #3640

- Added validation rules
  - Throwing `DataError` if an election with `ElectionCategory::Municipal` has a value in `voter_card_count` 
  - Throwing `DataError` if an election with a different `ElectionCategory` has no value in  `voter_card_count`
  - Added `voter_card_count` for field F.201
  - Added tests for these cases
- Added comparison logic and tests
- Updated F.201 copy 
- Updated tests
- Validation rules documentation has been adjusted in https://github.com/kiesraad/abacus-documentatie/pull/95 / https://github.com/kiesraad/abacus-documentatie/pull/94

## How to test

- Go to data entry and test if the validation works as expected
- Validate that the tests cover the field `voters_count.voter_card_count` properly

## Reviewer notes

- Does this cover everything?
- Since the field is an `Option<Count>`, the logic is different for comparison. Please review carefully.
- Does the validation behave like expected

### DIFF
--- a/backend/src/domain/results/voters_counts.rs
+++ b/backend/src/domain/results/voters_counts.rs
@@ -6,7 +6,7 @@ use utoipa::ToSchema;
 use super::count::Count;
 use crate::domain::{
     compare::Compare,
-    election::ElectionWithPoliticalGroups,
+    election::{ElectionCategory, ElectionWithPoliticalGroups},
     field_path::FieldPath,
     validate::{DataError, Validate, ValidationResult, ValidationResultCode, ValidationResults},
 };
@@ -34,6 +34,18 @@ impl AddAssign<&VotersCounts> for VotersCounts {
     fn add_assign(&mut self, other: &Self) {
         self.poll_card_count += other.poll_card_count;
         self.proxy_certificate_count += other.proxy_certificate_count;
+
+        // If both are Some, add the values
+        if let (Some(self_voter_card_count), Some(other_voter_card_count)) =
+            (self.voter_card_count, other.voter_card_count)
+        {
+            self.voter_card_count = Some(self_voter_card_count + other_voter_card_count);
+        }
+        // If self is None and other is Some, set self to other's value
+        else if self.voter_card_count.is_none() && other.voter_card_count.is_some() {
+            self.voter_card_count = other.voter_card_count;
+        }
+
         self.total_admitted_voters_count += other.total_admitted_voters_count;
     }
 }
@@ -51,6 +63,22 @@ impl Compare for VotersCounts {
             different_fields,
             &path.field("proxy_certificate_count"),
         );
+
+        match (&self.voter_card_count, &first_entry.voter_card_count) {
+            // If both Some, compare the values
+            (Some(voter_card_count), Some(first_entry_voter_card_count)) => {
+                voter_card_count.compare(
+                    first_entry_voter_card_count,
+                    different_fields,
+                    &path.field("voter_card_count"),
+                );
+            }
+            // If both none, do nothing
+            (None, None) => {}
+            // If one is Some and the other is None, add to different_fields
+            _ => different_fields.push(path.field("voter_card_count").to_string()),
+        }
+
         self.total_admitted_voters_count.compare(
             &first_entry.total_admitted_voters_count,
             different_fields,
@@ -75,16 +103,38 @@ impl Validate for VotersCounts {
             self.proxy_certificate_count
                 .validate(election, &path.field("proxy_certificate_count"))?,
         );
+
+        match (election.category, self.voter_card_count) {
+            (ElectionCategory::Municipal, None) => {}
+            (ElectionCategory::Municipal, Some(_)) => {
+                return Err(DataError::new(
+                    "Voter card count is not allowed for municipal elections",
+                ));
+            }
+            (_, None) => {
+                return Err(DataError::new(
+                    "Voter card count is required for non-municipal elections",
+                ));
+            }
+            (_, Some(voter_card_count)) => {
+                validation_results
+                    .join(voter_card_count.validate(election, &path.field("voter_card_count"))?);
+            }
+        }
+
         validation_results.join(
             self.total_admitted_voters_count
                 .validate(election, &path.field("total_admitted_voters_count"))?,
         );
 
-        if self.poll_card_count + self.proxy_certificate_count != self.total_admitted_voters_count {
+        if self.poll_card_count + self.proxy_certificate_count + self.voter_card_count.unwrap_or(0)
+            != self.total_admitted_voters_count
+        {
             validation_results.errors.push(ValidationResult {
                 fields: vec![
                     path.field("poll_card_count").to_string(),
                     path.field("proxy_certificate_count").to_string(),
+                    path.field("voter_card_count").to_string(),
                     path.field("total_admitted_voters_count").to_string(),
                 ],
                 code: ValidationResultCode::F201,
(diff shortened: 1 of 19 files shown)


praseodym:
> ⚠️ Based on #3643, first merge that PR. Changes for this PR can be found in [4833551](https://github.com/kiesraad/abacus/commit/4833551d24c2cc8141c0fa7939e914efe6244132)

Changed the PR target to `3632-add-voter-cards-to-data-structure` so that this PR only includes changes relative to that branch. The PR target will switch to `main` automatically once #3643 is merged (but this branch will still need rebasing because we use squash merges).

backend/src/domain/results/voters_counts.rs
```diff
@@ -75,16 +103,38 @@ impl Validate for VotersCounts {
...
+
+        match (election.category, self.voter_card_count) {
+            (ElectionCategory::Municipal, None) => {}
+            (ElectionCategory::Municipal, Some(_)) => {
+                return Err(DataError::new(
+                    "Voter card count is not allowed for municipal elections",
+                ));
+            }
+            (_, None) => {
+                return Err(DataError::new(
```
jorisleker, line 115: Shouldn't we make this check more explicit for the other supported election categories? If we add new election categories that also don't have voter cards (ie eilandsraad), it would be nice that we automatically notice that this code needs to be updated as well. If we just use else on the municipal election category, this code will just work, but lead to unwanted results.

Lionqueen94: In the pdf models I added `is_local_election` with a comment that we need to add "Island" when we add this category. We could do something similar in the backend: create a separate function with a comment/TODO to add the "Island" category in the future

This also means we can write "local" instead of "municipal" in all current comments

stacktraceghost: I've added `is_local_election` to `impl ElectionCategory`, so we can use that in the backend. In the frontend I've added the same as util so we can keep the comparison in one place in the future. They are both exhaustive, so if we add a value to the enum, we are required to add the case to those functions. 

Changed in https://github.com/kiesraad/abacus/pull/3666/commits/4fd46f7ba39dbc4b20db3eed5d54e488a84791d8

frontend/src/features/data_entry/components/voters_and_votes/VotersAndVotesForm.test.tsx
```diff
@@ -48,6 +48,7 @@ function renderForm() {
 const votersFieldIds = {
   pollCardCount: "data.voters_counts.poll_card_count",
   proxyCertificateCount: "data.voters_counts.proxy_certificate_count",
+  voterCardCount: "data.voters_counts.voter_card_count",
   totalAdmittedVotersCount: "data.voters_counts.total_admitted_voters_count",
 };
```
oliver3, line 54: Should we also check that voter_card_count is disabled in this test suite?

backend/src/domain/summary.rs
```diff
@@ -628,6 +628,7 @@ mod tests {
         assert_eq!(totals.voters_counts.total_admitted_voters_count, 94);
         assert_eq!(totals.voters_counts.poll_card_count, 89);
         assert_eq!(totals.voters_counts.proxy_certificate_count, 5);
+        assert_eq!(totals.voters_counts.voter_card_count, None);
```
oliver3, line 631: I was expecting a Some here. Why is it None?

praseodym: Looks like `election_fixture` creates a munipal election so this should be `None`?

oliver3: Ah yes absolutely, somehow was momentarily reversed in my head.

frontend/src/features/data_entry/components/voters_and_votes/VotersAndVotesForm.test.tsx
oliver3: Not really part of this PR, nvm

oliver3 (dismissed):

jorisleker (dismissed):
Tested:
- data entry for GSB and CSB, typist errors for voter cards displayed correctly
- resolve errors pages for GSB and CSB, coordinator errors for voter cards displayed correctly

Not tested:
- non-municipal elections with voter cards fields enabled

backend/src/domain/results/voters_counts.rs
```diff
@@ -75,16 +103,38 @@ impl Validate for VotersCounts {
...
 
-        if self.poll_card_count + self.proxy_certificate_count != self.total_admitted_voters_count {
+        if self.poll_card_count + self.proxy_certificate_count + self.voter_card_count.unwrap_or(0)
+            != self.total_admitted_voters_count
+        {
             validation_results.errors.push(ValidationResult {
                 fields: vec![
                     path.field("poll_card_count").to_string(),
                     path.field("proxy_certificate_count").to_string(),
+                    path.field("voter_card_count").to_string(),
```
praseodym: How does the frontend render this? Not sure if we should render an error on a disabled field.

jorisleker: It shows an error on the disabled field.

praseodym: And is that what we expect, also from an a11y standpoint? I'm fine with it either way

stacktraceghost: This renders an error on the disabled field indeed, this is a matter of choice. The validation rule is about A + B + C = D, so I think it is okay to show the error on that field. It also saves some conditionals based on the election category.

Example
image

jorisleker: It's not perfect, so if it's a small fix I would suggest removing it. 

Somewhat hesitant because the error message in the read only view in resolve errors (for the coordinator) does reference field C. But then again, we know for sure the error can't be in field C, so okay to not highlight that field as well.

stacktraceghost: Adjusted in https://github.com/kiesraad/abacus/pull/3666/commits/c3cb3d7c7ebdfb6bfc09e77901d63447a19272a6

jorisleker: Re-tested, LGTM 👌

frontend/src/features/data_entry/components/voters_and_votes/VotersAndVotesForm.test.tsx
stacktraceghost: Had to fix an e2e test, so decided to add this anyway: https://github.com/kiesraad/abacus/pull/3666/commits/5986dcfaa45de2af095f31d8feff84f99611fc9a

backend/src/domain/results/voters_counts.rs
stacktraceghost: @jorisleker How about the validation copy for the coordinator? It refers to field C, which we don't highlight now. Is it okay this way?

jorisleker: yes that's okay, as mentioned before:
> Somewhat hesitant because the error message in the read only view in resolve errors (for the coordinator) does reference field C. But then again, we know for sure the error can't be in field C, so okay to not highlight that field as well.

We know for sure the error can't be in C

oliver3 (dismissed):

oliver3 (approved):

Lionqueen94 (approved):
Validation works as expected and tests look good

PR #3667, opened 2026-07-23
Corrected `is_municipality` function and added `is_local_election` function

### LINKED ISSUE #3633: Adjust `is_municipality` in typst templates
In the typst templates we use `is_municipality` to decide if we show "kiezerspassen" or not, but also to decide between displaying "gemeente" and "openbaar lichaam". Showing of "kiezerspassen" should be based on `election.category` with as little logic as possible in the templates themselves.

We could also move this generic function to `scripts.typ`.

_Originally posted by @jschuurk-kr  in https://github.com/kiesraad/abacus/pull/3628#issuecomment-5042803232_

Lionqueen94:
Closes #3633 

## What & why

- Corrected `is_municipality` function and moved to scripts
- Added `is_local_election` function to scripts
- Use `is_local_election` function where kiezerspassen are involved
- Use `is_local_election` function on the signature pages

## How to test

- PDF diff should not show any changes
- Adjust one of the `election.location`s in one of the input json files to "Bonaire", "Saba" or "Sint Eustatius" and check that you get the "openbaar lichaam" version of the document (as discussed with @jschuurk-kr I didn't add any separate diffs for the public bodies yet, since they will not be using Abacus for now)

### DIFF
--- a/backend/templates/model-n-10-2.typ
+++ b/backend/templates/model-n-10-2.typ
@@ -2,16 +2,12 @@
 #import "common/scripts.typ": *
 #let input = json("inputs/model-n-10-2.json")
 
-#let is_municipality = (municipal, public_body) => if (
-  input.election.category == "Municipal"
-) { municipal } else { public_body }
-
 #let is_mobile = "polling_station_type" in input.polling_station and input.polling_station.polling_station_type == "Mobile"
 
-#let location_name = is_municipality[Gemeente #input.election.domain_id #input.election.location][Openbaar lichaam #input.election.location]
-#let location_type = is_municipality[gemeentelijk stembureau][stembureau voor het openbaar lichaam]
-#let this_location = is_municipality[deze gemeente][dit openbaar lichaam]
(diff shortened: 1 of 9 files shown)


backend/templates/model-n-10-2.typ
```diff
@@ -2,16 +2,12 @@
 #import "common/scripts.typ": *
 #let input = json("inputs/model-n-10-2.json")
 
-#let is_municipality = (municipal, public_body) => if (
```
praseodym, line 5: We can keep the code a lot more readable (and the diff minimal) if we keep defining `is_municipality` in each file but as a partially applied version of the shared function, fixing the location parameter for the entire file.

Lionqueen94: Thanks for the tip! Changed in f3cb48589e832e273e366fe2058f36f03a900315

praseodym (approved):
lgtm, PDF diff reports no changes

stacktraceghost (approved):

PR #3704, opened 2026-07-28
Allow importing polling stations with invalid number of voters

### LINKED ISSUE #3703: Abacus does not accept Polling station EMLs (110b) with MaxVotes of 0
The Polling station EMLs (110b) of GR26 contain 87 files in which `MaxVotes` is `0`. (There are 359 files in total, because some municipalities uploaded several versions. So the number of municipalities that have such an EML might be a little lower than 87.)

The rust-eml-nl eml-validator gives a warning for this case:
```
2026-07-28T08:05:42.144656Z  WARN eml_validator: - At position 1277 until 1299: Invalid value for {urn:oasis:names:tc:evs:schema:eml}MaxVotes: number would be zero for non-zero type
```

When trying to use this file in election creation, Abacus shows "Ongeldig stembureaubestand":
```
2026-07-28T08:06:48.683214Z ERROR request{method=POST uri=/api/elections/import/validate version=HTTP/1.1}: abacus::error: Error importing EML file: EMLError(UnknownPosition { kind: ValueConversionError(ParseIntError { kind: Zero }) })
```

Changing value of `MaxVotes` to `1` results in successful upload and election creation.

Issue found in #3524 Import stembureau EMLs checken.

---

Question is how we want so solve this. I think we should accept `MaxVotes` of `0` whem importing the EML, but the question is what happens after that. I see several options:
1. Validate nowhere it's been updated to the correct value. Risk is that it never gets updated to the correct value.
2. Force an update in the value during election creation. Risk is that the person creating the election does not have the correct number yet, so fills in something as incorrect as `0`.
3. Show an error icon in "Over deze verkiezing" as long as it's 0, but do not force it being updated. Risk is it won't get updated.
4. Do not allow the coordinator to "Start invoer" as long as the value is 0.

rnijveld:
## What & why

@jschuurk-kr [found out](https://github.com/kiesraad/abacus/issues/3703) that we don't import files in Abacus where the number of voters is invalid in EML110b files. It was decided that we want to allow importing such files and then fall back to zero whenever the MaxVotes element (containing the number of voters) in the EML110b file could not be properly read.

Resolves #3703.

## How to test

Try importing a new election through the import flow, make sure you have a 110b file that has an invalid MaxVotes element (i.e. with a value of 0, or with some random string that is not a number).

## Reviewer notes

The `InvalidNumberOfVoters` error variant was removed as it can no longer occur.


jschuurk-kr:
Created an election with GR26, using an EML110b with MaxVotes of 0.

Importing the EML file works. No warnings or errors in the backend logs.

Frontend has an empty field in the step "Aantal kiesgerechtigden"
image

Both leaving the field empty and filling a `0` result in an error in the frontend:
image
image

Filling in `1` lets you complete election creation.

Using an EML with `<MaxVotes>asdasdasd</MaxVotes>` has the same result.

jschuurk-kr:
@rnijveld We have `backend/src/eml/tests/eml110b_invalid_number_of_voters.eml.xml` with `<MaxVotes>test</MaxVotes>`. If that's used in a test, I would have expected that test to fail due to this change. Or did we fail to clean up that directory when introducing the rust-eml-nl crate?

In any case, if not on a lower level, we could at least create an e2e test for the `MaxVotes` `0` case.

jschuurk-kr:
> @rnijveld We have `backend/src/eml/tests/eml110b_invalid_number_of_voters.eml.xml` with `<MaxVotes>test</MaxVotes>`. If that's used in a test, I would have expected that test to fail due to this change. Or did we fail to clean up that directory when introducing the rust-eml-nl crate?
> 
> In any case, if not on a lower level, we could at least create an e2e test for the `MaxVotes` `0` case.

I can add the e2e test to this PR. And add the cleaning up of the test EML files to my todo list.

rnijveld:
> @rnijveld We have `backend/src/eml/tests/eml110b_invalid_number_of_voters.eml.xml` with `<MaxVotes>test</MaxVotes>`. If that's used in a test, I would have expected that test to fail due to this change. Or did we fail to clean up that directory when introducing the rust-eml-nl crate?

@jschuurk-kr This is a test that is now in the rust-eml-nl crate! I think we never properly cleaned up all the unused test files in that directory.

jschuurk-kr:
I added the e2e test in commit 8bf0740b8848736e8065b57cc5238e1a8a074991. Also fixed the naming of the page object for the page where you upload a polling station file.

frontend/e2e-tests/page-objects/election/create/UploadPollingStationsFilePgObj.ts
```diff
@@ -1,6 +1,6 @@
 import type { Locator, Page } from "@playwright/test";
 
-export class UploadPollingStationDefinitionPgObj {
+export class UploadPollingStationsFilePgObj {
```
stacktraceghost, line 3: File has not been renamed

backend/src/eml/mod.rs
```diff
@@ -973,12 +973,13 @@ pub fn number_of_voters_from_polling_stations_eml(
...
         .first()
         .ok_or(EMLImportError::PollingStationsWithoutContest)?;
-    contest
-        .max_votes
-        .copied_value()?
-        .get()
-        .try_into()
-        .map_err(|_| EMLImportError::InvalidNumberOfVoters)
+
+    Ok(match contest.max_votes.copied_value() {
```
stacktraceghost: We could add a backend test for this method to test if `123`, `0` and `randomstring` gives the expected result. That way we stay closer to testing the implementation than an e2e test.

frontend/e2e-tests/page-objects/election/create/UploadPollingStationsFilePgObj.ts
jschuurk-kr: fixed in 6044990c3906ff6bc042e6cdcabf8c0e79866a22

backend/src/eml/mod.rs
jschuurk-kr: @rnijveld said in Mattermost that it would take quite some effort to get different PollingStations structs to write such a  test.

backend/src/eml/mod.rs
```diff
@@ -973,12 +973,13 @@ pub fn number_of_voters_from_polling_stations_eml(
...
+        // If max_votes is too large we fall back to zero
+        Ok(v) => v.get().try_into().unwrap_or(0u32),
+        // If max_votes is not present or an invalid value we fall back to zero
+        Err(_) => 0u32,
+    })
```
praseodym: We have similar logic in the [`from_eml` function](https://github.com/kiesraad/abacus/blob/6044990c3906ff6bc042e6cdcabf8c0e79866a22/backend/src/eml/mod.rs#L135-L142), can we deduplicate it?

backend/src/eml/mod.rs
stacktraceghost: There are three ways to do it in a more compact way:
1. Take `tests/eml110b_test.eml.xml` and string replace `<MaxVotes>612694</MaxVotes>` with the value you want to test, pull it through `parse_polling_stations_eml_str` and then pass that into `number_of_voters_from_polling_stations_eml` and check the output
2. Use `election_fixture` to get an election, then use `election.as_polling_stations_eml(&polling_stations_fixture(...)))` to get the polling stations eml and manipulate `max_votes`. Then pass it into `number_of_voters_from_polling_stations_eml` and check the output
3. Abstract the logic to an `impl` and test that, but that doesn't guarantee that `number_of_voters_from_polling_stations_eml` actually uses it

Depending on what we decide in https://github.com/kiesraad/abacus/pull/3704#pullrequestreview-4816711501

rnijveld: I've decided that testing the new separate `max_votes` function should be good enough here to make sure we actually test the conversion, see 6f9a090b2ed58b83d428504c869f4ac921316528. That also prevents us from having to create any of these convoluted workarounds.

backend/src/eml/mod.rs
rnijveld: Fixed in 6f9a090b2ed58b83d428504c869f4ac921316528

stacktraceghost (approved):

praseodym (approved):

PR #3706, opened 2026-07-28
Resolving differences for second entry with errors

### LINKED ISSUE #3675: Add special case for second data entry with errors
If a second data entry has errors and it is chosen as the correct one, the first data entry is discarded and the coordinator gets redirected to the first entry resolve errors screen. However, this flow doesn't work when correction gets chosen.

@jorisleker updated the design of the resolve differences flow to block the correction option when the second data entry has errors. A information message is shown and the submit button label is updated.

[Design in Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=24183-12477&t=9QSAsPozF4GV3xZt-4)

praseodym:
## What & why

Add special case for resolving differences for second entry with errors: only discarding the first entry should be allowed. Resolves #3675.

## How to test

Create a data entry with differences using two different typist users, and make sure the second data entry contains errors. Then check the resolve differences page using the coordinator user.

It should look like this:
image

### DIFF
--- a/backend/src/domain/data_entry.rs
+++ b/backend/src/domain/data_entry.rs
@@ -36,6 +36,8 @@ pub enum DataEntryTransitionError {
     CannotTransitionUsingDifferentUser,
     /// The second data entry needs to be claimed by a user other than the one who claimed the first entry
     SecondEntryNeedsDifferentUser,
+    /// Correction is not allowed because the second entry has errors
+    CorrectionNotAllowed,
     ValidatorError(DataError),
     ValidationError(ValidationResults),
 }
@@ -745,6 +747,21 @@ impl DataEntryStatus {
         }
     }
 
+    /// Keep second entry and let original typist correct first entry.
+    /// This is only allowed when the second entry has no errors.
+    pub fn keep_second_and_correct_first(
+        self,
+        election: &ElectionWithPoliticalGroups,
+    ) -> Result<Self, DataEntryTransitionError> {
+        if let DataEntryStatus::EntriesDifferent(state) = &self
+            && state.second_entry.start_validate(election)?.has_errors()
+        {
+            return Err(DataEntryTransitionError::CorrectionNotAllowed);
+        }
+
+        self.keep_second_entry(election)
+    }
+
     /// Get the progress of the first entry (if there is a first entry), from 0 to 100
     pub fn get_first_entry_progress(&self) -> Option<u8> {
         match self {
@@ -888,6 +905,9 @@ impl Display for DataEntryTransitionError {
                 )
             }
             DataEntryTransitionError::Invalid => write!(f, "Invalid state transition"),
+            DataEntryTransitionError::CorrectionNotAllowed => {
+                write!(f, "Correction not allowed: second entry has errors")
+            }
             DataEntryTransitionError::ValidatorError(data_error) => {
                 write!(f, "Validator error: {data_error}")
             }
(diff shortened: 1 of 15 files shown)


backend/src/domain/data_entry.rs
```diff
@@ -36,6 +36,8 @@ pub enum DataEntryTransitionError {
     CannotTransitionUsingDifferentUser,
     /// The second data entry needs to be claimed by a user other than the one who claimed the first entry
     SecondEntryNeedsDifferentUser,
+    /// Correction is not allowed because the second entry has errors
```
stacktraceghost, line 39: Nitpick: the docstring is a too specific for a generic error named `CorrectionNotAllowed`. I would make the message more generic or the name more specific.

documentatie/state-machines/data-entry-state.md
```diff
@@ -14,6 +14,14 @@ Note the difference between `discard` and `reset`:
 When resolving differences between the first and second data entry (`EntriesDifferent` state), one of the options for the coordinator is to
 discard one entry. In this case, the remaining entry will from then on be the first entry, and the data entry is open for a new second entry.
 
+Instead of discarding an entry, the coordinator can also have it corrected by
```
stacktraceghost, line 17: Nitpick: line length differs from rest of markdown

frontend/src/features/resolve_differences/components/ResolveDifferencesForm.stories.tsx
```diff
@@ -87,13 +90,13 @@ export const Default: Story = {
     await expect(args.setWrongEntryAction).toHaveBeenLastCalledWith("discard");
     await expect(discardWrongEntry).toBeChecked();
 
-    // Choosing "neither" disables the second question again and clears its answer
+    // Choosing "neither" disables the second question again and sets discarding as the (disabled) answer
```
stacktraceghost, line 93:
```suggestion
    // Choosing "neither" disables the second question again and sets discarding as the checked answer
```

frontend/src/features/resolve_differences/components/ResolveDifferencesForm.tsx
```diff
@@ -19,12 +20,13 @@ export function ResolveDifferencesForm({
   setCorrectEntry,
```
stacktraceghost, line 20: Unrelated to this PR, but the `formState` (=`ResolveDifferencesFormState`) is being spread over `<ResolveDifferencesForm>` in `ResolveDifferencesPage.tsx`. Maybe better to just simply pass the entire form state? Then we can also drop the extends.

stacktraceghost (commented):
Tested by:
- Created a GSB GR election with first entries 100% and second entries 0%
- Entered a second entry with differences (A)
- Entered a second entry with differences and errors (B)

A did not block the option
B did block the option

Question: as a coordinator I was unable to see that B had errors (it has differences + errors), I only saw this:

Image

Is that clear enough?

jorisleker:
> Question: as a coordinator I was unable to see that B had errors (it has differences + errors), I only saw this:
> 
> Image
> Is that clear enough?

That the second data entry contains errors is only relevant when you decide to keep it, it should not be something the coordinator takes into account when comparing both entries: they should primarily compare both entries with the paper report. 

Once you select the second entry as the one to keep, you are informed that it has errors. Ideally you could resolve those errors immediately, but that will be a quite a lot work to design and implement. As this for sure is an edge case, we should not invest too much time in designing and building something for that use case.

documentatie/state-machines/data-entry-state.md
praseodym: The rest of the file is inconsisent so I just went for whatever line length Zed's rewrap function used.

backend/src/domain/data_entry.rs
praseodym: The docstring clarifies when this error can happen. I didn't want to write `CorrectionIsNotAllowedBecauseTheSecondEntryHasErrors` as the enum variant. Open to suggestions if we think this is important enough to fix.

jorisleker (approved):

frontend/src/features/resolve_differences/components/ResolveDifferencesForm.stories.tsx
stacktraceghost: Now I am reading this again and it actually does work.

stacktraceghost (approved):

frontend/src/features/resolve_differences/components/ResolveDifferencesForm.tsx
praseodym: I can make a separate PR for this

Issue #3705, opened 2026-07-28
Add input for PS and WS elections for Na 31-2 (Bijlage 1) and Na 14-2 (Bijlage 1) models


Lionqueen94:
- Add input for PS and WS elections for Na 31-2 model
- Add input for PS and WS elections for Na 31-2 Bijlage 1 model
- Add input for PS and WS elections for Na 14-2 model
- Add input for PS and WS elections for Na 14-2 Bijlage 1 model
- Adjust Na 14-2 model to show Kiezerspassen when provided
- Adjust Na 14-2 Bijlage 1 model to show Kiezerspassen when provided

PR #3707, opened 2026-07-28
Add valid voter cards for EML

### LINKED ISSUE #3672: Add valid_voter_cards for EML
`UncountedVotesReason::ValidVoterCards` should be added to `as_eml_count_contest` in `backend/src/eml/mod.rs`

Check if there are more places where this needs to be done.

Related comment: https://github.com/kiesraad/abacus/pull/3643#issuecomment-5068404796

stacktraceghost:
## What & why

Resolves #3672

- Set `voter_card_count` in `ElectionSummary::zero` based on `is_local_election`
- Add `total_uncounted_votes` for `ValidVoterCards` to `ElectionCountContest` and `ReportingUnitVotes` builder
  - Kept the same order as fields in `voters_counts`
  - Added a small helper to add this conditionally inline for `Some(...)`
- Adjusted tests

## How to test

- On abacus-test, download documents for a GR election and a WS/PS election.
- In the EML, look for `UncountedVotes ReasonCode="geldige kiezerspassen">`

The added tests check for the same. 

## Reviewer notes

- Typst models and inputs will be adjusted in https://github.com/kiesraad/abacus/issues/3705
- Is everything covered?


Lionqueen94 (dismissed):
Works as described

backend/src/eml/mod.rs
```diff
@@ -1146,26 +1198,88 @@ mod tests {
...
+                .first()
+                .unwrap()
+                .uncounted_votes
+                .contains_key(&UncountedVotesReason::ValidVoterCards)
+        );
+    }
+
+    #[test]
+    fn test_as_eml_count_contest_with_voter_cards() {
+        let election = election_fixture(CommitteeCategory::GSB, &[2]);
```
praseodym, line 1241: `election_fixture` creates a municipal election on which then we set `voter_card_count`, which is technically an invalid state. It's ok for now because `as_eml_count_contest` and the EML crate don't validate this, but it's not ideal.

praseodym (dismissed):
lgtm! The `ApplyIfSome` trait is interesting, although I'm not convinced if I prefer it over more straightforward Rust.

backend/src/eml/mod.rs
stacktraceghost: I can extend `election_fixture` with the parameter `election_category`, so we can also re-use it in the future. Since it touches a lot of files, I will do this in a separate PR.

stacktraceghost:
> lgtm! The `ApplyIfSome` trait is interesting, although I'm not convinced if I prefer it over more straightforward Rust.

I agree that we should not overcomplicate/abstract things, so I've decided to revert to normal syntax. https://github.com/kiesraad/abacus/pull/3707/commits/796b5cf80333ba943236aa192012c47bc91cfe39

praseodym (approved):

Lionqueen94 (approved):

backend/src/eml/mod.rs
stacktraceghost: Done in https://github.com/kiesraad/abacus/pull/3736

Issue #3658, opened 2026-07-23
Implement data entry correction discard and reset


oliver3:
See flowchart from #3656

Issue #3657, opened 2026-07-23
Show data entry warnings and empty fields


oliver3:
- Return differences in `ClaimDataEntryResponse` endpoint
- Store differences and clear fields in data entry reducer initially and persist using ClientState
- Add `W002` to `ValidationResultCode` and translations
- Update components and hooks to show warnings
- Add tests

oliver3:
@jorisleker Should W.001 ("De gemarkeerde velden zijn anders dan in de eerste invoer.") be shown at all when correcting first or second entry?

jorisleker:
Interesting question. Would be weird showing W.001 and W.002 at the same time. So if we wanted to show W.001, we would need some logic as to when to show W.001 and when to show W.002. Could get complicated.

Thinking about it a bit more: I would suggest to not show W.001 when correcting a first or second data entry _because of differences_. Reasoning: you know where the differences were (they are marked by W.002 and cleared), and you should copy it from paper and not be guided by how someone else interpreted the handwriting. Showing that your input isn't the same as the other typists, would be steering too much. Not showing W.001 is also best in line with the gist of discussions we had with Melanie and Arjen about correcting differences.

oliver3:
> Showing that your input isn't the same as the other typists, would be steering too much.

Agreed! Let's not show W.001 then for correcting differences in first or second entry.

PR #3713, opened 2026-07-29
Add DataEntryStatus correction states

### LINKED ISSUE #3655: Add new DataEntryStatus states
Add `DataEntryStatus` states and minimal implementation to be able to start data entry in correction state and display state correctly (data entry home, coordinator overview, more?)

No discard, no W.002 warnings or clearing fields yet

oliver3:
## What & why

Resolves #3655
- Code cleanup to make the change possible and more clear (see also Reviewer notes below)
- Add `FirstEntryCorrection` and `SecondEntryCorrection` states to `DataEntryStatus` according to the [state machine diagram
](https://github.com/kiesraad/abacus/blob/main/documentatie/state-machines/data-entry-state.md)
- Add state transitions to be able to resolve differences with all options and correct it as a typist
- Update frontend to handle the new states correctly
- NB: not implemented yet: [show warning and empty fields](https://github.com/kiesraad/abacus/issues/3657), and [discard (typist)](https://github.com/kiesraad/abacus/issues/3658)

## How to test
Go through the data entry flow:
- Create two different data entries with two typists
- Resolve differences by letting a typist correct their data entry
- Check that you can now re-enter your data entry as a typist

Check that all resolve differences options work as expected

Check that the states are displayed and handled correctly:
- Typist sees the data entry on the data entry home page (dropdown, errors)
- Coordinator sees the data entry on the election status page with the correct status, typist, percentage, ...
- Other things that I might have missed?

## Reviewer notes

Best to be reviewed per commit:
- Rename remaining "delete" to "discard" in data entry api and domain
- Combine `get_first_entry_progress()`, `get_second_entry_progress()` and `get_progress()` to be able to show the correct percentage in the frontend for correction states
- Move `initial_current_data_entry` logic to domain, and rename struct `CurrentDataEntry` used in claim and update transitions to `DataEntryUpdate` and use in update only.
- Add `DataEntryStatus` correction states and transitions


backend/src/domain/data_entry.rs
```diff
@@ -748,36 +748,16 @@ impl DataEntryStatus {
...
-        }
-    }
-
-    /// Get the progress of the second entry (if there is a second entry), from 0 to 100
-    pub fn get_second_entry_progress(&self) -> Option<u8> {
-        match self {
-            DataEntryStatus::Empty
-            | DataEntryStatus::FirstEntryInProgress(_)
-            | DataEntryStatus::FirstEntryFinalised(_) => None,
+            DataEntryStatus::FirstEntryHasErrors(_) => None,
```
stacktraceghost, line 935: Before we considered `FirstEntryHasErrors` / `EntriesDifferent` as 100%, now as `None`, is that correct / why?

stacktraceghost (commented):
Reviewed code.

Checked codebase for `delete` for inconsistencies:
- `ErrorReference::DataEntryCannotBeDeleted` is being used for reset, suggestion: change to `ErrorReference::DataEntryCannotBeReset`
- `DataEntryDeletedAuditData` is being used for
  - Actual deletion of a data entry for a polling station, matches what it does
  - Data entry discard, would suggest a separate audit event since it changes the state (not always `Empty`)
  - Data entry reset, would suggest a separate audit event since it always goes back to `Empty`
- Update `data_entry_integration_test.rs`
  - `test_data_entry_deletion` -> `test_data_entry_discard` + `delete_data_entry` -> `data_entry_discard`

In the frontend there might also be some inconsistencies in names. Could also be solved in a separate PR.

backend/src/domain/data_entry.rs
oliver3: `get_first_entry_progress()` and `get_second_entry_progress()` were used for showing the percentage in the election status, in `CategoryRow.tsx`:
```tsx
percentage: entry.second_entry_progress ?? entry.first_entry_progress ?? 0,
```

This was only shown for entries that had category "in_progress". Therefore the value for other states such as `FirstEntryHasErrors` and `EntriesDifferent` is never shown, and although you could argue that the value is 100% in these cases, I prefer None to indicate that it is not applicable when there is no data entry in progress.

oliver3: Also, the idea to have a "total progress", as the docstring for the original `get_progress()` implies (was used for audit log only), is not really a practical number now that you can go back to the first entry while the second entry is already done. This is the other reason why I think it is a good idea to only talk about the progress of a typist doing a data entry.

oliver3:
> Checked codebase for `delete` for inconsistencies:
> 
> * `ErrorReference::DataEntryCannotBeDeleted` is being used for reset, [..]
> * `DataEntryDeletedAuditData` is being used for [..]
> * Update `data_entry_integration_test.rs`

Changed in https://github.com/kiesraad/abacus/pull/3713/commits/21fce48dbdebfe1ee31323eb109f3d4693acf918

> In the frontend there might also be some inconsistencies in names. Could also be solved in a separate PR.

I'd prefer not to change new files indeed and would suggest leaving it as is for now

backend/src/domain/data_entry.rs
```diff
@@ -557,6 +630,36 @@ impl DataEntryStatus {
...
+
+                    Ok(Self::Definitive(Definitive {
+                        first_entry_user_id: state.first_entry_user_id,
+                        second_entry_user_id: state.second_entry_user_id,
+                        finished_at: Utc::now(),
+                        finalised_with_warnings: validation_results.has_warnings(),
+                        results: state.finalised_second_entry.clone(),
+                    }))
+                } else {
+                    Ok(Self::EntriesDifferent(EntriesDifferent {
```
praseodym, line 653: I think through this branch the data entry can get to a `EntriesDifferent` state with the first entry containing errors, which was previously not possible (only the second entry could contain errors, a first entry with errors would have been handled right after the first entry was finalised).

backend/src/domain/data_entry.rs
praseodym, line 1003: I think we also need to add the new states here

praseodym: This function should return `first_entry_finished_at` for the `SecondEntryCorrection` state

praseodym (changes requested):
See commends. In addition, the `ElectionProgress` component also needs to be updated (`"second_entry_correction"` added to `firstEntryFinished`).

backend/src/domain/data_entry.rs
oliver3: I think so too, I've created https://github.com/kiesraad/abacus/issues/3720 to find a solution. 

At this moment, if you proceed with a first correction with errors, and then try to correct the second entry with the same error, `finalise_second_entry` will stop you from reaching Definitive.

backend/src/domain/data_entry.rs
oliver3: I actually do not see the logic of this function especially now that we can mix and match the order of first and second entry. It is used only for audit logging, which contains timestamps that can be used to reconstruct all the different moments in time.

I suggest to remove this function and the corresponding audit log field.

oliver3: Updated in https://github.com/kiesraad/abacus/pull/3713/commits/55a7cbcd41985eee3a9daa75543d90c4d14f08e0

oliver3:
> In addition, the `ElectionProgress` component also needs to be updated (`"second_entry_correction"` added to `firstEntryFinished`).

Updated in https://github.com/kiesraad/abacus/pull/3713/commits/59f42c128f1c04b5e4c226fd63b4558e6c2ce591

backend/src/domain/data_entry.rs
praseodym: Fine by me!

praseodym (dismissed):

backend/src/domain/data_entry.rs
oliver3: Found out that it was also used for election status, so I've still removed the function and moved the logic to the election status, so that the context is more clear what these timestamps mean.

Done in https://github.com/kiesraad/abacus/pull/3713/commits/01cea1d9784e4d872c984883457030466a083d56

praseodym (approved):

stacktraceghost (approved):

PR #3725, opened 2026-07-30
Backend implementation radio buttons

### LINKED ISSUE #3684: Backend implementation for Data Entry Radio buttons
- Use Radio Button created in #3683
- Decide on backend/API representation

stacktraceghost:
## What & why

Resolves #3684

As proposed in https://github.com/kiesraad/abacus/issues/3684#issuecomment-5119878263

- Adjusted radio fields to be required, but nullable
- Adjusted test `check_openapi_json_for_null_type` to ignore nullable fields marked as required
- Adjusted OpenAPI TS generator to handle `null` and improved logic to get descriptions

## How to test

- Check if the types are correct

## Reviewer notes

N/A

### DIFF
--- a/frontend/src/types/generated/openapi.ts
+++ b/frontend/src/types/generated/openapi.ts
@@ -407,12 +407,8 @@ export type USER_DELETE_REQUEST_PATH = `/api/users/${UserId}`;
  * Information about the report ("Over het proces-verbaal")
  */
 export interface AboutReport {
-  /** Whether the extra page "controles en correcties" is inserted in the report
-("Is voorin het proces-verbaal de extra pagina controles en correcties ingevoegd?") */
-  checks_and_corrections_present?: ChecksAndCorrectionsPresent;
-  /** Whether a corrigendum accompanies the report
-("Is er een corrigendum bij het papieren proces-verbaal aanwezig?") */
-  corrigendum_present?: CorrigendumPresent;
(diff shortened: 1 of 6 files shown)


frontend/src/types/generated/openapi.ts
praseodym, line 1: nit: why were the comments dropped?

stacktraceghost: Because in the previous version, `tsType` received an object, in which case it went to loop over the properties and added a description if available.

~~In the new version, `tsType` receives `oneOf`, then recursively gets `null` (which is `null`) and `$ref` (where it takes the last bit/object name).~~

~~I've shortly tried to add the description within `if ("$ref" in s) {`, however this caused some duplication and I decided not to pursue that. If we really want to add it, I can spend some more time.~~

Edit: it actually receives an object as well, but the line of code that tries to get the description can't find it, because it is now a level deeper (inside `oneOf`).

stacktraceghost: Of course I still tried. Found a solution, simply made the generator look into the `oneOf` when it gets the object. https://github.com/kiesraad/abacus/pull/3725/commits/31e523dddea5a44b7b577673d806ed4fceb82714

Lionqueen94 (approved):
Looks good

praseodym (approved):

PR #3726, opened 2026-07-30
Enable full navigation for data entry corrections

### LINKED ISSUE #1990: When the typist restarts data entry after resolve errors, the navigation should be completely enabled
After the coordinator resolves errors by selecting "Invoer bewaren en correcties laten invoeren", and the typist starts data entry again, the navigation menu should show all the correct icons for the sections and the typist should be able to navigate through all the sections. See #3037 for a more thorough explanation.

This could probably be solved by:
- Making a backend struct for the `clientState`, with a more appropriate name
- Generate a state with `furthest` set to the last page and `current` to the first page

praseodym:
## What & why

Enable full navigation when a data entry is returned to the typist for corrections. Resolves #1990.

## How to test

Either:
- Create a data entry with errors, then as coordinator choose to return to typist for corrects 
- Create two data entries with differences, then as coordinator choose to return either for corrections
And then verify that the typist has the entire navigation enabled when correcting the data entry.

## Reviewer notes

@oliver3 and I had some discussion about the implementation, we chose to keep the client state opaque for the backend and go for this solution (see comments in #1990).

### DIFF
--- a/backend/src/domain/data_entry.rs
+++ b/backend/src/domain/data_entry.rs
@@ -296,6 +296,9 @@ pub struct FirstEntryInProgress {
     #[schema(value_type = Object)]
     /// Client state for the data entry (arbitrary JSON)
     pub client_state: ClientState,
+    /// Whether this entry was returned for correction
+    #[serde(default)]
+    pub is_correction: bool,
 }
 
 #[derive(Debug, Default, Serialize, Deserialize, Clone, Type, Eq, PartialEq)]
@@ -451,6 +454,7 @@ impl DataEntryStatus {
                 first_entry_user_id: user_id,
                 first_entry: initial_results,
                 client_state: ClientState::default(),
+                is_correction: false,
             })),
             DataEntryStatus::FirstEntryInProgress(_) | DataEntryStatus::FirstEntryCorrection(_) => {
                 if user_id == self.get_first_entry_user_id().expect("user id is present") {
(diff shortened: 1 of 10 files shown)


jorisleker (commented):
One thing that is a bit confusing (but acceptable):
- if you have a data entry with several sections that contain errors, 
- and fix the last section first,
- you are redirected to the first section with errors when clicking 'volgende'

But again: acceptable behavior, explicitly directing the user to what needs to be fixed might even be a good idea.

jorisleker (dismissed):
Tested on abacus-test, for GSB (normal flow and corrigendum), and in both cases the navigation is enabled ✅

backend/src/domain/data_entry.rs
```diff
@@ -296,6 +296,9 @@ pub struct FirstEntryInProgress {
     #[schema(value_type = Object)]
     /// Client state for the data entry (arbitrary JSON)
     pub client_state: ClientState,
+    /// Whether this entry was returned for correction
+    #[serde(default)]
```
oliver3: Why is this needed? It seemed to work without it as well.

praseodym: I added it because I expected fixtures to break, but they don't, so removed in eaedb0ca5c37f62013ed792bb056897c6e4ab059

oliver3 (approved):

jorisleker (approved):

PR #3709, opened 2026-07-28
Add input for PS and WS elections for Na 31-2 (Bijlage 1) and Na 14-2 (Bijlage 1) models

Lionqueen94:
Closes #3705 
Closes #3620 

## What & why

- Updated Na 14-2 and Na 14-2 Bijlage 1 typst models to render "kiezerspassen" when provided
- Added PS and WS variations of test inputs for Na 14-2, Na 14-2 Bijlage 1, Na 31-2 and Na 31-2 Bijlage 1 models in separate folders
- Added PS and WS variants for Na 14-2, Na 14-2 Bijlage 1, Na 31-2 and Na 31-2 Bijlage 1 models in pdf diff

### Misc
- Renamed `extra-model-p-22-2-variations` folder to `model-p-22-2-variations`
- Removed `model-p-22-2.json` and updated references since this is a copy of `model-p-22-2-variations/lt-19-seats-and-p9-and-p10.json`
- Corrected `election_id` in input test files
- Corrected `Test` location and `domain_id` in original Na 14-2 Bijlage 1 input json file
- All input files for Juinen now use the same `domain_id` (0035)
- Removed `counting_method` from one input test file with `committee_category: "CSB"`
- Renamed `bijlage1` in typst and json files to `bijlage-1`

## How to test

- Run `cargo run --bin gen-pdf` locally and verify the contents of the added variations for these models:
  - Na 14-2
  - Na 14-2 Bijlage 1
  - Na 31-2
  - Na 31-2 Bijlage 1

## Reviewer notes

- Do we want to add more tests to `typst_tests.rs` and `typst_smoke_tests.rs`?

PR #3730, opened 2026-07-30
Update visual-diff.sh after file renames

sanderboom:
## What & why

Update `visual-diff.sh` after file renames in https://github.com/kiesraad/abacus/pull/3709. Could be expanded further (e.g. not hardcoding filepaths), but this change at least makes it work again.

## How to test

Run `visual-diff.sh` from `backend/templates`.

## Reviewer notes

None.


Lionqueen94 (approved):
Sorry forgot about this!

praseodym (approved):

PR #3731, opened 2026-07-30
Improve and make footers consistent across all 2027 CSO models

sanderboom:
## What & why

Combined PR for models mentioned in Epic https://github.com/kiesraad/abacus/issues/3497

Add `(versie 2027)` to the footer (so different from the solution proposed in the supplied PDFs). 

Change wording in footers with SHAs:
- P 22-2 (and Bijlage)
- Na 14-2
- Na 31-2

Changed:
- Add model info (consistent with other models)
- Add `Datum:`
- Remove `(SHA-256)` and replace with `Digitale vingerafdruk van EML-telbestand bij dit PV`
- Ensure `margin-bottom: 3.2cm` (taken from `model-p-22-2.typ` that already had a 4 line footer)

Result:
image

## How to test

Create the PDFs and use `diff-pdf`. 

## Reviewer notes

Note: the larger footer means less content fits per page, so diff-pdf flags many pages as changed.


jorisleker:
Do we need to add the hash on all pages of the report? It takes up a lot of space (mainly an issue in the votes per list/candidate sections), and is relevant only once. **I propose to only show the hash on the first page**. And in future versions of the model, consider adding an extra field in the reports contents instead of in the footer (that way, we can add a bit more context about the hash, and explain it's role in the process).

@jschuurk-kr @arjen-ba @ring-ring-ring what do you think?

jorisleker (changes requested):
- ✅ Version looks good to me
- ✅ Footer with date and hash: contents and design OK, but: 
- ❌ can we flip the order? So the model name / description is always at the bottom, and the hash + timestamp are above that, and
- ❌ would not show it on all pages. Suggest removing this element from all pages but the first

ring-ring-ring:
> Do we need to add the hash on all pages of the report? It takes up a lot of space (mainly an issue in the votes per list/candidate sections), and is relevant only once. **I propose to only show the hash on the first page**. And in future versions of the model, consider adding an extra field in the reports contents instead of in the footer (that way, we can add a bit more context about the hash, and explain it's role in the process).
> 
> @jschuurk-kr @arjen-ba @ring-ring-ring what do you think?

Nice idea for the next iteration of the models. It makes sense to have it on the pages with data that also goes into the EML. @arjen-ba what would be your preference?

jorisleker:
> It makes sense to have it on the pages with data that also goes into the EML

Why? Does not really add additional security?

praseodym:
Can we move the requirements discussion to a separate issue so that we can unblock this PR? We can always look at this again once we have built the other required features for the PS/WS elections.

Lionqueen94 (changes requested):
Can you remove one of the extra signature fields from the Na 14-2 that now moves to the next page?
Image
Same for the Na 31-2:
Image

Otherwise it looks fine to me

praseodym:
@sanderboom there is no need to merge `main` into this branch unless there are merge conflicts that need to be resolved

sanderboom:
> Can you remove one of the extra signature fields from the Na 14-2 that now moves to the next page? 

Good point. Fixed in https://github.com/kiesraad/abacus/pull/3731/commits/4a43d8f5be1b4aad3c429610e696360d8ccb8f12

ring-ring-ring:
> Can we move the requirements discussion to a separate issue so that we can unblock this PR? We can always look at this again once we have built the other required features for the PS/WS elections.

I agree, it is a good idea to discuss further but for after PS/WS. I made an issue for it: https://github.com/kiesraad/abacus/issues/3737

Lionqueen94 (dismissed):

sanderboom:
#### Footer flip order
> can we flip the order? So the model name / description is always at the bottom, and the hash + timestamp are above that, and

Implemented in https://github.com/kiesraad/abacus/pull/3731/commits/4002f4a82caf260d2b5a405827f947297a6f9f9c @jorisleker 

#### Remove timezone offset 
Remove timezone offset from default date time format in model templates (and by that from the date time in the footer), implemented in: https://github.com/kiesraad/abacus/pull/3731/commits/2d37d975ddfb5a986d364ce09d62e1dac0d7b83a

Lionqueen94 (dismissed):
Looks good, extra "extra handtekeningen" compared to the model cannot hurt IMO

sanderboom:
Because of the footer change, some sections for 'Ondertekening door andere aanwezige leden van het stembureau' looked like this (title/description and first signature block on different pages):

image

In these cases an extra pagebreak is now inserted. 

Also made sure the last page fills out with signature boxes.

Not yet implemented for P 22-2, as this will receive more edits that will shift the layout - [added a reminder in the issue](https://github.com/kiesraad/abacus/issues/3540#issuecomment-5178685741).

Lionqueen94 (dismissed):
Looks good to me

sanderboom:
Problem for Na 31-2:

The combination of chapter title (Stemmen per lijst en per kandidaat), 50+ candidates and the FatFooter™️ resulted in "De lijst gaat verder op de volgende pagina" to be displayed **on** that _volgende pagina_.

It was decided to place the chapter title on a separate page which fixes the issue. Commit:  e1c15460fb22a00f8e7d5068d0ea44aad7adc1dd

Lionqueen94 (approved):

jorisleker (approved):

PR #3732, opened 2026-07-30
Show correction warning and clear fields

### LINKED ISSUE #3657: Show data entry warnings and empty fields

oliver3:
## What & why

Resolves https://github.com/kiesraad/abacus/issues/3657

When a typist starts correcting differences in their data entry, a warning should be shown per section with differences, and the fields that were different should be emptied. See [Figma design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=21910-10295&t=zsI6gOwSEsHto50q-1).

## How to test

With two different typists, create differences between the two data entries. As a coordinater, resolve the differences by giving one entry back to the typist. See that that typist can redo their entry and that the warnings are shown and fields are emptied.

## Reviewer notes
I initially started in separate commits, but subsequent improvements were hard to interactive rebase, so I ended up with one commit unfortunately.

- In the claim endpoint, add `correction_warnings` for either of the correction states, which is a list of fields that are different
  - New method `compare_entries()` for `DataEntryStatus` that compares the two entries for the states that have them
- On first claim (no client_state), add `correctionWarning` and reset field values for the correct sections
  - Functions `addCorrectionWarnings()` and `resetFieldValues()` added
  - Function `resetValueAtPath()` added in `dataEntryMapping.ts`
  - Update test data in `dataEntryMapping.test.ts` to have more useful paths
- Store and retrieve the `correctionWarnings` in the `ClientState` when saving and claiming again
- Add W002 validation and translations
- Show only the section `correctionWarning` if present, else show the normal errors and warnings and errorsAndWarningsAccepted

And also:
- Fix flaky test `BackupsPage.test.tsx` (was happening only locally?)

### DIFF
--- a/backend/src/domain/validate.rs
+++ b/backend/src/domain/validate.rs
@@ -114,6 +114,8 @@ pub enum ValidationResultCode {
 
     /// GSB CSO, GSB DSO, CSB: (Bij tweede invoer) Niet alle ingevoerde waardes van de tweede invoer zijn gelijk aan die van de eerste invoer
     W001,
+    /// GSB CSO, GSB DSO, CSB: (Bij verschil met andere invoer) Een coördinator heeft beide invoeren vergeleken, en aangegeven dat in deze invoer fouten zijn gemaakt
+    W002,
     /// GSB CSO, GSB DSO: 'Aantal kiezers en stemmen': Aantal blanco stemmen is groter dan of gelijk aan 3% van het totaal aantal uitgebrachte stemmen
     W201,
     /// GSB CSO, GSB DSO: 'Aantal kiezers en stemmen': Aantal ongeldige stemmen is groter dan of gelijk aan 3% van het totaal aantal uitgebrachte stemmen
--- a/backend/src/domain/validate.rs
+++ b/backend/src/domain/validate.rs
@@ -115,6 +115,7 @@ pub enum ValidationResultCode {
     /// GSB CSO, GSB DSO, CSB: (Bij tweede invoer) Niet alle ingevoerde waardes van de tweede invoer zijn gelijk aan die van de eerste invoer
     W001,
     /// GSB CSO, GSB DSO, CSB: (Bij verschil met andere invoer) Een coördinator heeft beide invoeren vergeleken, en aangegeven dat in deze invoer fouten zijn gemaakt
+    /// This warning is triggered in the frontend, not in the backend.
     W002,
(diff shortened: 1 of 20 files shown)


jorisleker (dismissed):
Tested by creating entries with differences and fixing one. All fields with differences were cleared and numeric fields were marked 👌

frontend/src/features/data_entry/utils/reducer.ts
```diff
@@ -43,6 +45,8 @@ export default function dataEntryReducer(state: DataEntryState, action: DataEntr
 
       let formState: FormState;
       let targetFormSectionId: FormSectionId | undefined;
+      const results = action.dataEntry.data;
```
praseodym: I think this should clone the `action.dataEntry.data` to avoid mutating it in place

oliver3: Done in https://github.com/kiesraad/abacus/pull/3732/commits/fcbb0a0d60b43d7a2073efefdfb0bada213bff1b

backend/src/domain/validate.rs
```diff
@@ -114,6 +114,8 @@ pub enum ValidationResultCode {
 
     /// GSB CSO, GSB DSO, CSB: (Bij tweede invoer) Niet alle ingevoerde waardes van de tweede invoer zijn gelijk aan die van de eerste invoer
     W001,
+    /// GSB CSO, GSB DSO, CSB: (Bij verschil met andere invoer) Een coördinator heeft beide invoeren vergeleken, en aangegeven dat in deze invoer fouten zijn gemaakt
+    W002,
```
praseodym, line 119: Maybe add a small doc comment that W.002 is triggered in the frontend?

praseodym (dismissed):
Looks good!

ring-ring-ring:
I think it is really cool. Big improvement <3

backend/src/domain/validate.rs
praseodym: Done in b280fde72dc7ed6ba3e17077c049115614e03060

praseodym:
Resolved merge conflict with #3763 in e9f4d9f28f7433192429aab03cc1b46cd2584f21.

praseodym (approved):

stacktraceghost (approved):
Code looks good. Tested according to description and it works as described.

PR #3736, opened 2026-07-31
Extend election_fixture with election category

stacktraceghost:
## What & why

Election category is becoming relevant for tests, so this PR extends `election_fixture` with the parameter `election_category` for usage in tests. 

## How to test

- Review code
- See if tests succeed

## Reviewer notes

N/A

### DIFF
--- a/backend/src/domain/election.rs
+++ b/backend/src/domain/election.rs
@@ -213,6 +213,8 @@ impl ElectionCategory {
         }
     }
 
+    /// Get the sub category for test elections, only available for tests and test data generation
+    #[cfg(any(test, feature = "dev-database"))]
     pub fn sub_category(&self, number_of_seats: u32) -> ElectionSubCategory {
         match self {
             ElectionCategory::Municipal => {
--- a/backend/src/domain/election.rs
+++ b/backend/src/domain/election.rs
@@ -409,6 +409,7 @@ pub(crate) mod tests {
     /// The number of political groups is the length of the `political_groups_candidates` slice.
     /// The number of candidates in each political group is equal to the value in the slice at that index.
     pub fn election_fixture_with_given_number_of_seats(
+        election_category: ElectionCategory,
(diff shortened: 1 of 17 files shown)


backend/src/domain/election.rs
Lionqueen94: The sub_category should now also be based on the given election category, you can use something similar to the `get_election_sub_category` function created for the generator

Lionqueen94 (changes requested):

backend/src/domain/election.rs
stacktraceghost: Thanks, missed that! I've moved `get_election_sub_category` to `impl ElectionCategory` as `sub_category` and applied it wherever we still had such a ternary.  https://github.com/kiesraad/abacus/pull/3736/commits/aeadad5f010148bbe30d7499b12199468b72c937

backend/src/domain/election.rs
```diff
@@ -212,6 +212,27 @@ impl ElectionCategory {
             ElectionCategory::WaterAuthority => "AB",
         }
     }
+
+    pub fn sub_category(&self, number_of_seats: u32) -> ElectionSubCategory {
```
Lionqueen94, line 218: I suggest giving this a test macro, since for the election import we need to base the Provincial sub category on the number of electoral districts and do not want to default to PS1 outside of test (data generator) code. The other code is as it should be (for Municipal and WaterAuthority)

stacktraceghost: Done in https://github.com/kiesraad/abacus/pull/3736/commits/e79cb4ee03c2d0e34f7e54f333c7a6ea2fe55f39. Since `test_data_gen` uses feature `dev-database`, I had to use any.

Lionqueen94 (approved):

praseodym (approved):

cikzh (approved):

PR #3739, opened 2026-07-31
Resolving differences for first entry with errors

### LINKED ISSUE #3720: Make sure that we cannot introduce errors when correcting differences
I think through this branch the data entry can get to a `EntriesDifferent` state with the first entry containing errors, which was previously not possible (only the second entry could contain errors, a first entry with errors would have been handled right after the first entry was finalised).

_Originally posted by @praseodym in https://github.com/kiesraad/abacus/pull/3713#discussion_r3677537090_

praseodym:
## What & why

Also special case for resolving differences for first entry with errors: only discarding the second entry should be allowed. Similar to #3706. Resolves #3720.

Screenshot:

## How to test

Create a data entry with differences and errors in the first entry: first first+second data entries with differences, resolve by correcting the first entry, then introducing errors in the first entry. This should lead the data entry through these state transitions: `Empty` -> `FirstEntryInProgress` -> `FirstEntryFinalised` -> `SecondEntryInProgress` -> `EntriesDifferent` -> `FirstEntryCorrection` -> introduce errors in first entry -> `EntriesDifferent` with errors in first entry.

### DIFF
--- a/backend/fuzz/fuzz_targets/data_entry_status.rs
+++ b/backend/fuzz/fuzz_targets/data_entry_status.rs
@@ -230,7 +230,11 @@ fn is_as_expected(
         }
         // KeepFirstEntry
         (DataEntryStatus::EntriesDifferent(_), Transition::KeepFirstEntry) => {
-            matches!(resulting_state, Ok(DataEntryStatus::FirstEntryFinalised(_)))
+            if first_entry_correct {
+                matches!(resulting_state, Ok(DataEntryStatus::FirstEntryFinalised(_)))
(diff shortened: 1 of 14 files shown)


jorisleker (approved):
Tested on abacus-test by creating conflicting entries, choosing the second one as correct, and then introducing an error in the first one. 

- The resolve differences page correctly states that the second data entry will be deleted if the first one is selected as the correct one.
- After resolving differences I am correctly redirected to resolve errors

oliver3 (approved):

PR #3751, opened 2026-08-03
Add DSO to election create

### LINKED ISSUE #3694: Add DSO to Election creation
Make DSO enabled and use in backend

Lionqueen94:
Closes #3694 

## What & why

- Make DSO in election create not disabled
- Process selected counting method
- Add test
- Remove (now) unused i18n (that was somehow even duplicated)
- Set no default for counting method and added validation

## How to test

- Import a test election
- Do not select a counting method and check the error message
- Select DSO
- Verify that the election is saved with counting method DSO

### DIFF
--- a/frontend/e2e-tests/page-objects/election/create/CountingMethodTypePgObj.ts
+++ b/frontend/e2e-tests/page-objects/election/create/CountingMethodTypePgObj.ts
@@ -3,11 +3,13 @@ import type { Locator, Page } from "@playwright/test";
 export class CountingMethodTypePgObj {
   readonly header: Locator;
   readonly cso: Locator;
+  readonly dso: Locator;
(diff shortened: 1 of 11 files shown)


praseodym (dismissed):
Looks good! Not sure if we should select CSO by default but it does avoid some validation logic and error message.

jorisleker:
> Looks good! Not sure if we should select CSO by default but it does avoid some validation logic and error message.

As there is no way to undo this choice, and no way to delete an election that was created with CSO selected by accident, I would make this an explicit choice for now, so nothing selecter by default.

praseodym (approved):
Looks good, no option is selected by default and an error message appears when continuing without selection

cikzh (approved):

PR #3754, opened 2026-08-03
Build Windows installer in weekly e2e workflow

praseodym:
## What & why

Build Windows installer in weekly e2e workflow so that we always have a recent Windows installer build available, like when @marjoleintamis wanted a new build for documentation work earlier today. I chose to extend the weekly e2e workflow because it already builds for Windows and runs the testsuite so we can be  sure that the build works like expected.

## How to test

[CI passes](https://github.com/kiesraad/abacus/actions/runs/30849099225)

### DIFF
--- a/.github/workflows/weekly-e2e-tests.yml
+++ b/.github/workflows/weekly-e2e-tests.yml
@@ -87,6 +87,40 @@ jobs:
         with:
           name: abacus-${{ matrix.target.os }}-weekly
           path: ${{ matrix.target.binary }}
+      - name: Build with airgap detection and TLS (for the installer)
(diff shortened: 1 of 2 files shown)


sanderboom (approved):

jschuurk-kr:
> I chose to extend the weekly e2e workflow because it already builds for Windows and runs the testsuite so we can be sure that the build works like expected.

Arguably we should then build the installer only after the testsuite passes, but you can't get to the download url without seeing the status of the tests, so am fine with the current setup.

jschuurk-kr (approved):

PR #3763, opened 2026-08-04
Implement data entry correction discard

### LINKED ISSUE #3658: Implement data entry correction discard and reset

praseodym:
## What & why

Implement data entry correction discard. When discarding a first data entry correction the second data entry becomes the new first entry. When discarding a second data entry correction, the first entry stays the first entry. Both transitions go to the `FirstEntryFinalised` state. Also see the [data entry state diagram](https://github.com/kiesraad/abacus/blob/d10737654569fa9facd993701b841edbd40b1e78/documentatie/state-machines/data-entry-state.md).

Resolves #3658, because the reset transition already works. 

## How to test

Create a data entry with differences, resolve by choosing correction, test discarding as typist and check new state.


oliver3 (approved):
Looks good and seems to be working!
Thanks for also adding the reset api tests.

jorisleker (approved):
Tested as described, works as expected 👌

PR #3782, opened 2026-08-05
Implement data entry disabled_when logic for DSO "Controles en correcties"

### LINKED ISSUE #3746: DSO first entry skip checks and corrections if there are none
During data entry, we should skip the second section if the page is missing. See [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=9942-34955&m=dev)

stacktraceghost:
## What & why

Resolves #3746 

- Added `disabled_when` to `DataEntrySection`, which takes `path` and `equal_to` as parameters
- Changed `dataEntryUtils`
   - Added `isSectionDisabled`: checks if section should be disabled based on `disabled_when` property
   - Added `updateDisabledSections`: updates `isDisabled` property in the formState
   - Added `resetDisabledSectionValues`: resets values of sections that are disabled
   - Adjusted `getNextSectionID` to find the next section which is not disabled
- `onSubmitForm` calls `resetDisabledSectionValues` to reset disabled sections on submit
- Reducer changed:
  - `DATA_ENTRY_CLAIMED`: call `resetDisabledSectionValues` on `results`
  - `DATA_ENTRY_CLAIMED`: initializes the initial form state instead of within `buildFormState`, so we can directly call `updateDisabledSections` instead on a deeper level
  - `FORM_SAVED`: calls `updateDisabledSections`
  - Renamed `buildFormState`/`buildCorrectionFormState` to `restoreFormState`/`restoreCorrectionFormState`
- Added DSO mockdata to `mock-data.ts` and extended typing for `test.utils.ts`
- Adjusted `DataEntryProgress` to support disabled sections.
- Added/adjusted tests

## What to test

In DSO, the section "Controles en correcties" will be skipped if you answer "Nee, de pagina ontbreekt" on the first section and go to next section, and will show greyed-out in the progress sidemenu with icon "empty". If you navigate to the section when it is disabled, all input fields should be disabled.

- Check if the section is disabled when it should, also test hotkeys
  - Test if the sidemenu behaves as expected
  - Test what happens when you refresh the page
- Test if the section "Controles and correcties" gets reset if you choose "Nee, de pagina ontbreekt" after filling in the section.
  - Check if the data posted to the API is also reset in time (should reset right after submit on the first page)

And test in other possible ways to see if it behaves as expected. 

## How to test

- Test either in Abacus test or locally

## Reviewer notes

- Validations will be done in https://github.com/kiesraad/abacus/issues/3687


praseodym:
Managed to get into an invalid state:
1. Check Ja / Ja on "Over het proces-verbaal" page and click next 
2. Check some boxes on the "Controles en correcties" page
3. Go back to "Over het proces-verbaal" with left-hand navigation or browser back button
4. Check No for the second question
5. Go to the "Controles en correcties" page with left-hand navigation or browser forward button, click "Wijzigingen opslaan"
6. "Controles en correcties" page is shown, but "Aantal kiezers en stemmen" page was expected

Screenshot:

image

praseodym:
> Test if the section "Controles and correcties" gets reset if you choose "Nee, de pagina ontbreekt" after filling in the section.
> 
>     * Check if the data posted to the API is also reset in time (should reset right after submit on the first page)

Should we make `checks_and_corrections` an `Option<ChecksAndCorrections>` so that we can omit it entirely when not filled out?

oliver3:
> Should we make checks_and_corrections an Option<ChecksAndCorrections> so that we can omit it entirely when not filled out?

Currently the frontend expects the DataEntryResults object to contain the whole tree of sections, subsections and fields and fails if it cannot update a value in this structure. I think it will be quite some effort to be able to remove and add a section from that structure at will, so I am not sure if this is worth it.

oliver3:
> Managed to get into an invalid state:

Implemented solution:

- Do not prevent navigation but show the section as disabled (done in https://github.com/kiesraad/abacus/pull/3782/commits/0bc20995daf19c2edb893fefbdbcdcfc466cf50b)
- Reset cached values when its corresponding section has become disabled (done in https://github.com/kiesraad/abacus/pull/3782/commits/c38251b070fa186c21b105fd86ac17336809e129)
- Reset `hasChanges` flag to prevent a new modal when navigating away from the disabled section where you previously made some changes (done in https://github.com/kiesraad/abacus/pull/3782/commits/cb41fcd187676dfb1bb7a7fe5880b486279bfb88)

I believe that this covers it now.

oliver3 (dismissed):

jorisleker (dismissed):
Created a DSO DataEntry, filled it in, and toggled between with and without "controles en correcties" several times. 
- menu item is disabled in nav when appropriate
- checkboxes are disabled and reset if nav item is disabled

praseodym (dismissed):

oliver3 (approved):

praseodym (approved):

Issue #3381, opened 2026-06-05
Epic: streamline resolving data entry differences


praseodym:
_Epic refinement: 23 July 2026 13:00-14:00_

# Context

Streamline resolving data entry differences, avoiding a full data re-entry for a single difference.

The requested solution is to let the coordinator decide whether the invalid data entry should be deleted or corrected. If correction is chosen, the conflicting fields in the data entry are cleared and the same typist has to correct that data entry. 

The sections with conflicting fields are marked with a warning. The typist that makes the correction gets to see all sections but is initially shown the first section (not the section with conflicting fields) for simplicity in implementation., e.g. errors for total count are not shown if there are conflicting fields on a section.

See [coordinator resolve page](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=21130-78594&t=WWL1s4GgD7c8c6rg-4) and [warning copy](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=21910-10302&t=WWL1s4GgD7c8c6rg-4) in Figma.

# Architecture

Refined:
- Extra states in state machine for correction (`FirstEntryCorrection` and `SecondEntryCorrection` OR `Correction` with extra 1/2 entry field)
- Backend returns field paths that conflicting when claiming data entry in correction state
- `ClientState` is extended with saved form section ids (array), which keep tracks of which sections have been saved
- Frontend renders warning message on sections with conflicting fields and clears those conflicting fields during the initial load from backend. Conflicting fields are marked with a warning icon.
  - Frontend does not render the conflict warning message and does not clear fields when that section has already been saved (i.e. section is included in saved form section ids in client state).
  - The conflicting fields warning takes precedent over all other errors/warnings on the section.

# Scope

PR #3793, opened 2026-08-06
Add data entry correction e2e tests

### LINKED ISSUE #3659: Review end to end tests
Tests are probably updated already in earlier issues, check for consistency

oliver3:
## What & why

Add end-to-end tests to check data entry correction, resolves https://github.com/kiesraad/abacus/issues/3659

## How to test

- See that the tests are green in the CI.
- Think about if the e2e test suite is sufficient for the [Epic streamline resolve differences](https://github.com/kiesraad/abacus/issues/3381).

## Reviewer notes

First commit moves existing discard tests to a new file.


frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
```diff
@@ -0,0 +1,247 @@
...
+import { DifferencesPage } from "e2e-tests/page-objects/data_entry/DifferencesPgObj";
+import { VotersAndVotesPage } from "e2e-tests/page-objects/data_entry/VotersAndVotesPgObj";
+import { ElectionStatus } from "e2e-tests/page-objects/election/ElectionStatusPgObj";
+import { dataEntryRequest, dataEntryWithDifferencesRequest } from "e2e-tests/test-data/request-response-templates";
+import type { Results } from "@/types/generated/openapi";
+import { type DataEntry, test } from "../../fixtures";
+
+test.use({
+  storageState: "e2e-tests/state/coordinator1-GSB.json",
+});
```
jschuurk-kr: I think the tests will read more easily, for example when calling `testCorrection()`, if you use the coordinator fixture instead of this storageState. (It's also on my todo-list to replace all storageStates with fixtures, so we have one consistent way of working with different users.)

frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    const abortInputModal = new AbortInputModal(typistPage);
+    await expect(abortInputModal.heading).toBeVisible();
+    await abortInputModal.discardInput.click();
+
+    const dataEntryHomePage = new DataEntryHomePage(typistPage);
+    await expect(dataEntryHomePage.fieldset).toBeVisible();
+
+    // the finalised first entry is kept, so a new second entry can be started
+    await page.goto(`/elections/${dataEntry.election_id}/status`);
```
jschuurk-kr: Alternatively you could do a [`page.reload()`](https://playwright.dev/docs/api/class-page#page-reload), because the coordinator is still on this page. Not sure if that's better because then you need to look higher in the test to see what page the coordinator was on.

(Same comment applies to the last `goto()` in "typist discards a first entry correction".)

frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
oliver3: Updated in https://github.com/kiesraad/abacus/pull/3793/commits/54812e2fe6dbcb71dfc59943858c9c0391b5f387 and also renamed `typistPage` to `typist`.

frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
oliver3: I have no preference for either :)

frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
jschuurk-kr: I was expecting this change would allow us to get rid of
```
test.use({
  storageState: "e2e-tests/state/coordinator1-GSB.json",
});
```

jschuurk-kr: There's still a `page` on line 141 in "first entry correction differences remain".

oliver3: Oops thanks, removed remaining page in https://github.com/kiesraad/abacus/pull/3793/commits/dfe48748389acfc80a5e5d24e6913c9085b8f50f and use storageState in https://github.com/kiesraad/abacus/pull/3793/commits/22e1f98769c626afec3bc24fa5d7474a2a3a0ce1

frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
```diff
@@ -0,0 +1,245 @@
...
+      await expect(checkAndSavePage.fieldset).toBeVisible();
+      await checkAndSavePage.save.click();
+
+      // Assert message (typist) and status (coordinator)
+      const dataEntryHomePage = new DataEntryHomePage(typist);
+      await coordinator.goto(`/elections/${dataEntry.election_id}/status`);
+      const electionStatusPage = new ElectionStatus(coordinator);
+
+      if (expectResolved) {
+        await expect(dataEntryHomePage.alertDataEntrySaved).toBeVisible();
```
praseodym: There could be a race condition here where the election status page is (re)load before the typist's data entry page is saved, because we don't wait for the data entry saved confirmation before loading the election status page.

praseodym: `dataEntryHomePage.alertDataEntrySaved` matches both first and second data entry, do we want to be more specific?

praseodym (dismissed):

frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
oliver3: Split up typist and coordinator assertions in https://github.com/kiesraad/abacus/pull/3793/commits/51db339e47e0e65a1f45de329e5eae0d3d89269d

oliver3: By being more specific I found that the text "Let op: verschil met eerste invoer" (`alertDataEntryDifferent`) was always used. Updated test and fixed notification text in https://github.com/kiesraad/abacus/pull/3793/commits/51db339e47e0e65a1f45de329e5eae0d3d89269d

praseodym: Good find and thanks for fixing!

praseodym (approved):

Lionqueen94 (approved):
Tests are comprehensive and work well 👍 and nice bugfix on the copy!

PR #3798, opened 2026-08-06
Backend adjust claim to return DSO model

### LINKED ISSUE #3728: Backend adjust claim to return DSO model
Endpoint `data_entry_claim` calls `Results::new`, which does not return DSO models yet. Also update `get_previous_results` (and make the match exhaustive, don't use `_ => None`)

There might be more adjustments to be made in the codebase, please research.

Add/expand tests where necessary.

Lionqueen94:
Closes #3728 

## What & why

- Added `DSOFirstSession` and `DSONextSession` to `Results::new` and `get_previous_results` (the latter is used in `claim`)
- Implemented `PollingstationResults` for `DSOFirstSessionResults`
- Added DSO election fixture without results
- Added DSO election fixture with results
- Added test for claim in first session and claim in next session
- Added tests for all cases of `Results::new`

## How to test

- Verify the added tests
- Claim a data entry for a DSO election

### DIFF
--- a/backend/src/api/data_entry.rs
+++ b/backend/src/api/data_entry.rs
@@ -304,7 +304,7 @@ async fn get_previous_results(
                 Results::DSOFirstSession(r) => Some(r.as_common()),
                 Results::CSONextSession(r) => Some(r.as_common()),
                 Results::DSONextSession(r) => Some(r.as_common()),
-                _ => None,
+                Results::GSB(_) => None,
             },
             _ => None,
         },
--- a/backend/src/api/data_entry.rs
+++ b/backend/src/api/data_entry.rs
@@ -1383,7 +1383,6 @@ mod tests {
         assert_eq!(status, DataEntryStatusName::FirstEntryInProgress);
     }
 
-    // TODO: This will work once #3687 is implemented
     #[test(sqlx::test(fixtures(path = "../../fixtures", scripts("election_11_dso"))))]
     async fn test_claim_data_entry_dso_ok(pool: SqlitePool) {
         let data_entry_id = DataEntryId::from(1101);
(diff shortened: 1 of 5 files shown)


backend/src/api/data_entry.rs
```diff
@@ -301,7 +301,9 @@ async fn get_previous_results(
             // Match polling station results only and get the common results
             DataEntryStatus::Definitive(d) => match d.results {
                 Results::CSOFirstSession(r) => Some(r.as_common()),
+                Results::DSOFirstSession(r) => Some(r.as_common()),
                 Results::CSONextSession(r) => Some(r.as_common()),
+                Results::DSONextSession(r) => Some(r.as_common()),
                 _ => None,
```
oliver3: Let's make match statements exhaustive (no wildcard `_ =>`) if we are already defining most variants, so that we do not forget new variants when they are added.

```suggestion
                Results::GSB(_) => None,
```

backend/src/domain/results/mod.rs
```diff
@@ -87,7 +89,11 @@ impl Results {
             committee_session.is_next_session(),
         ) {
             (CommitteeCategory::GSB, false) => {
-                Results::CSOFirstSession(CSOFirstSessionResults::empty(election))
+                if election.counting_method == Some(VoteCountingMethod::CSO) {
+                    Results::CSOFirstSession(CSOFirstSessionResults::empty(election))
+                } else {
+                    Results::DSOFirstSession(DSOFirstSessionResults::empty(election))
+                }
```
oliver3: Unfortunately the `ElectionWithPoliticalGroups` can be invalid with category `GSB` and counting_method `None`. I think it is better to panic! in this case instead of letting it become DSO.

Same for the if-else control flows below.

oliver3 (changes requested):

backend/src/api/data_entry.rs
Lionqueen94: Changed in 44987a01e1d8a41ca3290cd17220e68379968c42

backend/src/domain/results/mod.rs
Lionqueen94: Good one, changed and tested in 44987a01e1d8a41ca3290cd17220e68379968c42

oliver3 (dismissed):

oliver3 (approved):

praseodym (approved):

PR #3802, opened 2026-08-07
Resolve differences save section

### LINKED ISSUE #3780: Data entry "Controleren en opslaan" does not show W.002 warning
The W.002 warning is not considered on the save ("Controleren en opslaan") section of data entry. This can lead to confusing combinations where an error is masked because of the W.002 in the navigation, but the save screen does show that error.

One other question is should the typist be able to finalise the data entry without saving all sections with differences first?

Image

oliver3:
## What & why

Resolves https://github.com/kiesraad/abacus/issues/3780

When there is some section with a correction warning:
- Show different check and save section
- Show in the progress list only correction warnings and "you are here", no other progress status icons

## How to test

Create a difference with two typists, return one data entry for correction as a coordinator, see as that typist that it works as described above.

## Reviewer notes

To avoid changing the logic in the existing check and save form, I've wrapped it in a new CheckAndSaveSection and moved the functionality of showing errors and navigating away to that section (including one test).

### DIFF
--- a/frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
+++ b/frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
@@ -75,6 +75,7 @@ test.describe("data entry - correct differences", () => {
     // Save section with warning
     await dataEntryPage.progressList.votersAndVotes.click();
     const votersAndVotesPage = new VotersAndVotesPage(typist);
+    await expect(votersAndVotesPage.fieldset).toBeVisible();
     await votersAndVotesPage.next.click();
 
     // All progress list icons are shown again
--- a/frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
+++ b/frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
@@ -19,18 +19,8 @@ test.describe("data entry - correct differences", () => {
     const typist = typistOneGSB.page;
     await typist.goto(`/elections/${dataEntry.election_id}/data-entry/${dataEntry.id}/1`);
 
-    // Assert progress icons
-    const dataEntryPage = new DataEntryBasePage(typist);
-    await expect(dataEntryPage.progressList.extraInvestigationIcon).toHaveAccessibleName("je bent hier");
-    await expect(dataEntryPage.progressList.countingDifferencesPollingStationIcon).toHaveAccessibleName("opgeslagen");
-    await expect(dataEntryPage.progressList.votersAndVotesIcon).toHaveAccessibleName("bevat een waarschuwing");
-    await expect(dataEntryPage.progressList.differencesIcon).toHaveAccessibleName("opgeslagen");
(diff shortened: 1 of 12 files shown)


praseodym:
Rebased onto `main`

frontend/src/features/data_entry/components/check_and_save/CheckAndSaveSection.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Rendered when some section has a correction warning.
+ */
+function CheckAndSaveCorrectWarnings({ sections }: { sections: CorrectWarningSection[] }) {
+  return (
+    <>
+      <legend>
+        <h2>{t("check_and_save.title")}</h2>
+      </legend>
```
praseodym: `<legend>` needs to have a parent `<fieldset>` but has an `<article>` directly above it now

frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
```diff
@@ -63,6 +53,41 @@ test.describe("data entry - correct differences", () => {
...
+    await expect(dataEntryPage.progressList.differencesIcon).toBeHidden();
+    await expect(dataEntryPage.progressList.listIcon(1)).toBeHidden();
+    await expect(dataEntryPage.progressList.listIcon(2)).toBeHidden();
+    await expect(dataEntryPage.progressList.listIcon(3)).toBeHidden();
+    await expect(dataEntryPage.progressList.checkAndSaveIcon).toBeHidden();
+
+    // Save section with warning
+    await dataEntryPage.progressList.votersAndVotes.click();
+    const votersAndVotesPage = new VotersAndVotesPage(typist);
+    await votersAndVotesPage.next.click();
```
praseodym, line 79: We should check that the votersAndVotesPage has loaded (e.g. `await expect(votersAndVotesPage.fieldset).toBeVisible();`) before clicking next, otherwise Playwright could click next on the initial page

praseodym (dismissed):
lgtm with two minor comments

frontend/src/features/data_entry/components/check_and_save/CheckAndSaveSection.tsx
oliver3: There was no styling on the legend so I've removed it in https://github.com/kiesraad/abacus/pull/3802/commits/397df6db4594b85f05018e5bb3756457b0d8a072

frontend/e2e-tests/tests/data-entry-GSB/data-entry.correct-differences.e2e.ts
oliver3: Did not realise that the locator could still be the previous next button, thanks! Updated in https://github.com/kiesraad/abacus/pull/3802/commits/46334fae8f49bf65662a443d8008754e8f83f47e

praseodym (approved):

jorisleker (commented):
Tested the steps described in the PR. 
- not able to save as long as there are non-submitted sections with cleared input
- able to save after all sections have been submitted

jorisleker (approved):

PR #3808, opened 2026-08-10
DSO Backend validations

### LINKED ISSUE #3687: Add backend validation
See [validatieregels](https://github.com/kiesraad/abacus-documentatie/blob/main/use-cases/validatieregels-plausibiliteitschecks-tellingen.md)

Also implement tests

| code  | regel | Done? | 
| ----- | ----- | :-----: |
| F.121 | Over het proces-verbaal: Niet alle vragen bij 'Over het proces-verbaal' zijn beantwoord                      |X| 
| F.122 | Over het proces-verbaal: Ongeldige combinatie van antwoorden: `wel corrigendum, geen inlegvel`               | X |
| F.131 | 'Controles en correcties': geen vinkjes bij de eerste twee vragen                                            |X|
| F.132 | 'Controles en correcties': Ongeldige set documenten (vraag 'gecorrigeerde telresultaten' = 'nee')            | X |
| F.133 | 'Controles en correcties': Ongeldig antwoord in eerste zitting (vraag 'op verzoek van het CSB' is ingevuld)  | X  |
| F.134 | 'Controles en correcties': meer dan 1 antwoord op vraag 'zijn er gecorrigeerde telresultaten'                | X  |
| F.135 | 'Controles en correcties - Op verzoek van het centraal stembureau': 'controles en correcties aanwezig' = 'ja' EN ongeldig antwoord in eerste zitting (vraag is ingevuld)                |X   |

Also check if values in `checks_and_corrections` are all false if `about_report.checks_and_corrections_present` equals `PageMissing`.

cikzh:
## What & why
Implements https://github.com/kiesraad/abacus/issues/3687

## How to test
- Run the backend tests (`cargo test` or run individual tests)
- Or test in the frontend by triggering the errors in the DSO data entry

## Reviewer notes
- Please do a close review of the unit tests added in this PR. Especially the `cases` of each unit test. They pretty much determine the correctness of the implementation. I'd love feedback on interesting extra cases to test. Commits welcome!
- Check the fields in ValidationResultMockData.ts


backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +81,338 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+                corrected_results_csb_request,
+            )?;
+            let has_f131 = result.errors.iter().any(|e| e == &f131);
+            assert_eq!(has_f131, expect_f131,);
+        }
+
+        Ok(())
+    }
+
+    /// GSB DSO | F.132: 'Controles en correcties': Ongeldige set documenten (vraag 'gecorrigeerde telresultaten' = 'nee')
```
praseodym: Unfortunately I think the logic for F.132 is specified incorrectly, because in the text it assumes "Er is een corrigendum" but the check "vraag 'gecorrigeerde telresultaten' = 'nee'" does not cover this. Right now answering the "gecorrigeerde telresultaten" question as "no" always gives a validation error, which I don't think is correct. CC @jorisleker

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +81,338 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+        let validation_results = checks_and_corrections.validate(
+            &election_fixture(ElectionCategory::Municipal, committee_category, &[]),
+            &"checks_and_corrections".into(),
+        )?;
+
+        assert_eq!(validation_results.warnings.len(), 0);
         Ok(validation_results)
     }
+
+    /// GSB DSO | F.131: 'Controles en correcties': geen vinkjes bij de eerste twee vragen
```
praseodym: I think F.131 should only be checked when the inlegvel is present.

frontend/src/testing/api-mocks/ValidationResultMockData.ts
```diff
@@ -21,6 +21,36 @@ export const validationResultMockData: ErrorWarningsMap<ValidationResultCode> =
...
+    fields: ["data.checks_and_corrections.corrected_results_csb_request"],
+    code: "F133",
+  },
+  F134: {
+    fields: [
+      "data.checks_and_corrections.corrected_results_own_initiative",
+      "data.checks_and_corrections.corrected_results_csb_request",
+    ],
+    code: "F134",
+  },
```
praseodym, line 47: The frontend mocks seem to highlight different fields than the backend

praseodym (changes requested):
Not a complete review yet but wanted to leave these comments for discussion. Also this PR doesn't yet resolve some TODOs that should be done for issue #3687:
```rs
> rg 3687
backend/src/domain/results/mod.rs
459:                // TODO: https://github.com/kiesraad/abacus/issues/3687

backend/src/domain/results/checks_and_corrections.rs
88:        // TODO: https://github.com/kiesraad/abacus/issues/3687

backend/src/domain/results/about_report.rs
64:        // TODO: https://github.com/kiesraad/abacus/issues/3687
```

frontend/src/testing/api-mocks/ValidationResultMockData.ts
cikzh: I misunderstood this field. Both backend and mock are wrong. @Lionqueen94 cleared it up, will change it

cikzh: [5b5142d](https://github.com/kiesraad/abacus/pull/3808/commits/5b5142d801e65b1e8f57c12f00cc06dc3aa251ce)

cikzh:
> Also this PR doesn't yet resolve some TODOs that should be done for issue #3687:
> ```rust
> rg 3687
> backend/src/domain/results/mod.rs
> 459:                // TODO: https://github.com/kiesraad/abacus/issues/3687
> ```

Implemented in [91df406](https://github.com/kiesraad/abacus/pull/3808/commits/91df406b80d640186602c4d6437cdff13ae0126f)

backend/src/domain/results/checks_and_corrections.rs
jorisleker: I think you're right @praseodym. 

Looking at [F.132](https://github.com/kiesraad/abacus-documentatie/blob/main/use-cases/validatieregels-plausibiliteitschecks-tellingen.md#f132-controles-en-correcties-ongeldige-set-documenten-vraag-gecorrigeerde-telresultaten--nee), it seems obvious that the text implies that the user indicated that a corrigendum is present. That refers to the question `Is er een corrigendum bij het papieren proces-verbaal aanwezig?` in the previous section. 

I think we should update the validation logic of F.132 to `Is er een corrigendum bij het papieren proces-verbaal aanwezig? = Ja AND 'gecorrigeerde telresultaten' = 'nee'`. We probably missed this under the assumption that the section `Controles en correcties` would only be shown when the corrigendum question was answered with 'yes', but this is incorrect.

Also: we have two 'zijn er gecorrigeerde telresultaten' questions. F.132 is not clear to which of those questions the validation refers. I think it should be that any of these questions was answered with 'yes'. 

@jschuurk-kr @Lionqueen94 do you agree?

stacktraceghost:
Now that https://github.com/kiesraad/abacus/pull/3782 was merged, I have updated this branch with main. I was curious how the validations would work in the following situation:

1. Choose page missing, submit and the "Controles and correcties" section will be skipped
2. Enter something in "Aantal kiezers en stemmen"
3. Go back and choose for page available, then submit

It will go to the "Controles and correcties" section without errors. If you would submit or refresh, you would see errors. However, you can also just continue to the next step in the menu and finish the rest of the data entry. In that case, you will not see any errors. The last page will say everything is fine. When you save however, you will get the alert "Let op: fouten in het proces-verbaal". 

The last page should always report any error that is received from the backend. And the "Controles and correcties" section should also show errors if the user is past that section. 

Since this is a frontend issue, lets fix this in a separate PR.

backend/src/domain/results/checks_and_corrections.rs
jschuurk-kr: I think we should handle the GSB and CSB sections of the inlegvel "Controles en correcties" for "Zijn er gecorrigeerde telresultaten?" very differently.

The only check we need for the CSB section is that neither checkboxes are checked, because they should never be checked in a first session. In a next session, this whole page is not shown to the typist, since the question about corrected results is answered by the coordinator in "Onderzoek naar telresultaten" of the Na 14-1 versie 2. See also in Figma "DSO - volgende zitting".

For the GSB section we need two checks:
1. `Is er een corrigendum bij het papieren proces-verbaal aanwezig? = Ja AND 'gecorrigeerde telresultaten' = 'nee'` (mentioned by Joris above, update to F.132)
2. `Is er een corrigendum bij het papieren proces-verbaal aanwezig? = Nee AND 'gecorrigeerde telresultaten' = 'ja'` (new check)

praseodym: The CSB data entry does not have any CSO/DSO destinction and skips all investigation checkbox sections, it starts directly with the "Aantal kiezers en stemmen page" ([see Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=17951-10710&m=dev)).

jschuurk-kr: @praseodym With GSB and CSB I'm referring to the sections "Op eigen initiatief van het gemeentelijk stembureau" and "Op verzoek van het centraal stembureau" of the inlegvel "Controles en correcties". I made a small edit to my original comment to clarify this.

Lionqueen94:
> Now that #3782 was merged, I have updated this branch with main. I was curious how the validations would work in the following situation:
> 
>     1. Choose page missing, submit and the "Controles and correcties" section will be skipped
> 
>     2. Enter something in "Aantal kiezers en stemmen"
> 
>     3. Go back and choose for page available, then submit
> 
> 
> It will go to the "Controles and correcties" section without errors. If you would submit or refresh, you would see errors. However, you can also just continue to the next step in the menu and finish the rest of the data entry. In that case, you will not see any errors. The last page will say everything is fine. When you save however, you will get the alert "Let op: fouten in het proces-verbaal".
> 
> The last page should always report any error that is received from the backend. And the "Controles and correcties" section should also show errors if the user is past that section.
> 
> Since this is a frontend issue, lets fix this in a separate PR.

I added this to https://github.com/kiesraad/abacus/issues/3757

backend/src/domain/results/checks_and_corrections.rs
jorisleker: @jschuurk-kr can you prepare a PR for the validation rules in the use cases to reflect this change as well?

jschuurk-kr: https://github.com/kiesraad/abacus-documentatie/pull/116

backend/src/domain/results/checks_and_corrections.rs
cikzh: Discussed with @Lionqueen94: [08e7baf](https://github.com/kiesraad/abacus/pull/3808/commits/08e7bafcd1268095ccbda6d6acc1e96c2662acdf) implements doing all F.13x checks when the inlegvel is present

cikzh: Resolving this to reduce noise. Will implement as documented in https://github.com/kiesraad/abacus-documentatie/pull/116

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +92,429 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+                });
+            }
+
+            if !self.corrected_results_csb_request.is_empty() {
+                validation_results.errors.push(ValidationResult {
+                    fields: vec![path.to_string()],
+                    code: ValidationResultCode::F133,
+                    context: None,
+                });
+            }
```
Lionqueen94: This is the check for F135 that is done below

Lionqueen94 (changes requested):
Still reviewing, but already commenting this one so you can already pick this up

backend/src/domain/results/checks_and_corrections.rs
Lionqueen94: The check for F133 should be done next to the check for F132 I believe (one level higher)

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +92,429 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+            assert_eq!(
+                has_f131, expect_f131,
+                "Failed: reason_investigation_own_initiative: {reason_investigation_own_initiative:?}, corrected_results_own_initiative: {corrected_results_own_initiative:?}, corrected_results_csb_request: {corrected_results_csb_request:?}"
+            );
+        }
+
+        Ok(())
+    }
+
+    /// GSB DSO | F.132: 'Controles en correcties - Op eigen initiatief': 'controles en correcties aanwezig' = 'ja' EN één of beide vragen niet beantwoord
```
Lionqueen94:
```suggestion
    /// GSB DSO | F.132: 'Controles en correcties - Op eigen initiatief': 'controles en correcties aanwezig' = 'ja' EN 'gecorrigeerde telresultaten' = 'nee' EN 'Over het proces-verbaal: Is er een corrigendum?' = 'ja'
```

backend/src/domain/results/checks_and_corrections.rs
cikzh: Fixed in [57ef864](https://github.com/kiesraad/abacus/pull/3808/commits/57ef864ab49003f8eb741bef9aadbdbe434b2757)

backend/src/domain/results/checks_and_corrections.rs
cikzh: I've changed the implementation of the errors as defined by https://github.com/kiesraad/abacus-documentatie/pull/116 See commits after [306489f](https://github.com/kiesraad/abacus/pull/3808/commits/306489f0094a753b1f4080ad162ce40993963c75)

backend/src/domain/results/mod.rs
```diff
@@ -448,17 +455,71 @@ impl Compare for Results {
...
+
+                    validation_results.join(
+                        results
+                            .checks_and_corrections
+                            .validate(election, &path.field("checks_and_corrections"))?,
+                    );
+                };
+
+                validation_results.join(results.as_common().validate(election, path)?);
+                Ok(validation_results)
```
praseodym: Why not move this into an `impl Validate for DSOFirstSessionResults`, so that this function ends up under the line limit again

Lionqueen94 (dismissed):
Tested all validation errors 👍

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +92,437 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+            ),
+        ) in cases.into_iter().enumerate()
+        {
+            let result = validate(
+                corrigendum_present,
+                reason_investigation_own_initiative.clone(),
+                corrected_results_own_initiative.clone(),
+                corrected_results_csb_request.clone(),
+            )?;
+            dbg!(&result.errors);
```
praseodym: Remove?

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +92,437 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+        )?;
+
+        // We allow one, because we expect W.204 to trigger with no political_group_votes.
+        assert_eq!(validation_results.warnings.len(), 1);
+        Ok(validation_results)
+    }
+
+    /// GSB DSO | F.131: 'Controles en correcties - Op eigen initiatief': 'controles en correcties aanwezig' = 'ja' EN één of beide vragen niet beantwoord
+    #[test]
+    #[allow(clippy::too_many_lines)]
```
praseodym: Can be removed (use expect instead of allow to get a warning when the annotation is no longer needed)

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +92,437 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+                "Case #{case_index} failed: reason_investigation_own_initiative: {reason_investigation_own_initiative:?}, corrected_results_own_initiative: {corrected_results_own_initiative:?}, corrected_results_csb_request: {corrected_results_csb_request:?}"
+            );
+        }
+
+        Ok(())
+    }
+
+    /// GSB DSO | F.132: 'Controles en correcties - Op eigen initiatief': 'controles en correcties aanwezig' = 'ja' EN 'gecorrigeerde telresultaten' = 'nee' EN 'Over het proces-verbaal: Is er een corrigendum?' = 'ja'
+    #[test]
+    #[allow(clippy::too_many_lines)]
```
praseodym: Can be removed

backend/src/domain/results/about_report.rs
```diff
@@ -57,12 +57,144 @@ impl Compare for AboutReport {
...
-        _election: &ElectionWithPoliticalGroups,
-        _path: &FieldPath,
+        election: &ElectionWithPoliticalGroups,
+        path: &FieldPath,
     ) -> Result<ValidationResults, DataError> {
-        let validation_results = ValidationResults::default();
-        // TODO: https://github.com/kiesraad/abacus/issues/3687
+        let mut validation_results = ValidationResults::default();
 
+        if election.committee_category == CommitteeCategory::GSB {
```
praseodym: Do we need this check, isn't `AboutReport` GSB-only?

praseodym (changes requested):
I think the Definitive data entry in `election_12_dso_with_results.sql` now contains errors, which is an impossible state during normal use.

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +92,437 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+            dso_first_session_results::DSOFirstSessionResults,
+            voters_counts::VotersCounts,
+            votes_counts::VotesCounts,
+            yes_no::YesNo,
+        },
+        validate::{DataError, ValidationResult, ValidationResultCode, ValidationResults},
+    };
+
+    fn validate(
+        corrigendum_present: Option<CorrigendumPresent>,
```
stacktraceghost: This file does not contain the logic to check if ` 'controles en correcties aanwezig' = 'ja'`, that logic is found in `mod.rs`. I don't think we have to test the entire results struct here, we could just test `ChecksAndCorrections`  itself.

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +92,437 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+            );
+        }
+
+        Ok(())
+    }
+
+    /// GSB DSO | F.132: 'Controles en correcties - Op eigen initiatief': 'controles en correcties aanwezig' = 'ja' EN 'gecorrigeerde telresultaten' = 'nee' EN 'Over het proces-verbaal: Is er een corrigendum?' = 'ja'
+    #[test]
+    #[allow(clippy::too_many_lines)]
+    fn test_f132() -> Result<(), DataError> {
```
stacktraceghost: This rule is not implemented in this file, I would expect it in `mod.rs`. Same for F133.

In `mod.rs` I see that F132 checks the value of `results.about_report.checks_and_corrections_present` (PagePresent), `results.about_report.corrigendum_present` (TwoDocuments) and `results.checks_and_corrections.corrected_results_own_initiative` (no). The test however, never checks variants of `results.about_report.checks_and_corrections_present`. Similarly for F133.

backend/src/domain/results/about_report.rs
cikzh: The intent of the check was indeed to ensure `AboutReport` is GSB-only, but I see now the distinction is already made in the `impl Results` match. Removed in [4de597a](https://github.com/kiesraad/abacus/pull/3808/commits/4de597adca5f347e33832c1ad7a63b06a6c9bcaf)

backend/src/domain/results/checks_and_corrections.rs
cikzh: The lint still triggers for me, but I've changed it to `expect` in https://github.com/kiesraad/abacus/pull/3808/commits/b9250ffec436ca2f669df5beef8abe670ce274a8

backend/src/domain/results/checks_and_corrections.rs
cikzh: Same here, lint still triggers, changed to `expect` in The lint still triggers for me, but I've changed it to `expect` in https://github.com/kiesraad/abacus/pull/3808/commits/b9250ffec436ca2f669df5beef8abe670ce274a8

backend/src/domain/results/checks_and_corrections.rs
cikzh: Removed in https://github.com/kiesraad/abacus/pull/3808/commits/b9250ffec436ca2f669df5beef8abe670ce274a8

backend/src/domain/results/mod.rs
cikzh: Good idea, moved in https://github.com/kiesraad/abacus/pull/3808/commits/b9250ffec436ca2f669df5beef8abe670ce274a8

backend/src/domain/results/checks_and_corrections.rs
cikzh: As discussed on mattermost, only constructing the ChecksAndCorrections struct again. [ae92cb1](https://github.com/kiesraad/abacus/pull/3808/commits/ae92cb14b60dc47952a215950dfa37f87cd0433f)

backend/src/domain/tabulation.rs
praseodym, line 1: I think this removed test coverage for the `checks.corrected_results_csb_request.as_bool() == Some(true)` branch of `DSOInvestigations::append_result`

cikzh: That is because the PR sets `checks.corrected_results_csb_request` to `None` for `ps2_result` on line 1137. If `checks.corrected_results_csb_request` is something other than `None`, F.135 will trigger. 
So either the check is unnecessary, or the F.135/F.131 definition or logic is flawed. Let's discuss on Mattermost

backend/src/domain/results/checks_and_corrections.rs
cikzh: Added checks in 1fede176e9b259a907dc2ad94771602df551707d

backend/src/domain/tabulation.rs
cikzh: As discussed on Mattermost, a PR will be created to remove the check

Lionqueen94 (dismissed):

Lionqueen94:
> I think the Definitive data entry in `election_12_dso_with_results.sql` now contains errors, which is an impossible state during normal use.

This one is fixed in 2387904bccae726ad448cd9c37b732289cfc0646 (corrected category and checkbox for corrected results)

stacktraceghost:
Please avoid rebasing after setting up the PR for review. I can't see the changes since my last review this way, which costs a lot of time since I need to review the whole file again.

Lionqueen94 (dismissed):

Lionqueen94 (dismissed):

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +92,269 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
     fn validate(
         &self,
-        _election: &ElectionWithPoliticalGroups,
-        _path: &FieldPath,
+        election: &ElectionWithPoliticalGroups,
+        path: &FieldPath,
+    ) -> Result<ValidationResults, DataError> {
+        let mut validation_results = ValidationResults::default();
+
+        if election.committee_category == CommitteeCategory::GSB {
```
stacktraceghost: Similar to [this comment](https://github.com/kiesraad/abacus/pull/3808/files#diff-278038b7b36dae28806c5bc86563542cd570f99b70a3ce48f7dd8d1112f3089f), this struct is GSB-only.

backend/src/domain/results/checks_and_corrections.rs
```diff
@@ -81,12 +92,269 @@ impl Compare for ReasonInvestigationOwnInitiative {
...
+            if (!self
```
stacktraceghost: If a validation rule spans multiple structs, we usually put it one level higher (like F401). I would expect `F131/F134/F135` in `mod.rs`, because all those rules are also dependent on `'controles en correcties aanwezig' = 'ja'` which is `about_report.checks_and_corrections_present`. 

That way, the entire logic of the rule is in one place. At this moment it has a dependency on the if-statement in `mod.rs` and the tests do not cover the whole rule.

stacktraceghost (commented):
Note that [F311](https://github.com/kiesraad/abacus-documentatie/blob/main/use-cases/validatieregels-plausibiliteitschecks-tellingen.md#f311-232-verklaring-voor-verschil--nee-en-vanwege-een-onverklaard-verschil-in-stap-controles-en-correcties-is-niet-aangevinkt) has not been implemented yet. According to the documentation, that should be implemented for GSB DSO?

cikzh:
Good catch about F.311, I had not spotted that one yet. I'll implement that one in a separate PR
Issue: https://github.com/kiesraad/abacus/issues/3870

backend/src/domain/results/checks_and_corrections.rs
cikzh: Done in [69b9af4](https://github.com/kiesraad/abacus/pull/3808/commits/69b9af45a5d0a80c4f894755de99d8668936a25e)

backend/src/domain/results/checks_and_corrections.rs
cikzh: Removed during the move in [69b9af4](https://github.com/kiesraad/abacus/pull/3808/commits/69b9af45a5d0a80c4f894755de99d8668936a25e)

backend/src/domain/results/checks_and_corrections.rs
cikzh: I must say that we're making a trade-off here. This change greatly increases the LoC in the `validate` function for `DSOFirstSessionResults`

stacktraceghost: The amount of lines in total are still the same, we have some other files of similar size. In my opinion it is better to keep the rules in one piece, so we can also test them fully in the same file.

The moved tests are currently missing variations for `ChecksAndCorrectionsPresent`.

cikzh: Testing for `ChecksAndCorrectionsPresent::PageMissing` always results in a `DataError` when `ChecksAndCorrections` is non-empty. That's why I added the `test_error_when_checks_and_validations_are_not_empty` in [9fdb835](https://github.com/kiesraad/abacus/pull/3808/commits/9fdb835736ce9a0723af5a7ca2e758f6ed6124cd) to cover the `ChecksAndCorrectionsPresent::PageMissing` path.

stacktraceghost (approved):

praseodym (approved):

PR #3810, opened 2026-08-10
Clean up EML test files

jschuurk-kr:
## What & why

The folder `backend/src/eml/tests` contains EML test files that are used by unit, integration, and/or e2e tests. And it contains a few EML test files that are no longer used, because the related tests moved to rust-eml-nl.

This PR addresses most of these issues. The changes made, matching the commits:
- remove unused EML test files due to introduction of rust-eml-nl
- move eml files only used in e2e tests to e2e test data folder
- split `eml110b_test.eml.xml` in backend and e2e version + fix e2e test because filename change

This leaves two files used by both the integration and e2e tests: `eml110a_test.eml.xml` and `eml230b_test.eml.xml`. I considered adding copies of these files to `frontend/e2e-tests/test-data/eml-files`, but didn't see a convincing reason to do so. There's also no meaningful way to split these files into a backend and e2e version, as was possible for `eml110b_test.eml.xml`.

## How to test

CI pipeline passes.

## Reviewer notes

Review per commit.


praseodym (approved):

cikzh (approved):
Lekker opruimen 👍 !

PR #3814, opened 2026-08-11
Let Biome exit with an error code on warnings

praseodym:
## What & why

Let Biome exit with an error code on warnings to prevent mistakes like a stale `console.log` (e.g. https://github.com/kiesraad/abacus/pull/3782/commits/30c2b16308a6c9a308e218ae472e02418d33c1f0). This gets logged by Biome as a warning, but Lefthook and CI don't fail on this.

## How to test

Try to commit a `console.log` and check that Lefthook prevents you from doing so.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -146,7 +146,7 @@ jobs:
       - name: Install dependencies
         run: pnpm install
       - name: Run Biome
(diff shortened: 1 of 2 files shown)


oliver3 (approved):
Works as advertised

rnijveld (approved):

PR #3815, opened 2026-08-11
Extend data entry fuzzer with correction states

### LINKED ISSUE #3642: Update data_entry_status fuzzer
With the extra states in the state machine

praseodym:
## What & why

Extend data entry fuzzer with the `FirstEntryCorrection` and `SecondEntryCorrection` correction states, including the claim, save, finalise and discard transitions. Resolves #3642.

Also:
- Replace the `correct_entry` bool with a three-variant `EntryValue` to support both entries with errors and entries with differences.
- Replace the `users` and `*_entry_correct` variables with a new `Model` struct.
- List every state explicitly in the catch-all match arm, so a new `DataEntryStatus` state fails to compile until it is added to the fuzzer.

## How to test

Run fuzzer for a while, should not give any errors.

### DIFF
--- a/backend/fuzz/fuzz_targets/data_entry_status.rs
+++ b/backend/fuzz/fuzz_targets/data_entry_status.rs
@@ -28,8 +28,13 @@ use libfuzzer_sys::{
     fuzz_target,
 };
 
-fn valid_result() -> Results {
-    Results::CSOFirstSession(CSOFirstSessionResults {
+/// A valid result without any votes
+fn valid_empty_result() -> Results {
+    Results::CSOFirstSession(valid_empty_cso_result())
+}
+
+fn valid_empty_cso_result() -> CSOFirstSessionResults {
+    CSOFirstSessionResults {
         extra_investigation: ExtraInvestigation {
             extra_investigation_other_reason: YesNo::default(),
             ballots_recounted_extra_investigation: YesNo::default(),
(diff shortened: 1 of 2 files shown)


backend/fuzz/fuzz_targets/data_entry_status.rs
```diff
@@ -316,122 +429,189 @@ fn is_as_expected(
...
+fn is_kept_as_expected(
+    resulting_state: &Result<DataEntryStatus, DataEntryTransitionError>,
+    entry: EntryValue,
+) -> bool {
+    if entry.has_errors() {
+        matches!(resulting_state, Ok(DataEntryStatus::FirstEntryHasErrors(_)))
+    } else {
+        matches!(resulting_state, Ok(DataEntryStatus::FirstEntryFinalised(_)))
+    }
+}
```
jschuurk-kr: I think we should expand the comment to say that keeping a second entry discards the first entry and the second entry becomes the first one. That makes it easier to understand this part of the fuzzer:
```
        // KeepSecondEntry
        (DataEntryStatus::EntriesDifferent(_), Transition::KeepSecondEntry) => {
            is_kept_as_expected(resulting_state, second_entry)
        }
```

jschuurk-kr (commented):
Did a partial review:
- ran the fuzzer for a while, no errros
- ran the fuzzer after deliberately breaking it by putting an incorrect state in a `matches!`, resulted in an error
- reviewed `is_as_expected()`, `is_kept_as_expected()` and `is_finalised_as_expected()` comparing it to the flowchart

backend/fuzz/fuzz_targets/data_entry_status.rs
praseodym: Expanded in 8de99ba3edc2ca0bfe13b3fa9a50fc505555accb

jschuurk-kr (approved):

oliver3 (approved):

PR #3816, opened 2026-08-11
Integration tests in one binary

cikzh:
## What & why
Every category of integration tests  (e.g. apportionment_integration_test) compiles into its own binary. This eats a lot of memory, as they are compiled simultaneously, often resulting in OOM issues on my laptop. Even when running a single unit test with `cargo test name_of_unit_test`, all binaries would build.

## How to test
Run cargo test, see that only an integration_test binary is built, instead of a binary per test category.


praseodym (approved):

praseodym:
This is a good improvement mostly because it reduces disk space and linker CPU time, but for your OOM problem I'd look into setting the [`CARGO_BUILD_JOBS` environment variable](https://doc.rust-lang.org/cargo/reference/config.html#buildjobs) instead.

jschuurk-kr (approved):

sanderboom:
Cool, 13 testbins: 233.9GB / 66.4s -> 1 testbin: 2.43GB / 10.3s. Learned a lot while reviewing!

cikzh:
> This is a good improvement mostly because it reduces disk space and linker CPU time, but for your OOM problem I'd look into setting the [`CARGO_BUILD_JOBS` environment variable](https://doc.rust-lang.org/cargo/reference/config.html#buildjobs) instead.

Yeah, I already have it in place (via .cargo/config.toml). Using the Wild linker also helps somewhat.

PR #3828, opened 2026-08-13
Show 50 candidates in Na 14-2 and adjust votes table explainer texts for corrigenda

### LINKED ISSUE #3827: Show 50 candidates in Na 14-2 and adjust votes table explainer texts for corrigenda

Lionqueen94:
Closes #3827 

## What & why

- Show 50 instead of 40 candidates in Na 14-2, to do this the margin-bottom for this footer had to be slightly adjusted from 3.2cm to 2.9cm
- Adjusted votes table explainer texts for all 4 corrigendum models
- Always have the chapter title on a separate page (also in Na 14-1 versie 1 and 2) to make sure 50 or less candidates on list 1 also fit on the page

## How to test

- Verify the changes in the pdf diffs

## Reviewer notes

- Decisions noted here: https://github.com/kiesraad/abacus-documentatie/pull/119
- Final footer changes made in: https://github.com/kiesraad/abacus/pull/3837

### DIFF
--- a/backend/templates/model-na-14-1-versie-2.typ
+++ b/backend/templates/model-na-14-1-versie-2.typ
@@ -201,6 +201,10 @@ verschil.]
 
 == Stemmen per lijst en per kandidaat <per_list_and_candidate>
 
+Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere telling. Getallen die niet zijn veranderd, hoeven niet ingevuld te worden in de kolom ‘gecorrigeerd’. Onder ‘oorspronkelijk’ staan de getallen die door het stembureau of door het #location_type in een eerdere zitting zijn vastgesteld.
+
+#pagebreak(weak: true)
+
 #for political_group in input.votes_tables {
   votes_table(
     title: [#political_group.number #political_group.name],
@@ -212,9 +216,7 @@ verschil.]
     column_total: "Subtotaal kolom",
     sum_total: columns => [Totaal lijst (kolom #columns)],
     total_instruction: [Neem dit totaal over in rubriek #ref(<cast_votes>) bij de juiste lijst.],
-    explainer_text: [
-      Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere telling. Getallen die niet zijn veranderd, hoeven niet ingevuld te worden in de kolom ‘gecorrigeerd’. Onder ‘oorspronkelijk’ staan de getallen die door het stembureau of door het #location_type in een eerdere zitting zijn vastgesteld.
-    ]
+    explainer_text: [Vul alléén de getallen in die veranderd zijn ten opzichte van de oorspronkelijke telling.]
   )
 }
 
--- a/backend/templates/model-na-14-2-bijlage-1.typ
+++ b/backend/templates/model-na-14-2-bijlage-1.typ
@@ -198,5 +198,6 @@ Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere tell
     column_total: "Subtotaal kolom",
     sum_total: columns => [Totaal lijst (kolom #columns)],
     total_instruction: [Neem dit totaal over in rubriek #ref(<cast_votes>) van deze bijlage bij de juiste lijst.],
+    explainer_text: [Vul alléén de getallen in die veranderd zijn ten opzichte van de oorspronkelijke telling.]
   )
 }
--- a/backend/templates/model-na-14-2.typ
+++ b/backend/templates/model-na-14-2.typ
@@ -16,7 +16,7 @@
 
     Corrigendum van een #location_type \
     Model Na 14-2 (versie 2027)
-  ], margin-bottom: 3.2cm
+  ], margin-bottom: 2.9cm
 )
 
 #set heading(numbering: none)
@@ -167,6 +167,10 @@ Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere tell
 
 == Stemmen per lijst en per kandidaat
 
+De kolom ‘gecorrigeerd’ toont alléén de getallen die veranderd zijn ten opzichte van een eerdere telling. Onder ‘oorspronkelijk’ staan de getallen die in een eerdere zitting door het #location_type zijn vastgesteld.
+
+#pagebreak(weak: true)
+
 #for political_group in input.votes_tables {
   votes_table(
     title: [#political_group.number #political_group.name],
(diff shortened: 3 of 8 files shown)


backend/templates/model-na-14-2-bijlage-1.typ
```diff
@@ -198,5 +198,6 @@ Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere tell
     column_total: "Subtotaal kolom",
     sum_total: columns => [Totaal lijst (kolom #columns)],
     total_instruction: [Neem dit totaal over in rubriek #ref(<cast_votes>) van deze bijlage bij de juiste lijst.],
+    explainer_text: [Vul alléén de getallen in die veranderd zijn ten opzichte van de oorspronkelijke telling.]
```
sanderboom, line 201:
```suggestion
    explainer_text: [Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere telling.]
```

As suggested in the email and consistent with the larger explainer on the title page (not sure if this was changed afterwards).

backend/templates/model-na-14-2.typ
```diff
@@ -16,7 +16,7 @@
 
     Corrigendum van een #location_type \
     Model Na 14-2 (versie 2027)
-  ], margin-bottom: 3.2cm
+  ], margin-bottom: 2.9cm
```
sanderboom, line 19: Should we change this in all FatFooter declarations?

backend/templates/model-na-14-2.typ
```diff
@@ -179,9 +183,7 @@ Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere tell
     column_total: "Subtotaal kolom",
     sum_total: columns => [Totaal lijst (kolom #columns)],
     total_instruction: [Neem dit totaal over in rubriek #ref(<cast_votes>) van deze bijlage bij de juiste lijst.],
-    explainer_text: [
-      Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere telling. Getallen die niet zijn veranderd, hoeven niet ingevuld te worden in de kolom ‘gecorrigeerd’. Onder ‘oorspronkelijk’ staan de getallen die in een eerdere zitting door het #location_type zijn vastgesteld.
-    ],
+    explainer_text: [De kolom ‘gecorrigeerd’ toont alléén de getallen die veranderd zijn ten opzichte van de oorspronkelijke telling.],
```
sanderboom, line 185:
```suggestion
    explainer_text: [De kolom ‘gecorrigeerd’ toont alléén de getallen die veranderd zijn ten opzichte van een eerdere telling.],
```

As suggested in the email and consistent with the larger explainer on the title page (not sure if this was changed afterwards).

backend/templates/model-na-14-1-versie-2.typ
```diff
@@ -212,9 +216,7 @@ verschil.]
     column_total: "Subtotaal kolom",
     sum_total: columns => [Totaal lijst (kolom #columns)],
     total_instruction: [Neem dit totaal over in rubriek #ref(<cast_votes>) bij de juiste lijst.],
-    explainer_text: [
-      Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere telling. Getallen die niet zijn veranderd, hoeven niet ingevuld te worden in de kolom ‘gecorrigeerd’. Onder ‘oorspronkelijk’ staan de getallen die door het stembureau of door het #location_type in een eerdere zitting zijn vastgesteld.
-    ]
+    explainer_text: [Vul alléén de getallen in die veranderd zijn ten opzichte van de oorspronkelijke telling.]
```
sanderboom, line 219:
```suggestion
    explainer_text: [Vul alléén de getallen in die veranderd zijn ten opzichte van een eerdere telling.]
```

As suggested in the email and consistent with the larger explainer on the title page (not sure if this was changed afterwards).

sanderboom (changes requested):

backend/templates/model-na-14-2-bijlage-1.typ
Lionqueen94: This has been discussed after the email and has been approved by Arjen as implemented. The word 'oorspronkelijk' needed to be used to make it more clear.

backend/templates/model-na-14-2.typ
Lionqueen94: Same as above

backend/templates/model-na-14-1-versie-2.typ
Lionqueen94: Same as above

backend/templates/model-na-14-2.typ
Lionqueen94: I was thinking this, but this would cause more changes, so I refrained from it for now. We can then adjust the number of signatures again as well haha. What do other reviewers think?

jorisleker (approved):
Reviewed diffs, LGTM

backend/templates/model-na-14-2.typ
jorisleker: Number of signutures is quite insane now, with that full page of extra signature fields to compensate for one missing field on the first page

Lionqueen94: It seems like the number of signatures as intended in the model (6 total) does fit on one page with the 2.9cm margin-bottom. So this would be a plus to do it on all typst models and get rid of the extra page of signatures.

jorisleker: Okay with me if we can add a todo somewhere that we will revert this if we do not show the hash on all pages anymore.

jschuurk-kr: Two questions:
- Are we sure a smaller margin will be printed correctly? (no idea what a minimal required margin is for printers)
- Same question as @sanderboom (I think): if we do this, shouldn't we do this in all models with a footer that contains a hash? (Might be best to do that in a separate PR, though.)

Lionqueen94: @jorisleker Do you have any idea about Joep's first point?
For the second point, yes it would then be for all files with hash footers, because those all have signature pages which we can then reduce to one page instead of 2 (as the model is designed)

Lionqueen94: Ok we have figured out a fix for this, I will make a separate PR for this. Resolving this now.

Lionqueen94: See https://github.com/kiesraad/abacus/pull/3837

sanderboom (approved):

PR #3837, opened 2026-08-17
Final footer update to ensure minimum printer bottom margin of 10mm

### LINKED ISSUE #3836: Final footer update to ensure minimum printer bottom margin of 10mm

Lionqueen94:
Closes #3836 

## What & why

- Adjust footer-descent to ensure minimum margin of 10mm
- Remove extra signature pages.
- Added P 2a variations.

## How to test

- Verify the changes in the pdf diffs
- Run `cargo run --bin gen-pdf` locally and verify the contents of the added variations for Model P 2a

## Reviewer notes

- Model P 22-2 is not (fully) adjusted yet to the updated model, since this will be done in https://github.com/kiesraad/abacus/issues/3540


backend/templates/model-p-22-2.typ
sanderboom, line 579:
```suggestion
#stack(spacing: 0.5em, ..range(0, 6).map(_ => textbox[Naam:][Handtekening:]))
```

I checked, it fits ;)

sanderboom (approved):
Looks good!

backend/templates/model-p-22-2.typ
Lionqueen94: This one will be adjusted to the updated model further in https://github.com/kiesraad/abacus/issues/3540

Lionqueen94: And apparently the updated model only shows 5 even though there is space for one more haha

jorisleker (changes requested):
Footer looks good, but I think this should still be fixed:

- Model N-10-2: signature page has been adjusted so it has 7 signuture fields on the first page. We did this to prevent having a full additional page with signature fields to make up for the one missing compared to the official model. The Abacus generated model however still has the additional second page with 12 signature fields. This can be removed (please check other signature pages as well)
Image

When fixing this, it might be a good idea to also take a look at the _presentielijst_ in the N 10-1, which now wraps over two pages, making the 'bezwaren tijdens het stemmen' table to wrap two pages as well (no caused by this PR, so okay to create a separate issue for this as well):

Image
Image

jorisleker:
> * Model N-10-2: signature page has been adjusted so it has 7 signuture fields on the first page. We did this to prevent having a full additional page with signature fields to make up for the one missing compared to the official model. The Abacus generated model however still has the additional second page with 12 signature fields. This can be removed (please check other signature pages as well)

Or am I interpreting the pdf-diff output wrongly? Is the additional page actually removed without pdf-diff marking it with a red line?

jorisleker (approved):
Interpreted pdf-diff output incorrectly, additional signature pages are gone now.
Changed to approval. @Lionqueen94 can you fix the other issue I mentioned in my previous review in a separate PR? Tx!

PR #3824, opened 2026-08-13
Add election tree based committee extraction

rnijveld:
## What & why

Fixes #3845 This introduces the initial types needed for extracting the information needed by Abacus from the election tree from the election definition EML file. In general the idea is:

- Abacus needs to know what committees there are in an election
- Committees have a responsible region, region they are seated in and tell something about the district they are in.
- Districts come in three variants: None (this election does not contain any districts), All (this election contains districts, but this specific committee concerns all of them at the same time), or Specific to point to a specific district

## How to test

Best tested by running the unit tests from the committees.rs file. Other than that this does not integrate into anything Abacus yet.

## Reviewer notes

Most changes are in the new file committees.rs, most of the added lines are due to me adding several existing election definitions from previous elections verbatim.

PR #3843, opened 2026-08-18
Use election tree data within Abacus

### LINKED ISSUE #3846: Store election tree information in abacus election struct
We need to store the information extracted from the election tree in our Election structs and then apply that information all across Abacus.

rnijveld:
## What & why
Fixes #3846 This builds upon #3824 and integrates the information that Abacus requires into the different Election structs and then the rest of Abacus starts using that information instead. This is quite a major change and introduces 5 new fields on the election struct and removes one (domain_id). Anywhere that used domain id previously we now use authority_id instead. For municipal elections there is no difference between the two, so everything GR related should keep working as is. For GSB level this now fixes the incorrect usage of domain id (it would point to the id of the Water Authority, not the id for the GSB). For CSB we still need some changes, but those are outside the scope of this.

Note: despite previous indications, this branch should be fully functional, including importing of elections.

## How to test

This reaches pretty much everywhere in the application, but most immediately this touches:

- Any frontend page that used the domain id or location fields previously where it now uses the authority_id and authority_region fields instead.
- Anything related to any of the generated PDFs
- Anything related to importing/exporting EMLs

## Reviewer notes
I've moved around a couple of the types that were defined in `committees.rs` in #3824 given that they are now part of the domain model for elections, but that code should have largely remained the same, your review should probably focus on if I've captured all usages of the old fields.


oliver3:
The two PDF changes are that they now show the correct title whereas before it also showed "Gemeenteraad Hovenerwoud 2022" for the PS and WS variant of model P 2a
image

Edit: and are changed because the names in the input changed

backend/src/api/election.rs
```diff
@@ -636,9 +654,12 @@ fn check_hash(
 /// Parse EML_NL 110 election definition and EML_NL 230 candidate list into a [`NewElection`]
 fn parse_election_candidates_eml(
     election_eml_data: &str,
+    selected_committee: Option<(CommitteeCategory, Option<RegionKey>)>,
     candidate_eml_data: Option<&str>,
 ) -> Result<NewElection, APIError> {
-    let mut election = NewElection::from_eml_str(election_eml_data)?;
+    let (mut election, _election_tree) =
+        NewElection::from_eml_str(election_eml_data, selected_committee)?;
+    // TODO: need to pick the right district first before adding candidates
```
oliver3: Please refer to the issue where this will be done

backend/src/api/election.rs
```diff
@@ -703,8 +724,14 @@ async fn create_sub_committees(
 ) -> Result<(), APIError> {
     match (election.committee_category, election.category) {
         (CommitteeCategory::CSB, ElectionCategory::Municipal) => {
+            // TODO: this is not right, we should use the election tree instead
```
oliver3: Please refer to the issue where this will be solved

oliver3 (approved):
There are some TODO's added for which it is not immediately clear when they will be done. 

As discussed with @rnijveld : some of them will be picked up in a following PR and to the remaining ones an issue number will be added.

Lionqueen94 (approved):
Looks good, small comment about writing `eerste kamer` in a comment without capitalization, but that's definitely not a blocker

PR #3850, opened 2026-08-19
Document Election structs and their fields

rnijveld:
## What & why

The different election structs and the fields within them might have become a little bit confusing. I thought it would be a good idea to add some documentation so that we later understand what each field is intended for.

## How to test

No behavioral changes, only doc comments.

## Reviewer notes

There is a little bit of repetition in the struct docs, but I've decided that repeating all field docs for all of the three structs is inefficient and decided to put in references instead. Most editors with Rust support will allow some form of 'go to definition' to allow easy access to the full descriptions.

### DIFF
--- a/backend/src/domain/election.rs
+++ b/backend/src/domain/election.rs
@@ -12,62 +12,135 @@ use crate::domain::identifier::id;
 
 id!(ElectionId);
 
-/// Election without political groups
+/// Election without political groups.
+///
+/// Note: an election within Abacus does not represent the entire election, but
+/// rather a single committee (i.e. stembureau at the CSB, HSB or GSB level)
+/// within the election.
+///
+/// When access to the political groups and their candidates is required, use
+/// [`ElectionWithPoliticalGroups`] instead. When creating a new election, use
+/// [`NewElection`] instead.
 #[derive(Serialize, Deserialize, ToSchema, Clone, Debug, PartialEq, Eq, Hash, FromRow)]
 #[serde(deny_unknown_fields)]
 pub struct Election {
+    /// See [`ElectionWithPoliticalGroups::id`]
     pub id: ElectionId,
+    /// See [`ElectionWithPoliticalGroups::name`]
     pub name: String,
+    /// See [`ElectionWithPoliticalGroups::committee_category`]
     pub committee_category: CommitteeCategory,
+    /// See [`ElectionWithPoliticalGroups::counting_method`]
     #[serde(skip_serializing_if = "Option::is_none")]
     #[schema(nullable = false)]
     pub counting_method: Option<VoteCountingMethod>,
+    /// See [`ElectionWithPoliticalGroups::election_id`]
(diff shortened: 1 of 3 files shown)


backend/src/domain/election.rs
```diff
@@ -12,62 +12,135 @@ use crate::domain::identifier::id;
...
+/// When you do not need access to political groups and their candidates, you
+/// can use [`Election`] instead. When creating a new election, use
+/// [`NewElection`] instead.
 #[derive(Serialize, Deserialize, ToSchema, Clone, Debug, PartialEq, Eq, Hash, FromRow)]
 #[serde(deny_unknown_fields)]
 pub struct ElectionWithPoliticalGroups {
+    /// Identifier of the election within Abacus
     pub id: ElectionId,
+    /// Name of the election
     pub name: String,
```
jschuurk-kr: This made me wonder how the `name` is set, so maybe add "as defined in the EML_NL election definition"?

backend/src/domain/election.rs
```diff
@@ -12,62 +12,135 @@ use crate::domain::identifier::id;
...
+    /// Examples include "Municipal" for municipal elections or "WaterAuthority"
+    /// for the water authority elections.
     pub category: ElectionCategory,
+    /// The sub-category of the election, as defined by the EML_NL election
+    /// definition.
     pub sub_category: ElectionSubCategory,
+    /// The number of seats that are to be elected for. For example, this is 150
+    /// for elections for the House of Representatives (tweede kamer).
     pub number_of_seats: u32,
+    /// How many voters are registered for this election.
```
praseodym:
```suggestion
    /// How many voters are registered for this election committee district.
```

backend/src/domain/election.rs
rnijveld: Fixed in latest commit

backend/src/domain/election.rs
```diff
@@ -12,62 +12,135 @@ use crate::domain::identifier::id;
...
     pub domain: Option<ElectionDomain>,
+    /// The category of the election, as defined by EML_NL election definition.
+    /// Examples include "Municipal" for municipal elections or "WaterAuthority"
+    /// for the water authority elections.
     pub category: ElectionCategory,
+    /// The sub-category of the election, as defined by the EML_NL election
+    /// definition.
     pub sub_category: ElectionSubCategory,
+    /// The number of seats that are to be elected for. For example, this is 150
+    /// for elections for the House of Representatives (tweede kamer).
```
Lionqueen94:
```suggestion
    /// for elections for the House of Representatives (Tweede Kamer).
```

Lionqueen94 (dismissed):

backend/src/domain/election.rs
rnijveld: Fixed!

Lionqueen94 (approved):

praseodym (approved):

Issue #3840, opened 2026-08-18
Check if correct candidate list is uploaded for a district


Lionqueen94:
> @Lionqueen94 @stacktraceghost Based on OSV test data from the previous PS election there's a difference in the "kandidatenlijsten" between PS1 and PS2 elections.
> 
> In PS1 elections (no voting districts) the EML 230b contains `<ContestIdentifier Id="geen"/>`. In PS2 elections (voting districts) the EML 230b contains `<ContestIdentifier Id="4"><ContestName>Leiden</ContestName></ContestIdentifier>`. I believe that refers to the following section of the election definition:
> ```
>                 <kr:Region RegionNumber="4" RegionCategory="KIESKRING" SuperiorRegionNumber="9" SuperiorRegionCategory="PROVINCIE">
>                     <kr:RegionName>Leiden</kr:RegionName>
>                     <kr:Committee CommitteeCategory="HSB"/>
>                 </kr:Region>
> ```
> 
> The election definition also contains:
> ```
>                 <kr:Region RegionNumber="537" RegionCategory="GEMEENTE" SuperiorRegionNumber="4" SuperiorRegionCategory="KIESKRING">
>                     <kr:RegionName>Katwijk</kr:RegionName>
>                 </kr:Region>
> ```
> Which would allows us to validate the correct "kandidatenlijsten" is being uploaded by the municipality. I don't think this is in scope yet of this epic? (And it should be a separate issue from this one.) cc @jorisleker  

 _Originally posted by @jschuurk-kr in [#3618](https://github.com/kiesraad/abacus/issues/3618#issuecomment-5326537540)_

Also make sure a descriptive frontend error is shown @jorisleker do you want to create an error in Figma and link it here?

jorisleker:
Copy/design for [Frontend error](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=25179-14399&t=kxi67R0iOlTMpXge-11)

rnijveld:
Backend is fixed in #3862

Issue #3618, opened 2026-07-21
Create new PS/WS election


jschuurk-kr:
- [x] ~Read election tree from election definition EML via rust-eml-nl~ Done in #3845 
- [ ] Save election name for water authority election as: `ElectionName.stripPrefix("⁨Algemeen bestuur van het ")` and subsequently uppercase the first letter (so it will start with `Waterschap`, `Wetterskip` or `Hoogheemraadschap`) (see [design](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=23924-13557&m=dev))
- [ ] Show list of municipalities to select GSB from
- [x] ~different import flow (i.e. definition -> select committee -> upload candidate list) to support PS2 GSBs~ Done in #3858
- [x] ~(check first if/what is needed) Store name and domain id of all higher levels~ Done in #3846
  - WS: GSB - CSB
  - PS: GSB - (HSB -) CSB (Note: not HSB right now, as we don't do PS2 elections beyond the GSB yet)
- [x] ~Make sure `domain_id` and potential other fields are corrected in election fixtures 11 and 12~ Done in #3846 and as part of #3764 

This should be done in multiple PRs.

jschuurk-kr:
@Lionqueen94 @stacktraceghost Based on OSV test data from the previous PS election there's a difference in the "kandidatenlijsten" between PS1 and PS2 elections.

In PS1 elections (no voting districts) the EML 230b contains `<ContestIdentifier Id="geen"/>`. In PS2 elections (voting districts) the EML 230b contains `<ContestIdentifier Id="4"><ContestName>Leiden</ContestName></ContestIdentifier>`. I believe that refers to the following section of the election definition:
```
                <kr:Region RegionNumber="4" RegionCategory="KIESKRING" SuperiorRegionNumber="9" SuperiorRegionCategory="PROVINCIE">
                    <kr:RegionName>Leiden</kr:RegionName>
                    <kr:Committee CommitteeCategory="HSB"/>
                </kr:Region>
```

The election definition also contains:
```
                <kr:Region RegionNumber="537" RegionCategory="GEMEENTE" SuperiorRegionNumber="4" SuperiorRegionCategory="KIESKRING">
                    <kr:RegionName>Katwijk</kr:RegionName>
                </kr:Region>
```
[...]

PR #3862, opened 2026-08-22
Final gsb import steps in backend

rnijveld:
## What & why
This adds the final parts of being able to select a specific GSB in the backend during importing of elections. It also fixes a couple of related tiny issues. Commits should all fix a specific part.

- authority_region was missing from audit events, added that
- Fixed a small typo in docs for identifier of domain
- During importing, we now check that the candidate list is for the same district region as the selected committee
- Validation endpoint can now return a list of GSBs for selecting one in the frontend
- GSB endpoints for import now accept a 'selected GSB' that should be given the RegionKey of the selected region
- Sub committee creation now is based upon the election tree instead of the election category.

Part of #3618 
District validation is backend part of #3840

## How to test

Run through the import process. Note: we can't yet select a GSB, that needs frontend work. It should however still work as it did for GR GSBs/CSBs.


cikzh (dismissed):
Reviewed code and did a test import

backend/src/eml/mod.rs
```diff
@@ -246,12 +253,27 @@ impl NewElection {
...
+        let contest_id = contest.identifier.id.cloned_value()?;
+        match (district_region, contest_id) {
+            // Only allow none-district if the election district is none
+            (CommitteeDistrict::None, c) if c.is_geen() => {}
+            // If the district region is all, allow any contest id (except geen)
+            (CommitteeDistrict::All, c) if !c.is_geen() => {}
+            // Otherwise, the contest id must match the district number
+            (CommitteeDistrict::Specific(s), other)
+                if let Some(num) = s.key.number
+                    && other.value() == num.to_string() => {}
```
praseodym: Does this also work with the roman numerals district numbers?

rnijveld: Nope! Changed in b10847fc26dd01c9733f2a302903b65a4240cc0e

backend/src/api/election.rs
```diff
@@ -719,37 +751,23 @@ async fn create_election_with_committee_session(
...
-            )
-            .await?;
-        }
-        (
-            CommitteeCategory::CSB,
-            ElectionCategory::Provincial | ElectionCategory::WaterAuthority,
-        ) => {
-            todo!("Will be implemented in the future")
-        }
-        (CommitteeCategory::GSB, _) => {}
```
stacktraceghost, line 752: It should not be possible to create subcommittees for elections with committee category GSB. Should we keep an extra check here or just rely on not passing subcommittees for GSB's?

stacktraceghost (approved):

praseodym (approved):

backend/src/api/election.rs
rnijveld: This being a local function within this file only, I would not be worried about this.

cikzh (approved):
Approved, but not merging, as I'm not sure if the previous conversation is resolved.

Issue #3818, opened 2026-08-12
Update test_data_gen to generate DSO entries


stacktraceghost:
Currently only `CSOFirstSession` data entries are being generated in `test_data_gen`. Extend this to also generate DSO data entries.

Found this by running into an `error: "Invalid state transition"` when:
1. Generating a DSO election with 0% second entries
2. Trying to claim a second entry

Additionally, it is possible to generate a DSO election with all entries finished and complete the data entry, while it has the wrong data entry models. We should think about what checks to add and maybe do an exploratory test to see where no mismatch checks are in place (the data entry model vs committee_category/counting_method). 

Related https://github.com/kiesraad/abacus/issues/3695

Issue #3402, opened 2026-06-10
Introduce report service abstraction


praseodym:
#3320 resolved #3241 but still left some domain->api usages that can be fixed by introducing a report service abstraction as already noted in #3241.

> Good improvement! There are some domain->api usages that we should get rid of, but we can do that in a follow up PR.

_Originally posted by @praseodym in https://github.com/kiesraad/abacus/pull/3320#pullrequestreview-4466512665_

PR #3863, opened 2026-08-24
Results variant must match election and committee session

oliver3:
While investigating https://github.com/kiesraad/abacus/issues/3818 we found out that we could generate files for a completed election where the Results variant/model did not match the election and committee session type.

In this PR two extra checks have been added to check if the Results model matches the election and committee session:
- In the data entry endpoints
- when generating files

## How to test

Currently when generating a completed DSO election, the generator generates CSO Results. Before, Abacus would generate files for this combination which should not be possible. Now, you will get an error instead.

When generating a DSO election with no data entry, it was already not possible to start (claim) data entry, but now this will be explicitly stopped with a `DataIntegrityError`.

## Reviewer notes

Is this too much for a scenario that cannot occur with normal Abacus usage?


backend/src/domain/report/structs.rs
praseodym, line 206: I think we should also check `previous_results`

praseodym (commented):
I think we should add the check to every place that uses `list_results_for_committee_session`, so aside from my other comment that also includes here:
https://github.com/kiesraad/abacus/blob/e16de4c12a6e038382e415bb51c32fb4a060b825/backend/src/service/apportionment.rs#L218-L219

It also bothers me that all this logic is in `src/domain/report/structs.rs` but we already have #3402 to tackle that.

backend/src/domain/report/structs.rs
oliver3: Added verification after all `list_results_for_committee_session()` calls in https://github.com/kiesraad/abacus/pull/3863/commits/dc5b862e94b74581f3d59f2b83aab5fcb2019d84 after changing the methods into verify methods returning `Result` type in https://github.com/kiesraad/abacus/pull/3863/commits/b168ca38036014ef1f4a53ca6cebf835eea7b9b0

Lionqueen94 (dismissed):

backend/src/domain/report/structs.rs
```diff
@@ -201,6 +205,7 @@ impl ResultsInputData {
         {
             let previous_results =
                 list_results_for_committee_session(conn, previous_committee_session.id).await?;
+            previous_results.verify_models(&election, &committee_session)?;
```
stacktraceghost: This is checking the previous results against the current committee session

backend/src/domain/results/mod.rs
```diff
@@ -132,6 +134,42 @@ impl Results {
...
+        election: &ElectionWithPoliticalGroups,
+        session: &CommitteeSession,
+    ) -> Result<(), IncorrectResultsModel> {
+        let category = election.committee_category;
+        let method = election.counting_method;
+        let next = session.is_next_session();
+
+        let correct = match self {
+            Results::DSOFirstSession(_) => {
+                category == CommitteeCategory::GSB && method == Some(VoteCountingMethod::DSO)
```
stacktraceghost, line 148: We could add `&& !next` here and to the `CSOFirstSession` /  `GSB` to catch the previous comment.

backend/src/domain/results/mod.rs
```diff
@@ -476,11 +514,40 @@ impl Validate for Results {
...
+        for (_, results) in self {
+            results.verify_model(election, session)?;
+        }
+        Ok(())
+    }
+}
+
 #[cfg(test)]
 pub mod tests {
+    use core::assert_matches;
```
stacktraceghost:
```suggestion
    use std::assert_matches;
```

In other places we only use `use std::assert_matches;`

backend/src/domain/results/mod.rs
oliver3: I had that first, until an e2e let me realise that a xSOFirstSession can occur in a next committee session, when a new polling station has been added.

oliver3: Checking this again, I see that it should be a CSONextSession indeed. I'll investigate what was going on before.

Edit: this is when an investigation plus new data entry has been added in the new committee session

oliver3: Ah it was a different scenario: when you have a next committee session with a polling station that does not have new data entry, it will retrieve the previous data entry from the database which can be a `CSOFirstSessionResults`.
An example would be polling station with id `722` in `election_7_four_sessions.sql`.

backend/src/domain/report/structs.rs
oliver3: Fixed in https://github.com/kiesraad/abacus/pull/3863/commits/c0bc578ff40f102212d14a9871a3673df3bbb435

backend/src/domain/results/mod.rs
oliver3: Let's use `std::` only indeed, changed in https://github.com/kiesraad/abacus/pull/3863/commits/c0bc578ff40f102212d14a9871a3673df3bbb435

backend/src/domain/results/mod.rs
oliver3: To conclude: FirstSession results can also occur for committee next sessions, so we cannot check for `!next` in those cases

stacktraceghost (approved):

praseodym (approved):

PR #3868, opened 2026-08-25
Generate DSOFirstSessionResults for DSO counting method

### LINKED ISSUE #3818: Update test_data_gen to generate DSO entries

oliver3:
## What & why

Resolves [Update test_data_gen to generate DSO entries](https://github.com/kiesraad/abacus/issues/3818)

## How to test

Generate some municipal elections and see that they generate correct data. Abacus now has extra verifications added in https://github.com/kiesraad/abacus/issues/3818 so it should fail pretty quickly otherwise.

## Reviewer notes

Instead of `generate_cso_first_session_results()` and `generate_gsb_results()` (and the missing dso variant), there is now a `generate_results()` that uses election and committee session to generate the correct `Results`, using `generate_common_results()` for the common parts (returned as `CommonPollingStationResults`).


backend/src/test_data_gen/generators.rs
```diff
@@ -732,14 +739,70 @@ async fn generate_csb_data_entry(
...
+            extra_investigation: generate_extra_investigation(rng),
+            counting_differences_polling_station: generate_counting_differences_ps(rng),
+            voters_counts: common.voters_counts,
+            votes_counts: common.votes_counts,
+            differences_counts: common.differences_counts,
+            political_group_votes: common.political_group_votes,
+        }),
+        (GSB, Some(DSO)) => Results::DSOFirstSession(DSOFirstSessionResults {
+            about_report: Default::default(),
+            checks_and_corrections: Default::default(),
```
praseodym: I think results with these questions unanswered will be invalid once validation is merged

oliver3: Added generator to create valid about_report and checks_and_corrections sections in https://github.com/kiesraad/abacus/pull/3868/commits/09a334a113e0efd5976774a0c0cf586a2d80f165

stacktraceghost (dismissed):

praseodym (dismissed):

stacktraceghost (approved):

praseodym (approved):

PR #3876, opened 2026-08-27
Update to pnpm 11.21.0

praseodym:
## What & why

Update to pnpm 11.21.0, see [release notes](https://pnpm.io/blog/releases/11.21-11.22). Not updating past 11.21.0 because nixpkgs (that I personally use) doesn't have a more recent version yet.

## How to test

CI still passes and pnpm works locally after updating.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -8,7 +8,7 @@
     "pnpm": "^11",
     "npm": "please-use-pnpm"
   },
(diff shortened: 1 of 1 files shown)


Lionqueen94 (approved):

oliver3 (approved):
Works on my machine

PR #3878, opened 2026-08-27
Add validation F.311

### LINKED ISSUE #3870: Implement F.311
See https://github.com/kiesraad/abacus-documentatie/blob/main/use-cases/validatieregels-plausibiliteitschecks-tellingen.md#f311-232-verklaring-voor-verschil--nee-en-vanwege-een-onverklaard-verschil-in-stap-controles-en-correcties-is-niet-aangevinkt

oliver3:
## What & why

Implements https://github.com/kiesraad/abacus/issues/3870

## How to test

Start a DSO data entry with differences between D & H. Check dat when you select "Nee, er is een onverklaard verschil" in 2.3.2, you have to have a corrigendum with "Op eigen initiatief": "Vanwege een onverklaard verschil".

See that otherwise, you will get validation F.311:
- No corrigendum on "Over het proces-verbaal"
- Corrigendum, but "Vanwege (het vermoeden van) een andere fout" instead of "Vanwege een onverklaard verschil"

Check that this indeed is how it [should be implemented](https://github.com/kiesraad/abacus-documentatie/blob/main/use-cases/validatieregels-plausibiliteitschecks-tellingen.md#f311-232-verklaring-voor-verschil--nee-en-vanwege-een-onverklaard-verschil-in-stap-controles-en-correcties-is-niet-aangevinkt)

Check the copy for both the typist and the coordinator


stacktraceghost (approved):
Tested, checked copy and implementation + tests look good.

One note: `Verklaring voor verschil = nee` could be interpreted two ways:
- If only `no` is checked (current situation)
- If `no` is checked (this would be only `no` or `both`)

When [proposed F.313](https://github.com/kiesraad/abacus-documentatie/pull/124) is implemented, this does not matter anymore, because:
- F.313 will be shown when D = H and one or both is checked
- F.310 will be shown when D <> H and nothing or both is checked

cikzh (approved):

PR #3880, opened 2026-08-27
Added DSO specific validation feedback texts

### LINKED ISSUE #3871: DSO specific validation feedback texts
Add coordinator DSO overrides in a feedback_GSB_DSO.json file and add logic in getTranslations()

This is for existing CSO coordinator validation rules have different texts for DSO: 
- F.201 F.202 F.203 F.204 
- F.301 F.302 F.303 F.304 F.305 F.306 F.307 F.308 F.309 
- F.401 F.402 F.403

Lionqueen94:
Closes #3871 

## What & why

- Add DSO overrides for validation results and make sure they are used if applicable
- Update Feedback stories to add separate GSB CSO and DSO elections in the args to make testing easier

## How to test

You can verify the texts via [storybook](https://link-nederland-informatie.abacus-test.nl/storybook/?path=/story/components-ui-feedback--gsb-coordinator&args=election:GSB+DSO;validationResults[0]:F101;validationResults[1]:F102;validationResults[2]:F111;validationResults[3]:F112;validationResults[4]:F121;validationResults[5]:F122;validationResults[6]:F131;validationResults[7]:F132;validationResults[8]:F133;validationResults[9]:F134;validationResults[10]:F135;validationResults[11]:F201;validationResults[12]:F202;validationResults[13]:F203;validationResults[14]:F204;validationResults[15]:F301;validationResults[16]:F302;validationResults[17]:F303;validationResults[18]:F304;validationResults[19]:F305;validationResults[20]:F306;validationResults[21]:F307;validationResults[22]:F308;validationResults[23]:F309;validationResults[24]:F310;validationResults[25]:F312;validationResults[26]:F401;validationResults[27]:F402;validationResults[28]:F403;validationResults[29]:W001;validationResults[30]:W002;validationResults[31]:W201;validationResults[32]:W202;validationResults[33]:W203;validationResults[34]:W204;validationResults[35]:W205;validationResults[36]:W206) by switching between election `GSB CSO` and `GSB DSO` in the args

## Reviewer notes

Please carefully review against https://github.com/kiesraad/abacus-documentatie/blob/main/use-cases/validatieregels-plausibiliteitschecks-tellingen.md if anything is missing/wrong

### DIFF
--- a/frontend/src/i18n/feedback-structure.test.tsx
+++ b/frontend/src/i18n/feedback-structure.test.tsx
@@ -26,4 +26,18 @@ describe("feedback translations structure", () => {
       }
     }
   });
+
+  test("validate feedback_GSB_DSO.json", () => {
+    const feedback = translations.feedback_GSB_DSO;
+    const feedbackItems = Object.values(feedback);
+
+    for (const entry of feedbackItems) {
+      const keys = Object.keys(entry.coordinator);
+
+      // No other keys than below allowed
+      for (const key of keys) {
+        expect(["title", "content", "actions"]).toContain(key);
+      }
+    }
+  });
 });
--- a/frontend/src/i18n/locales/nl/feedback_GSB_DSO.json
+++ b/frontend/src/i18n/locales/nl/feedback_GSB_DSO.json
@@ -1,0 +1,86 @@
+{
+  "F201": {
+    "coordinator": {
+      "content": "Controleer in rubriek 2.3 of er een onverklaard verschil opgelost wordt als het juiste getal bij D wordt ingevuld."
+    }
+  },
+  "F202": {
+    "coordinator": {
+      "content": "Controleer in rubriek 2.3 of er een onverklaard verschil opgelost wordt als de juiste getallen bij E en H worden ingevuld."
+    }
+  },
+  "F203": {
+    "coordinator": {
+      "content": "Controleer in rubriek 2.3 of er een onverklaard verschil opgelost wordt als het juiste getal bij H wordt ingevuld."
+    }
+  },
+  "F204": {
+    "coordinator": {
+      "content": "Controleer in rubriek 2.3 of er een onverklaard verschil opgelost wordt als de juiste getallen bij E en E.1 t/m E.n worden ingevuld."
+    }
+  },
+  "F301": {
+    "coordinator": {
+      "actions": "<ul><li>Herstel de fout door op papier het juiste getal in te (laten) vullen.</li><li>Controleer ook of er een onverklaard verschil ontstaat.</li><li>Pas zo nodig rubriek 2.3.2 aan, en volg de instructies over hertellen die daar staan.</li></ul>"
+    }
+  },
+  "F302": {
+    "coordinator": {
+      "actions": "<ul><li>Maak een corrigendum waarin de juiste optie geselecteerd wordt.</li></ul>"
+    }
+  },
+  "F303": {
+    "coordinator": {
+      "actions": "<ul><li>Maak een corrigendum waarin de juiste optie geselecteerd wordt.</li></ul>"
+    }
+  },
+  "F304": {
+    "coordinator": {
+      "content": "Is op het proces-verbaal duidelijk aangegeven welk van de opties bedoeld is?",
+      "actions": "<ul><li>Zo ja: laat dat dan overnemen in Abacus.</li><li>Zo nee: maak een corrigendum waarin de juiste optie geselecteerd wordt.</li></ul>"
+    }
+  },
+  "F305": {
+    "coordinator": {
+      "actions": "<ul><li>Corrigeer de rekenfout door een corrigendum te (laten) maken waarop I en/of J leeg zijn.</li><li>Kijk ook of het corrigeren van de fout een onverklaard verschil introduceert (zie 2.3.2).</li><li>Hertel dan de stembiljetten en het aantal toegelaten kiezers. Hertel tot de fout gevonden is, of alles één keer herteld is.</li><li>Vul nieuwe telresultaten op het corrigendum in.</li></ul>"
+    }
+  },
+  "F306": {
+    "coordinator": {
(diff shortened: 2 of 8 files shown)


frontend/src/utils/ValidationResults.test.ts
```diff
@@ -176,21 +176,40 @@ describe("dottedCode", () => {
...
+  test("should return CSO coordinator translations for GSB validation result for CSO election with DSO override", () => {
+    expect(getTranslations(GSBCSOElection, validationResultMockData.F201, "coordinator")).toEqual({
+      code: "F.201",
+      title: t("feedback_GSB.F201.coordinator.title"),
+      content: tx("feedback_GSB.F201.coordinator.content"),
+      actions: tx("feedback_GSB.F201.coordinator.actions"),
+    });
+  });
+
+  test("should return DSO coordinator translations for GSB validation result for DSO election with DSO override", () => {
```
stacktraceghost, line 232: This test only validates `content` for its behavior. We don't have a DSO translation which covers all three properties, but we could use `F401` to also cover `actions`.

frontend/src/i18n/locales/nl/feedback_GSB_DSO.json
```diff
@@ -0,0 +1,86 @@
+{
```
stacktraceghost, line 1: This doesn't include `F121/F122/F131/F132/F133/F134/F135/F311`. I think those should be moved over from `feedback_GSB` since those are DSO-only.

jorisleker: A, that's something we did not cover in our [discussion on mattermost](https://digilab.overheid.nl/chat/kiesraad/pl/buy81hteejgw3yngc94gsbekzh) 🙃 

Depends on how you interpret 'override'. Should it be an override if the actual _error_ differs (ie: only for DSO), of if the _text_ differs? In the first case we should follow your suggestion, in the second case the current implementation is defendable.

jorisleker (dismissed):
Reviewed all errors/warnings where feedback differs for coordinator GSB CSO and DSO, and compared them to the validation rules doc. All looks good 👌

frontend/src/i18n/locales/nl/feedback_GSB_DSO.json
Lionqueen94: As I understood it from the discussion, only texts that are different should be in the DSO file, and not the specific DSO only feedback. I could see those move to the file though, it's just about preference I think.

frontend/src/utils/ValidationResults.test.ts
Lionqueen94: Good point, changed the tests to test F401 in 1064b099a13d3df5a5aac3fb3ce2c97915bfb3b8

frontend/src/i18n/locales/nl/feedback_GSB_DSO.json
stacktraceghost: I agree it is a matter of preference.

stacktraceghost (dismissed):
Reviewed the code and compared the texts with the validation documentation.

jorisleker (dismissed):
Re-approved since last commit only changed a test. Did not review that

praseodym (changes requested):
String interpolation for F.403 is broken in the coordinator UI:

Image

Please add a test for this.

Also `frontend/src/i18n/feedback-structure.test.tsx` needs to be updated to validate the `feedback_GSB_DSO.json` file too.

Lionqueen94:
> String interpolation for F.403 is broken in the coordinator UI:
> Image
> 
> Please add a test for this.
> 
> Also `frontend/src/i18n/feedback-structure.test.tsx` needs to be updated to validate the `feedback_GSB_DSO.json` file too.

Fixed in 7d70a47a3f9a0e9de0908fb9f38fb65ba88b8e53, thanks for the check!

frontend/src/i18n/feedback-structure.test.tsx
```diff
@@ -26,4 +26,18 @@ describe("feedback translations structure", () => {
       }
     }
   });
+
+  test("validate feedback_GSB_DSO.json", () => {
```
praseodym, line 30: why copy-paste instead of adding `GSB_DSO` to the `test.each` above?

Lionqueen94: Because the types then become unknown, causing lots of lint issues. The DSO file only has coordinator keys and no titles, and Typescript doesn't understand it anymore

frontend/src/utils/ValidationResults.test.ts
```diff
@@ -176,21 +176,48 @@ describe("dottedCode", () => {
...
+  });
+
+  test("should return CSO coordinator translations for GSB validation result for CSO election with DSO override", () => {
+    expect(getTranslations(GSBCSOElection, validationResultMockData.F401, "coordinator")).toEqual({
+      code: "F.401",
+      title: t("feedback_GSB.F401.coordinator.title", { political_group_number: 1 }),
+      content: tx("feedback_GSB.F401.coordinator.content", undefined, { political_group_number: 1 }),
+      actions: tx("feedback_GSB.F401.coordinator.actions", undefined, { political_group_number: 1 }),
+    });
+  });
```
praseodym, line 230: I find it hard to reason about these tests because they compare a function call to the result of other function calls, wouldn't it be better to test against the actual copy instead?

Lionqueen94: Shall I do that just for the new tests or change it for all tests? They were already tested like this, that's why I continued using this testing format

praseodym: Just changing these new tests is fine by me

Lionqueen94: Changed in 5d242f78db4d2e34a519ef8b3228ac81c1fd03b8

praseodym (approved):

jorisleker (approved):
Same comment as previous approval :-)

PR #3881, opened 2026-08-27
Prevent more Vite warnings

Lionqueen94:
## What & why

Prevent more warnings about the Vite config using features that are unsupported by `configLoader: 'native'`

Prevents these warnings:
```
(!) Your Vite config uses features that are unsupported by `configLoader: 'native'`, which is planned to become the default in a future major version of Vite:                                                                                                
  - `__dirname` (vitest.config.ts:9:24). Use `import.meta.dirname` instead                                                     
  - import "./vite.config" without a file extension (vitest.config.ts:7:24). Add the file extension                            
Set `VITE_CONFIG_NATIVE_IGNORE_WARNING=true` to suppress this warning.
```

## How to test

Verify that the Vite warnings are gone when running the tests for example

### DIFF
--- a/frontend/vitest.config.ts
+++ b/frontend/vitest.config.ts
@@ -4,9 +4,10 @@ import { storybookTest } from "@storybook/addon-vitest/vitest-plugin";
 import { playwright } from "@vitest/browser-playwright";
 import { coverageConfigDefaults, defineConfig, mergeConfig } from "vitest/config";
 
-import viteConfig from "./vite.config";
+import viteConfig from "./vite.config.ts";
 
-const dirname = typeof __dirname !== "undefined" ? __dirname : path.dirname(fileURLToPath(import.meta.url));
+const dirname =
+  typeof import.meta.dirname !== "undefined" ? import.meta.dirname : path.dirname(fileURLToPath(import.meta.url));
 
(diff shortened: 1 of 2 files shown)


frontend/vitest.config.ts
```diff
@@ -4,9 +4,10 @@ import { storybookTest } from "@storybook/addon-vitest/vitest-plugin";
 import { playwright } from "@vitest/browser-playwright";
 import { coverageConfigDefaults, defineConfig, mergeConfig } from "vitest/config";
 
-import viteConfig from "./vite.config";
+import viteConfig from "./vite.config.ts";
 
-const dirname = typeof __dirname !== "undefined" ? __dirname : path.dirname(fileURLToPath(import.meta.url));
+const dirname =
```
praseodym: `import.meta.dirname` is never undefined so we can just inline it as `import.meta.dirname` below (`path.join(import.meta.dirname, ".storybook")`)

Lionqueen94: Even better, changed in ce0f09af2361644cc80a83b4fed47746770c666a

praseodym (approved):

stacktraceghost (approved):

PR #3883, opened 2026-08-27
Create election PS/WS

### LINKED ISSUE #3618: Create new PS/WS election

stacktraceghost:
## What & why

Resolves #3618 

Adds frontend support for adding PS/WS elections, including selecting the GSB for non-municipal GSB elections. See [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=23891-14974&m=dev) for the entire flow.

Commits and changes:
- [Add GSB selection](https://github.com/kiesraad/abacus/commit/83c1be5a95621dedc8483eab987c811aa5dcd99c)
  - Contains the flow changes for the frontend
  - `CheckAndSave.tsx` now has a typed request object
- [Tests SelectGSB, refactor ElectionMockData](https://github.com/kiesraad/abacus/commit/6e16909fe8f2513ee78fda85bb890d2f1d79a0ed)
  - Adds tests for the new GSB selection screen
  - Refactor of `ElectionMockData.ts` to prevent repetition
- [Add full flow tests and other tests](https://github.com/kiesraad/abacus/commit/b233a0a5451becd1bbebda086e19c5f93d97ac79)
  - Adds full flow tests for every combination
  - Adds PS/WS election mocks
- [Fix lint for ElectionCreateContextProvider](https://github.com/kiesraad/abacus/commit/4369b0914fb19d26ab6968fa51147cb7b095c265)
  - Removed some repetition
- [isOneOf helper](https://github.com/kiesraad/abacus/commit/44af611fddf16d83f5f34ef5f71331de16ee998d)
  - Handy helper that type narrows
  - Put in place where possible, allowed removal of some now obsolete code
- [Added eml_name to Election structs](https://github.com/kiesraad/abacus/pull/3883/changes/331b257878d02d06299b819d4748829acfb637db)
  - Adds eml_name field to Election structs that is not serialized
  - Adds this field to all instances of Elections

Note: `ElectionMockData.ts` could use further refactoring to support the different combinations more logically, however in this PR I had to stop somewhere.

## How to test

- Go through the flow by uploading several different verkiezingdefinities:
  - GR GSB/CSB
  - PS GSB + check that CSB is disabled
  - WS GSB/CSB
- Check if the flow matches [Figma](https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=23891-14974&m=dev)

## Reviewer notes

- Review per commit
- Verify if the test data is correct


backend/src/eml/mod.rs
```diff
@@ -61,6 +63,30 @@ fn max_votes(max_votes: &StringValue<NonZeroU64>) -> u32 {
     }
 }
 
+/// Format the EML election name for Abacus
+///+
```
Lionqueen94:
```suggestion
///
```

stacktraceghost: Applied in https://github.com/kiesraad/abacus/pull/3883/commits/3207d21841aef8f4a8abc2f6acedd7486a308dbc

backend/src/eml/mod.rs
```diff
@@ -189,7 +215,7 @@ impl NewElection {
 
         Ok((
             Self {
-                name: identifier.name.to_string(),
+                name: format_election_name(&identifier.name, category),
```
praseodym, line 218: I think also affects the election name that is written back to EML files that Abacus outputs, is that correct?

frontend/src/features/election_create/components/ElectionCreateContextProvider.tsx
```diff
@@ -73,17 +82,38 @@ export interface ElectionCreateState {
   isNumberOfVotersUserEdited?: boolean;
 }
 
-// biome-ignore lint/complexity/noExcessiveLinesPerFunction: TODO function should be refactored
+const resetPollingStations = {
+  pollingStations: undefined,
+  pollingStationDefinitionData: undefined,
+  pollingStationDefinitionFileName: undefined,
+  pollingStationDefinitionMatchesElection: undefined,
+} satisfies ElectionCreateState;
```
praseodym, line 92: Shouldn't this also reset `numberOfVoters` and `isNumberOfVotersUserEdited`?

frontend/src/testing/api-mocks/ElectionMockData.ts
```diff
@@ -375,142 +402,140 @@ export const getCSBElectionMockData = (
...
-      ],
-      redacted_indexes: [2, 9],
-    },
-    gsb_list: [],
-    election: newElectionMockData,
-    polling_stations: pollingStationMockData,
-    polling_station_definition_matches_election: matching_election,
-    number_of_voters: number_of_voters,
+    counting_method: "CSO",
+    number_of_voters: 2000,
```
praseodym, line 444: These should be unset for CSB elections

backend/src/eml/mod.rs
stacktraceghost: EML uses `election.name` for `<ElectionName>`, just checked. So yes, indirectly it affects the name that is written to the EML. 

Should we keep that name the same?

frontend/src/testing/api-mocks/ElectionMockData.ts
stacktraceghost: CSB elections get returned above these lines and the `baseElections` do not contain `counting_method`. Do you want to add it as an extra certainty?

frontend/src/features/election_create/components/ElectionCreateContextProvider.tsx
stacktraceghost: Changed in https://github.com/kiesraad/abacus/pull/3883/commits/14924e2503d1df42b7385f179a0234db36d7f18c. Also added a reset for `committeeCategory` and `countingMethod` when you upload a different election.

frontend/src/testing/api-mocks/ElectionMockData.ts
```diff
@@ -253,46 +258,68 @@ export const politicalGroupsMockData: PoliticalGroup[] = [
...
-      sub_category: "GR2",
-      number_of_seats: 29,
-      number_of_voters: 0,
-      election_date: "2024-11-30",
-      nomination_date: "2024-11-01",
-    },
+    { ...baseElectionGR, id: 1, committee_category: "GSB", counting_method: "CSO", number_of_voters: 2000 },
+    { ...baseElectionGR, id: 2, committee_category: "CSB", number_of_voters: 0 },
+    { ...baseElectionAB, id: 3, committee_category: "GSB", counting_method: "CSO", number_of_voters: 2000 },
+    { ...baseElectionAB, id: 4, committee_category: "CSB", counting_method: "CSO", number_of_voters: 2000 },
```
praseodym: CSB shouldn't have a counting method or number of voters

frontend/src/testing/api-mocks/ElectionMockData.ts
praseodym: This function is fine, but see the other comment: https://github.com/kiesraad/abacus/pull/3883/changes/BASE..3207d21841aef8f4a8abc2f6acedd7486a308dbc#r3880305604

frontend/src/testing/api-mocks/ElectionMockData.ts
stacktraceghost: Caught that this morning as well and fixed it in https://github.com/kiesraad/abacus/pull/3883/commits/a18333030fd6181eed48efd74c10e1e556c51b2d

backend/src/eml/mod.rs
praseodym: Yes in the output documents it should be identical to the election definition. So if we want to display a different name we either have to save that as a separate field or use `format_election_name` for display.

frontend/src/testing/api-mocks/ElectionMockData.ts
```diff
@@ -318,7 +318,7 @@ export const electionListMockResponse: ElectionListResponse = {
     { ...baseElectionGR, id: 1, committee_category: "GSB", counting_method: "CSO", number_of_voters: 2000 },
     { ...baseElectionGR, id: 2, committee_category: "CSB", number_of_voters: 0 },
     { ...baseElectionAB, id: 3, committee_category: "GSB", counting_method: "CSO", number_of_voters: 2000 },
-    { ...baseElectionAB, id: 4, committee_category: "CSB", counting_method: "CSO", number_of_voters: 2000 },
+    { ...baseElectionAB, id: 4, committee_category: "CSB", number_of_voters: 2000 },
```
praseodym: number_of_voters should also be 0

praseodym: I feel like we should refactor the election type similar to the `CommitteeSpecificTotals` in `ElectionTotals` so that we avoid setting fields that don't exist for this type of election, but that's for another time.

stacktraceghost: Fixed in https://github.com/kiesraad/abacus/pull/3883/commits/53692c7d7495d82a766229b690ada558c3f4868f

I agree the mocked data can use more refactoring. I have tried to go down the path of making everything more type-aware and restructure the way it is organized, but that would go too far out of scope for this PR.

praseodym: Created #3885 for the refactoring

backend/src/eml/mod.rs
Lionqueen94: My suggestion would be to add an `eml_name` and save the original name in it and use it for the eml files. Would that be a good solution @praseodym ?

praseodym: Yes that would work for now. Maybe we can skip serialization for that field so it doesn't get exposed to the frontend.

Lionqueen94: Added in 331b257878d02d06299b819d4748829acfb637db

rnijveld (dismissed):
Attempted creation of PS1, PS2 and WS GSB, all worked properly. Backend changes look ok, haven't looked in great detail at frontend changes, but looks good from my side!

backend/src/eml/mod.rs
praseodym: It looks like `get_eml_election_identifier_builder` is still using the display name instead of the new `eml_name` field?

backend/templates/inputs/model-n-10-1-variations/model-n-10-1-GR.json
```diff
@@ -2,6 +2,7 @@
   "election": {
     "id": 2,
     "name": "Gemeenteraad Juinen 2035",
+    "eml_name": "Gemeenteraad Juinen 2035",
```
praseodym: The `eml_name` field has `#[serde(skip_serializing)]`, so how can it end up in a serialized JSON?

backend/src/eml/mod.rs
Lionqueen94: Yes I missed this, fixed and tested in 331b257878d02d06299b819d4748829acfb637db 
Edit: Nomination date in election fixture needed to be updated to pass the eml validator - nomination date needs to be before election date

backend/templates/inputs/model-n-10-1-variations/model-n-10-1-GR.json
Lionqueen94: Fixed in 69b594a8f2341235fdd3c7c7563cc87a57d0a91c

praseodym (approved):

rnijveld (approved):
Looking at the `#[serde(skip_serializing)]` attributes on `eml_name` a little bit longer, I'm not convinced those attributes should have been added, as serializing and then deserializing one of those structs now results in dropping the eml_name attribute. But I'll create a separate issue to discuss that!

PR #3892, opened 2026-08-31
Full flow end to end tests for PS/WS GSB

### LINKED ISSUE #3619: Full flow end-to-end tests for PS/WS GSB

Lionqueen94:
Closes #3619 

## What & why

- Renamed existing full flow tests to specify they are for GR elections
- Added check for `voter_card_count` to be disabled if not provided to fill in
- Fixed CSB election import integration tests by now using correct and matching 110a and 230b files
- Converted existing full flow GSB and CSB tests into using test variants
  - Added full flow test for WS GSB CSO
  - Added full flow test for WS GSB DSO
  - Added full flow test for PS1 GSB CSO
  - Added full flow test for PS2 GSB DSO

## How to test

Run the tests with: `pnpm run test:e2e-ui tests/full-flow`


jschuurk-kr:
> Do we want to run all full flow tests for every commit? The newly added tests use real EMLs from the 2023 elections with 18-20 political groups. This causes them to run significantly longer (playwright test shards 2/3 and 3/3 now both take around 5.5m since they each contain half of the full flow tests). Or adding another shard might help reduce the runtime some more?

I want to, because these tests tell us if the basic flow for each election works.

Best solution I think is to reduce the number of candidates and political groups. That also means we need to replace the names of political groups and of candidates with fictional names. Because for testing we should either use un-edited historical data or fictional data.

frontend/e2e-tests/page-objects/election/create/CountingMethodTypePgObj.ts
```diff
@@ -7,9 +7,13 @@ export class CountingMethodTypePgObj {
...
-    this.header = page.getByRole("heading", { level: 2, name: "Type stemopneming in Test" });
+    this.header = page.getByRole("heading", { level: 2, name: /Type stemopneming in \w/ });
     this.cso = page.getByRole("radio", { name: "Centrale stemopneming (CSO)" });
     this.dso = page.getByRole("radio", { name: "Decentrale stemopneming (DSO)" });
     this.next = page.getByRole("button", { name: "Volgende" });
   }
+
+  async checkHeaderContainsName(name: string) {
+    await expect(this.page.getByRole("heading", { level: 2, name: `Type stemopneming in ${name}` })).toBeVisible();
+  }
```
jschuurk-kr: Nitpick (so ok not to change): in cases like this I prefer the page object to only get the element and the test to still do the assert.

frontend/e2e-tests/helpers-utils/e2e-test-browser-helpers.ts
```diff
@@ -147,9 +147,9 @@ export async function uploadPollingStations(page: Page, eml = eml110b) {
 function getCorrigendumFilename(countingMethod: VoteCountingMethod): RegExp {
   switch (countingMethod) {
     case "CSO":
-      return /Model_Na14-2_GR2022_Stembureau_\d+_Bijlage_1.pdf/;
+      return /Model_Na14-2_[A-Z]{2}\d{4}_Stembureau_\d+_Bijlage_1.pdf/;
     case "DSO":
-      return /Model_Na14-1_versie_2_GR2022_Stembureau_\d+.pdf/;
+      return /Model_Na14-1_versie_2_[A-Z]{2}\d{4}_Stembureau_\d+.pdf/;
```
jschuurk-kr, line 154: The test loses specificity through this change. I expect that we test the filename in our unit and/or our integration test, though, so not an issue for now.

In the near future I do want to review/refactor our full flow tests, to make sure we deal with the recent increase in the number of elections we support in a good way. That'll also allow us to extend them more easily with all the remaining elections later on.

After going through more of this PR, I'd like to do two kinds of refactor:
1. introducing more helpers so the tests in the full flow test get shorter and more readable;
2. trying to parametrize the full flow tests.

This refactor can wait until after our deadline for PS/WS27.

frontend/e2e-tests/page-objects/election/create/SelectGSBPgObj.ts
```diff
@@ -0,0 +1,15 @@
...
+  readonly header: Locator;
+  readonly regions: Locator;
+
+  constructor(protected readonly page: Page) {
+    this.header = page.getByRole("heading", { level: 2, name: "Selecteer het gemeentelijk stembureau" });
+    this.regions = page.getByTestId("region_list").locator("tbody").getByRole("row");
+  }
+
+  async clickRegionFromList(number: string) {
+    await this.page.getByTestId(`region-${number}`).click();
```
jschuurk-kr: Would be slightly better to have a `this.region_list = page.getByTestId("region_list")` which can then be used by `this.regions` and by `clickRegionFromList()`.

jschuurk-kr (dismissed):
We should reduce and ficitonalize the test data as I mentioned [here](https://github.com/kiesraad/abacus/pull/3892#issuecomment-5505710511).

frontend/e2e-tests/page-objects/election/create/SelectGSBPgObj.ts
Lionqueen94: Changed in 66feafd501d1b167ce8dde46b5dc31c59cb9bb68

frontend/e2e-tests/page-objects/election/create/CountingMethodTypePgObj.ts
Lionqueen94: Changed in 66feafd501d1b167ce8dde46b5dc31c59cb9bb68

stacktraceghost:
It would be great if we could limit the amount of repeating code in the full flow tests. Maybe we can make helpers for the parts that are similar?

frontend/e2e-tests/helpers-utils/e2e-test-browser-helpers.ts
Lionqueen94: I still refactored it now, since there was so much duplicate code, see ed5095f72e17e1839cb580e71a27e5fee3de8374

Lionqueen94:
The tests now all use fictional data with only 3 lists, so they are significantly faster, also they are now using TestVariants so the duplication is non-existent anymore. Please have a look at the changes @jschuurk-kr @stacktraceghost :)

jschuurk-kr:
> The tests now all use fictional data with only 3 lists, so they are significantly faster, also they are now using TestVariants so the duplication is non-existent anymore. Please have a look at the changes @jschuurk-kr @stacktraceghost :)

Looks like the test variants do not play nice with the sharding.

[e2e tests for this PR](https://github.com/kiesraad/abacus/actions/runs/34844218258):
image

[e2e tests on main](https://github.com/kiesraad/abacus/actions/runs/34830055400):
image

stacktraceghost:
> > The tests now all use fictional data with only 3 lists, so they are significantly faster, also they are now using TestVariants so the duplication is non-existent anymore. Please have a look at the changes @jschuurk-kr @stacktraceghost :)
> 
> Looks like the test variants do not play nice with the sharding

I think moving `test.describe.configure({ mode: "serial" });` into ```test.describe(`full flow ${variant.name}`, () => {```  might help here.

Lionqueen94:
> > > The tests now all use fictional data with only 3 lists, so they are significantly faster, also they are now using TestVariants so the duplication is non-existent anymore. Please have a look at the changes @jschuurk-kr @stacktraceghost :)
> > 
> > 
> > Looks like the test variants do not play nice with the sharding
> 
> I think moving `test.describe.configure({ mode: "serial" });` into `` test.describe(`full flow ${variant.name}`, () => { `` might help here.

That worked! :D
image

frontend/e2e-tests/test-data/eml-files.ts
```diff
@@ -158,3 +249,107 @@ export const eml230b_more_than_45_candidates: Eml230b = {
...
+    "014f",
+    "0e61",
+    "e737",
+    "1d0d",
+    "1639",
+    "999d",
+  ],
+};
+
+export const eml230b_AB_more_than_45_candidates: Eml230b = {
```
stacktraceghost, line 279: This test data seems unused.

frontend/e2e-tests/page-objects/election/create/SelectGSBPgObj.ts
```diff
@@ -0,0 +1,17 @@
+import type { Locator, Page } from "@playwright/test";
+
+export class SelectGSBPgObj {
+  readonly header: Locator;
+  readonly region_list: Locator;
```
stacktraceghost: nit: we use camelCase elsewhere

frontend/e2e-tests/tests/full-flow/full-flow-GSB.e2e.ts
```diff
@@ -0,0 +1,795 @@
...
+  },
+  {
+    name: "WS DSO",
+    electionDefinition: eml110a_AB,
+    candidateDefinition: eml230b_AB,
+    countingMethod: "DSO",
+    regionNumber: "0126",
+    regionName: "'s-Gravenveen",
+    electionName: "Waterschap Rivier en Polder 2023",
+    filename: "ab2023_'s-gravenveen_gemeente_'s-gravenveen",
```
stacktraceghost, line 127: Question: is it agreed that we allow apostrophe in filenames? In [another PR](https://github.com/kiesraad/abacus/pull/3947/changes#diff-bdbb772ccd9d2680a1ba35ffc1ef0ee873a3ea490e21cc448d64e29346c551d1R170) for example we strip them.

frontend/e2e-tests/test-data/request-response-templates.ts
```diff
@@ -325,6 +217,23 @@ export const noChecksAndCorrectionsDataEntryDSO: DSOFirstSessionResults & { mode
   ...structuredClone(commonDataEntry),
 };
 
+export const noRecountNoDifferencesWithVoterCardCountDataEntry: CSOFirstSessionResults & { model: "CSOFirstSession" } =
```
stacktraceghost: For typing, I'd like to suggest `Extract<Results, { model: "CSOFirstSession" }>` so we stay linked to the generated types.

stacktraceghost (commented):
Few comments, looks good otherwise!

Reviewed `frontend/e2e-tests/tests/full-flow/full-flow-CSB.e2e.ts` by diffing vs one of the removed ones.

frontend/e2e-tests/test-data/eml-files.ts
Lionqueen94: I added this for @sanderboom so he can use it for the CSB full flow test. We don't have enough candidates in the regular file (also for GR) and I thought to help him by adding this already for WS as well.

frontend/e2e-tests/tests/full-flow/full-flow-GSB.e2e.ts
Lionqueen94: Interesting, for now it downloads with apostrophes in the filenames. If we want to change this I suggest we add a new issue for this to align it with the PR you're referring to.

jschuurk-kr: They need to be stripped to match what OSV does and what the platform teluitslagen expects.

Examples from OSV test data: `definitieve-documenten_tk2025_gemeente_s-gravenhage-20240927-112123.zip` and `correctie_tk2025_gemeente_s-gravenhage-20240927-110202.zip`.

Lionqueen94: Okay then we need to add an issue for this, because that's not happening now. Do you want to add one @jschuurk-kr ?

frontend/e2e-tests/page-objects/election/create/SelectGSBPgObj.ts
Lionqueen94: Fixed in 762147d06f343216065169652753249822171308

frontend/e2e-tests/test-data/request-response-templates.ts
Lionqueen94: Changed in 762147d06f343216065169652753249822171308

stacktraceghost (dismissed):

frontend/e2e-tests/helpers-utils/e2e-test-browser-helpers.ts
Lionqueen94: This specific filename regex has been fixed in 5870ed50409d7fcf3b3bfd760855b211f2ac716a

stacktraceghost (approved):

frontend/e2e-tests/tests/full-flow/full-flow-GSB.e2e.ts
```diff
@@ -0,0 +1,802 @@
...
+  },
+  {
+    name: "WS CSO",
+    electionDefinition: eml110a_AB,
+    candidateDefinition: eml230b_AB,
+    countingMethod: "CSO",
+    regionNumber: "0123",
+    regionName: "Heemdamseburg",
+    electionName: "Waterschap Rivier en Polder 2023",
+    filename: "ab2023_heemdamseburg_gemeente_heemdamseburg",
```
jschuurk-kr, line 115: For a WS election, the first "heemdamseburg" should not be "heemsdamseburg" but the name of the water authority. Similar thing applies to PS elections, where it should be the name of the province.

__PS27__
Image

__WS27__
Image

frontend/e2e-tests/tests/full-flow/full-flow-GSB.e2e.ts
jschuurk-kr: new issue for stripping apostrophes: https://github.com/kiesraad/abacus/issues/3951

frontend/e2e-tests/tests/full-flow/full-flow-GSB.e2e.ts
jschuurk-kr: new issue to fix this in the application: https://github.com/kiesraad/abacus/issues/3955

oliver3 (approved):

Issue #3374, opened 2026-06-05
Epic: RSA signing GSB counts EML


praseodym:
Refinement: 2026-08-05 11u. 

# Context

OSV2020 compatible EML signing using an RSA keypair (public and private key). Validation on the CSB side is part of #3630.

When an election is created an RSA keypair is created. The public key can be downloaded from Abacus and uploaded to the [platform Uitwisseling](https://www.kiesraad.nl/documenten/2025/09/11/instructie-overdrachtsplatform). EML files created by Abacus are signed with the RSA private key. The higher level committees (HSB/CSB) can validate this signature with the public key published on the platform Uitwisseling.

CSB files do not need to be signed because there is no higher level committee to validate the signature (the signatures and public keys are not public).

# Architecture

RSA cryptography operations can reuse dependencies used for TLS ([rcgen](https://crates.io/crates/rcgen) + [aws-lc-rs](https://crates.io/crates/aws-lc-rs)). Consider putting the RSA code to its own crate and using a Rust file in the service layer to do the EML_NL file signing.

Keypair stored in a separate database table (untested):
```sql
CREATE TABLE signing_keypair
(
    election_id      INTEGER PRIMARY KEY NOT NULL,
    certificate      TEXT                NOT NULL, -- PEM-encoded certificate (contains public key)
    private_key      BLOB                NOT NULL, -- DER-encoded private key
    show_reminder    INTEGER             NOT NULL DEFAULT TRUE, -- boolean value

    FOREIGN KEY (election_id) REFERENCES election (id)
    CONSTRAINT election UNIQUE (election_id) 
) STRICT;
```

# Scope

For GSB elections (and later also HSB):
- Update use cases
- Write RSA operations code (keypair creation, signing, validation)
- Create RSA keypair during election creation and write to database
- Make RSA public key available for download, see [design in Figma](
https://www.figma.com/design/zZlFr8tYiRyp4I26sh6eqp/Kiesraad---Abacus-optelsoftware?node-id=21130-80921&p=f&t=heBcrWJSb2GRqj8O-0)
  - Use filename similar to OSV2020 for now (s/osv/abacus/)
- Need track if public key was uploaded to the platform ("Ik heb dit (al) gedaan" buttons)
- Sign created EML_NL files with the private key and write signature file to the ZIP file
  - Change copy on report page to include new signature file in file listing
- e2e tests to download public key and to check if EML signature is in ZIP file
- Test if importing signed EML_NL file works with OSV2020

[...]

PR #3896, opened 2026-09-02
Add eml_signature crate with keypair read/write

### LINKED ISSUE #3768: EML signature RSA keypair operations library
EML signature RSA operations library (crate?) for creating keypair. Signing and verifying signatures is part of #3912.

praseodym:
## What & why

Resolves #3768 (epic #3374). Add a new workspace crate `backend/eml_signature` to generate OSV2020-U-compatible RSA-4096 signing keypairs and certificates, as well as functions to parse certificates.

Signing/verifying EML files will follow in a second PR (#3912). Using this crate from Abacus itself is also future work.

## How to test

```sh
cd backend
cargo nextest run -p eml_signature
cargo run -p eml_signature --example create -- AB2027_Aardenboezem 9977 Nieuwstrand 2027-03-17
cargo run -p eml_signature --example inspect -- PSB9977.crt
```

`tests/osv2020.rs` has an integration test to verify that the crate can parse OSV2020-U-compatible certificate and re-serialize them in the same encoding.

## Reviewer notes

- Suggested review order: `lib.rs` → `README.md` → `subject.rs` → `certificate.rs` → `keypair.rs` → tests.
- Private keys should never leak: keys are `Zeroizing` and `SigningKeyPair` has no `Debug` impl that would print the key to stdout.


backend/eml_signature/src/keypair.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// The metadata certificate. It is exported as the `.crt` file and, in the
+    /// OSV2020-U format, embedded in every `.signature` file.
+    pub fn certificate(&self) -> &Certificate {
+        &self.certificate
+    }
+}
+
+/// Generate a fresh RSA-4096 keypair ([`RSA_KEY_BITS`]).
+fn generate_key_pair() -> Result<Zeroizing<KeyPair>, EmlSignatureError> {
+    const _: () = assert!(RSA_KEY_BITS == 4096, "RsaKeySize::_4096 below must match");
```
oliver3: Please add a small comment how this assertion works

backend/eml_signature/src/certificate.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+impl Certificate {
+    /// Read a certificate from X.509 DER.
+    pub fn from_der(der: &[u8]) -> Result<Self, EmlSignatureError> {
+        let certificate = X509Certificate::from_der(der).map_err(invalid_certificate)?;
+        let tbs = certificate.tbs_certificate();
+        let spki_der = tbs
+            .subject_public_key_info()
+            .to_der()
```
oliver3: Seems from the code that it is not necessary to do the DER encoding if you then only retrieve the public_key again from the DER?

backend/eml_signature/src/certificate.rs
```diff
@@ -0,0 +1,336 @@
...
+            public_key: PublicKey::from_der(&spki_der)?,
+        })
+    }
+
+    /// Read a certificate from PEM, the format of the `.crt` file that
+    /// OSV2020-U exports.
+    pub fn from_pem(pem: &[u8]) -> Result<Self, EmlSignatureError> {
+        let der = X509Certificate::from_pem(pem)
+            .map_err(invalid_certificate)?
+            .to_der()
```
oliver3: Not sure how costly DER encoding is, but could you also do `from_der()` and `from_pem()` that use a common helper `from_x509()` or something?

backend/eml_signature/src/keypair.rs
praseodym: This was there to prevent a mismatch between `RSA_KEY_BITS` and `RsaKeySize::_4096` at compile time, but because it would also be caught later in the certificate creation code I've dropped the assertion in 9be01cf7a6edd4e90b4a67377a1b85ff9badfa06.

backend/eml_signature/src/certificate.rs
praseodym: Indeed, changed in 9be01cf7a6edd4e90b4a67377a1b85ff9badfa06

backend/eml_signature/src/certificate.rs
praseodym: PEM is base64-encoded DER, so 9be01cf7a6edd4e90b4a67377a1b85ff9badfa06 changes `from_pem` to decode to DER and then use `from_der`.

oliver3 (approved):

rnijveld (approved):

PR #3921, opened 2026-09-08
Update and deny yanked dependencies

praseodym:
## What & why

- Update the chacha20 crate to version 0.10.2 because version 0.10.1 was yanked, see upstream issue https://github.com/RustCrypto/stream-ciphers/pull/583.
- Let cargo-deny also deny yanked dependencies, so that cargo-deny reports usage of yanked dependencies as an error instead of a warning.

## How to test

[Install cargo-deny](https://github.com/embarkstudios/cargo-deny#usage), then `cargo deny check` in the `backend` directory should no longer report the yanked chacha20 dependency.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -644,9 +644,9 @@ checksum = "f079e83a288787bcd14a6aea84cee5c87a67c5a3e660c30f557a3d24761b3527"
 
 [[package]]
 name = "chacha20"
-version = "0.10.1"
+version = "0.10.2"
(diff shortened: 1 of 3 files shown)


Lionqueen94 (approved):
LGTM! Cargo deny reports ok now

oliver3 (approved):
On main, `cargo-deny check` results in exit code 0 (`echo $?`) after warning about yanked crate `chacha20`

After adding `yanked = "deny"`, it gives an error and returns exit code 1

Issue #3771, opened 2026-08-05
Public key upload reminder


praseodym:
Reminder on election page to upload public key to platform, suppressed by "Dit heb ik (al) gedaan" button

Issue #3769, opened 2026-08-05
Generate and store keypair per election


praseodym:
For GSB only (and HSB in the future)

Generate a new keypair per election and store private key and certificate in the database. Integrate with API designed in #3770.

Investigate if we want to generate it lazily (service function), to try to reduce test execution times, and to prevent having to add it to the database fixtures.

RSA key generation takes 0.5-1.2 seconds on @praseodym's machine:  
```
  Time (mean ± σ):     468.5 ms ± 244.0 ms    [User: 452.3 ms, System: 3.2 ms]
  Range (min … max):   148.8 ms … 1194.9 ms    50 runs
```

PR #3926, opened 2026-09-08
Add signing_keypair db table and repository

### LINKED ISSUE #3923: Create database table and repository
Create database table and backend repository. See table description in architecture #3374.

oliver3:
## What & why

Resolves https://github.com/kiesraad/abacus/issues/3923

Database table to store the [certificate/private key](https://github.com/kiesraad/abacus/issues/3769) and [show_reminder](https://github.com/kiesraad/abacus/issues/3771) data

Build failed with clippy warning `cargo::non_kebab_case_bins`, so this is also fixed with a [separate commit](https://github.com/kiesraad/abacus/pull/3926/commits/f64779851def9cf62fd3bfa034cdd80dd9f300f0).

## How to test

Tests are successful. Further testing can be done in subsequent issues.

### DIFF
--- a/backend/apportionment/fuzz/Cargo.toml
+++ b/backend/apportionment/fuzz/Cargo.toml
@@ -20,14 +20,14 @@ tracing-subscriber = { version = "0.3.23", default-features = false, features = 
 path = ".."
 
 [[bin]]
-name = "simple-apportionment"
+name = "simple_apportionment"
 path = "fuzz_targets/simple_apportionment.rs"
 test = false
 doc = false
 bench = false
 
 [[bin]]
-name = "differential-osv2020"
+name = "differential_osv2020"
 path = "fuzz_targets/differential_osv2020.rs"
 test = false
 doc = false
@@ -55,14 +55,14 @@ doc = false
 bench = false
 
 [[bin]]
-name = "house-monotonicity"
+name = "house_monotonicity"
 path = "fuzz_targets/house_monotonicity.rs"
 test = false
 doc = false
 bench = false
 
 [[bin]]
-name = "population-monotonicity"
+name = "population_monotonicity"
 path = "fuzz_targets/population_monotonicity.rs"
 test = false
 doc = false
--- a/backend/apportionment/fuzz/Cargo.toml
+++ b/backend/apportionment/fuzz/Cargo.toml
@@ -20,14 +20,14 @@ tracing-subscriber = { version = "0.3.23", default-features = false, features = 
 path = ".."
 
 [[bin]]
-name = "simple_apportionment"
+name = "simple-apportionment"
 path = "fuzz_targets/simple_apportionment.rs"
 test = false
 doc = false
 bench = false
 
 [[bin]]
-name = "differential_osv2020"
+name = "differential-osv2020"
 path = "fuzz_targets/differential_osv2020.rs"
 test = false
 doc = false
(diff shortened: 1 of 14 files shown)


backend/migrations/11_signing_keypair.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE signing_keypair
+(
+    election_id   INTEGER PRIMARY KEY NOT NULL,
+    certificate   TEXT                NOT NULL,              -- PEM-encoded certificate (contains public key)
+    private_key   BLOB                NOT NULL,              -- DER-encoded private key
+    show_reminder INTEGER             NOT NULL DEFAULT TRUE, -- show upload reminder
+
+    FOREIGN KEY (election_id) REFERENCES elections (id),
+    CONSTRAINT election UNIQUE (election_id)
```
stacktraceghost: This constraint is redundant, since `election_id` is already a primary key.

backend/src/repository/signing_keypair_repo.rs
```diff
@@ -0,0 +1,122 @@
+use sqlx::SqliteConnection;
+use zeroize::Zeroizing;
+
+use crate::domain::election::ElectionId;
+
+pub async fn get_certificate(
+    conn: &mut SqliteConnection,
+    election_id: &ElectionId,
```
stacktraceghost: In other repos we are passing ids without `&`

oliver3: Ah yes, it is Copy. Removed in https://github.com/kiesraad/abacus/pull/3926/commits/e67e58d162e16723c606dfbd5b956ecb9b5f10f5

backend/migrations/11_signing_keypair.sql
oliver3: Well spotted thanks, removed from this and one other migration file in https://github.com/kiesraad/abacus/pull/3926/commits/060f084fdc685ff03392d7840e6b4fb742cbbf05

stacktraceghost (dismissed):

praseodym:
> Build failed with clippy warning `cargo::non_kebab_case_bins`, so this is also fixed with a [separate commit](https://github.com/kiesraad/abacus/pull/3926/commits/f64779851def9cf62fd3bfa034cdd80dd9f300f0).

Do we know why this suddenly popped up? The binary names haven't changed and neither did our toolchain version.

backend/apportionment/fuzz/Cargo.toml
praseodym, line 1: These names are also used in the nightly CI fuzz job, so I think that'll need to be updated too. But also see my other comment, I'm a bit surprised that we need these changes at all.

praseodym:
> > Build failed with clippy warning `cargo::non_kebab_case_bins`, so this is also fixed with a [separate commit](https://github.com/kiesraad/abacus/pull/3926/commits/f64779851def9cf62fd3bfa034cdd80dd9f300f0).
> 
> Do we know why this suddenly popped up? The binary names haven't changed and neither did our toolchain version.

I checked the [failed CI run](https://github.com/kiesraad/abacus/actions/runs/34236720839/job/102096217003) and it turns out that the `cargo::non_kebab_case_bins` lint is just a warning, the actual failure was that the lock file was outdated:
```
error: cannot update the lock file /home/runner/work/abacus/abacus/backend/fuzz/Cargo.lock because --locked was passed to prevent this
help: to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
```

You can see that [another CI run from earlier today](https://github.com/kiesraad/abacus/actions/runs/34240174842/job/102108008750) has the same warnings but passes overall.

I've created #3930 to make a more complete fix for this, @oliver3 can you remove the kebab-case changes from this PR?

praseodym (changes requested):
Please remove the kebab-case changes, see previous comment.

oliver3:
> Please remove the kebab-case changes, see previous comment.

Thanks for looking into it more thoroughly! Reverted the bin renaming (but not the updated lock file) in https://github.com/kiesraad/abacus/pull/3926/commits/bbbdbc0f4574b582eb386129a0612a275cd2e31c

praseodym (approved):

stacktraceghost (approved):

PR #3928, opened 2026-09-08
Add Water Authority support of test election generator

### LINKED ISSUE #3918: Add CSB WS to test data generator

cikzh:
## What & why

Closes #3918
Add support for Water Authority to test election generator

## How to test
Create elections via `/dev` "Genereer testverkiezing"

## Reviewer notes
Check for correct visibility of input field "Aantal GSB's"
Check that multiple GSB's are generated as provided by "Aantal GSB's"
Duplicate GSB names are very much possible, I'm not sure if we need to bother avoiding them though, let me know.

### DIFF
--- a/backend/src/test_data_gen/generators.rs
+++ b/backend/src/test_data_gen/generators.rs
@@ -1,14 +1,14 @@
 use std::error::Error;
 
 use chrono::{Datelike, Days, NaiveDate, TimeDelta};
-use rand::{SeedableRng, rngs::StdRng, seq::IndexedRandom};
+use rand::{RngExt, SeedableRng, rngs::StdRng, seq::IndexedRandom};
 use sqlx::{SqliteConnection, SqlitePool};
 use tracing::{info, warn};
 
 use crate::{
     SqlitePoolExt,
     domain::{
-        committee_session::{CommitteeSession, CommitteeSessionCreateRequest},
+        committee_session::{CommitteeSession, CommitteeSessionCreateRequest, CommitteeSessionId},
         committee_session_status::CommitteeSessionStatus,
         data_entry::{
             DataEntryId, DataEntrySource, DataEntryStatus, Definitive, FirstEntryFinalised,
@@ -40,7 +40,7 @@ use crate::{
             votes_counts::VotesCounts,
             yes_no::YesNo,
         },
-        sub_committee::SubCommitteeFirstSession,
+        sub_committee::{SubCommitteeFirstSession, SubCommitteeNumber},
         validate::Validate,
     },
     repository::{
@@ -111,6 +111,42 @@ async fn generate_csb_data_entries(
     Ok(second_entries > 0)
 }
 
+#[expect(clippy::too_many_arguments)]
+async fn generate_csb_sub_committee(
+    conn: &mut SqliteConnection,
+    rng: &mut StdRng,
+    args: &GenerateElectionArgs,
+    committee_session_id: CommitteeSessionId,
+    number: SubCommitteeNumber,
+    name: &str,
+    election: &ElectionWithPoliticalGroups,
+    votes: Option<Vec<Vec<u32>>>,
+) -> Result<bool, Box<dyn Error>> {
+    let sub_committee_first_session = create_sub_committee(
+        conn,
+        committee_session_id,
+        number,
+        name,
+        CommitteeCategory::GSB,
+    )
+    .await
+    .map_err(|e| format!("{e:?}"))?;
+
+    if args.with_data_entry {
+        generate_csb_data_entries(
+            conn,
+            rng,
+            args,
+            sub_committee_first_session,
+            election,
+            votes,
+        )
+        .await
+    } else {
+        Ok(false)
+    }
+}
+
 /// Generate polling stations and data entries for a GSB election
 async fn generate_gsb_election_data(
     rng: &mut StdRng,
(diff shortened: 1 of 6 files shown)


frontend/src/features/dev/components/GenerateTestElectionForm.tsx
```diff
@@ -22,6 +22,7 @@ import { StringFormData } from "@/utils/stringFormData";
 const RANGE_HINT = "Gebruik notatie zoals 10..50 of 9..=45 of een enkel getal zoals 40";
 
 const RANGE_FIELDS = [
+  { key: "gsbs", label: "Aantal GSB's", placeholder: "5..10" },
```
jschuurk-kr, line 25: If I leave this field empty and un-check "Inclusief invoer", I get a CSB with 0 GSBs. (If I leave the field checked, a number of GBSs are generated.)

jschuurk-kr: Same issue when I do fill the field "Aantal GSB's" and un-check "Inclusief invoer": no GSBs.
image

backend/src/test_data_gen/generators.rs
```diff
@@ -160,34 +196,53 @@ async fn generate_csb_election_data(
...
+            .await?
+        }
+        ElectionCategory::WaterAuthority => {
+            // We need at least one GSB
+            let gsbs = rng.random_range(args.gsbs.clone()).max(1);
+            let mut data_entry_complete = args.with_data_entry;
+            for number in 1..=gsbs {
+                let name = locality(rng);
+                data_entry_complete = data_entry_complete
+                    && generate_csb_sub_committee(
```
stacktraceghost, line 229: Without `args.with_data_entry` set to `false`, no sub committee is generated (issue of @jschuurk-kr). 

Note that `generate_csb_sub_committee` might return false on random chance as well, resulting in no data entries being generated after that point. The test doesn't catch this, because it only runs with data entry and chances at 100%.

backend/src/test_data_gen/generators.rs
```diff
@@ -160,34 +196,53 @@ async fn generate_csb_election_data(
...
+                        name,
+                        election,
+                        votes.clone(),
+                    )
+                    .await?;
+            }
+            data_entry_complete
+        }
+        ElectionCategory::Provincial => {
+            todo!("Provincial CSB election generation not supported")
```
stacktraceghost, line 244: When you submit the form with PS + CSB, it now throws an error. We could disable the option in the form or simply allow this arm as before (so PS+CSB gets generated without sub committees).

backend/src/test_data_gen/generators.rs
```diff
@@ -1116,18 +1172,25 @@ mod tests {
...
-            create_test_election(&municipal(GSB, Some(DSO)), &pool, None).await,
-            create_test_election(&municipal(CSB, None), &pool, None).await,
+            create_test_election(&election(GSB, Municipal, Some(CSO)), &pool, None).await,
+            create_test_election(&election(GSB, Municipal, Some(DSO)), &pool, None).await,
+            create_test_election(&election(CSB, Municipal, None), &pool, None).await,
+            create_test_election(&election(CSB, Municipal, None), &pool, None).await,
+            create_test_election(&election(GSB, WaterAuthority, Some(CSO)), &pool, None).await,
+            create_test_election(&election(GSB, WaterAuthority, Some(DSO)), &pool, None).await,
+            create_test_election(&election(CSB, WaterAuthority, None), &pool, None).await,
+            create_test_election(&election(CSB, WaterAuthority, None), &pool, None).await,
```
stacktraceghost, line 1193: CSB WaterAuthority is duplicated

jschuurk-kr:
Noticed three things that don't need addressing:
1. Test data generation does not result in audit log events. IIRC, this has been noticed and discussed in the past.
2. When a CSB does not have sub-committees the coordinator status page does not show a message similar to what we have for GSBs ("Er zijn nog geen stembureaus toegevoegd voor deze verkiezing.") Such a CSB should never exist, so there's no need to show a message.
3. Creating a CSB for a PS election results in a `502 Bad Gateway` and the following browser dialog:
Image

The backend output shows `not yet implemented: Provincial CSB election generation not supported`.
Since it might be a while before we implement CBS for PS, [you could argue](https://github.com/kiesraad/abacus/pull/3928#discussion_r3965744686) for a better error message or for not allowing this option in the frontend. We can consider that if the current solution leads to questions or confusion.

jschuurk-kr:
~~Created an election with data from AB23 and compared it with an election from the test data generator. Did not find any differences.~~

~~edit: no differences in the database. Checking a difference in the front-end.~~

jschuurk-kr (dismissed):
Election id, location, name should match.

### AB23 - Rivierenland

Image

### Test data generator

Image

Lionqueen94:
> Election id, location, name should match.
> ### AB23 - Rivierenland
> Image
> ### Test data generator
> Image

Regarding the first screenshot, "Gemeente Rivierenland" does not exist, because that is the water authority. We might still be making a mistake here in the frontend that we add "Gemeente" where this is not applicable for water authority CSB. Something to have a look into in this epic, if there are any more places where this happens.

jschuurk-kr:
> Regarding the first screenshot, "Gemeente Rivierenland" does not exist, because that is the water authority. We might still be making a mistake here in the frontend that we add "Gemeente" where this is not applicable for water authority CSB. Something to have a look into in this epic, if there are any more places where this happens.

https://github.com/kiesraad/abacus/issues/3936

backend/src/test_data_gen/generators.rs
cikzh: CSB Municipal was duplicate as well, removed them in [3377f83](https://github.com/kiesraad/abacus/pull/3928/commits/3377f833260993281b6d210794b1671faa327e2f)

backend/src/test_data_gen/generators.rs
cikzh: Yeah, I tried to be too smart with this one, creating a bug in the process. Fixed in [e812c22](https://github.com/kiesraad/abacus/pull/3928/commits/e812c22671b8ee5e0d00ced7901fc6156bf139df)

backend/src/test_data_gen/generators.rs
cikzh: Went for allowing the match arm in [949d7b4](https://github.com/kiesraad/abacus/pull/3928/commits/949d7b4880d1cbfde15df882974bccd2e6365a86) and adding a TODO comment.

frontend/src/features/dev/components/GenerateTestElectionForm.tsx
cikzh: This was a bug, fixed in [e812c22](https://github.com/kiesraad/abacus/pull/3928/commits/e812c22671b8ee5e0d00ced7901fc6156bf139df)

backend/src/test_data_gen/generators.rs
stacktraceghost: https://github.com/kiesraad/abacus/pull/3928/commits/e812c22671b8ee5e0d00ced7901fc6156bf139df undoes this change again

cikzh: Weird... I think I did an oopsie while rebasing. Fixed it now

stacktraceghost (dismissed):

backend/src/test_data_gen/generators.rs
jschuurk-kr: The link to the commit does not lead anywhere.

More importantly, when I now create a CSB for PS, no sub-committees are created. And if we want to allow creation of CSBs for PS, we should also distinguish PS1 (sub-committees are GSBs) from PS2 (sub-committees are HSBs). So I strongly prefer the old situation with a clear error message than the current situation where we create something that's not correct. Also since the scope of this PR is WS, not PS.

oliver3: Generator now fails again (using `todo!()`) for CSB + PS, and I've disabled the radio button which is a bit subtle but a hint what you're not supposed to do. I'll also have a look at the error message outside of this PR's scope.

sanderboom: Nice solution! :ok:

sanderboom (approved):
As mentioned I can trigger 

Image

when selecting PS before I select CSB (no prio at all though).

oliver3:
> As mentioned I can trigger
> Image
> 
> when selecting PS before I select CSB (no prio at all though).

[Added proper feedback](https://github.com/kiesraad/abacus/pull/3971) sent from the backend, in lieu of disabling the option.

backend/src/test_data_gen/generators.rs
oliver3: Thanks :) but I've updated it again with backend error handling in a separate PR again.

oliver3 (approved):

PR #3937, opened 2026-09-10
Report endpoints check committee session is part of election

### LINKED ISSUE #3924: `download_zip_results` does not check that committee session belongs to election
The `/api/elections/{election_id}/committee_sessions/{committee_session_id}/download_zip_results` API endpoint does not validate that the committee session identified by `committee_session_id` belongs to the election identified by `election_id`. However, the election id is used to generate the filenames for the download. This means that by changing the ids in the URL, a file can be downloaded with the name of one election and containing the results of the committee session from another election.

We should validate that the committee session with id `committee_session_id` belongs to the election with id `election_id`, and return an error otherwise. This needs to be fixed on `main` and then backported to the 1.1 release branch.

stacktraceghost:
## What & why

Resolves #3924

The endpoints in `report.rs` were not checking if the given `CommitteeSessionId` was part of the election given by `ElectionId`. This PR adds a function which validates if `committee_session.election_id` matches the given `election_id`. A test has also been added.

## How to test

- Validate the code
- Try to call one of these endpoints with a mismatching `election_id`

## Reviewer notes

N/A

### DIFF
--- a/backend/src/api/report.rs
+++ b/backend/src/api/report.rs
@@ -393,6 +393,7 @@ mod tests {
         response::{IntoResponse, Response},
     };
     use chrono::NaiveDateTime;
+    use http_body_util::BodyExt;
     use sqlx::SqlitePool;
     use test_log::test;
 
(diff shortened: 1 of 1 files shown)


oliver3 (approved):

Lionqueen94 (approved):
Looks good!

PR #3939, opened 2026-09-10
Use local AppData for Windows installation

### LINKED ISSUE #3925: Use local AppData for Windows installation
For the Windows installation, use local AppData (`%localappdata%`) instead of roaming AppData (`%appdata%`).

There is also a reference to this directory in the user documentation, which should also be updated.

This needs to be fixed on `main` and then backported to the 1.1 release branch.

stacktraceghost:
## What & why

Resolves #3925

## How to test

Run the installer on Windows, verify:
- The installation folder should be `%localappdata%\Abacus\`
- Application runs
- You are able to create backups
- When uninstalling, verify that all files are deleted when choosing that option (since the parameter changed)

## Reviewer notes

- I can provide a prebuilt installer.
- [InnoSetup constants](https://jrsoftware.org/ishelp/index.php?topic=consts)
  - `DefaultDirName={localappdata}\{#MyAppName}` is the installation folder. It makes `{app}` point to `%localappdata%\Abacus\`

### DIFF
--- a/packaging/windows/abacus.iss
+++ b/packaging/windows/abacus.iss
@@ -28,7 +28,7 @@ AppPublisher={#MyAppPublisher}
 AppPublisherURL={#MyAppURL}
 AppSupportURL={#MyAppURL}
 AppUpdatesURL={#MyAppURL}
-DefaultDirName={userappdata}\{#MyAppName}
+DefaultDirName={localappdata}\{#MyAppName}
 ; Setup will not show the Select Destination Location wizard page.
 DisableDirPage=yes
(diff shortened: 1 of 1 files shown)


sanderboom (approved):
Tested on Windows 10
- installs to `C:\Users\Sander\AppData\Local\Abacus`
- app runs
- desktop shortcut 'Abacus data' points to the correct folder
- backup creation works
- uninstall with and without 'keep files' works as expected

oliver3 (approved):
Works on Win11Pro (VM) too!

(install, run, backup, uninstall)

PR #3940, opened 2026-09-10
Adjust GSB list sorting

Lionqueen94:
## What & why

- Now ignoring leading `'s-` in gsb names when sorting gsb list, to make sure the sorting is improved.
- Replaced all instances of `'s Gravenveen` in testdata with `'s-Gravenveen` to match existing cities `'s-Hertogenbosch` and `'s-Gravenhage`

Main:
image

Now:
image

## How to test

Verify the changed table content in the updated test and the screenshots above.

## Reviewer notes

Can also solve this by ignoring everything before the firs capital letter in the name. Please comment your preference or any other solution that will fix this. There should be no other options other than `'s` according to https://nl.wikipedia.org/wiki/Lijst_van_Nederlandse_gemeenten
Another more official source for the alphabetical ordering is: https://www.cbs.nl/nl-nl/onze-diensten/methoden/classificaties/overig/gemeentelijke-indelingen-per-jaar/indeling-per-jaar/gemeentelijke-indeling-op-1-januari-2024


jorisleker:
On the linked Wikipedia page, `'s-Hertogenbosch` is spelled with a hyphen after `'s`. Not sure what the official spelling is, but we might want to ignore optional hyphens as well. 

If we decide to ignore everything before the first capital, then please take into account that the first capital might be a diacritical character as well.

Lionqueen94:
> On the linked Wikipedia page, `'s-Hertogenbosch` is spelled with a hyphen after `'s`. Not sure what the official spelling is, but we might want to ignore optional hyphens as well.
> 
> If we decide to ignore everything before the first capital, then please take into account that the first capital might be a diacritical character as well.

Good one, I changed it to `'s-` in all of our testdata and made sure that is ignored when sorting the GSBs

cikzh (approved):

oliver3 (approved):

PR #3942, opened 2026-09-13
Create e2e test: election creation for CSB WS

### LINKED ISSUE #3941: Create e2e test: election creation for CSB WS
(Initially https://github.com/kiesraad/abacus/issues/3915 was created for this task, but that was changed to creation of the full flow test - later it was decided that we still also want an e2e test for GSB WS, so hence this issue).

sanderboom:
## What & why

Resolves https://github.com/kiesraad/abacus/issues/3941. 

Test does:
- create a WS CSB election via the UI (using `eml110a_AB` and `eml230b_AB`)
- ... run similar tests used for 'create GR election'
- check whether a subcommittee is created for every GSB

## How to test

`pnpm test:e2e-ui tests/election/election-create.e2e.ts` > 'WS CSB election creation'

## Reviewer notes

Used `await test.step("check ...` so a failure shows which part of the test broke. AFAICS `test.step` was not used before in the codebase. Note that when using `pnpm run test:e2e-ui`, the step does not show up in the sidebar on the left, but as a collapsible section under 'Actions'. It does not change how the tests run. We probably want to split up `election-create.e2e.ts` at a certain moment.


frontend/e2e-tests/tests/election/election-create.e2e.ts
```diff
@@ -385,6 +386,59 @@ test.describe("Election creation", () => {
...
+      await expect(overviewPage.getAlertElectionCreated("CSB", "Waterschap Rivier en Polder 2023")).toBeVisible();
+
+      // The election overview page
+      const electionRow = overviewPage.findElectionRowById(election.id);
+      await expect(electionRow).toBeVisible();
+      await expect(electionRow).toContainText("Waterschap Rivier en Polder 2023");
+      await expect(electionRow).toContainText("CSB - Rivier en Polder (10)");
+      await expect(electionRow).toContainText("Klaar voor invoer— Zitting CSB");
+
+      await test.step("check it creates a subcommittee for every GSB", async () => {
```
jschuurk-kr, line 431: We have a [dependabot PR](https://github.com/kiesraad/abacus/pull/3962) that upgrades Playwright to v1.63. That version adds [Step params and subtitles](https://playwright.dev/docs/release-notes#-step-params-and-subtitles). Not sure if that would add anything of value, but worth a look.

frontend/e2e-tests/tests/election/election-create.e2e.ts
```diff
@@ -385,6 +386,59 @@ test.describe("Election creation", () => {
...
+        await page.goto(`/elections/${election.id}/status`);
```
jschuurk-kr, line 432: It's better to navigate through buttons or links than using `page.goto()`, because it gives us a little extra coverage on that navigation. I also suspect that `page.goto()` does a full reload, so using buttons/links should be faster. 🤞

frontend/e2e-tests/tests/election/election-create.e2e.ts
sanderboom: Experiment with the mentioned feature: https://github.com/kiesraad/abacus/commit/0d02b082a257cdbe6757b6ec2bc0f5e5e88a50fe 

`subtitle` is useful for a step that is repeated in a loop with a stable title and a varying value in the subtitle. I also moved logical blocks to separate steps. Pro of both methods: improve identification on failure. Con: extra code, the collapsed view in the Playwright UI and currently nested steps needs `eslint-disable-next-line playwright/no-nested-step`. Win some/lose some :thinking:

frontend/e2e-tests/tests/election/election-create.e2e.ts
sanderboom: Good point! Although in this case there is no reasonable clickable path available (there is no 'Bekijk voortgang' yet) :)

frontend/e2e-tests/tests/election/election-create.e2e.ts
jschuurk-kr: It's great you looked into this, but I'm not seeing much value in addding the steps. There's plenty of cons and the pro doesn't add much.

frontend/e2e-tests/tests/election/election-create.e2e.ts
jschuurk-kr: Ah, I see! Didn't realize that page was accessible through the url, but not yet through the UI.

frontend/e2e-tests/tests/election/election-create.e2e.ts
```diff
@@ -385,6 +386,59 @@ test.describe("Election creation", () => {
...
+        const statusPage = new ElectionStatus(page);
+        await expect(statusPage.notStartedRows).toHaveCount(4);
+        for (const name of ["Heemdamseburg", "Juinen", "Middelgein", "'s-Gravenveen"]) {
```
stacktraceghost: Instead of a loop, this could work as well:
```ts
const names = ["Heemdamseburg", "Juinen", "Middelgein", "'s-Gravenveen"];
expect(statusPage.notStartedRows).toContainText([...names]);
```

It requires names in that specific order.

frontend/e2e-tests/tests/election/election-create.e2e.ts
```diff
@@ -385,6 +386,59 @@ test.describe("Election creation", () => {
     });
   });
 
+  test.describe("WS CSB election creation", () => {
```
stacktraceghost: It would be nice to parameterize this test and combine it with the one above (like `full-flow-GSB/CSB.e2e.ts`) if it doesn't take too much time.

stacktraceghost (commented):
Some comments.

Generic question: it seems like we are only covering GR for GSB and not PS/WS in this file. But the full-flow tests also create an election. Isn't there a lot of overlap between those two?

jschuurk-kr:
> Generic question: it seems like we are only covering GR for GSB and not PS/WS in this file. But the full-flow tests also create an election. Isn't there a lot of overlap between those two?

There is indeed overlap between the full flow e2e tests and the election creation e2e tests. That's fine, though, since the scope/focus of the tests is different. As such, I would expect an election creation e2e test to have some additional asserts on the election having been created correctly, compared to the full flow test. Because the full flow (mostly) asserts implicitly on election creation, as in: you can complete the full flow with the created election.

frontend/e2e-tests/tests/election/election-create.e2e.ts
sanderboom: Fixed in a similar way in 81926c268db4945b882ab02c8e59c44897c369c9

frontend/e2e-tests/tests/election/election-create.e2e.ts
sanderboom: Good call - I'll take a look, should not take too much time.

sanderboom: Merged the two CSB tests in 25e5de861332ad9f4b6bdff1263a5d4feadb00c8

stacktraceghost (approved):

jschuurk-kr (approved):

Issue #3938, opened 2026-09-10
Determine what to do with certificate validity for expired elections


oliver3:
Generating the RSA keypair fails for old elections, because we check in the code that `not_before` (today) is not after `not_after` (election date plus three months).

For production this should never be a problem, but during development and trying out Abacus with old elections, this can lead to confusion as the current error message is "Abacus is stuk".

I've currently commented out the code that checks these dates, to be able to test the viewing/downloading certificate (publieke sleutel) functionality, but this should somehow be reinstated.

PR #3947, opened 2026-09-15
Generate and store election signing keypair

### LINKED ISSUE #3769: Generate and store keypair per election

oliver3:
## What & why

Resolves https://github.com/kiesraad/abacus/issues/3769

## How to test

- Generate a new GSB election
- As an administrator, go to the election's home page
- Go to the certificate ("publieke sleutel") page
- See that the public key appears after ~1 second
- Check that the details shown are as expected
- Download the certificate
- Determine if the filename is as expected
- You can check the content of the certificate with one of the following commands:

```sh
# Using Abacus source
cd backend
cargo run -p eml_signature --example inspect /path/to/public_key_abacus_gr2026_juinen_gemeente_juinen.crt

# Using OpenSSL (MacOS command, should work similarly on other OSes)
openssl x509 -inform pem -noout -text -in /path/to/public_key_abacus_gr2026_juinen_gemeente_juinen.crt

```

- Refresh the page and see that the certificate is there instantly. You could check a local database to see that the certificate is generated the first time and stored in the `signing_keypair` table.
## Reviewer notes

- ~~The tests in `service/signing.rs` call the actual `SigningKeyPair::generate()` function that will take a second or so. I will have a look in a separate PR if I can decouple that to improve the test execution time.~~ --> decided that this is not worth it for one keypair generation
- Created #3938 to determine what to do with checking if the certificate validity not_before is before not_after, otherwise existing fixture will throw an error when using the public/private keypair.


oliver3:
--> Draft, tests are failing

backend/src/service/signing.rs
```diff
@@ -0,0 +1,182 @@
...
+        )))?,
+    };
+
+    let subject = CertificateSubject::new(
+        election.election_id.clone(),
+        env!("ABACUS_GIT_VERSION"),
+        committee,
+    );
+
+    SigningKeyPair::generate(&subject, Utc::now().date_naive(), election.election_date)
```
stacktraceghost: This could be wrapped in `tokio::task::spawn_blocking` as hinted by the method doc
```rs
/// RSA-4096 key generation takes about a second, so call this from a
/// blocking thread when in async code.
```

oliver3: Good one, thanks. Fixed in https://github.com/kiesraad/abacus/pull/3947/commits/37481c0ef0c158911fcedd118138898e3f960a96

stacktraceghost (dismissed):

stacktraceghost (commented):
When clicking on "Ik heb dit al gedaan" on the election homepage or before the keypair has been generated on the keypair page, the `dismiss_public_key_upload_reminder` will return a 404 since the record doesn't exist in the table yet.

Reproduction steps:
1. Generate test GSB election
2. Log in as beheerder and go to election
3. Click on "Ik heb dit al gedaan"

oliver3:
> When clicking on "Ik heb dit al gedaan" on the election homepage or before the keypair has been generated on the keypair page, the `dismiss_public_key_upload_reminder` will return a 404

Created issue https://github.com/kiesraad/abacus/issues/3973 to fix this

stacktraceghost (dismissed):

frontend/src/i18n/locales/nl/log.json
```diff
@@ -44,6 +44,7 @@
     "PollingStationsImported": "Stembureaus geïmporteerd",
     "PollingStationsExported": "Stembureaus geëxporteerd",
     "PublicKeyUploadReminderDismissed": "Publieke sleutel uploadherinnering afgevinkt",
+    "SigningKeypairCreated": "Sleutel-paar voor digitale handtekeningen aangemaakt",
```
Lionqueen94:
```suggestion
    "SigningKeypairCreated": "Sleutelpaar voor digitale handtekeningen aangemaakt",
```
I think this is more correct in Dutch without the hyphen, but this is a very small nitpick

Lionqueen94 (dismissed):

frontend/src/i18n/locales/nl/log.json
oliver3: Applied suggestion in https://github.com/kiesraad/abacus/pull/3947/commits/f5db2b9494d23005ac81ec9aecde4b2165219651

stacktraceghost (approved):

Lionqueen94 (approved):

PR #3949, opened 2026-09-15
Fixed incorrect gemeente in copy frontend and duplicate CSB in NavBarLinks

### LINKED ISSUE #3936: Check for and fix incorrect "gemeente" in copy frontend
> Regarding the first screenshot, "Gemeente Rivierenland" does not exist, because that is the water authority. We might still be making a mistake here in the frontend that we add "Gemeente" where this is not applicable for water authority CSB. Something to have a look into in this epic, if there are any more places where this happens.

_Originally posted by @Lionqueen94 in https://github.com/kiesraad/abacus/issues/3928#issuecomment-5602088141_

Image

Lionqueen94:
Closes #3936 

## What & why

- Corrected testdata to have correct `authority_id` and other fields for CSB, and corrected `election_id`, `name`, `domain`, etc. where needed
- Removed `t("municipality")` from the 3 places where it's used
- Made sure committee category is only rendered for GSB in NavBarLinks, to prevent "CSB CSB"
- Adjusted tests to changes

## How to test

- Verify the changes in the code
- Verify there is no mention of "Gemeente" on the report pages and the election information table
- When creating a CSB election, verify that the navbar shows only a single "CSB" (see the screenshot in https://github.com/kiesraad/abacus/issues/3936#issuecomment-5615000396 to compare)

## Reviewer notes

When looking at the screenshots, the code and actual election definitions, I noticed some test data inconsistencies I thought important to solve, hence the many changed files, however there are not a lot of changed lines of code, so it should be easy to review.
The `NavBarLinks` issue was not spotted yet, but since it's present in the screenshot in this issue (https://github.com/kiesraad/abacus/issues/3936#issuecomment-5615000396), I thought to fix it now instead of later in a separate PR.


Lionqueen94:
@jorisleker Do we need an ID for CSBs in the `NavBarLinks` as well? In that case we can switch to using the `domain.id` instead of the `authority_id`

oliver3:
So this means that `"municipality"` can be removed from the `generic.json` as it is not used anymore

jorisleker:
> @jorisleker Do we need an ID for CSBs in the `NavBarLinks` as well? In that case we can switch to using the `domain.id` instead of the `authority_id`

After yesterdays discussion about the election definition: I think we could simplify by removing all ID's in the navbar because numbering is not consistently present for all domains (/levels) For now: okay with leaving them out for CSB.

jorisleker (dismissed):

oliver3 (dismissed):

jorisleker (dismissed):
Have not checked wether translation is truly unused. Approval based on previous review of UI

oliver3 (approved):

jorisleker (approved):

PR #3971, opened 2026-09-17
Add and use GenerateError in test generator

oliver3:
## What & why

The test generator used a non-specific `Box<dyn std::error::Error>` for its error type, which did not make it easy to return error messages to the user. This PR introduces a GenerateError enum that is currently either an SQL error or a unsupported error, that can be used to inform the user.

You will get the following error when selecting CSB + PS:

image

The `disabled` from the PS option (when CSB is selected) has been removed again, to avoid duplicating logic in the frontend and it was also not fool-proof as @sanderboom pointed out in https://github.com/kiesraad/abacus/pull/3928#pullrequestreview-5224324285

## How to test

Try to generate an election with CSB and PS from the frontend on the `/dev` page, or using the CLI with 

`cargo run --bin gen-test-election -- csb provincial`


oliver3:
Draft: waiting for other PR to merge first

stacktraceghost (dismissed):

backend/src/test_data_gen/generators.rs
```diff
@@ -243,7 +240,9 @@ async fn generate_csb_election_data(
             !data_entry_completes.contains(&false)
         }
         ElectionCategory::Provincial => {
-            todo!("Provincial CSB election generation not yet supported");
+            return Err(GenerateError::unsupported(
+                "CSB voor Provinciale Statenverkiezing is nog niet ondersteund",
+            ));
```
Lionqueen94, line 245:
```suggestion
            return Err(GenerateError::unsupported(
                "CSB voor Provinciale Statenverkiezing wordt nog niet ondersteund",
            ));
```
We so far didn't have Dutch language error messages I believe. I'm fine with it as is, since it's dev functionality anyway. But it is a bit different from the usual practise of adding the error translations in the frontend.

Lionqueen94 (approved):
Nice to see the error in the tracing as well now

stacktraceghost (approved):

PR #3975, opened 2026-09-17
Adjust election overview committee number

### LINKED ISSUE #3957: The elections overview page should show number/id of the committee, not the of the election domain
Follow-up of [this finding](https://github.com/kiesraad/abacus/issues/3913#issuecomment-5695548416) and [this comment](https://github.com/kiesraad/abacus/issues/3913#issuecomment-5695792627).

In the elections overview page, we currently show the domain id (if available) in the column "Type stembureau". That is confusing (see screenshot below) and incorrect. We should show the number of the actual committee, so e.g. for a GSB the municipal code (0106 for Assen).

I suspect this can be done by using the `RegionNumber` from the election definition, although we then need to be mindful for a CSB to use the `RegionNumber` of the `RegionCategory="WATERSCHAP"`, not of `RegionCategory="KIESKRING"`. That's also the region number that matches the `ElectionDomain Id` for WS. (PS election definitions do not have a ElectionDomain Id.)

Image

stacktraceghost:
## What & why

Resolves #3957 

Adjusted the number in the column `Type stembureau` in `OverviewPage`:
- GSB will show `authority_id`
- CSB will show `domain_id` if available

Made the function that generates the label exhaustive. Adding a new category in the future will force us to adjust this function.

## How to test

- Check if the correct number is shown
- Test it by importing an election

## Reviewer notes

In the process of deciding what it should be, I've imported a GSB GR/PS/WS and CSB GR/WS, as seen in the screenshow below.

image

Old situation:
image

New situation:
image


Lionqueen94 (approved):
Looks good

jschuurk-kr (approved):
Looks good!

Image

PR #3977, opened 2026-09-17
Ensure leading zeros for municipality codes in frontend

### LINKED ISSUE #3959: Gemeentecodes should always be prefixed with leading zeroes
Follow-up on [this finding](https://github.com/kiesraad/abacus/issues/3913#issuecomment-5695633639) and [this comment](https://github.com/kiesraad/abacus/issues/3913#issuecomment-5695837993).

In several CSB screens we do not show leading zeroes for municipal codes, while we should. However, we must never prefix polling station numbers, or voting district numbers.

The three occurrences found during testing (there might be more):
- the data entry status screen for the coordinator
- the list where a typist chooses which polling station to do data entry for
- the polling station number shown in the top left of the data entry screens

We do show these leading zeroes when creating a PS/WS GSB and you have to select a municipality.

sanderboom:
## What & why

Resolves https://github.com/kiesraad/abacus/issues/3959

Locations: 

- the data entry status screen for the coordinator                                                                                                                                                                                                                                                            
<img width="699" height="320" alt="image" src="https://github.com/user-attachments/assets/a7299d97-fb79-4437-8054-7ac522e41943" />

- the list where a typist chooses which polling station to do data entry for
<img width="669" height="306" alt="image" src="https://github.com/user-attachments/assets/267e02e9-a81d-48bc-8169-ff20f76b968a" />

- the polling station number shown in the top left of the data entry screens
<img width="401" height="88" alt="image" src="https://github.com/user-attachments/assets/dada6420-3862-4f4d-b7d2-c1431e6ff91c" />

- html title on the data entry screens
<img width="309" height="39" alt="image" src="https://github.com/user-attachments/assets/44021840-ab58-4c3b-8512-77de81ec0aa1" />

- the unfinished data entry link on the data entry home page
<img width="721" height="269" alt="image" src="https://github.com/user-attachments/assets/9a7ab7ed-ad09-46cc-9e5d-39e7cd36b6ef" />

- the number in the top left of the data entry detail screen
<img width="449" height="112" alt="image" src="https://github.com/user-attachments/assets/090276d9-5af6-4264-9c54-3492e00402f9" />

- deletion modal
<img width="475" height="261" alt="image" src="https://github.com/user-attachments/assets/7341ac67-3f56-4ff4-8a25-15572426d4ff" />

- confirmation message deletion
<img width="655" height="139" alt="image" src="https://github.com/user-attachments/assets/90e990c7-27a1-4f77-9da0-8253cda301f6" />

- after handling erroneous input
<img width="481" height="201" alt="image" src="https://github.com/user-attachments/assets/fd764b71-03e5-40fd-a129-adea88d8c200" />

Not screenshotted: 
- the confirmation messages shown after resuming or discarding a first entry
- the number in the top left of the resolve differences screen
- messages shown after resolving differences
- warning messages in data entry picker

Unchanged:
`id` and `key` attributes in the data entry list as those are data and not display values

```ts
                <Table.Row
                  id={`data-entry-row-${source.number}`}
                  key={source.number}
                  to={getUrlForDataEntry(electionId, dataEntry.statusEntry)}
                >
```

## How to test

Run the tests. New test file: `dataEntrySource.test.ts`.

## Reviewer notes

None.


Lionqueen94 (approved):
Looks good to me

rnijveld:
What we're actually printing here is the `authority_id` of the source as defined in PR #3981. It would be better if we don't do this formatting in the frontend because that creates an additional place where we're trying to format these numbers, which I think we should try to prevent. I think we should instead wait until that PR is merged and then directly print the authority_id if it is available or fall back to the number if it is a regular polling station. That also immediately fixes it for the future HSB implementation, because there the number formatting will fail again and that would introduce another duplicated number formatting if we would do that in the frontend as well.

sanderboom:
> What we're actually printing here is the `authority_id` of the source as defined in PR #3981. It would be better if we don't do this formatting in the frontend because that creates an additional place where we're trying to format these numbers, which I think we should try to prevent. I think we should instead wait until that PR is merged and then directly print the authority_id if it is available or fall back to the number if it is a regular polling station. That also immediately fixes it for the future HSB implementation, because there the number formatting will fail again and that would introduce another duplicated number formatting if we would do that in the frontend as well.

Good call - definitely the better solution. Update: 799b6dab62b6d86f49b3bd5a12e1a0f48b428f0f.

Also updated Storybook: 656a1b5b496391251ab08092a431d2c536fae7e7

Lionqueen94 (approved):

jorisleker (approved):
LGTM,

I think it is a bit confusing to use the municipality code in feedback and alerts. Numbers are the way to refer to polling stations, but when referring to GSBs (or HSBs in future releases), it makes more sense to use their names. We made a similar decision when sorting lists in #3979. Will create a separate issue for this.

jorisleker:
> LGTM,
> 
> I think it is a bit confusing to use the municipality code in feedback and alerts. Numbers are the way to refer to polling stations, but when referring to GSBs (or HSBs in future releases), it makes more sense to use their names. We made a similar decision when sorting lists in #3979. Will create a separate issue for this.

Created #3985 for this

PR #3979, opened 2026-09-18
Now sorting lists of GSBs on name instead of number of data entry source

### LINKED ISSUE #3958: Lists of GSBs should be sorted on name, not on number.
Follow-up on [this finding](https://github.com/kiesraad/abacus/issues/3913#issuecomment-5695599095) and [this comment](https://github.com/kiesraad/abacus/issues/3913#issuecomment-5695817391).

Currently we sort the list of GSBs for the CSB on "Nummer", not "Stembureau". This sorting makes sense for GSBs, where you do use the polling station number rather than the polling station name. For a list of GSBs it makes more sense to use the name to find the right GSB, so we should sort them on name instead.

So more generally: only lists of polling stations should be sorted on their number. Lists of GSBs and HSBs should be sorted on name.

The two occurrences found during testing (there might be more):
- list shown to typists where they pick a polling station to start data entry for.
- data entry status overview for the coordinator.

Note that in the election creation screens, we do sort the GSBs by name, so we should re-use that code. Also see PR https://github.com/kiesraad/abacus/pull/3940 for how that was implemented.

Lionqueen94:
Closes #3958

## What & why

- Now sorting lists of GSBs on name instead of number of data entry source

## How to test

- Generate a CSB WS test election without data entry
- Check that the CSB typist gets an alphabetically sorted list when they go to the data entry home page
- Check that the CSB coordinator gets an alphabetically sorted list when they go to the election status page
- Check that no change in ordering occurs for GSB elections (should remain sorted on number)

## Reviewer notes

@jschuurk-kr I would like to check the content of the data entries table in the e2e test, but am I correct that there's no easy way to check the content of a table in playwright? We have our custom `toHaveTableContent()` function for vitest which is great. Do you have a suggestion what I can use in playwright?


jschuurk-kr:
> @jschuurk-kr I would like to check the content of the data entries table in the e2e test, but am I correct that there's no easy way to check the content of a table in playwright? We have our custom `toHaveTableContent()` function for vitest which is great. Do you have a suggestion what I can use in playwright?

Playwright doesn't have anything for this, so you'd need to write a custom function like  have for vitest.

I don't think we really need to cover this in our e2e tests, though. The coverage you've added through the other tests should be sufficient.

frontend/src/features/election_status/components/ElectionStatus.stories.tsx
```diff
@@ -138,6 +138,88 @@ export const ElectionStatusNoLinks: StoryObj<StoryProps> = {
...
+  args: {
+    addLinks: false,
+  },
+  render: ({ addLinks, buttonNavigate }) => {
+    return (
+      <TestUserProvider userRole={"coordinator_csb"}>
+        <ElectionStatus
+          statuses={sortList(electionStatusesCSBMock, (item) => item.source)}
+          committeeSession={committeeSessionMockData}
+          election={electionMockData}
```
stacktraceghost: This is using GSB election mock data, we can use `csbElectionMockData` and `csbCommitteeSessionMockData` here

frontend/src/hooks/election/useElectionStatus.ts
```diff
@@ -9,5 +9,12 @@ export function useElectionStatus() {
     throw new Error("useElectionStatus must be used within an ElectionStatusProvider");
   }
 
-  return context;
+  const { statuses, refetch } = context;
+  // Sort statuses by source number (if source type is PollingStation) or source name (if source type is SubCommittee)
```
stacktraceghost: I just realized that components calling `useContext(ElectionStatusProviderContext);` directly, will get unsorted `statuses`. We do have cases that use the context like that, but they don't use the statuses.

To make it more consistent, we could also choose to call `sortList` inside `ElectionStatusProvider.tsx`.

frontend/src/features/election_status/components/ElectionStatus.stories.tsx
Lionqueen94: Changed in 43fd9fced884bde3c7d39e2322acf9c7bcba35ad

frontend/src/hooks/election/useElectionStatus.ts
Lionqueen94: Good one, moved sorting to the provider in 43fd9fced884bde3c7d39e2322acf9c7bcba35ad

stacktraceghost (dismissed):

sanderboom (dismissed):
Tested via the UI, the tests and double checked previous definitions for possible `’-s` (instead of `'-s`) which would mess up sorting (result: non-existent). Looks good!

stacktraceghost (approved):

sanderboom (approved):

PR #3981, opened 2026-09-18
Add `authority_id` and `authority_name` to `SubCommittee`

### LINKED ISSUE #3904: Store authority_id/authority_name for each sub-committee
While importing EML 510b/c files we need to match the file to be imported with the list of sub-committees for our election. This matching is only possible based on the `ManagingAuthority` element defined in 510b/c files, thus we need to store the information to match against for each sub-committee.

We can extract this information from the election tree when importing a new election (i.e. from the 110a file).

- Add authority_id and authority_name to `SubCommittee` struct.
- When creating new `SubCommittees` take the authority_id and authority_name from the `CommitteeDetails` struct available during importing.

Lionqueen94:
Closes #3904

## What & why

- Added `authority_id` and `authority_name` to `SubCommittee`

## How to test

- Review the code
- Check that the authority id and authority name are added to each subcommittee when you create a CSB for WS for example


backend/src/api/election.rs
```diff
@@ -771,6 +771,8 @@ async fn create_sub_committees(
             region_number,
             &committee.responsible_region.name,
             committee.category,
+            &committee.managing_authority_id,
+            committee.name.clone(),
```
rnijveld: You should use `committee.managing_authority_name()`. Maybe you can update the documentation for `CommitteeDetails::name` to indicate that you probably want to use `Self::managing_authority_name()` if you actually want to get the name of the committee?

backend/migrations/9_sub_committees.sql
```diff
@@ -6,6 +6,8 @@ CREATE TABLE sub_committees
     number                INTEGER                           NOT NULL,
     name                  TEXT                              NOT NULL,
     category              TEXT                              NOT NULL,
+    authority_id          TEXT                              NOT NULL,
+    authority_name        TEXT                              ,
```
rnijveld: Should also be required, we always have an authority_name.

backend/src/repository/sub_committee_repo.rs
```diff
@@ -47,13 +49,16 @@ pub async fn list_first_session(
 }
 
 /// Create a single sub electoral committee for a committee session
+#[expect(clippy::too_many_arguments)]
 pub async fn create(
     conn: &mut SqliteConnection,
```
rnijveld, line 52: I think we should add a `NewSubCommittee` instead of adding more arguments here

backend/src/repository/data_entry_repo.rs
```diff
@@ -97,8 +97,9 @@ pub async fn update(
     .await
 }
 
-/// Find which entity (polling station or sub committee) is the source of a data entry.
-/// We expect to only find one match: either a polling station or a sub committee, not both.
+/// Find which entity (polling station or subcommittee) is the source of a data entry.
+/// We expect to only find one match: either a polling station or a subcommittee, not both.
+#[expect(clippy::too_many_lines)]
```
rnijveld: Maybe we could split off the two actual queries into `fn try_fetch_polling_station_for_source()` and `fn try_fetch_sub_committee_for_source()` or something like that?

rnijveld (changes requested):

backend/src/api/election.rs
Lionqueen94: Changed in dfad770c8197ea78093fc3cc0e845971ef067990 also expanded the comment on `CommitteeDetails::name` as you suggested

backend/migrations/9_sub_committees.sql
Lionqueen94: Changed in dfad770c8197ea78093fc3cc0e845971ef067990

backend/src/repository/sub_committee_repo.rs
Lionqueen94: Done in dfad770c8197ea78093fc3cc0e845971ef067990

backend/src/repository/data_entry_repo.rs
Lionqueen94: Good idea, I extracted the queries in 787d4b3d4876cdbd406be32b87fc13a49719a705

rnijveld (dismissed):

backend/src/eml/committees.rs
```diff
@@ -171,7 +171,8 @@ pub struct CommitteeDetails {
 
     /// If this committee has a specific name, it will be set here.
     ///
-    /// Note: most committees do not have specific names.
+    /// Note: most committees do not have specific names, and you probably want to use
+    /// Self::managing_authority_name() if you actually want to get the name of the committee
```
stacktraceghost, line 175: We can remove `pub` if we should use `managing_authority_name()` and don't use `name`.

Lionqueen94: Done in 6c5f3b848568eab048a31ca2660f0ca6af65bb28

stacktraceghost (approved):

oliver3 (approved):

